PR #449, opened 2022-08-22
SW-1320 Handle input field on-change callbacks for whitespaces

karthikbtf:
If user enters nothing but spaces we don't want to trigger the on change callback.


src/components/common/Select/Select.tsx
```diff
@@ -131,8 +131,12 @@ export default function Select(props: SelectProps): JSX.Element {
   };
 
   const onChangeHandler = (e: ChangeEvent<HTMLInputElement>) => {
+    const textValue = e.target.value;
+    if (textValue.match(/^\s+$/)) {
+      return;
+    }
     if (onChange) {
```
karthikbtf, line 138: i will extract this into a utility

constanzauanini (approved):

PR #451, opened 2022-08-23
Replace old cancel button with new button component

constanzauanini:
Screen Shot 2022-08-23 at 11 37 46
Screen Shot 2022-08-23 at 11 38 01
Screen Shot 2022-08-23 at 11 38 23
Screen Shot 2022-08-23 at 11 38 35
Screen Shot 2022-08-23 at 11 38 48


karthikbtf (approved):
Thanks, for the remaining cleanup maybe we can ignore seed management that will get replaced?

constanzauanini:
> Thanks, for the remaining cleanup maybe we can ignore seed management that will get replaced?

I think that all the "old components" that we are cleaning up are used in seed management. New sections use new components

PR #456, opened 2022-08-24
Make entire box clickable on Dashboard

constanzauanini:


src/components/seeds/summary/AccessionByStatus.tsx
```diff
@@ -20,12 +22,13 @@ export default function AccessionByStatus({ label, status, quantity }: Props): J
       textAlign='center'
       width={isMobile ? '100%' : '20%'}
       padding='22px 10px 40px 10px'
+      onClick={() => history.push(`${APP_PATHS.ACCESSIONS}?stage=${status}`)}
```
karthikbtf, line 25: Link is better so we can support cmd+shift to open in a new tab.. can we move Link above the box?

karthikbtf: ok we can try later if it's not doable now.

karthikbtf (approved):

karthikbtf (approved):
Thanks!

PR #457, opened 2022-08-24
Use dialog box from web-components

constanzauanini:
Screen Shot 2022-08-24 at 11 34 25
Screen Shot 2022-08-24 at 11 34 37


karthikbtf (commented):
can we remove the style dictionary variables as well?
also a screenshot to confirm this works if possible?

karthikbtf (approved):
Thanks

PR #459, opened 2022-08-24
Use table component from web-components

constanzauanini:
Screen Shot 2022-08-24 at 13 54 45


karthikbtf (commented):
Deep paths are maybe not a good thing.
Lets make these exportable from web-components.
I can do that I guess.
Order will be renamed to SortOrder

karthikbtf (approved):
🎉  thanks @constanzauanini for updating web-components and terraware-web, that was full circle.

PR #461, opened 2022-08-29
SW-1528 Basic details view

constanzauanini:
Desktop:
Screen Shot 2022-08-29 at 11 19 49

Mobile:
Screen Shot 2022-08-29 at 11 20 28


src/components/accession2/DetailPanel.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const gridLeftSide = isMobile ? 12 : 4;
+
+  const gridRightSide = isMobile ? 12 : 8;
+  const today = moment();
+  const seedCollectionDate = accession?.collectedDate ? moment(accession?.collectedDate, ' YYYY-MM-DD') : undefined;
+
+  const age = seedCollectionDate ? today.diff(seedCollectionDate, 'months') : undefined;
+
+  console.log(age);
```
karthikbtf, line 36: remove?

src/components/accession2/DetailPanel.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const displayAge = () => {
+    if (age! < 1) {
```
karthikbtf, line 39: looks like age can be undefined according to code above, we should maybe handle the undefined case?

constanzauanini: It is managed before calling this method

karthikbtf: Hmm, doesn't make the code very readable then.
Can we at least have a check for age and if it is undefined output some error like "this should not happen" and return empty string?
I think we should make it obvious with some code, otherwise only the person that wrote this code would know.

karthikbtf (approved):
Looks good over all. Some comments if you want to follow up, I know this code is still evolving so we can always clean up later.

karthikbtf (approved):

karthikbtf:
Thanks, looks good 🎉

PR #464, opened 2022-08-30
reorg code for accession2

karthikbtf:
@constanzauanini I missed catching this earlier.
Can we use `/components/accession2` for all accession2 components? Directory is a bit generic for just Accession2View.


src/components/accession2/Accession2View.tsx
```diff
@@ -0,0 +1,67 @@
+import { TabContext, TabList, TabPanel } from '@mui/lab';
```
karthikbtf, line 1: i just renamed the previous index.tsx to Accession2View.tsx, this is not new code

constanzauanini: Sure

constanzauanini (approved):

PR #471, opened 2022-09-02
SW-1524 Intermediate work for creating new Accession2 

karthikbtf:
- Wanted to do this in smaller chunks
- Current effort will not save successfully yet

Terraware App 2022-09-01 16-29-32

Terraware App 2022-09-02 11-07-22

Terraware App 2022-09-02 11-06-51

### DIFF
--- a/src/api/accessions2/accession.ts
+++ b/src/api/accessions2/accession.ts
@@ -1,6 +1,7 @@
 import axios from 'axios';
 import { paths } from '../types/generated-schema';
 
+const ACCESSIONS2_ROOT_ENDPOINT = '/api/v2/seedbank/accessions';
 const ACCESSIONS2_ENDPOINT = '/api/v2/seedbank/accessions/{id}';
 
 type ListAutomationsResponsePayload =
@@ -37,4 +38,16 @@ export const updateAccession2 = async (accession: Accession2): Promise<UpdateAcc
   }
 
   return response;
+};
+
+type AccessionPostResponse =
+  paths[typeof ACCESSIONS2_ROOT_ENDPOINT]['post']['responses'][200]['content']['application/json'];
+
+export type AccessionPostRequestBody =
+  paths[typeof ACCESSIONS2_ROOT_ENDPOINT]['post']['requestBody']['content']['application/json'];
+
+export const postAccession = async (accession: AccessionPostRequestBody): Promise<string> => {
+  const response: AccessionPostResponse = (await axios.post(ACCESSIONS2_ROOT_ENDPOINT, accession)).data;
+
+  return response.accession.accessionNumber;
 };
--- a/src/api/accessions2/accession.ts
+++ b/src/api/accessions2/accession.ts
@@ -9,6 +9,21 @@ type ListAutomationsResponsePayload =
 
 export type Accession2 = ListAutomationsResponsePayload['accession'];
 
+export const Accession2Status = [
+  'Awaiting Check-In',
+  'Pending',
+  'Awaiting Processing',
+  'Processing',
+  'Cleaning',
+  'Processed',
+  'Drying',
+  'Dried',
+  'In Storage',
+  'Withdrawn',
+  'Used Up',
+  'Nursery',
+];
+
 export const getAccession2 = async (accessionId: number): Promise<Accession2> => {
   const endpoint = ACCESSIONS2_ENDPOINT.replace('{id}', `${accessionId}`);
   const response: ListAutomationsResponsePayload = (await axios.get(endpoint)).data;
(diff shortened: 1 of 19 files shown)


src/api/accessions2/accession.ts
```diff
@@ -38,3 +39,15 @@ export const updateAccession2 = async (accession: Accession2): Promise<UpdateAcc
...
 };
+
+type AccessionPostResponse =
+  paths[typeof ACCESSIONS2_ROOT_ENDPOINT]['post']['responses'][200]['content']['application/json'];
+
+export type AccessionPostRequestBody =
+  paths[typeof ACCESSIONS2_ROOT_ENDPOINT]['post']['requestBody']['content']['application/json'];
+
+export const postAccession = async (accession: AccessionPostRequestBody): Promise<string> => {
+  const response: AccessionPostResponse = (await axios.post(ACCESSIONS2_ROOT_ENDPOINT, accession)).data;
```
constanzauanini, line 50: We should add try, catch here

src/components/accession2/Collectors2.tsx
```diff
@@ -0,0 +1,108 @@
...
+      onChange(id, getNonEmptyCollectors(updatedCollectors));
+      return updatedCollectors;
+    });
+  };
+
+  return (
+    <>
+      {collectorsList.map((collector, index) => (
+        <Box key={index} mb={2} display='flex' alignItems='center' sx={{ display: 'block', position: 'relative' }}>
+          <Select
```
constanzauanini, line 68: In figma this field doesn't look like a Select, just a textfield

src/components/accession2/Accession2Create.tsx
```diff
@@ -0,0 +1,176 @@
...
+              {strings.SEED_COLLECTION_DETAIL}
+            </Typography>
+          </Grid>
+          <Grid item xs={12} sx={marginTop}>
+            <Species2Dropdown record={record} organization={organization} setRecord={setRecord} />
+          </Grid>
+          <Grid item xs={12} sx={datePickerStyle}>
+            <DatePicker
+              id='collectedDate'
+              label={strings.COLLECTED_DATE_REQUIRED}
```
constanzauanini, line 129: This label should be "Collecting Date" according to design

src/components/accession2/Accession2Create.tsx
```diff
@@ -0,0 +1,176 @@
...
+              aria-label={strings.COLLECTED_DATE_REQUIRED}
+              value={dates.collectedDate}
+              onChange={changeDate}
+              error={dateErrors.collectedDate === true}
```
constanzauanini, line 133: What does this error do in the UI? I tested and I don't see anything

karthikbtf: Yeah.. it doesn't work, that's sad. I hoped it would, same for maxDate property. I will create a Jira for us to clean up Date picker properties (make them work or remove them). I will remove this code here for now. Thanks for pointing that out.

karthikbtf:
@constanzauanini I am going to fill out the second half of the form so we can save, in this review. The code isn't much and I'm almost done.

karthikbtf:
@constanzauanini this is ready - i added TODOs for what I will do later.
I still need to add the expandable sections for address, gps coords, plant/site description.. which i will do next week.

src/components/accession2/Accession2Create.tsx
karthikbtf: I'm keeping labels closer to the payload properties for now, I believe the figma is wrong here

src/components/accession2/Collectors2.tsx
karthikbtf: we need someway of supporting autosuggest.. our current autosuggest may need some tweaks then.
let me keep this select for now until we have a new autosuggest component.

constanzauanini (approved):

src/components/accession2/Collectors2.tsx
karthikbtf: i have asked this question in figma, will have a follow up PR accordingly if needed

karthikbtf: Allison wants same behavior we currently have for collectors. I will make it a textfield later after we upgrade the textfield component.

karthikbtf:
@constanzauanini this one is ready to land as-is. Since you approved I made 3 changes:
1. use Autocomplete instead of Select for collectors (as per your feedback)
2. updated types using latest BE API (to set storageLocation during create accession2)
3. made the selectors for location/sub-location reusable

src/components/seeds/nursery/NewTest.tsx
```diff
@@ -58,7 +58,6 @@ function initTest(test?: ViabilityTest): ViabilityTest {
   return (
     test ?? {
       testType: 'Nursery',
-      selected: false,
```
karthikbtf, line 61: this change is from latest API changes

constanzauanini (approved):

PR #472, opened 2022-09-02
SW-1538 Clear suggested species names when less than 2 chars

alexTerraformation:
In the add species modal if the user has entered fewer than 2 characters, clear the suggested species list.

Screen Shot 2022-09-02 at 9 23 14 AM
Screen Shot 2022-09-02 at 9 23 40 AM
Screen Shot 2022-09-02 at 9 23 51 AM

### DIFF
--- a/src/components/Species/AddSpeciesModal.tsx
+++ b/src/components/Species/AddSpeciesModal.tsx
@@ -69,7 +69,7 @@ export default function AddSpeciesModal(props: AddSpeciesModalProps): JSX.Elemen
     const getOptionsForTyped = async () => {
       const requestId = Math.random().toString();
       setRequestId('names', requestId);
-      const response = await getSuggestedSpecies(debouncedSearchTerm);
+      const response = await listSpeciesNames(debouncedSearchTerm);
       if (response.requestSucceeded) {
         if (getRequestId('names') === requestId) {
           console.log(`Using species names response for value ${debouncedSearchTerm}`);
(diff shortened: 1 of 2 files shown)


src/components/Species/AddSpeciesModal.tsx
```diff
@@ -143,6 +141,18 @@ export default function AddSpeciesModal(props: AddSpeciesModalProps): JSX.Elemen
     }
   };
 
+  const getSuggestedSpecies = async (search: string) => {
```
karthikbtf, line 144: Nice optimization. We could add this code at the source https://github.com/terraware/terraware-web/blob/main/src/api/species/species.ts#L145  - if we want this behavior to reflect across usages?

karthikbtf: Got it, yes your original approach would make sense - my suggestion would be to apply this to all future usages (since this is backend implementation related).

constanzauanini: For me it is ok to have it here, because do we always want to return empty when searching with one character? I was thinking that if we use this listSpecies in a different search, we could want all the species that start with 'a' for example. But, as Alex said, this is the only usage, so we can move it to the api if you think is better Karthik

karthikbtf: My thought was we should have consistent search behavior in our app (also I thought BE cannot handle < 2 chars, could be a misunderstanding on my part).
Having said that - you are correct, we have only one use-case. It doesn't matter at this point.
@alexTerraformation if you haven't already made changes related to my suggestion - we can keep the existing version.

karthikbtf (approved):
🎉

src/components/Species/AddSpeciesModal.tsx
constanzauanini: Oh, I thought it was a FE behaviour only. If BE cannot handle, then you are totally right. Thanks @karthikbtf for the clarification

karthikbtf (approved):
Thanks 🎉

PR #473, opened 2022-09-02
Show photos on details view

constanzauanini:
Screen Shot 2022-09-02 at 17 21 20

### DIFF
--- a/src/components/accession2/DetailPanel.tsx
+++ b/src/components/accession2/DetailPanel.tsx
@@ -139,7 +139,15 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
           <Grid item xs={gridLeftSide} sx={categoryStyle}>
             {strings.PHOTOS}
           </Grid>
-          <Grid item xs={gridRightSide} />
+          <Grid item xs={gridRightSide} display='flex'>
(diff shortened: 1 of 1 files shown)


karthikbtf:
Oh wow, `Los Gigantes` !!
I believe we need to show thumbnails otherwise it will be too slow to download.

If you use a `maxHeight` and/or `maxWidth` query param you will get thumbnail.
```
/api/v1/seedbank/accessions/${accession.id}/photos/${file}?maxHeight=50
```

Clicking thumbnail should be an href to original image file.
Thumbnails should all be same height (not width, width will scale)

constanzauanini:
> Oh wow, `Los Gigantes` !! I believe we need to show thumbnails otherwise it will be too slow to download.
> 
> If you use a `maxHeight` and/or `maxWidth` query param you will get thumbnail.
> 
> ```
> /api/v1/seedbank/accessions/${accession.id}/photos/${file}?maxHeight=50
> ```
> 
> Clicking thumbnail should be an href to original image file. Thumbnails should all be same height (not width, width will scale)

Thanks @karthikbtf, I was not sure about the behaviour and asked Ming on the figma file. But I will do it like this then.

karthikbtf (approved):
🎉  nice!

PR #474, opened 2022-09-05
Add view photos modal

constanzauanini:
Screen Shot 2022-09-05 at 12 52 26

### DIFF
--- a/src/components/accession2/DetailPanel.tsx
+++ b/src/components/accession2/DetailPanel.tsx
@@ -1,11 +1,13 @@
 import { Box, Grid, Typography, useTheme } from '@mui/material';
 import { Button } from '@terraware/web-components';
 import moment from 'moment';
+import { useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import { Accession2 } from 'src/api/accessions2/accession';
 import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
+import ViewPhotosModal from './ViewPhotosModal';
 
 type DetailPanelProps = {
   accession?: Accession2;
@@ -34,6 +36,9 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
 
   const age = seedCollectionDate ? today.diff(seedCollectionDate, 'months') : undefined;
 
+  const [photosModalOpened, setPhotosModalOpened] = useState(false);
+  const [selectedSlide, setSelectedSlide] = useState(0);
+
   const displayAge = () => {
     if (age === undefined) {
       return '';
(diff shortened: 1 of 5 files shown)


src/components/accession2/DetailPanel.tsx
```diff
@@ -51,115 +56,129 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
...
+          </Grid>
+          <Grid item xs={12} sx={gridRowStyle}>
+            <Grid item xs={gridLeftSide} sx={categoryStyle}>
+              {strings.PHOTOS}
+            </Grid>
+            <Grid item xs={gridRightSide} display='flex'>
+              {accession.photoFilenames?.map((file, index) => {
+                return (
+                  <Box paddingRight={theme.spacing(2)} key={`photo-${index}`}>
+                    <img
```
constanzauanini, line 159: I added a fragment at the top so that I could add the <Modal> and it looks like a lot of changes, but changes are here. Basically replacing the old anchor with opening the new modal

src/components/accession2/ViewPhotosModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+      open={open}
+      title={strings.PHOTOS}
+      size='large'
+      middleButtons={[
+        <Button
+          label={strings.PREVIOUS_ARROW}
+          priority='secondary'
+          onClick={() => {
+            if (myCarousel.current) {
+              const prevSlide = myCarousel.current.state.currentSlide - 1;
```
karthikbtf, line 46: guessing the left and right buttons are disabled when they are at the beginning and last? i don't have pictures in staging so don't know how to test vercel.

karthikbtf (approved):
This looks good. Just curious, is this the requirement? Not sure I saw it in figma or PRD. Last I remembered we want to open the image on click in new tab. This looks good too, we may need to support a full resolution option,, like at least clicking on this carousel image should open in new tab. Otherwise we are only showing some fixed resolution. Just a thought.

src/components/accession2/ViewPhotosModal.tsx
constanzauanini: They are not disabled, but they don't do anything if you click them

constanzauanini:
> This looks good. Just curious, is this the requirement? Not sure I saw it in figma or PRD. Last I remembered we want to open the image on click in new tab. This looks good too, we may need to support a full resolution option,, like at least clicking on this carousel image should open in new tab. Otherwise we are only showing some fixed resolution. Just a thought.

Yes, this was what Ming answered when I asked what should happen when clicking the image. I will mention you in the figma file so you can follow up.

src/components/accession2/ViewPhotosModal.tsx
karthikbtf: Ah ok, we should disable if currentSlide is 0 and same for the other one if currentSlide is max?

karthikbtf:
> > This looks good. Just curious, is this the requirement? Not sure I saw it in figma or PRD. Last I remembered we want to open the image on click in new tab. This looks good too, we may need to support a full resolution option,, like at least clicking on this carousel image should open in new tab. Otherwise we are only showing some fixed resolution. Just a thought.
> 
> Yes, this was what Ming answered when I asked what should happen when clicking the image. I will mention you in the figma file so you can follow up.

Ok, that's good then. Can we add a click on photo to open in new tab without any width/height params?

karthikbtf:
> > > > This looks good. Just curious, is this the requirement? Not sure I saw it in figma or PRD. Last I remembered we want to open the image on click in new tab. This looks good too, we may need to support a full resolution option,, like at least clicking on this carousel image should open in new tab. Otherwise we are only showing some fixed resolution. Just a thought.
> > > 
> > > 
> > > Yes, this was what Ming answered when I asked what should happen when clicking the image. I will mention you in the figma file so you can follow up.
> > 
> > 
> > Ok, that's good then. Can we add a click on photo to open in new tab without any width/height params?
> 
> Sure, we can add that. I don't know why I can't mention you in figma, but this is the link to the comment: https://www.figma.com/file/ZqqFm4xaA5dtQJFHdyCujN?node-id=1929:397124#259261633

Thanks 🙏 I added a comment

karthikbtf (approved):

PR #482, opened 2022-09-07
SW-1525 Address support for Accession2 Create

karthikbtf:
- The UX has country last, maybe we should have that up ahead for logical flow of selection.. but... lets go according to design for now

Terraware App 2022-09-06 17-19-15

Terraware App 2022-09-06 17-22-08

Terraware App 2022-09-06 17-22-25

Terraware App 2022-09-06 17-22-42

### DIFF
--- a/src/components/accession2/Accession2Address.tsx
+++ b/src/components/accession2/Accession2Address.tsx
@@ -76,7 +76,7 @@ export default function Accession2Address(props: Accession2AddressProps): JSX.El
     return (
       <Grid item xs={12} marginTop={theme.spacing(2)}>
         <Box display='flex' justifyContent='flex-start'>
-          <Link sx={{ textDecoration: 'none' }} href='#' id='addCollectorButton' onClick={() => setIsOpen(true)}>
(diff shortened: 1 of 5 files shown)


src/components/accession2/Accession2Address.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const onChangeCountry = (newValue: string) => {
+    const found = countries?.find((country) => country.name === newValue);
+    if (found) {
+      onChange('collectionSiteCountryCode', found.code.toString());
+    }
+  };
+
+  const onChangeSubdivision = (newValue: string) => {
+    if (countries && record.collectionSiteCountryCode) {
+      const selectedCountry = getCountryByCode(countries, record.collectionSiteCountryCode);
```
constanzauanini, line 45: Could you use the getSelectedCountry here?

constanzauanini:
Love the screenshots with Argentina "che che che" hahaha

src/components/accession2/Accession2Address.tsx
karthikbtf: good catch, i copy pasted from add new org.. will fix in both places.

constanzauanini (approved):

PR #485, opened 2022-09-07
Disallow console.* unless explicitly permitted with a pragma directive

karthikbtf:
Lets catch this in builds in case we cannot in code review.
If we need a console.* we can use the `// tslint:disable-next-line: no-console` directive.


constanzauanini (approved):
We used to have this on before, I don't remember why we disabled

karthikbtf:
> We used to have this on before, I don't remember why we disabled

Thanks! We didn't have many anyways, but for future.

PR #488, opened 2022-09-08
SW-1568 Extract snackbar as an easier to use utility

karthikbtf:
- added some replaced examples to illustrate
- we can replace the other use-cases in smaller chunks

### DIFF
--- a/src/components/EditOrganization/index.tsx
+++ b/src/components/EditOrganization/index.tsx
@@ -7,8 +7,6 @@ import { ServerOrganization } from 'src/types/Organization';
 import TextField from '../common/Textfield/Textfield';
 import useForm from 'src/utils/useForm';
 import Select from '../common/Select/Select';
-import { useSetRecoilState } from 'recoil';
-import snackbarAtom from 'src/state/snackbar';
 import FormBottomBar from '../common/FormBottomBar';
 import { Country, Subdivision } from 'src/types/Country';
 import { searchCountries } from 'src/api/country/country';
@@ -17,6 +15,7 @@ import { getCountryByCode, getSubdivisionByCode } from 'src/utils/country';
 import { makeStyles } from '@mui/styles';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
 import PageSnackbar from 'src/components/PageSnackbar';
+import useSnackbar from 'src/utils/useSnackbar';
 
 interface StyleProps {
   isMobile: boolean;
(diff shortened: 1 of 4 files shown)


src/utils/useSnackbar.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // page helpers
+  const pageInfo = (msg: Message, title?: Title, closeCallback?: OnCloseCallback) =>
+    page(msg, title, 'info', closeCallback);
+  const pageWarning = (msg: Message, title?: Title, closeCallback?: OnCloseCallback) =>
+    page(msg, title, 'warning', closeCallback);
+  const pageSuccess = (msg: Message, title?: Title, closeCallback?: OnCloseCallback) =>
+    page(msg, title, 'success', closeCallback);
+  const pageError = (msg: Message, title?: Title, closeCallback?: OnCloseCallback) =>
+    page(msg, title, 'critical', closeCallback);
```
karthikbtf, line 41: wanted to use typed props here but figured this is easier to use in most of the use-cases

src/components/EditOrganization/index.tsx
```diff
@@ -113,18 +112,10 @@ export default function OrganizationView({ organization, reloadOrganizationData
     } else {
       const response = await updateOrganization(organizationRecord);
       if (response.requestSucceeded) {
-        setSnackbar({
-          type: 'toast',
-          priority: 'success',
-          msg: 'Changes saved',
-        });
+        snackbar.toastSuccess('Changes saved');
```
constanzauanini, line 115: This should have use string files, but I know it was like that before haha. So no need to change it now

karthikbtf: yeah, i saw that. i will move it to strings.

src/utils/useSnackbar.ts
```diff
@@ -0,0 +1,53 @@
...
+      onCloseCallback,
+      priority,
+    });
+  };
+
+  // toast helpers
+  const toastInfo = (msg: Message, title?: Title) => toast(msg, title, 'info');
+  const toastWarning = (msg: Message, title?: Title) => toast(msg, title, 'warning');
+  const toastSuccess = (msg: Message, title?: Title) => toast(msg, title, 'success');
+  const toastError = (msg?: Message, title?: Title) => toast(msg || strings.GENERIC_ERROR, title, 'critical');
```
constanzauanini, line 31: I like this default of generic error :)

constanzauanini (approved):

karthikbtf:
Thanks @constanzauanini i will update usages of other places in follow up PRs.

PR #490, opened 2022-09-09
SW-1536 Add new withdrawal

constanzauanini:
Screen Shot 2022-09-09 at 11 13 07
Screen Shot 2022-09-09 at 11 13 20


src/components/accession2/WithdrawModal.tsx
```diff
@@ -0,0 +1,273 @@
...
+  user: User;
+}
+
+export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element {
+  const { onClose, open, accession, reload, organization, user } = props;
+
+  const newWithdrawal: Withdrawal2 = {
+    purpose: 'Nursery',
+    withdrawnByUserId: user.id,
+    date: moment().format('YYYY-MM-DD'),
```
karthikbtf, line 33: maybe we could add this to src/utils/date.ts since we use it in a few places

src/strings/index.tsx
```diff
@@ -562,6 +567,11 @@ const strings = new LocalizedStrings({
     DELETE_ACCESSION_MESSAGE: 'You’re about to delete accession {0}.',
     ARE_YOU_SURE: 'Are you sure?',
     LATITUDE_LONGITUDE: 'Latitude, Longitude',
+    WITHDRAW: 'Withdraw',
+    AMOUNT_REMAINING: 'Amount ({0} remaining)',
+    WITHDRAW_ALL: 'Withdraw all',
+    CT: 'ct',
```
karthikbtf, line 573: i have a feeling this was not intentional, we should probably spell out 'count'..

src/components/accession2/WithdrawModal.tsx
```diff
@@ -0,0 +1,273 @@
...
+              options={['Nursery Media', 'Agar Petri Dish', 'Paper Petri Dish', 'Other']}
+              onChange={(value: string) => onChangeViabilityTesting('substrate', value)}
+              selectedValue={viabilityTesting.substrate}
+              fullWidth={true}
+              readonly={true}
+            />
+            <Select
+              label={strings.TREATMENT}
+              placeholder={strings.SELECT}
+              options={['Soak', 'Scarify', 'GA3', 'Stratification', 'Other']}
```
karthikbtf, line 203: best to export this as a type from src/types? same for other hardcoded lists in this file like test types

src/components/accession2/WithdrawModal.tsx
```diff
@@ -0,0 +1,273 @@
...
+      }
+      if (accession.withdrawals) {
+        accession.withdrawals?.push(record);
+      } else {
+        accession.withdrawals = [record];
+      }
+      const response = await updateAccession2(accession);
+      if (response.requestSucceeded) {
+        reload();
+      }
```
karthikbtf, line 76: else toast error

src/components/accession2/WithdrawModal.tsx
```diff
@@ -267,6 +273,17 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
             onChange={onChangeDate}
           />
         </Grid>
+        <Grid item xs={12} sx={{ marginTop: theme.spacing(2) }}>
+          {isNotesOpened ? (
+            <Textfield id='notes' value={record.notes} onChange={onChange} type='textarea' label={strings.NOTES} />
```
karthikbtf, line 278: oh no, these notes are used for the plant description. yikes, we need to ask Steve maybe.

constanzauanini: No, no. My record here is a withdrawal, so this field would be accession.withdrawal[0].notes

karthikbtf: ugh, my bad.. thanks for clarifying :)

karthikbtf (approved):
Looks good for a first version, we can iterate/clean-up as we go if there are any unanswered questions from BE.

src/strings/index.tsx
constanzauanini: I asked Ming in figm

karthikbtf (approved):
hope this works

PR #492, opened 2022-09-09
SW-1597 Update number of plants collected from to be single number

karthikbtf:
Terraware App 2022-09-09 11-35-04

Terraware App 2022-09-09 11-34-32


src/components/accession2/DetailPanel.tsx
```diff
@@ -143,7 +143,7 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
               {strings.PLANT_AND_SITE}
             </Grid>
             <Grid item xs={gridRightSide}>
-              {`${strings.COLLECTED_FROM} ${accession.plantsCollectedFromMax}-${accession.plantsCollectedFromMin} ${strings.WILD_PLANT}`}
+              {`${strings.COLLECTED_FROM} ${accession.plantsCollectedFrom} ${strings.WILD_PLANT}`}
```
karthikbtf, line 146: guessing WILD_PLANT will be changed to read from the record?

constanzauanini: Yes, we can update it now if it's in the api

constanzauanini (approved):

src/components/accession2/DetailPanel.tsx
karthikbtf: yes it is the collectionSource, i will update.

src/components/accession2/DetailPanel.tsx
```diff
@@ -143,7 +147,9 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
               {strings.PLANT_AND_SITE}
             </Grid>
             <Grid item xs={gridRightSide}>
-              {`${strings.COLLECTED_FROM} ${accession.plantsCollectedFrom} ${strings.WILD_PLANT}`}
+              {`${strings.COLLECTED_FROM}${numPlants === undefined ? '' : ' ' + numPlants}${
```
constanzauanini, line 150: Thanks for doing this!

PR #493, opened 2022-09-12
SW-1535 Add check in and new substrates/treatments for wthdrawal

constanzauanini:


src/components/accession2/Accession2View.tsx
```diff
@@ -73,6 +74,13 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
     textDecoration: 'none',
   };
 
+  const checkInAccession = async () => {
+    if (accession) {
+      await checkIn(accession.id);
```
karthikbtf, line 79: we should probably show a spinner somewhere while this is happening? Or at least some modal saying checking-in done - click ok. Otherwise the user will barely notice what happened.
I will ask in figma.

src/components/accession2/Accession2View.tsx
```diff
@@ -112,7 +120,11 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
             <IconButton sx={{ marginLeft: 3, height: '24px' }} onClick={() => setOpenDeleteAccession(true)}>
               <Icon name='iconTrashCan' />
             </IconButton>
-            <Button onClick={() => setOpenWithdrawModal(true)} label={strings.WITHDRAW} />
+            {accession && accession.state === 'Awaiting Check-In' ? (
```
karthikbtf, line 123: we have` if (accession) ` code in this file.. this component does not make sense without an accession. it should not be an optional property.

karthikbtf (commented):
looks good, i have a few comments.

src/components/accession2/Accession2View.tsx
karthikbtf: we can follow up separately maybe. there are no check-in design screens.

constanzauanini: I guess it's just like this because in the previous flow it was also like this (no modal or spinner)

src/components/accession2/Accession2View.tsx
constanzauanini: The problem here is that we get the accession from the accession id in the url, and if the id doesn't exist, then the accession is null/undefined

src/components/accession2/Accession2View.tsx
karthikbtf: > I guess it's just like this because in the previous flow it was also like this (no modal or spinner)

Yes, I understand. I asked a question on Figma.. we can follow up later if they respond.

karthikbtf (approved):
Just one request to make the accession not-optional.

src/components/accession2/Accession2View.tsx
karthikbtf: > The problem here is that we get the accession from the accession id in the url, and if the id doesn't exist, then the accession is null/undefined

We should then make the accession view a separate component, and the code that looks for accession id a parent component. This is the ideal way to design here.

karthikbtf: can we at least follow up separately? we don't need to put all the code in one component, we should separate by intent when we can.

karthikbtf (approved):
looks good for now, we can follow up on comments separately.

src/components/accession2/Accession2View.tsx
constanzauanini: Yes, of course

PR #495, opened 2022-09-12
SW-1605 Show in-situ/ex-situ suffixes for collection source

karthikbtf:
- added a new Select utility that has a { label, value } type for options but string for selectedValue and onChange
- we could maybe move this into web-components as a new Dropdown component? the Current one is outdated

Terraware App 2022-09-12 12-48-17

Terraware App 2022-09-12 12-47-59

Terraware App 2022-09-12 12-47-36

### DIFF
--- a/src/components/accession2/CollectionSource.tsx
+++ b/src/components/accession2/CollectionSource.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import strings from 'src/strings';
+import { SelectValue } from 'src/components/common/Select/Select';
+
+interface Props {
+  placeholder: string;
+  id: string;
+  label: string;
+  onChange: (value: string) => void;
+  selectedValue?: string;
+}
+
+export default function CollectionSource(props: Props): JSX.Element {
+  return (
+    <SelectValue
+      {...props}
+      fullWidth={true}
+      readonly={true}
+      options={[
+        {
+          label: strings.WILD_IN_SITU,
+          value: 'Wild',
+        },
+        {
+          label: strings.REINTRODUCED,
+          value: 'Reintroduced',
+        },
+        {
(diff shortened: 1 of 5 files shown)


src/components/accession2/CollectionSource.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+interface Props {
+  placeholder: string;
+  id: string;
+  label: string;
+  onChange: (value: string) => void;
+  selectedValue?: string;
+}
+
+export default function CollectionSource(props: Props): JSX.Element {
```
karthikbtf, line 13: made this a reusable component

src/components/common/Select/Select.tsx
```diff
@@ -1,2 +1,58 @@
...
+ * here. Not a candidate for web-components at this point.
+ *
+ * Example:
+ * <SelectValue
+ *   options=[{ label: 'label', value: 'value'}, { label: 'label1': value: 'value1'} ]
+ *   onChange={(value: string) => setSomeValue(value)}
+ *   selectedValue={'value1'}
+ * />
+ */
+export function SelectValue(props: SelectValueProps): JSX.Element {
```
karthikbtf, line 43: we could copy it to web-components as the new Dropdown.. or leave it here

constanzauanini: I could have use this select on my quantity PR, and also withdrawal.

constanzauanini (approved):

src/components/common/Select/Select.tsx
karthikbtf: ha ha yes, i added a comment in your PR. I noticed we have this use-case a lot.
i can merge this in sooner.. i will move it to Dropdown/web-components later and make relevant updates..

PR #499, opened 2022-09-13
SW-1610 Error validation for create-accession2 flow

karthikbtf:
- Added error validation for required fields and for dates
- Fixed the accession2 states list for create

Terraware App 2022-09-13 16-26-11

Terraware App 2022-09-13 16-25-42


karthikbtf:
@constanzauanini this PR will conflict with your PR.. I can handle the conflict resolution after you merge your work.

src/components/accession2/Accession2Create.tsx
```diff
@@ -67,28 +80,48 @@ export default function CreateAccession(props: CreateAccessionProps): JSX.Elemen
...
     const date = new Date(value).getTime();
     const now = Date.now();
-    if (isNaN(date) || date > now) {
+
+    setDateError(id, '');
+
+    if (isNaN(date)) {
+      setDateError(id, strings.INVALID_DATE);
+      return;
+    } else if (date > now) {
```
karthikbtf, line 100: Not using the server's time here.. but can make that change if needed.

constanzauanini: I don't think is necessary

src/components/accession2/Accession2Create.tsx
```diff
@@ -67,28 +80,48 @@ export default function CreateAccession(props: CreateAccessionProps): JSX.Elemen
...
+  const hasErrors = () => {
+    const missingRequiredField = MANDATORY_FIELDS.some((field: MandatoryField) => record[field] === undefined);
+    const hasDateErrors = dateErrors.recordedDate || dateErrors.collectedDate;
+    return missingRequiredField || hasDateErrors;
+  };
+
   const saveAccession = async () => {
-    // TODO data validation and show errors
+    if (hasErrors()) {
+      setDateError('collectedDate', record.collectedDate ? dateErrors.collectedDate : strings.REQUIRED_FIELD);
```
constanzauanini, line 120: Because we are changing the record only when the value exists, it can happen that if date was not set (undefined/null) and the you set and invalid date (5555-55-55), you get "invalid date" but after pressing save, this line changes the "Invalid date" error to "required field". Not very serious

constanzauanini: Excellent! Thank you!

karthikbtf:
@constanzauanini rebased off your latest changes and added support in edit view, if you can test.

PR #501, opened 2022-09-14
Display quantity correctly

constanzauanini:
Screen Shot 2022-09-14 at 11 35 59
Screen Shot 2022-09-14 at 11 36 18

### DIFF
--- a/src/components/accession2/Accession2View.tsx
+++ b/src/components/accession2/Accession2View.tsx
@@ -99,6 +99,47 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element 
     }
   };
 
+  const getAbsoluteQuantity = () => {
+    if (accession && accession.remainingQuantity) {
+      if (accession.remainingQuantity.units === 'Seeds') {
+        return (
+          <Typography display='flex' sx={{ marginRight: '5px' }}>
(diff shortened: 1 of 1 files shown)


src/components/accession2/Accession2View.tsx
```diff
@@ -99,6 +99,47 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
     }
   };
 
+  const getAbsoluteQuantity = () => {
+    if (accession && accession.remainingQuantity) {
+      if (accession.remainingQuantity.units === 'Seeds') {
+        return (
+          <Typography display='flex' sx={{ marginRight: '5px' }}>
+            <Typography sx={{ fontWeight: 600, marginRight: '5px' }}>{accession.remainingQuantity.quantity}</Typography>
+            <Typography>{strings.CT}</Typography>
```
karthikbtf, line 108: we aren't using 'COUNT'?

karthikbtf (approved):

src/components/accession2/Accession2View.tsx
constanzauanini: In all designs shows "ct", but I can change it if you think it should be count

karthikbtf: That's fine, we can ask Ming later and update if needed.

PR #502, opened 2022-09-14
SW-1607 Show withdrawal button if quantity is set

constanzauanini:


src/components/accession2/Accession2View.tsx
```diff
@@ -184,6 +184,13 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
     return '';
   };
 
+  const showWithdrawalButton = () => {
```
karthikbtf, line 187: `renderWithdrawalButton`?

karthikbtf (approved):
one suggestion but good to go

karthikbtf (approved):
Thanks!

PR #503, opened 2022-09-14
SW-1613 Check-in dialog confirmation modal

karthikbtf:
I have to run this by Ming but we can iterate after getting it into staging maybe.

Terraware App 2022-09-14 09-49-43


karthikbtf:
@minqingzhou thoughts on this modal?

constanzauanini (approved):

karthikbtf:
I will follow up separately if Ming has any suggestions.

PR #509, opened 2022-09-15
Add age, weight and count columns to database

constanzauanini:
Screen Shot 2022-09-15 at 10 47 39


src/components/seeds/database/Filters.tsx
```diff
@@ -127,19 +127,22 @@ export default function Filters(props: Props): JSX.Element {
...
-                onClick={(event) => handleClick(event, col)}
-                icon={<ArrowDropDown />}
-                className={props.filters[col.key] ? classes.selectedFilter : ''}
-              />
-            </div>
-          );
+          if (col.filter?.type !== 'hidden') {
+            return (
+              <div key={col.key} className={classes.pill}>
+                <Chip
```
karthikbtf, line 133: we should move a chip component to web-components so designers can control it

karthikbtf: this is for later

constanzauanini: This chip is old, I think at some point we will remove them and use filters as in species page. We have a pill component there that should be moved into web-components

karthikbtf: Got it, the sooner we do this the better I think - so the visual updates can be controlled by design. 
Since we are working on this now.. ??? Or at least a fast follow up?

src/components/seeds/database/columns.ts
```diff
@@ -319,6 +320,24 @@ const COLUMNS: DatabaseColumn[] = [
     type: 'string',
     filter: { type: 'multiple_selection' },
   },
+  {
+    key: 'age',
+    name: strings.AGE,
+    type: 'string',
+    filter: { type: 'hidden' },
```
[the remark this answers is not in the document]
constanzauanini: They are different columns, one is AGE (yr) and the other is AGE (months).

src/components/seeds/database/Filters.tsx
constanzauanini: I understood we were not going to update this filters UI now

karthikbtf: We don't need to change the filters.. just need to export this from web-components for designer control. So they can update styles on their own. If you think it's ok to keep it here that's fine too.

src/components/seeds/database/columns.ts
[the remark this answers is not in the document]
karthikbtf: Ah, got it. So we have two new columns in the list of columns.. and user can pick which column they want? Any of them enabled by default?

karthikbtf (approved):

src/components/seeds/database/Filters.tsx
constanzauanini: What if move this to web-components we will need to create a **Chip** component that is going to be replaced by the **Pill** component later.

src/components/seeds/database/columns.ts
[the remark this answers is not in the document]
constanzauanini: Yes, I will enable months by default. We can change it if we want years by default

src/components/seeds/database/Filters.tsx
karthikbtf: If we have a concrete plan to replace it with Pill before the visual updates effort then maybe not worth it.
Otherwise, there are benefits to extracting this out - even if only for a few months. Again, we can keep it here as-is if there's a strong preference. Sometimes these plans get delayed and we will be stuck with this Chip.
Fwiw, the pill is not in web-components either. I created a ticket for this.

karthikbtf: Maybe best to get @chudo76 's thoughts on this :) I will ask in the Jira.

constanzauanini: Thanks @karthikbtf !

PR #510, opened 2022-09-15
SW-313 Update content metdata in public.html with Terraware info

karthikbtf:
I don't think there's anything else in there needing an update.


constanzauanini (approved):
Steve will be happy! 🎉

sgrimm:
Hooray!

PR #518, opened 2022-09-19
Edit viability rate

constanzauanini:
Screen Shot 2022-09-19 at 15 00 00

Screen Shot 2022-09-19 at 14 56 13

### DIFF
--- a/src/components/accession2/edit/ViabilityModal.tsx
+++ b/src/components/accession2/edit/ViabilityModal.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { ChangeEvent, useEffect } from 'react';
 import strings from 'src/strings';
 import Button from 'src/components/common/button/Button';
 import DialogBox from 'src/components/common/DialogBox/DialogBox';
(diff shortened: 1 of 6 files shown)


src/components/accession2/edit/ViabilityModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const snackbar = useSnackbar();
+
+  const saveQuantity = async () => {
+    const response = await updateAccession2(record);
+    if (response.requestSucceeded) {
+      reload();
+      onCloseHandler();
+    } else {
+      snackbar.toastError();
+      setRecord(accession);
```
karthikbtf, line 33: if we skip this and the closehandler, user can continue trying.
Also, we need to add error handling like in the mocks (if percentage is > 100 etc.)

src/components/accession2/edit/ViabilityModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }, [accession, setRecord]);
+
+  const onCloseHandler = () => {
+    setRecord(accession);
+    onClose();
+  };
+
+  const onAddViabilityTest = () => {
+    setNewViabilityTestOpened(true);
+    onCloseHandler();
```
karthikbtf, line 49: if user successfully created a new viability test, looks like the best UX is to switch tabs to viability-testing - maybe we can pass in an extra param with line 48 to control this behavior?

src/strings/index.tsx
```diff
@@ -612,6 +612,8 @@ const strings = new LocalizedStrings({
     TESTING_STAFF: 'Testing Staff',
     START_DATE_REQUIRED: 'Start Date *',
     NUMBER_OF_SEEDS_TESTED_REQUIRED: '# Seeds Tested *',
+    VIABILITY_RATE: 'Viability Rate',
+    ADD_A_VIABILITY_TESTING: 'Add a Viability Testing',
```
karthikbtf, line 616: should be `Add a Viability Test` ... even if it says `Testing` in the mocks that would be incorrect

constanzauanini:
> Thanks for doing this @constanzauanini I had a jira assigned to me.. will reassign it to you. A few comments.

Oh, I was trying to find the Jira for this and couldn't find it!

src/components/accession2/edit/ViabilityModal.tsx
```diff
@@ -15,23 +15,33 @@ export interface ViabilityDialogProps {
   onClose: () => void;
   reload: () => void;
   setNewViabilityTestOpened: React.Dispatch<React.SetStateAction<boolean>>;
+  changeTab: (event: React.SyntheticEvent, newValue: string) => void;
```
karthikbtf, line 18: since we aren't really using the event, maybe we should remove it here and in the Accession2View file? sorry for this last nit.

karthikbtf (approved):
Thanks for following up. One more comment .. and also the strings change to 'Add a Viability Test'. Feel free to merge after.

karthikbtf (approved):
Thank you! 🎉

PR #521, opened 2022-09-20
Add pending-tests indicator to tab label

karthikbtf:
- also show seed bank name instead of storage location name

Terraware App 2022-09-20 10-10-55

### DIFF
--- a/src/components/accession2/view/Accession2View.tsx
+++ b/src/components/accession2/view/Accession2View.tsx
@@ -24,6 +24,7 @@ import useSnackbar from 'src/utils/useSnackbar';
 import useQuery from 'src/utils/useQuery';
 import useStateLocation, { getLocation } from 'src/utils/useStateLocation';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
+import { getSeedBank } from 'src/utils/organization';
 import _ from 'lodash';
 
 const useStyles = makeStyles(() => ({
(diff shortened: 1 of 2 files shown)


src/components/accession2/view/Accession2View.tsx
```diff
@@ -343,11 +373,11 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
             </Box>
           </Box>
         )}
-        {accession?.storageLocation && (
+        {accession?.facilityId && (
           <Box sx={editableParentProps}>
             <Icon name='iconMyLocation' className={classes.iconStyle} />
             <Box sx={editableProps}>
-              <Typography paddingLeft={1}>{accession.storageLocation}</Typography>
+              <Typography paddingLeft={1}>{getSeedBank(organization, accession.facilityId)?.name}</Typography>
```
constanzauanini, line 380: We are showing the seed bank only now, and not the storage location. Is that what we want?

karthikbtf: i can change this to `seed bank name (italic storage location name)` and we can iterate  on that?

karthikbtf: Added storage location, we can change the display later if Ming has another suggestion.

Terraware App 2022-09-20 10-27-29

karthikbtf: Ming commented that we need to show both.

constanzauanini: Thanks!

constanzauanini (approved):

src/components/accession2/view/Accession2View.tsx
karthikbtf: I asked if my current proposal "seed bank name (storage location name)" is ok. if she suggests something else I will send a new PR

PR #523, opened 2022-09-20
Allow notes to be edited for completed test, other fields are read-only.

karthikbtf:
- also refactored code a bit to make properties more generic

Terraware App 2022-09-20 16-11-31


src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
```diff
@@ -34,19 +33,21 @@ export interface NewViabilityTestModalProps {
   reload: () => void;
   organization: ServerOrganization;
   user: User;
-  selectedTest: ViabilityTest | undefined;
+  viabilityTest: ViabilityTest | undefined;
```
karthikbtf, line 36: i renamed properties a bit to clean up scope.

src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
```diff
@@ -9,22 +9,19 @@ export interface ViewViabilityTestModalProps {
   open: boolean;
   accession: Accession2;
   onClose: () => void;
-  selectedTest: ViabilityTest;
-  setNewViabilityTestOpened: React.Dispatch<React.SetStateAction<boolean>>;
-  setSelectedTest: React.Dispatch<React.SetStateAction<ViabilityTest | undefined>>;
```
karthikbtf, line 14: i moved controlling code into the parent since the child component here is not  aware of what's internal to parent

karthikbtf:
Making the 'notes' an in-place-edit in the View component is a nice-to-have but we can do that later if needed. It introduces a new pattern that needs more styling from design.
For now went with original plan of having everything disabled except notes (for a test that is marked complete).

src/components/accession2/view/Accession2View.tsx
```diff
@@ -302,20 +302,25 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
...
             />
           )}
           <NewViabilityTestModal
             open={newViabilityTestOpened}
             reload={reloadData}
             accession={accession}
-            onClose={() => setNewViabilityTestOpened(false)}
+            onClose={() => {
+              setNewViabilityTestOpened(false);
+              setSelectedTest(undefined);
```
constanzauanini, line 319: I think this is not working correctly. If I click on add test, fill some fields and close the modal with Cancel, and then click "add test" again, the data I filled before is still in the modal

constanzauanini: But it's unrelated with this change, it wasn't cleaning the fields before either

constanzauanini (approved):

src/components/accession2/view/Accession2View.tsx
constanzauanini: Yes, that is a good idea. I didn't realize we could do that to clean up

karthikbtf: Thanks, I will make that change.

karthikbtf: Made the change, if you get a chance to test let me know. It works great now. Noticed we had this problem in other modals as well like edit-state etc. All good now.

karthikbtf: > I was testing and on edit quantity modal, the link "go to weight calculator" is not working

Looks like we are calling onClose here, maybe we should introduce an on-hide here.

karthikbtf: actually, i can reuse the isCalculatorOpened flag

karthikbtf: I will just remove this check for quantity modal since it is already doing a cleanup on close.

PR #524, opened 2022-09-21
Add observations chart

constanzauanini:
Screen Shot 2022-09-21 at 10 34 49

### DIFF
--- a/src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
+++ b/src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
@@ -5,6 +5,7 @@ import strings from 'src/strings';
 import { ViabilityTest } from 'src/api/types/accessions';
 import { getFullTestType } from 'src/utils/viabilityTest';
 import ObservationsChart from './ObservationsChart';
+import { useDeviceInfo } from '@terraware/web-components/utils';
 
 export interface ViewViabilityTestModalProps {
   open: boolean;
@@ -17,6 +18,7 @@ export interface ViewViabilityTestModalProps {
 
 export default function ViewViabilityTestModal(props: ViewViabilityTestModalProps): JSX.Element {
   const { onClose, open, selectedTest, setNewViabilityTestOpened, setSelectedTest } = props;
+  const { isMobile } = useDeviceInfo();
 
   const onCloseHandler = () => {
     setSelectedTest(undefined);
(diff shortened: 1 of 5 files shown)


src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
```diff
@@ -124,9 +125,11 @@ export default function ViewViabilityTestModal(props: ViewViabilityTestModalProp
             );
           })}
         </Grid>
-        <Grid item xs={12}>
-          Chart
-        </Grid>
+        {selectedTest.testResults && selectedTest.testResults.length > 0 ? (
```
karthikbtf, line 128: ternary not needed for these rendering use-cases?

```
{ selectedTest.testResults && selectedTest.testResults.length > 0  && (
   <Grid item xs={12}>
       <ObservationsChart observations={selectedTest.testResults} />
   </Grid>
)}
```

karthikbtf (approved):
lgtm, can you check mobile version and post screenshot?
one suggestion but up to you, feel free to merge

src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
constanzauanini: Sometimes when I do that, and testResults exists but it has length 0, it renders a 0

karthikbtf: that's because the first check needs to be
```
{ selectedTest.testResults !== undefined
```
ok ternary is fine if you are more comfortable that way.

karthikbtf: if you were seeing zero it's most likely because some other check was `array.length` instead of `array.length > 0` since both are truthy checks in javascript. you can experiment when you have time to understand this better.

constanzauanini: Yes. Thank you for the explanation. It must have been that. Not checking the `> 0 `

karthikbtf:
@constanzauanini when you get a chance look at the mobile view, follow up PR if needed?

Terraware App 2022-09-21 07-09-12

karthikbtf:
> @constanzauanini when you get a chance look at the mobile view, follow up PR if needed?
> 
> Terraware App 2022-09-21 07-09-12

Thanks for the mobile styles. Modal could use a little bit of margin? Maybe for a follow up cleanup later.

karthikbtf (approved):

PR #525, opened 2022-09-21
SW-1705 Use withdrawals API for new withdrawals

karthikbtf:
- updated to use new withdrawals endpoint
- having issues withdrawing-all need to test that more


constanzauanini (approved):
Thanks for updating this!

PR #528, opened 2022-09-21
SW-1717 Fetch build-version file with dynamic query param to bust the cache

karthikbtf:
- to help fix the cached build-version fetch in production
- also added the page message component to accessions2 view

I didn't test using an nginx setup but can do that.

```
Request URL: http://localhost:3000/build-version.txt?timestamp=1663793133662
Request Method: GET
Status Code: 200 OK
```

Terraware App 2022-09-21 13-48-52


sgrimm (approved):
No opinion on the React part of this, but adding the parameter seems good.

constanzauanini (approved):

PR #530, opened 2022-09-22
SW-1672 Set viability percentage in accession

constanzauanini:
Screen Shot 2022-09-22 at 11 36 25

### DIFF
--- a/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
+++ b/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
@@ -100,7 +100,7 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
       if (response.requestSucceeded) {
         reload();
         onCloseHandler();
-        if (testCompleted) {
+        if (testCompleted && !readOnly) {
           setSavedRecord({ ...record });
           setOpenViabilityResultModal(true);
         }
--- a/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
+++ b/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
@@ -25,6 +25,7 @@ import { Close } from '@mui/icons-material';
 import { preventDefaultEvent } from '@terraware/web-components/utils';
 import { getTodaysDateFormatted } from 'src/utils/date';
 import { ViabilityTest } from 'src/api/types/accessions';
+import ViabilityResultModal from './ViabilityResultModal';
 
 export interface NewViabilityTestModalProps {
   open: boolean;
@@ -43,6 +44,8 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
   const [users, setUsers] = useState<OrganizationUser[]>();
   const [testCompleted, setTestCompleted] = useState<boolean>(viabilityTest?.endDate !== undefined);
   const [totalSeedsTested, setTotalSeedsTested] = useState(0);
+  const [openViabilityResultModal, setOpenViabilityResultModal] = useState(false);
+  const [savedRecord, setSavedRecord] = useState<ViabilityTest>();
   const contributor = isContributor(organization);
   const snackbar = useSnackbar();
   const theme = useTheme();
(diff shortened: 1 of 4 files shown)


src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
```diff
@@ -184,221 +191,234 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
   };
 
   return (
-    <DialogBox
```
constanzauanini, line 187: I didn't change anything here just moved all this inside <> </>, so I could add the modal component also

karthikbtf: Got it, thanks

src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
```diff
@@ -97,6 +100,10 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
       if (response.requestSucceeded) {
         reload();
         onCloseHandler();
+        if (testCompleted) {
+          setSavedRecord({ ...record });
+          setOpenViabilityResultModal(true);
+        }
```
karthikbtf, line 106: if user opens an already completed test and only edits the notes, we may still end up calling this. maybe this should be `if (testCompleted && !readOnly)`

src/components/accession2/viabilityTesting/ViabilityResultModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+import strings from 'src/strings';
+import useSnackbar from 'src/utils/useSnackbar';
+import { ViabilityTest } from 'src/api/types/accessions';
+
+export interface ViabilityResultModalProps {
+  open: boolean;
+  accession: Accession2;
+  onClose: () => void;
+  reload: () => void;
+  viabilityTest: ViabilityTest;
```
karthikbtf, line 13: Thanks for keeping the props clean without introducing parent semantics 🙏

src/components/accession2/viabilityTesting/ViabilityResultModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+  };
+
+  const getViabilityResult = () => {
+    let sum = 0;
+    viabilityTest?.testResults?.forEach((tr) => {
+      sum = sum + Number(tr.seedsGerminated);
+    });
+    const result = (sum / Number(viabilityTest.seedsTested)) * 100;
```
karthikbtf, line 40: can we use the percent in the viability test returned from BE? it's ok to reload data.

src/components/accession2/view/Accession2View.tsx
```diff
@@ -312,20 +312,20 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
               onEdit={() => setOpenNewViabilityTest(true)}
             />
           )}
-          {openNewViabilityTest && (
```
karthikbtf, line 315: why remove this?

src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
constanzauanini: Good catch!

src/components/accession2/viabilityTesting/ViabilityResultModal.tsx
constanzauanini: I tried to do that, but the problem is that if we are creating a new test we don't have the id to get that value back from the backend

src/components/accession2/view/Accession2View.tsx
constanzauanini: Beacause the new modal can't open if that one don't exist (after closing it)

src/components/accession2/viabilityTesting/ViabilityResultModal.tsx
karthikbtf: hmm, ok got it, we have no way of back referencing the viability test since BE returns the entire accession. maybe we can bring this up with Steve.. for now this is ok maybe.

src/components/accession2/view/Accession2View.tsx
constanzauanini: Yes, also tried that, but the new modal is opening behind :(

karthikbtf (approved):
Thanks for the changes.

PR #531, opened 2022-09-22
SW-1726 show edit state only if accession is checked-in

karthikbtf:
- show edit state only if accession is checked-in
- noticed that some accessions are already checked in but status is still awaiting check-in, these are mostly v1 accessions
- added a box to fill up space when we are not showing the edit icon so the spacing looks even

Terraware App 2022-09-22 08-32-01

Terraware App 2022-09-22 08-31-35

Terraware App 2022-09-22 08-30-23

### DIFF
--- a/src/components/accession2/view/Accession2View.tsx
+++ b/src/components/accession2/view/Accession2View.tsx
@@ -431,9 +431,13 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element 
               >
                 {accession.state}
               </Typography>
-              <IconButton sx={{ marginLeft: 3, height: '24px' }} onClick={() => setOpenEditStateModal(true)}>
-                <Icon name='iconEdit' className={`${classes.editIcon} edit-icon`} />
-              </IconButton>
+              {accession.state !== 'Awaiting Check-In' ? (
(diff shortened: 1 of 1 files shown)


src/components/accession2/view/Accession2View.tsx
```diff
@@ -431,9 +431,13 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
...
               </Typography>
-              <IconButton sx={{ marginLeft: 3, height: '24px' }} onClick={() => setOpenEditStateModal(true)}>
-                <Icon name='iconEdit' className={`${classes.editIcon} edit-icon`} />
-              </IconButton>
+              {accession.state !== 'Awaiting Check-In' ? (
+                <IconButton sx={{ marginLeft: 3, height: '24px' }} onClick={() => setOpenEditStateModal(true)}>
+                  <Icon name='iconEdit' className={`${classes.editIcon} edit-icon`} />
+                </IconButton>
+              ) : (
+                <Box sx={{ marginLeft: 3, height: '24px', width: 2 }} />
```
karthikbtf, line 439: i could remove this box and just add the spacing to the parent Box conditionally if needed.

constanzauanini: This if fine, I think it's more self-explanatory

constanzauanini (approved):

PR #532, opened 2022-09-22
SW-1716 Add validations to create viability flow

constanzauanini:
There are too many validations here, and I'm still testing if this is working correctly

Screen Shot 2022-09-22 at 16 55 06
Screen Shot 2022-09-22 at 16 55 15
Screen Shot 2022-09-22 at 16 55 23
Screen Shot 2022-09-22 at 16 56 54
Screen Shot 2022-09-22 at 17 47 20

### DIFF
--- a/src/strings/index.tsx
+++ b/src/strings/index.tsx
@@ -646,6 +646,9 @@ const strings = new LocalizedStrings({
     TOTAL_SEEDS_TESTED: 'Total Seeds Tested',
     APPLY_RESULT_QUESTION: 'Do you want to apply this result to the accession?',
     APPLY_RESULT: 'Apply Result',
+    SEEDS_TESTED_ERROR: 'Seeds tested should be less than remaining seeds',
+    SEEDS_GERMINATED_ERROR: 'Germinated amount should be less than tested amount.',
+    INVALID_VALUE: 'Invalid Value',
   },
 });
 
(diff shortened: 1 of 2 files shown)


src/strings/index.tsx
```diff
@@ -646,6 +646,9 @@ const strings = new LocalizedStrings({
     TOTAL_SEEDS_TESTED: 'Total Seeds Tested',
     APPLY_RESULT_QUESTION: 'Do you want to apply this result to the accession?',
     APPLY_RESULT: 'Apply Result',
+    SEEDS_TESTED_ERROR: 'Seeds tested should be less than remaining seeds',
+    SEEDS_GERMINATED_ERROR: 'Germinated amount should be less than tested amount.',
```
karthikbtf, line 650: both these messages should probably say 'should not be more than' instead of 'should be less than' (since they don't have to be less than, they can be equal to)

constanzauanini: Yes, you are right, it could be equal, I will change it. I took the message from the figma file but this make sense

karthikbtf, line 650: also might help to rename the key to `TOTAL_SEEDS_TESTED_ERROR` same for germinated

karthikbtf (approved):
Looks good thanks for doing this, have 2 comments about strings. The message may need to be fixed since it is technically incorrect.

karthikbtf (approved):
looks good, one suggestion on returning true vs false for the validateXYZ functions.

PR #535, opened 2022-09-23
SW-1736 allow editing quantity only when status is Drying or In Storage

karthikbtf:
- this is WIP, waiting on clarification

### DIFF
--- a/src/components/accession2/view/Accession2View.tsx
+++ b/src/components/accession2/view/Accession2View.tsx
@@ -297,6 +297,10 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element 
     width: isMobile ? '100%' : 'auto',
   };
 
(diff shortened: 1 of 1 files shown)


src/components/accession2/view/Accession2View.tsx
```diff
@@ -297,6 +297,10 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
     width: isMobile ? '100%' : 'auto',
   };
 
+  const spaceFiller = () => <Box sx={{ marginLeft: 3, height: '24px', width: 2 }} />;
+
+  const quantityEditable = accession?.state === 'Drying' || accession?.state === 'In Storage';
```
karthikbtf, line 302: This is WIP, we may need to add 'Awaiting Check-In' here.

constanzauanini (approved):

PR #536, opened 2022-09-23
Validations and spacing to withdrawal modal

constanzauanini:
Screen Shot 2022-09-23 at 14 08 44

Screen Shot 2022-09-23 at 14 22 01

### DIFF
--- a/src/components/accession2/withdraw/WithdrawModal.tsx
+++ b/src/components/accession2/withdraw/WithdrawModal.tsx
@@ -130,13 +130,24 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element 
 
   const onSelectAll = (id: string, value: boolean) => {
     if (!withdrawAllSelected) {
-      setRecord({
-        ...record,
-        withdrawnQuantity: {
-          quantity: accession.remainingQuantity?.quantity || 0,
-          units: accession.remainingQuantity?.units || 'Grams',
-        },
-      });
+      if (record.purpose === 'Viability Testing') {
+        // if purpose is VT we use estimated count
+        setRecord({
+          ...record,
+          withdrawnQuantity: {
+            quantity: accession.estimatedCount ? accession.estimatedCount : 0,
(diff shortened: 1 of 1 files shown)


src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -162,6 +178,32 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
+        setIndividualError('withdrawnQuantity', strings.INVALID_VALUE);
+        return false;
+      } else {
+        if (Number(record.withdrawnQuantity.quantity) <= 0) {
+          setIndividualError('withdrawnQuantity', strings.INVALID_VALUE);
+          return false;
+        }
+      }
+    } else {
+      setIndividualError('withdrawnQuantity', strings.REQUIRED_FIELD);
```
karthikbtf, line 200: if quantity is 0 we will end up here? if so, maybe we need a different message for quantity=0?

karthikbtf (commented):
should there be space between the unit dropdown and the amount field?

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -162,6 +178,32 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
+      return false;
+    }
+    setIndividualError('', strings.REQUIRED_FIELD);
```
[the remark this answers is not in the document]
constanzauanini: Ups, correct. Thanks

src/components/accession2/withdraw/WithdrawModal.tsx
constanzauanini: 0 is '0' as string, so you end up in the invalid value message

karthikbtf: ah yes, forgot this is a string input

karthikbtf (approved):
Thanks 👍

PR #538, opened 2022-09-26
Import accessions data

constanzauanini:
This has a lot of repeated code. I could try to make a component to reuse code.

I don't see accessions created after import, and I couldn't test the repeated accession number either.

Screen Shot 2022-09-26 at 11 52 11
Screen Shot 2022-09-26 at 11 51 49
Screen Shot 2022-09-26 at 11 52 03

### DIFF
--- a/src/components/seeds/database/ImportAccessionsModal.tsx
+++ b/src/components/seeds/database/ImportAccessionsModal.tsx
@@ -1,0 +1,354 @@
+import { Link } from '@mui/material';
+import { Theme } from '@mui/material';
+import { makeStyles } from '@mui/styles';
+import { Button, DialogBox, Icon, ProgressCircle } from '@terraware/web-components';
+import React, { useEffect, useRef, useState } from 'react';
+import {
+  downloadAccessionsTemplate,
+  getAccessionsUploadStatus,
+  GetAccessionsUploadStatusResponsePayload,
+  resolveAccessionsUpload,
+  uploadAccessionsFile,
+} from 'src/api/accessions2/accession';
+import { Facility } from 'src/api/types/facilities';
+import strings from 'src/strings';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  spacing: {
+    marginRight: theme.spacing(2),
+  },
+  dropContainer: {
+    background: '#F9FAFA',
+    border: '1px dashed #A9B7B8',
+    borderRadius: '8px',
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center',
+    padding: '32px',
+  },
+  hiddenInput: {
+    display: 'none',
+  },
+  title: {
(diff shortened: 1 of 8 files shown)


src/components/seeds/database/index.tsx
```diff
@@ -407,33 +410,111 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
+          <SelectSeedBankModal
+            organization={organization}
+            open={selectSeedBankModalOpen}
+            onClose={onSeedBankSelected}
+          />
+          <SelectSeedBankModal
+            organization={organization}
+            open={selectSeedBankForImportModalOpen}
+            onClose={onSeedBankForImportSelected}
+          />
```
karthikbtf, line 516: lets try to use one modal? we should be able to use the correct callbacks based on which flow triggered the modal?

src/components/seeds/database/index.tsx
```diff
@@ -407,33 +410,111 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
+        id={id}
+        open={open}
+        anchorEl={anchorEl}
+        onClose={handleClose}
+        anchorOrigin={{
+          vertical: 'bottom',
+          horizontal: 'left',
+        }}
+      >
+        <Box>
```
karthikbtf, line 467: looks a bit strange? 
Terraware App 2022-09-26 08-02-29

src/components/seeds/database/ImportAccessionsModal.tsx
```diff
@@ -0,0 +1,354 @@
+import { Link } from '@mui/material';
```
karthikbtf, line 1: yes if you want to build out a reusable component that would be cool, if you want to do that as a fast follow-up so we can start testing, that's fine too

karthikbtf (commented):
not sure why you aren't seeing accessions,,, where did you get the accessions csv from? can you share with me?

constanzauanini:
> not sure why you aren't seeing accessions,,, where did you get the accessions csv from? can you share with me?

I downloaded the template and added data there

src/components/seeds/database/ImportAccessionsModal.tsx
constanzauanini: Yes, lets merge this first and I will create a component for the two cases in next pr

src/components/seeds/database/index.tsx
constanzauanini: Yes, I just moved the two buttons inside the menu, but how should it look? Like a simple menu?

karthikbtf: SBD_V2 – Figma 2022-09-26 08-25-13

constanzauanini: Thanks!

karthikbtf (approved):

PR #540, opened 2022-09-27
SW-1366 Replace founderId with plantId in web-app

karthikbtf:
- no changes in search as we currently don't search by founder/plant id
- tested in create/update of v2 accessions


sgrimm (approved):
Straightforward enough!

joshbuchea (approved):

PR #541, opened 2022-09-27
Create import modal component and use it for import species and import accessions

constanzauanini:


src/api/types/file.ts
```diff
@@ -0,0 +1,14 @@
+import { paths } from './generated-schema';
```
karthikbtf, line 1: call this `upload` or `uploadFile`?

karthikbtf (approved):
Thank you for this refactor!

PR #542, opened 2022-09-27
SW-1748 Use web-components theme with terraware token names in terraware-web

karthikbtf:
- this is a POC and won't break existing palette usage
- App.tsx shows how we would use a terraware palette color
- terraware-web will be updated with all token names once we have the revised design tokens
- this PR will be updated with ver 28 of web-components in yarn.lock once https://github.com/terraware/web-components/pull/65 is approved
- tested by copying over the distribution into node_modules for now


constanzauanini (approved):
Excellent!

PR #548, opened 2022-09-28
SW-1801 Use same default date for withdraw modal as other modals like create accession

karthikbtf:
See bug video here https://www.loom.com/share/538d106e11b84e9e9345a03a967ac19e

- update util to use same code that other modals use

### DIFF
--- a/src/utils/date.tsx
+++ b/src/utils/date.tsx
@@ -1,7 +1,6 @@
 import { getDateDisplayValue } from '@terraware/web-components/utils';
-import moment from 'moment';
 export default getDateDisplayValue;
 
 export const getTodaysDateFormatted = () => {
-  return moment().format('YYYY-MM-DD');
+  return getDateDisplayValue(Date.now());
 };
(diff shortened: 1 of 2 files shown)


src/utils/date.tsx
```diff
@@ -1,7 +1,6 @@
 import { getDateDisplayValue } from '@terraware/web-components/utils';
-import moment from 'moment';
 export default getDateDisplayValue;
 
 export const getTodaysDateFormatted = () => {
-  return moment().format('YYYY-MM-DD');
```
karthikbtf, line 6: not sure what moment was returning maybe not utc

karthikbtf: other modals use default today's date as `getDateDisplayValue(Date.now())`
open to other suggestions but prefer one code fragment for consistency

karthikbtf: @constanzauanini fyi

constanzauanini: We should delete this helper and just use getDateDisplayValue(Date.now()) everywhere. But this is fine for now if we don't want to replace it in this PR

constanzauanini (approved):

src/utils/date.tsx
karthikbtf: I was thinking we could replace all usages of `getDateDisplayValue(Date.now())` with `getTodaysDateFormatted()` -  to be more high level. Just want to make sure you are ok with this plan of not using the `moment` based code.

constanzauanini: Yes, Date is js native, so I think less conversions is better

PR #549, opened 2022-09-28
Fix menu styles

constanzauanini:
Screen Shot 2022-09-28 at 15 57 58

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -7,7 +7,9 @@ import {
   Popover,
   Typography,
   Button as ButtonMUI,
-  IconButton,
+  MenuList,
+  MenuItem,
+  useTheme,
 } from '@mui/material';
 import { Theme } from '@mui/material';
 import { makeStyles } from '@mui/styles';
@@ -149,6 +151,7 @@ type DatabaseProps = {
 export default function Database(props: DatabaseProps): JSX.Element {
   const { isMobile } = useDeviceInfo();
   const classes = useStyles({ isMobile });
+  const theme = useTheme();
   const history = useHistory();
   const query = useQuery();
   const location = useStateLocation();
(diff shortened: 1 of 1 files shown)


src/components/seeds/database/index.tsx
```diff
@@ -478,29 +481,16 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
-              sx={{ ':hover': { background: 'none' } }}
-              id='edit-columns'
-            >
-              <Icon name='iconColumns' />
-
-              <Typography sx={{ color: '#136BD3', paddingLeft: 1 }}>{strings.CUSTOMIZE_TABLE_COLUMNS}</Typography>
-            </IconButton>
-          </Box>
-        </Box>
+        <MenuList sx={{ padding: theme.spacing(2, 0) }}>
```
karthikbtf, line 484: remove spacing here?

src/components/seeds/database/index.tsx
```diff
@@ -478,29 +481,16 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
+          <MenuItem onClick={() => onDownloadReport()} id='download-report' sx={{ padding: theme.spacing(1, 2) }}>
```
karthikbtf, line 485: add more vertical spacing here and below?

src/components/seeds/database/index.tsx
constanzauanini: Removing parent padding and adding to children looks like this:

Screen Shot 2022-09-28 at 16 17 25

Is that correct?

karthikbtf: hmm..my mistake. looks like your original version is consistent with our menus

Screen Shot 2022-09-28 at 12 22 53 PM

karthikbtf (approved):
original fix looks good 🎉

PR #550, opened 2022-09-28
Make withdrawal and check-in buttons size medium

constanzauanini:


karthikbtf:
Looks good, tested in mobile 
Terraware App 2022-09-28 12-14-40

maybe we can align those elements? or another PR.. ?

karthikbtf (approved):
Thanks

constanzauanini:
> Looks good, tested in mobile Terraware App 2022-09-28 12-14-40
> 
> maybe we can align those elements? or another PR.. ?

Fixing it here

PR #551, opened 2022-09-28
SW-1799 Country/subdivision dropdown fixes

karthikbtf:
- switch country and subdivision ordering
- fix edit view to populate country (clearing of fields is broken, filed another ticket for that)
- fix edit view to populate GPS entries
- fix detail panel to show user friendly country/subdivision names

### DIFF
--- a/src/components/accession2/view/DetailPanel.tsx
+++ b/src/components/accession2/view/DetailPanel.tsx
@@ -1,10 +1,13 @@
 import { Box, Grid, IconButton, Typography, useTheme } from '@mui/material';
 import { makeStyles } from '@mui/styles';
 import { Icon } from '@terraware/web-components';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { Accession2 } from 'src/api/accessions2/accession';
 import strings from 'src/strings';
 import { ServerOrganization } from 'src/types/Organization';
+import { searchCountries } from 'src/api/country/country';
+import { getCountryByCode, getSubdivisionByCode } from 'src/utils/country';
+import { Country } from 'src/types/Country';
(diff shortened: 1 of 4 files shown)


src/components/accession2/view/DetailPanel.tsx
```diff
@@ -61,6 +75,19 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
...
   };
 
+  const getCollectionSiteAddress = () => {
+    const city = accession?.collectionSiteCity;
+    const country = accession?.collectionSiteCountryCode;
+    const subdivision = accession?.collectionSiteCountrySubdivision;
+    const data = [
+      city,
+      countries && country && subdivision ? getSubdivisionByCode(countries, country, subdivision)?.name : '',
+      countries && country ? getCountryByCode(countries, country)?.name : '',
```
constanzauanini, line 85: I don't think we need to show the subdivision and country names here. In figma, the example shows CA, USA. Maybe confirm with Ming?

karthikbtf: I don’t think she was being specific. Doesn’t make sense to show something else that user didn’t see when selecting. The codes are not understood by all. I can remove this for now and create a new bug for Allison and Ming to follow up on.

constanzauanini: Yes, reasonable!

constanzauanini (approved):

src/components/accession2/view/DetailPanel.tsx
karthikbtf: @constanzauanini I will get this in for now - if feedback on https://terraformation.atlassian.net/browse/SW-1822 is different I will follow up accordingly.

PR #552, opened 2022-09-29
SW-1797 Allow changing status from Used Up to any status

karthikbtf:
- require quantity edits at this point
- refactored code to reuse some components and modals

Terraware App 2022-09-28 16-41-49


src/components/accession2/edit/EditState.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
```
karthikbtf, line 1: this was extracted out from EditStateModal... and it now also instantiated in quantity modal when we edit Used Up status.

constanzauanini:
How can I end with a Used up status? I couldn't test this

karthikbtf:
> How can I end with a Used up status? I couldn't test this

Withdraw All

constanzauanini:
> > How can I end with a Used up status? I couldn't test this
> 
> Withdraw All

Thanks! Looking good. Not related to this, I don't see the pencil to edit quantity appearing on hover

constanzauanini (approved):

constanzauanini:
> > > How can I end with a Used up status? I couldn't test this
> > 
> > 
> > Withdraw All
> 
> Thanks! Looking good. Not related to this, I don't see the pencil to edit quantity appearing on hover

Oh, ignore my previous comment. I forgot it has to be drying or in storage!

PR #560, opened 2022-09-29
SW-1818 Allow user to apply viability percent from cut test on the accession.

karthikbtf:
- moved viability percent calculation to a util
- removed cut tests seedsTested validation (since this is implied and not user supplied)

### DIFF
--- a/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
+++ b/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
@@ -188,7 +188,7 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
 
   const hasErrors = () => {
     if (record) {
-      const seedTestedError = !validateSeedsTested(record.seedsTested);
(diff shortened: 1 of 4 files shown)


src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
```diff
@@ -221,7 +221,7 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
       if (response.requestSucceeded) {
         reload();
         onCloseHandler();
-        if (testCompleted && !readOnly) {
+        if ((testCompleted || record.testType === 'Cut') && !readOnly) {
```
constanzauanini, line 224: In cut test readOnly will always be false (we can't mark it as complete), so if we are editing the cut test and not changing any value we will see the viability message again. Not sure if that's correct

karthikbtf: good catch, i will update this to only check when the value is edited

karthikbtf: updated and tested

constanzauanini (approved):
🎉

PR #565, opened 2022-09-30
Accession 2 misc tiny UI fixes

karthikbtf:
- SW-1830 - end-drying reminder to be shown only when accession is in status 'Drying'
- SW-1829 - expand clickable area for accession info (status, location, etc.)
- SW-1826 - Get Started banner to be shown in empty state only
- 
Terraware App 2022-09-30 10-07-56

Terraware App 2022-09-30 10-07-30


constanzauanini (approved):
🎉

PR #566, opened 2022-09-30
SW-1839 Reorder list of withdrawal purposes

karthikbtf:
- Nursery first, Other last
- Waiting for PM decision on this list


karthikbtf:
@constanzauanini I can get this in and iterate if there's more feedback from PMs

constanzauanini (approved):
Ok, we can iterate this when we get feedback!

PR #568, opened 2022-09-30
SW-1827 Fix see all accessions results + related fixes

karthikbtf:
- Clicking on accessions by status in the dashboard works only once
- Subsequent clicks don't show the correct response because the previous filter isn't cleared
- Added a fix to clear filters if we have a new filter request
- Also added fix for SW-1833 (don't show notes icon if there are no notes)

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -52,7 +52,7 @@ import { useRecoilState } from 'recoil';
 import EmptyMessage from 'src/components/common/EmptyMessage';
 import { APP_PATHS } from 'src/constants';
 import TfMain from 'src/components/common/TfMain';
-import { ACCESSION_STATES } from '../../../types/Accession';
+import { ACCESSION_STATES, ACCESSION_2_STATES } from '../../../types/Accession';
 import SelectSeedBankModal from '../../SeedBank/SelectSeedBankModal';
 import { isAdmin } from 'src/utils/organization';
 import { LocalizationProvider } from '@mui/x-date-pickers';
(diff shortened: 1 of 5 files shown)


src/components/seeds/summary/index.tsx
```diff
@@ -189,7 +189,7 @@ export default function SeedSummary(props: SeedSummaryProps): JSX.Element {
                     <Typography fontSize='20px' color='#000000' paddingBottom='8px' paddingLeft='28px'>
                       {strings.ACCESSION_BY_STATUS}
                     </Typography>
-                    <Link className={classes.accessionsLink} to={APP_PATHS.ACCESSIONS}>
+                    <Link className={classes.accessionsLink} to={`${APP_PATHS.ACCESSIONS}?stage=}`}>
```
constanzauanini, line 192: Why is this necessary?

src/components/seeds/database/index.tsx
```diff
@@ -201,8 +201,9 @@ export default function Database(props: DatabaseProps): JSX.Element {
     const stage = query.get('stage');
     const facilityId = query.get('facilityId');
     let newSearchCriteria = searchCriteria || {};
-    if (stage) {
-      if (ACCESSION_STATES.indexOf(stage) !== -1) {
+    if (stage || query.has('stage')) {
```
constanzauanini, line 204: if stage is equal to query.get('stage'); why is this or necessary?

src/components/seeds/summary/index.tsx
karthikbtf: We need a way to say load all stages (since there might be an existing filter). Using `stage=` for that. Thought about `stage=*` but `stage=` would clear previous filters, and `stage=<something` would set the next filter. Open to suggestions.

src/components/seeds/database/index.tsx
karthikbtf: query.get('stage') will capture `stage=<something>` query.has('stage') will capture if the key stage exists - even if there is no value. query.get('stage') will only capture value.

constanzauanini: Oh, ok, thanks for explaining that

src/components/seeds/summary/index.tsx
constanzauanini: I think this is ok for now. I didn't realize this wasn't working more than 1 time

constanzauanini (approved):

src/components/seeds/summary/index.tsx
karthikbtf: Thanks. Yes I didn’t realize that either. I’ll update the web-components version too and update this PR.

PR #569, opened 2022-10-03
Add common name to accessions table columns

constanzauanini:


karthikbtf (approved):
Do you need to fix any tests here?

constanzauanini:
> Do you need to fix any tests here?

No, I ran them locally and they are ok

karthikbtf:
> > Do you need to fix any tests here?
> 
> No, I ran them locally and they are ok

Awesome thanks 🙏🏻

PR #570, opened 2022-10-03
SW-1820 Allow clearing address fields

constanzauanini:
Save typed value (before selecting the option) to allow editing State on Accession edit


karthikbtf (approved):
Thanks 🎉 
Just a reminder to add a few words in PR description about fix - so newer team members can review with some overview.

PR #573, opened 2022-10-03
SW-1904 Fix end-drying reminder date display on accessions view

karthikbtf:
- show relative time left/passed for end-drying reminder
- update padding on mobile view as per spec, current horizontal padding was too big
- update alignment to be `flex-start` on mobile, `center` on desktop - to match the spec

Terraware App 2022-10-03 14-35-26

Terraware App 2022-10-03 16-01-08

### DIFF
--- a/src/components/accession2/view/Accession2View.tsx
+++ b/src/components/accession2/view/Accession2View.tsx
@@ -301,7 +301,7 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element 
 
   const editableParentProps = {
     display: 'flex',
-    padding: themeObj.spacing(0, 2, isMobile ? 3 : 0),
+    padding: themeObj.spacing(0, isMobile ? 0 : 2, isMobile ? 3 : 0, isMobile ? 0 : 2),
     alignItems: 'center',
     width: isMobile ? '100%' : 'auto',
   };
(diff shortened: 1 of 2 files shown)


constanzauanini:
This looks good! Can we align the seeds icon of the state with the state name on mobile?

src/components/accession2/view/Accession2View.tsx
```diff
@@ -482,8 +482,19 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element
           <Box sx={editableParentProps}>
             <Icon name='notification' className={classes.iconStyle} />
             <Box sx={editableProps} onClick={() => setOpenEndDryingReminderModal(true)}>
-              <Typography paddingLeft={1}>
-                {accession?.dryingEndDate ? strings.END_DRYING_REMINDER_ON : strings.END_DRYING_REMINDER_OFF}
+              <Typography paddingLeft={1} lineHeight={isMobile ? 1.2 : 1.5}>
+                {accession?.dryingEndDate
+                  ? `${strings.END_DRYING_REMINDER} ${moment(accession.dryingEndDate).fromNow()}`
```
charlie-tf, line 487: this is cool! out of curiosity, does `moment` inherit its locale from global state somehow?

karthikbtf:
> This looks good! Can we align the seeds icon of the state with the state name on mobile?

Kinda looks like this in the spec (see mobile left hand side).. i do agree with you though. Will try something.

image-20221003-185610

karthikbtf:
> > This looks good! Can we align the seeds icon of the state with the state name on mobile?
> 
> Kinda looks like this in the spec (see mobile left hand side).. i do agree with you though. Will try something.
> 
> image-20221003-185610

Terraware App 2022-10-04 08-23-20

fixed

PR #585, opened 2022-10-05
SW-1585 Clean up Edit buttons and some newer Delete buttons

alexTerraformation:
Add pencil icon to all edit buttons and make edit buttons only pencil icon on mobile (except for empty states).  Also add trash can icon to some newer delete buttons from seeds v2.
Screen Shot 2022-10-04 at 8 50 21 PM
Screen Shot 2022-10-04 at 8 51 00 PM
Screen Shot 2022-10-04 at 8 51 11 PM
Screen Shot 2022-10-04 at 8 51 20 PM
Screen Shot 2022-10-04 at 8 51 41 PM
Screen Shot 2022-10-04 at 8 51 47 PM
Screen Shot 2022-10-04 at 8 52 27 PM
Screen Shot 2022-10-04 at 8 52 36 PM
Screen Shot 2022-10-04 at 8 53 57 PM
Screen Shot 2022-10-04 at 8 55 00 PM
Screen Shot 2022-10-04 at 8 55 10 PM
Screen Shot 2022-10-04 at 8 55 22 PM
Screen Shot 2022-10-04 at 8 55 52 PM

### DIFF
--- a/src/components/Organization/index.tsx
+++ b/src/components/Organization/index.tsx
@@ -109,11 +109,11 @@ export default function OrganizationView({ organization }: OrganizationViewProps
             <h2>{organization?.name}</h2>
           </Grid>
           <Grid item xs={4} className={classes.titleButtonContainer}>
-            <Button
-              label={isMobile ? strings.EDIT : strings.EDIT_ORGANIZATION}
-              priority='secondary'
-              onClick={goToEditOrganization}
-            />
+            {isMobile ? (
+              <Button icon='iconEdit' priority='secondary' onClick={goToEditOrganization} />
+            ) : (
+              <Button icon='iconEdit' label={strings.EDIT} priority='secondary' onClick={goToEditOrganization} />
+            )}
           </Grid>
         </Grid>
         <Grid item xs={gridSize()}>
--- a/src/components/MyAccount/DeleteOrgModal.tsx
+++ b/src/components/MyAccount/DeleteOrgModal.tsx
@@ -22,7 +22,7 @@ export default function DeleteOrgDialog(props: DeleteOrgDialogProps): JSX.Elemen
       size='medium'
       middleButtons={[
         <Button label={strings.CANCEL} priority='secondary' type='passive' onClick={onClose} key='button-1' />,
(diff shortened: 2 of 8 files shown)


rachel-tf-design:
Looks great! Thanks @alexTerraformation :)

constanzauanini (approved):

src/components/Organization/index.tsx
```diff
@@ -109,11 +109,11 @@ export default function OrganizationView({ organization }: OrganizationViewProps
...
-            <Button
-              label={isMobile ? strings.EDIT : strings.EDIT_ORGANIZATION}
-              priority='secondary'
-              onClick={goToEditOrganization}
-            />
+            {isMobile ? (
+              <Button icon='iconEdit' priority='secondary' onClick={goToEditOrganization} />
+            ) : (
+              <Button icon='iconEdit' label={strings.EDIT} priority='secondary' onClick={goToEditOrganization} />
+            )}
```
karthikbtf, line 116: Do we have a preference on how to handle cases like this - where the only difference is the label?
```
{ isMobile ? <mobile markup> : <desktop markup>}
```
vs
```
<markup label={isMobile ? '' : string} >
```

We have been mostly using the latter in the codebase.. however, the former does express intent clearly.

Just a question for all. Not requesting any changes at this point.

constanzauanini: What I try to do is that if only one attribute is different between the two elements option 2 is better, but if we have more than one attributed changed (for example label, priority, icon) it's better to create two elements

charlie-tf: I like the latter myself, it's more concise and also gets the intent across I think. Certainly if you have a lot of props conditional on `isMobile` the former would be the way to go. If you had other props that were conditional on other variables you might get into some kinda nasty nesting, but at that point you might want to rethink inlining the logic and maybe break out into another component 🤷

karthikbtf: I am leaning towards the former now that I see it in code. We have an impasse 😱 - we could evolve our codestyle over time maybe, now that the team is growing?

PR #588, opened 2022-10-05
SW-1931 Show notes icon inside the notes div

constanzauanini:
Put the notes icon inside the div so text can go under the icon.

Screen Shot 2022-10-05 at 10 46 56

### DIFF
--- a/src/components/accession2/view/DetailPanel.tsx
+++ b/src/components/accession2/view/DetailPanel.tsx
@@ -158,8 +158,10 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
 
               {accession.collectionSiteNotes && (
                 <Box marginTop={2} display='flex'>
-                  <Icon name='iconFile' className={classes.folderIcon} />
-                  <Typography>{accession.collectionSiteNotes}</Typography>
+                  <Typography>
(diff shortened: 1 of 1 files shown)


src/components/accession2/view/DetailPanel.tsx
```diff
@@ -158,8 +158,10 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
 
               {accession.collectionSiteNotes && (
                 <Box marginTop={2} display='flex'>
-                  <Icon name='iconFile' className={classes.folderIcon} />
-                  <Typography>{accession.collectionSiteNotes}</Typography>
+                  <Typography>
```
karthikbtf, line 161: center align?

karthikbtf (approved):

src/components/accession2/view/DetailPanel.tsx
constanzauanini: The typography is not display flex, so I can't set up align center. I could set up a line height, but I like more how it looks without line height

karthikbtf: Ok sounds good.

PR #589, opened 2022-10-05
Increase font size for +Add links

constanzauanini:
Screen Shot 2022-10-05 at 11 54 48


karthikbtf (commented):
For mobile as well? Can we check?

constanzauanini:
> For mobile as well? Can we check?

I would guess the change is for both, but will check with Ming

karthikbtf (approved):
I'm not a huge fan of using `sx={{}}` we seem to have a lot of duplicated style settings. 
With makeStyles we reused classes. 
Maybe we can do something similar here going forward at least.. capture common styles in an object and use those `sx={styleInfo}`.
Looks good for now.

PR #593, opened 2022-10-05
SW-1953 Fix race condition in accessions database search when coming from dashboard quick-links

karthikbtf:
- sometimes the slower search (of all accessions) finishes last and it's results get used in the table
- use a state that is reset on unmount, which indicates if a search result is to be used or not

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -252,6 +252,7 @@ export default function Database(props: DatabaseProps): JSX.Element {
   }, [query, location, history, setSearchCriteria, organization, searchCriteria]);
 
   useEffect(() => {
+    let activeRequests = true;
     const getFieldsFromSearchColumns = () => {
       let columnsNamesToSearch = Array<string>();
       if (searchColumns.includes('active')) {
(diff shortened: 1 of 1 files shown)


src/components/seeds/database/index.tsx
```diff
@@ -325,6 +344,9 @@ export default function Database(props: DatabaseProps): JSX.Element {
       populateFieldOptions();
       populateSearchSummary();
     }
+    return () => {
+      activeRequests = false;
```
karthikbtf, line 348: this happens on unmount, when the state filters change and new requests are issued.

charlie-tf:
So if I understand correctly, the idea here is to make sure that if any of the `await`ed methods in the `useEffect` returns after unmounting, they do not call the `set*` methods and modify state. Do those state variables affect the global state in some way? Or in some other carry over after the component unmounts?

charlie-tf:
Oh, I understand your comment now, and I now remember that the `useEffect` cleanup also runs before the `useEffect` runs again. So you can ignore most of my last question 😅

charlie-tf (approved):
This makes sense to me!

karthikbtf:
> 

@charlie-tf you are correct - this is to avoid setting those values - which will be set in the next useEffect run which will set the values again.
The problem we had was the earlier requests taking more time and overwriting newer requests - leading to incorrect behavior.
Good question.

PR #594, opened 2022-10-05
SW-1954 Replace sourcePlantOrigin with collectionSource

karthikbtf:
- use new list values
- not using (in-situ) / (ex-situ) for v1 unless we really want that


sgrimm (approved):
Seems correct to me.

PR #597, opened 2022-10-06
SW-1844 Show error when withdrawn quantity exceeds remaining quantity

constanzauanini:
Screen Shot 2022-10-06 at 14 33 12

### DIFF
--- a/src/components/accession2/withdraw/WithdrawModal.tsx
+++ b/src/components/accession2/withdraw/WithdrawModal.tsx
@@ -198,6 +198,15 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element 
           setIndividualError('withdrawnQuantity', strings.INVALID_VALUE);
           return false;
         }
+        if (
+          accession.remainingQuantity?.units &&
+          Number(record.withdrawnQuantity.units === accession.remainingQuantity.units)
+        ) {
+          if (record.withdrawnQuantity.quantity > accession.remainingQuantity?.quantity) {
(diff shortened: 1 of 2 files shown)


src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -198,6 +198,15 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
           setIndividualError('withdrawnQuantity', strings.INVALID_VALUE);
           return false;
         }
+        if (
+          accession.remainingQuantity?.units &&
+          Number(record.withdrawnQuantity.units === accession.remainingQuantity.units)
```
charlie-tf, line 203: so if I understand correctly, this makes sure the units of the remaining and withdrawn quantities are the same, e.g. `ct`. How do we validate when the units are different? It looks like we'll always drop down to the `return true`?

constanzauanini: Yes. You are correct Charlie. I wrote in the Jira ticket https://terraformation.atlassian.net/browse/SW-1844 that we will leave that validation for the backend, because if not we will need to do all conversions between units in the frontend. Backend is already doing the validation

charlie-tf: ooh, gotcha. so in that case this will validate and allow the withdrawal to be executed on the frontend, and if there is a backend validation error it will show a server error. cool, thanks!

charlie-tf (approved):

PR #601, opened 2022-10-06
SW-1837 / SW-1965 Misc UI fixes

karthikbtf:
- disallow editing status of accession that is awaiting check in (the edit icon was hidden but status badge was clickable)
- default viability start and recording dates to today


src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
```diff
@@ -70,6 +70,7 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
       withdrawnByUserId: user.id,
       testType: 'Lab',
       seedsTested: 0,
+      startDate: getTodaysDateFormatted(),
```
constanzauanini, line 73: Oh I just added the same change, I will close my pr

constanzauanini (approved):

PR #603, opened 2022-10-07
SW-1963 Disable hover on species table when there is no edit allowed

constanzauanini:
The hover style is added when the table receives a "onSelect" function. Don't pass on select function to species table when user is a contributor

### DIFF
--- a/src/components/Species/index.tsx
+++ b/src/components/Species/index.tsx
@@ -512,7 +512,7 @@ export default function SpeciesList({ organization, reloadData, species }: Speci
                   setSelectedRows={isUserContributor ? undefined : setSelectedSpeciesRows}
                   showTopBar={true}
                   Renderer={SpeciesCellRenderer}
-                  onSelect={selectAndEditSpecies}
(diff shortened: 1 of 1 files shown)


src/components/Species/index.tsx
```diff
@@ -512,7 +512,7 @@ export default function SpeciesList({ organization, reloadData, species }: Speci
                   setSelectedRows={isUserContributor ? undefined : setSelectedSpeciesRows}
                   showTopBar={true}
                   Renderer={SpeciesCellRenderer}
-                  onSelect={selectAndEditSpecies}
+                  onSelect={!isUserContributor ? selectAndEditSpecies : undefined}
```
karthikbtf, line 515: i made a mistake by calling this `isUserContributor` - we should really have a flag that says `userCanEdit`.. we can change it later i guess

karthikbtf (approved):
that was simple

src/components/Species/index.tsx
constanzauanini: I will change it

PR #606, opened 2022-10-10
SW-1750 Remove Accession Editing from Contributor Role

charlie-tf:
- add a userCanEdit const that checks for contributor role
  - is false if the role is contributor
- make all of the editing points conditional on userCanEdit
- check both desktop and mobile views
- TODO (followup): changes for viability test add/update/delete
- note: license-report and generated-schema were updated from 
`yarn license-report` and `yarn generate-types`


charlie-tf:
broke an integration test, apparently? converting to draft to investigate

charlie-tf:
nope, it was just flaky

constanzauanini:
> * userCanEdit

e2e tests are flaky. We have a task to review that and also when removing v1 of accessions we will have to delete most of them

constanzauanini (approved):
Thanks for doing this change @charlie-tf !

PR #607, opened 2022-10-11
SW-1973 Get correct substrate options on withdrawal modal

constanzauanini:
Substrates differ on lab test and nursery test. Show correct options for each type on withdrawal modal


charlie-tf (approved):
Nice refactor as well 👍

PR #609, opened 2022-10-11
Bump web-components version

constanzauanini:
Use new web-components version. Adding right/left margins on mobile dialog box


karthikbtf (approved):
🎉 
Update description of PR to mention picking up margin on mobile dialog-box changes? That way it shows up in the merge commit.

PR #613, opened 2022-10-11
SW-1980 Fix accessions onboarding page

karthikbtf:
- implement both species and seedbanks onboarding

Terraware App 2022-10-11 16-33-09

Terraware App 2022-10-11 16-32-43

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -147,6 +147,7 @@ type DatabaseProps = {
   displayColumnNames: string[];
   setDisplayColumnNames: (fields: string[]) => void;
   hasSeedBanks: boolean;
+  hasSpecies: boolean;
 };
 
 export default function Database(props: DatabaseProps): JSX.Element {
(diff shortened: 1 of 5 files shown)


src/components/seeds/database/index.tsx
```diff
@@ -671,10 +699,8 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
                   className={classes.message}
-                  title={emptyMessageStrings.NO_SEEDBANKS_ADMIN_TITLE}
-                  text={emptyMessageStrings.NO_SEEDBANKS_ADMIN_MSG}
-                  buttonText={strings.GO_TO_SEED_BANKS}
-                  onClick={goToSeedBanks}
+                  title={emptyMessageStrings.ACCESSIONS_ONBOARDING_ADMIN_TITLE}
+                  rowItems={getEmptyState()}
                 />
               ) : (
                 <EmptyMessage
```
constanzauanini, line 706: Should we also change this contributor message for no species case?

karthikbtf: Yes this will only be a text change no call to action buttons. I was going to do this as a follow up.

constanzauanini (approved):

PR #614, opened 2022-10-12
SW-1819 Remove some console errors from web-app

karthikbtf:
- this could be a never-ending effort, going to do this in smaller chunks

### DIFF
--- a/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
+++ b/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
@@ -373,7 +373,7 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
         scrolled={true}
       >
         <Grid container item xs={12} spacing={2} textAlign='left'>
-          <Grid xs={12} padding={theme.spacing(1, 3, 1, 5)}>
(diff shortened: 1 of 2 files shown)


src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
```diff
@@ -373,7 +373,7 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
         scrolled={true}
       >
         <Grid container item xs={12} spacing={2} textAlign='left'>
-          <Grid xs={12} padding={theme.spacing(1, 3, 1, 5)}>
+          <Grid item xs={12} padding={theme.spacing(1, 3, 1, 5)}>
```
constanzauanini, line 376: Adding this "item" is making all rows to be at the same level of the grey box of observations. We should add more padding to this rows

karthikbtf: Yes good point, I will revisit this.

karthikbtf:
Will revisit this later.

src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
karthikbtf: cc @joshbuchea

PR #617, opened 2022-10-12
SW-1810 Mobile view of add viability test

constanzauanini:
Screen Shot 2022-10-12 at 11 22 27

### DIFF
--- a/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
+++ b/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
@@ -15,7 +15,7 @@ import { postViabilityTest } from 'src/api/accessions2/viabilityTest';
 import useSnackbar from 'src/utils/useSnackbar';
 import { renderUser } from 'src/utils/renderUser';
 import { Close } from '@mui/icons-material';
-import { preventDefaultEvent } from '@terraware/web-components/utils';
+import { preventDefaultEvent, useDeviceInfo } from '@terraware/web-components/utils';
 import { getTodaysDateFormatted, isInTheFuture } from '@terraware/web-components/utils';
 import { ViabilityTest } from 'src/api/types/accessions';
 import ViabilityResultModal from './ViabilityResultModal';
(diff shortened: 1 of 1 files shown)


src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
```diff
@@ -513,8 +518,14 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
...
 
               {record?.testResults?.map((testResult, index) => (
-                <Box key={index} mb={2} display='flex' alignItems='baseline'>
-                  <Grid item xs={12}>
+                <Box
+                  key={index}
+                  mb={2}
+                  display={isMobile ? 'block' : 'flex'}
+                  alignItems='baseline'
+                  maxWidth={isMobile ? '90%' : '100%'}
```
karthikbtf, line 526: why 90%? this may not work for all mobile form factors?

karthikbtf (approved):
This is getting a little messy.. maybe we should think about separating out mobile vs desktop here? Going to be hard to read/maintain otherwise. Maybe as a follow up, I don't know.

src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
constanzauanini: This is because the box doesn't fill the whole space in mobile, and to avoid adding a new grid, but I could try with a new grid if you think that is better

karthikbtf: Or use a margin right that is the width of the close icon?

karthikbtf: This is just my observation - feels like we are forcing the desktop layout to also work for mobile - but they are quite different that maybe having two separate markups might be the better way to do this.
I could be wrong of course, so this is fine for now in my relative opinion :)

karthikbtf: I have accepted the review, feel free to merge - this does work well according to my testing.

constanzauanini: I will change the margin. I don't feel we need to split it in two markups. It's just displaying the content flex vs block and some spacings, but the elements are all the same. We will have a lot of repeated code if we separate it

karthikbtf: > I will change the margin. I don't feel we need to split it in two markups. It's just displaying the content flex vs block and some spacings, but the elements are all the same. We will have a lot of repeated code if we separate it

Now that you mention it.. yes. Margin fix looks good, thanks 🙏

PR #618, opened 2022-10-12
SW-1851 Add nursery menu item and route

constanzauanini:
Screen Shot 2022-10-12 at 16 27 29

### DIFF
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -35,6 +35,7 @@ export default function NavBar({ organization, setShowNavBar }: NavBarProps): JS
   const isOrganizationRoute = useRouteMatch(APP_PATHS.ORGANIZATION + '/');
   const isMonitoringRoute = useRouteMatch(APP_PATHS.MONITORING + '/');
   const isSeedbanksRoute = useRouteMatch(APP_PATHS.SEED_BANKS + '/');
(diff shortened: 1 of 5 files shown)


charlie-tf (approved):
Looks good!

src/components/NavBar.tsx
```diff
@@ -102,6 +103,16 @@ export default function NavBar({ organization, setShowNavBar }: NavBarProps): JS
...
+      <NavSection title={strings.SEEDLINGS.toUpperCase()} />
+      <NavItem
+        label={strings.INVENTORY}
+        icon='iconSeedling'
+        selected={!!isNurseryRoute}
+        onClick={() => {
+          closeAndNavigateTo(APP_PATHS.NURSERY);
+        }}
+        id='nursery'
+      />
```
karthikbtf, line 115: We don't want this displayed for everyone now I think... might be good to use the routeEnabled check.
https://github.com/terraware/terraware-web/blob/main/src/features.ts#L82
```
isRouteEnabled('Nursery management') && (...)
```

karthikbtf: oops should be featureEnabled not route enabled (basically if user opted in to see this feature).
routeEnabled check is for the routing in App.tsx

PR #620, opened 2022-10-12
SW-1991 Show bag IDs for accessions awaiting check-in only

karthikbtf:
- hide them if accession is already checked in


sgrimm (approved):
This seems about as safe a change as one could hope for.

PR #621, opened 2022-10-12
SW-1992 (WIP) upgrade node version for github actions

karthikbtf:
- this is a test
- See https://github.com/actions/setup-node
- See https://github.com/actions/checkout

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -45,7 +45,7 @@ jobs:
         id: yarn-cache-dir-path
         run: echo "::set-output name=dir::$(yarn cache dir)"
 
-      - uses: actions/cache@v2
+      - uses: actions/cache@v3
         id: yarn-cache
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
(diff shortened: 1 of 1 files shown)


constanzauanini (approved):
Are we updating just to be updated or any other reason?

karthikbtf:
> Are we updating just to be updated or any other reason?

Node version 12 is deprecated and no longer supported from early 2023. They are giving us time to upgrade to 16.
This PR is still WIP, current changes are not working.. need to investigate some more.

constanzauanini:
> > Are we updating just to be updated or any other reason?
> 
> Node version 12 is deprecated and no longer supported from early 2023. They are giving us time to upgrade to 16. This PR is still WIP, current changes are not working.. need to investigate some more.

Oh, ok! But we are not specifically setting node version to be 12, are we?

karthikbtf:
> > > Are we updating just to be updated or any other reason?
> > 
> > 
> > Node version 12 is deprecated and no longer supported from early 2023. They are giving us time to upgrade to 16. This PR is still WIP, current changes are not working.. need to investigate some more.
> 
> Oh, ok! But we are not specifically setting node version to be 12, are we?

Correct, it's in the existing github actions we use, they set node to 12. version 3 for some of them use 16.
Some of the other github actions don't have new versions, I am still figuring this out.

karthikbtf:
this was done

PR #622, opened 2022-10-13
Add nurseries empty state

constanzauanini:
Replace Inventory path name with inventory, and add new nurseries path

- [x] TODO: Replace navBar nurseries icon and empty state blobby icon

Screen Shot 2022-10-13 at 15 40 10


src/App.tsx
```diff
@@ -484,6 +502,9 @@ export default function App() {
                   {getSeedBanksView()}
                 </Route>
               )}
+              <Route exact path={APP_PATHS.NURSERIES}>
+                {getNurseriesView()}
+              </Route>
```
karthikbtf, line 507: we need the isRouteEnabled guard around this

src/App.tsx
```diff
@@ -317,13 +318,30 @@ export default function App() {
...
+  const selectedOrgHasNurseries = (): boolean => {
+    if (selectedOrganization && selectedOrganization.facilities) {
+      return selectedOrganization.facilities.some((facility) => {
+        return facility.type === 'Nursery';
+      });
+    } else {
+      return false;
+    }
+  };
+
```
karthikbtf, line 330: I understand this is how we did all other empty states - ideally empty state logic should be within the component for Nurseries. App should only do the routing.

karthikbtf (commented):
some comments

constanzauanini:
> Add the route guard and merge. I will create a ticket to move empty state content into individual components.

I still have to change the icons

karthikbtf:
> > Add the route guard and merge. I will create a ticket to move empty state content into individual components.
> 
> I still have to change the icons

and we have to wait for vercel to start working maybe

PR #623, opened 2022-10-13
SW-1848 Add a new Nursery

constanzauanini:
Screen Shot 2022-10-13 at 15 45 48


src/api/types/generated-schema.ts
```diff
@@ -512,6 +519,7 @@ export interface components {
       /** Default message to publish if the automation type isn't yet supported by the server. */
       message?: string;
     };
+    /** Details of ewly-created seedling batch. */
```
karthikbtf, line 522: we should let Steve know of this perhaps 'ewly' typo

src/components/NewNursery/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [descriptionError, setDescriptionError] = useState('');
+  const snackbar = useSnackbar();
+  const theme = useTheme();
+
+  const [record, , onChange] = useForm<Facility>({
+    name: '',
+    id: -1,
+    type: 'Nursery',
+    organizationId: organization.id,
+    connectionState: 'Not Connected',
```
karthikbtf, line 32: remove?

src/components/NewNursery/index.tsx
```diff
@@ -0,0 +1,111 @@
+import { Container, Grid, Typography, useTheme } from '@mui/material';
```
karthikbtf, line 1: reusable component for seedbank and nursery or this ok?
pretty much boilerplate

src/components/NewNursery/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    const response = await createFacility(record);
+    if (response.requestSucceeded) {
+      reloadOrganizationData();
+      snackbar.toastSuccess(strings.NURSERY_ADDED);
+    } else {
+      snackbar.toastError();
+    }
+
+    goToNurseries();
```
karthikbtf, line 68: only if save succeeded

karthikbtf (commented):
looks good a few comments

src/components/NewNursery/index.tsx
constanzauanini: connectionState is required, maybe I should ask Steve to make it optional ?

karthikbtf (approved):
looks good

src/components/NewNursery/index.tsx
karthikbtf: feel free to merge after this change

src/components/NewNursery/index.tsx
constanzauanini: Yes, we could create a common component. Not sure if it will be really useful because it will be used by this two pages only, what do you think?

karthikbtf: Yes, I agree. This is fine.

PR #626, opened 2022-10-14
SW-1849 Nursery table with search

constanzauanini:
Screen Shot 2022-10-14 at 11 52 19
Screen Shot 2022-10-14 at 11 52 38

### DIFF
--- a/src/components/Nurseries/index.tsx
+++ b/src/components/Nurseries/index.tsx
@@ -5,7 +5,7 @@ import { useDeviceInfo } from '@terraware/web-components/utils';
 import { useEffect, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import { search, SearchNodePayload } from 'src/api/seeds/search';
-import { Facility } from 'src/api/types/facilities';
+import { Facility, FacilityType } from 'src/api/types/facilities';
 import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
 import { ServerOrganization } from 'src/types/Organization';
@@ -81,7 +81,7 @@ export default function NurseriesList({ organization }: NurseriesListProps): JSX
           name: result.name as string,
           description: result.description as string,
           organizationId: parseInt(result.organization_id as string, 10),
-          type: result.type as 'Nursery' | 'Seed Bank' | 'Desalination' | 'Reverse Osmosis',
+          type: result.type as FacilityType,
           connectionState: result.connectionState as 'Not Connected' | 'Connected' | 'Configured',
         });
       });
--- a/src/components/Nurseries/index.tsx
+++ b/src/components/Nurseries/index.tsx
@@ -1,5 +1,159 @@
+import { Box, Grid, Typography } from '@mui/material';
+import { Button, Table, TableColumnType } from '@terraware/web-components';
+import TextField from '@terraware/web-components/components/Textfield/Textfield';
+import { useDeviceInfo } from '@terraware/web-components/utils';
+import { useEffect, useState } from 'react';
+import { useHistory } from 'react-router-dom';
+import { search, SearchNodePayload } from 'src/api/seeds/search';
+import { Facility } from 'src/api/types/facilities';
+import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
+import { ServerOrganization } from 'src/types/Organization';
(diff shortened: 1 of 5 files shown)


src/utils/organization.tsx
```diff
@@ -21,3 +21,11 @@ export const isAdmin = (organization: ServerOrganization | undefined) => {
 export const isContributor = (roleHolder: ServerOrganization | OrganizationUser | undefined) => {
   return roleHolder?.role === 'Contributor';
 };
+
+export const getAllNurseries = (organization: ServerOrganization): (Facility | undefined)[] => {
```
karthikbtf, line 25: small request, since getAllSeedBanks and getAllNurseries are similar code.. can we extract out the common code into a function?
```
getFacilitiesByType = (organization, type) => // returns facilities by type
getAllSeedBanks = (organization) => getFacilitiesByType(organziation, 'Seed Bank');
getAllNurseries = (organization) => getFacilitiesByType(organization, 'Nursery');
```

src/components/Nurseries/index.tsx
```diff
@@ -1,5 +1,159 @@
...
+  const clearSearch = () => {
+    setTemporalSearchValue('');
+  };
+
+  const onChangeSearch = (id: string, value: unknown) => {
+    setTemporalSearchValue(value as string);
+  };
+
+  useEffect(() => {
+    const refreshSearch = async () => {
```
karthikbtf, line 63: Something on my mind earlier from other pages like seed banks etc. we could do in memory search if we wanted - no BE request.
BE request is fine.. but we should fetch latest results on line 107 (otherwise we could technically end up showing latest results only when there's a search term - otherwise we show stale info that was fetched on page load).

karthikbtf (commented):
Looks great, I have two comments for follow up.

src/components/Nurseries/index.tsx
karthikbtf: my suggestion is to remove the `if` on line 64 and remove the `else` section from line 106 - so we are always fetching latest results. that would be a simple fix. thoughts?

constanzauanini: But in that case we would be always doing a BE search, not a search in memory, right?

karthikbtf: Yes - that's if we want to use BE for search.
We could always use in-memory or always use BE for search. If we use a mix of that - results may be inconsistent.
BE search should be better so we always show latest results?
My suggestion may lead to infinite useEffect trigger (because nurseries keeps changing) - we can use a lodash isEqual check before setting nursery results (don't set if the results in state are same as what we are trying to set).
My main suggestion is avoiding a mix of BE vs in-memory.

karthikbtf: We can revisit this later if you prefer - since this is how the other pages behave.

constanzauanini: I tested using only the search api (not getting the nurseries from the org anytime) and it works correctly! This way would always show latests results

karthikbtf: > I tested using only the search api (not getting the nurseries from the org anytime) and it works correctly! This way would always show latests results

Thanks, looks good on vercel. Sorry for the trouble.

src/components/Nurseries/index.tsx
```diff
@@ -1,5 +1,139 @@
...
+      const searchResults = await search(params);
+      const nurseriesResults: Facility[] = [];
+      searchResults?.forEach((result) => {
+        nurseriesResults.push({
+          id: result.id as number,
+          name: result.name as string,
+          description: result.description as string,
+          organizationId: parseInt(result.organization_id as string, 10),
+          type: result.type as 'Nursery' | 'Seed Bank' | 'Desalination' | 'Reverse Osmosis',
+          connectionState: result.connectionState as 'Not Connected' | 'Connected' | 'Configured',
```
karthikbtf, line 85: nit: these types should exported from the api/types using generated schema as source of truth,, and imported here. 
this list is already out dated in SeedBanks/index.tsx

karthikbtf (approved):
one nit but we can follow up later if you want

src/components/Nurseries/index.tsx
constanzauanini: Fixed it

karthikbtf (approved):

PR #630, opened 2022-10-14
SW-1997 enable a way to test production view in staging

karthikbtf:
- added a set/get property pair in opt-in-feature config
- set/get can be defined in specific context
- introduced a new feature 'show production view'
- useEnvironment provides set/get semantics for 'show production view'
- uses session storage for property persistence
- not using BE preferences here since we have no way of clearing it if production view is enabled, we don't show opt-in features in production view
- clearing "productionView" from session storage will restore default view
- this is only for testing and not production end-user facing

### DIFF
--- a/src/components/OptInFeatures/index.tsx
+++ b/src/components/OptInFeatures/index.tsx
@@ -27,6 +27,12 @@ export default function OptInFeatures({ refresh }: OptInFeaturesProps): JSX.Elem
             data[key] = prefs[key] || false;
           }
         });
+
+        OPT_IN_FEATURES.forEach((f) => {
+          if (f.get) {
+            data[f.preferenceName] = f.get();
+          }
+        });
       }
       setTimeout(() => {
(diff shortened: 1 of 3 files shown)


charlie-tf (approved):
Pretty cool! So is this used to, for example, show/hide Nursery features since those are WIP?

constanzauanini (approved):

karthikbtf:
> Pretty cool! So is this used to, for example, show/hide Nursery features since those are WIP?

Sort of related. We have an opt-in [feature page ](https://github.com/terraware/terraware-web/tree/main/src/components/OptInFeatures) only visible in staging - allowing users to see WIP work.
There is some extra `isProduction` code (such as to control displaying opt-in features - and some other legacy stuff before we had opt-in.. and whatever else we might need it for in the future). This PR is to mock `isProduction=true` to test that UI doing the right thing.. such as not showing opt-in options.

PR #635, opened 2022-10-17
SW-2008 Allow summary sub-section to wrap with count and type info

karthikbtf:
- was previously wrapping by the single word
- make the wrapping units a pair of count/type

Terraware App 2022-10-17 16-13-04

Terraware App 2022-10-17 16-13-59

Terraware App 2022-10-17 16-13-32

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -642,29 +642,33 @@ export default function Database(props: DatabaseProps): JSX.Element {
                         sx={{ background: '#F2F4F5', display: 'flex', color: '#000000', padding: 2, flexWrap: 'wrap' }}
                       >
                         <Typography>{strings.TOTAL}</Typography>
-                        <Typography sx={{ paddingLeft: '4px', fontWeight: 500 }}>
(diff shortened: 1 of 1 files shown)


src/components/seeds/database/index.tsx
```diff
@@ -642,29 +642,33 @@ export default function Database(props: DatabaseProps): JSX.Element {
                         sx={{ background: '#F2F4F5', display: 'flex', color: '#000000', padding: 2, flexWrap: 'wrap' }}
                       >
                         <Typography>{strings.TOTAL}</Typography>
```
constanzauanini, line 644: Why is total outside the box?

karthikbtf: Total <count, label> <count, label> <count, label>
The count/label pairs are in boxes because they need to be displayed together without any wrapping.

constanzauanini: Oh, ok. I thought "Total <num> accessions" was one thing

constanzauanini (approved):

PR #639, opened 2022-10-18
SW-1865 Add inventory summary section, first version

karthikbtf:
- add inventory details page, this is by species id `/inventory/<species-id>`
- tooltips will be added later

Terraware App 2022-10-18 16-24-28

Terraware App 2022-10-18 16-24-10


src/api/inventory/search.ts
```diff
@@ -0,0 +1,70 @@
+import { search } from 'src/api/search/search';
+
+const getSearchPayload = (speciesId: number | string, organizationId: number | string, count: number) => {
```
karthikbtf, line 3: this explicit search fields is intentional - as it is a static search for summaries.
we can refactor this once all other inventory search is built out - such as for seedlings batches etc.

src/api/search/search.ts
```diff
@@ -0,0 +1,17 @@
+import axios from '..';
```
karthikbtf, line 1: moved this here from seeds/search
we need to move more constructs such as search criteria in here but after removing 'seeds' from the names... that will be a follow up

karthikbtf:
Updated title and screenshots after redoing the search using `GET /api/v1/nursery/species/{id}/summary` - this is much simpler.

PR #642, opened 2022-10-19
SW-1603 Show Species Database Error to User

charlie-tf:
- Add snackbar to problem tooltip
- Get the error message from the database call
- Display in an error toast if exists
image

### DIFF
--- a/src/api/species/species.ts
+++ b/src/api/species/species.ts
@@ -97,7 +97,7 @@ type SimpleSuccessResponsePayload =
  */
 export type UpdateSpeciesResponse = {
   requestSucceeded: boolean;
-  responseMessage?: string;
+  error?: string;
 };
 
 export async function updateSpecies(species: Species, organizationId: number): Promise<UpdateSpeciesResponse> {
@@ -279,7 +279,7 @@ export async function acceptProblemSuggestion(problemId: number) {
   };
   await axios.post(SPECIES_PROBLEM.replace('{problemId}', problemId.toString())).catch((e) => {
     response.requestSucceeded = false;
-    response.responseMessage = e?.response?.data?.error?.message;
+    response.error = e?.response?.data?.error?.message;
   });
   return response;
 }
--- a/src/api/species/species.ts
+++ b/src/api/species/species.ts
@@ -277,10 +277,12 @@ export async function acceptProblemSuggestion(problemId: number) {
   const response: UpdateSpeciesResponse = {
     requestSucceeded: true,
   };
-  await axios.post(SPECIES_PROBLEM.replace('{problemId}', problemId.toString())).catch((e) => {
+  try {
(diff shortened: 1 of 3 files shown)


src/api/species/species.ts
```diff
@@ -97,6 +97,7 @@ type SimpleSuccessResponsePayload =
  */
 export type UpdateSpeciesResponse = {
   requestSucceeded: boolean;
+  responseMessage?: string;
```
karthikbtf, line 100: this is fine... here's an example of how we have been naming the property recently 
https://github.com/terraware/terraware-web/blob/main/src/api/accessions2/accession.ts#L27
only reason we added a util is to remove some boilerplate code

src/components/Species/ProblemTooltip.tsx
```diff
@@ -62,7 +64,10 @@ export default function ProblemTooltip({
   };
 
   const acceptFix = async (problemId: number) => {
-    await acceptProblemSuggestion(problemId);
+    const response = await acceptProblemSuggestion(problemId);
+    if (!(response && response.requestSucceeded)) {
```
karthikbtf, line 68: can just be `if (!response.requestSucceeded) ` since by contract you will get a response

karthikbtf (approved):
some suggestions

src/api/species/species.ts
```diff
@@ -276,11 +277,10 @@ export async function acceptProblemSuggestion(problemId: number) {
   const response: UpdateSpeciesResponse = {
     requestSucceeded: true,
   };
-  try {
-    await axios.post(SPECIES_PROBLEM.replace('{problemId}', problemId.toString()));
-  } catch {
+  await axios.post(SPECIES_PROBLEM.replace('{problemId}', problemId.toString())).catch((e) => {
```
karthikbtf, line 280: Not sure why I missed this.. Try/catch would be the better pattern with await.. curious why the promise like handling.

karthikbtf: one is an asynchronous pattern
```
  return promise.then().catch();
```
the other is a blocking pattern
```
  try {
     return await promise;
  } catch(e) {
     return error
  }
```
there is no hard and fast rule... its mostly for readability.
you can mix/match them for sure but no point in using await..
plenty of references online .. hate pointing to links from random sources but we all like stackoverflow.. so here you go https://stackoverflow.com/questions/69362121/when-should-i-use-try-catch-instead-of-then-catch
once again, its a matter of readability versus theoretical correctness.

karthikbtf: fwiw  - before introduction of `await` a lot of us were in promise chain hell.. hard to read/follow/debug code...

charlie-tf: Thank you. I will make the change ❤️

karthikbtf (approved):
awesome, thanks!

PR #643, opened 2022-10-19
Extract search out from seeds api directory

karthikbtf:
- now that we use it for more than just search
- tested across webpage


constanzauanini (approved):
Looks good! Thanks for doing this

PR #644, opened 2022-10-19
remove v1 accessions api from generated types

karthikbtf:
- now that it is removed from the BE


karthikbtf:
will add this in another PR

PR #646, opened 2022-10-19
SW-1863 Add search filters to inventory table

constanzauanini:
Screen Shot 2022-10-19 at 17 35 12

Screen Shot 2022-10-19 at 17 28 54

### DIFF
--- a/src/components/Inventory/InventoryTable.tsx
+++ b/src/components/Inventory/InventoryTable.tsx
@@ -9,6 +9,9 @@ import { useDeviceInfo } from '@terraware/web-components/utils';
 import { APP_PATHS } from 'src/constants';
 import TextField from '@terraware/web-components/components/Textfield/Textfield';
 import InventoryCellRenderer from './TableCellRenderer';
+import InventoryFilters from './InventoryFiltersPopover';
+import Pill from '../Species/Pill';
+import { getAllNurseries } from 'src/utils/organization';
 
 const columns: TableColumnType[] = [
   { key: 'species_scientificName', name: strings.SPECIES, type: 'string' },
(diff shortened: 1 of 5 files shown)


src/components/Inventory/index.tsx
```diff
@@ -81,61 +85,93 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
-                  },
-                ],
-              },
-              {
-                operation: 'field',
-                field: 'organization_id',
-                values: [organization.id],
-              },
-            ],
+  const getParams = useCallback(() => {
```
karthikbtf, line 88: wondering if all this is better pushed into the table component...
we would need some way to identify the empty state though.

karthikbtf (commented):
@constanzauanini looks good,,, i think the filters need to be multiple selection.. with search looking for results in any of the selected nurseries.

karthikbtf (approved):
approving for now, follow up with @chudo76 whether this needs to be multiple-select filter?

src/components/Inventory/InventoryTable.tsx
```diff
@@ -9,6 +9,9 @@ import { useDeviceInfo } from '@terraware/web-components/utils';
 import { APP_PATHS } from 'src/constants';
 import TextField from '@terraware/web-components/components/Textfield/Textfield';
 import InventoryCellRenderer from './TableCellRenderer';
```
karthikbtf, line 11: unrelated to change but might be good to rename this file to InventoryCellRenderer. The function name inside is InventoryCellRenderer.. and imported here as such as well.

src/components/Inventory/index.tsx
constanzauanini: Yes, I tried to do that but I was repeating code to see if the inventory has columns or not

PR #648, opened 2022-10-20
SW-1863 add multiple-select nursery filters for inventory

karthikbtf:
- centered spinner
- adjusted position of page snackbar
- updated filters to be multiple-select
- filters will next be used in the seedlings table (follow up)
- 
Terraware App 2022-10-20 10-03-29

Terraware App 2022-10-20 10-03-11

Terraware App 2022-10-20 09-57-48


karthikbtf:
Ran this by @chudo76 and he is ok with it.

constanzauanini (approved):
Looking good!

PR #650, opened 2022-10-20
SW-1870 Add nursery filters to seedlings batches table

karthikbtf:
- also moved Pill out from Species to components root
- tested in mobile and desktop

Terraware App 2022-10-20 12-03-11


src/components/Inventory/InventoryTable.tsx
```diff
@@ -10,7 +10,7 @@ import { APP_PATHS } from 'src/constants';
 import TextField from '@terraware/web-components/components/Textfield/Textfield';
 import InventoryCellRenderer from './InventoryCellRenderer';
 import InventoryFilters, { InventoryFiltersType } from './InventoryFiltersPopover';
-import Pill from '../Species/Pill';
+import Pill from 'src/components/Pill';
```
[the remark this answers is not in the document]
constanzauanini: Ok. And then @joshbuchea will move it to web-components?

src/components/Inventory/view/InventorySeedlingsTable.tsx
```diff
@@ -122,6 +145,23 @@ export default function InventorySeedslingsTable(props: InventorySeedslingsTable
     return;
   };
 
+  const getFilteredNurseryName = (facilityId: number) => {
```
constanzauanini, line 148: We have this same function and next one here and in InventoryTable. Maybe we can export it from there? Or create a util?

constanzauanini (approved):
One comment, but looks good!

src/components/Inventory/view/InventorySeedlingsTable.tsx
karthikbtf: what i can do is extract out the input+filter+pills as a separate component and re-use in both tables.
i can do that later as a cleanup. created a jira https://terraformation.atlassian.net/browse/SW-2033

src/components/Inventory/FilterUtils.ts
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { ServerOrganization } from 'src/types/Organization';
+import { getAllNurseries } from 'src/utils/organization';
+import { InventoryFiltersType } from './InventoryFiltersPopover';
+
+export const getNurseryName = (facilityId: number, organization: ServerOrganization) => {
```
constanzauanini, line 6: Thank you!!

PR #651, opened 2022-10-21
Don't disable save button if estimated count change

constanzauanini:
When changing the count value in the quantity modal (by using the calculator) when grams value is already set, save button is disabled. Enable it when count value change


karthikbtf (approved):
thanks, i noticed this and forgot to bring it up

PR #652, opened 2022-10-21
Add a way to reload data in seedlings view

karthikbtf:
- seedlings view does not lift state of summary and table into parent component
- need a way to reload data from parent view across child components
- mocked message passing using a modified timestamp


karthikbtf:
@constanzauanini can we use this as the baseline for future add/edit seedlings/batches work

constanzauanini (approved):

PR #653, opened 2022-10-21
SW-1876 support deletion of seedlings batches

karthikbtf:
- added a confirmation modal (which wasn't in the design - but this is an assumption)
- copied confirmation modal from species modal deletion
- we can iterate on the confirmation as needed
- deletion supports batch delete (client side, not BE side)

Terraware App 2022-10-21 10-35-28

Terraware App 2022-10-21 10-35-50

Terraware App 2022-10-21 10-35-07

Terraware App 2022-10-21 10-36-04

Terraware App 2022-10-21 10-36-39

### DIFF
--- a/src/components/Inventory/view/InventorySeedlingsTable.tsx
+++ b/src/components/Inventory/view/InventorySeedlingsTable.tsx
@@ -220,45 +220,6 @@ export default function InventorySeedslingsTable(props: InventorySeedslingsTable
             />
           ))}
         </Grid>
-        {selectedRows.length > 0 && (
-          <Box
-            sx={{
-              backgroundColor: '#F2F4F5',
-              display: 'flex',
-              flexDirection: 'row',
-              padding: `${theme.spacing(1)} ${theme.spacing(2)}`,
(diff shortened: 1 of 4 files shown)


src/components/Inventory/view/InventorySeedlingsTable.tsx
```diff
@@ -203,6 +220,45 @@ export default function InventorySeedslingsTable(props: InventorySeedslingsTable
             />
           ))}
         </Grid>
+        {selectedRows.length > 0 && (
```
constanzauanini, line 223: This is not necessary. You can use the showTopBar={true} on the table to do this and topBarButtons={ ... } to add the delete-batch button

karthikbtf: ah did not know.. will look at an example, thanks

karthikbtf: fixed

constanzauanini (approved):
:)

PR #654, opened 2022-10-21
SW-1739 Replace Hardcoded Colors

charlie-tf:
- Part 1 of ??

working spreadsheet: [here](https://docs.google.com/spreadsheets/d/1nAaoqbf8DoSoLI389HYXHcZya3NnzG1RFUeYSBr3GQA/edit#gid=324615937)

### DIFF
--- a/src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
+++ b/src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import { Box, Grid, Typography } from '@mui/material';
+import { Box, Grid, Theme, Typography } from '@mui/material';
 import { Button, DialogBox, IconTooltip } from '@terraware/web-components';
 import { Accession2 } from 'src/api/accessions2/accession';
 import strings from 'src/strings';
@@ -15,6 +15,18 @@ import TooltipLearnMoreModal, {
   LearnMoreModalContentTreatment,
   TooltipLearnMoreModalData,
 } from 'src/components/TooltipLearnMoreModal';
+import { makeStyles } from '@mui/styles';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  dialogContainer: {
+    textAlign: 'left',
+    color: theme.palette.TwClrTxt,
+    fontSize: '14px',
+  },
+  titleStyle: {
+    color: theme.palette.TwClrTxtSecondary,
+  },
+}));
 
 export interface ViewViabilityTestModalProps {
   open: boolean;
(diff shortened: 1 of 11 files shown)


src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
```diff
@@ -51,10 +64,6 @@ export default function ViewViabilityTestModal(props: ViewViabilityTestModalProp
     onEdit();
   };
 
-  const titleStyle = {
-    color: '#859799',
```
karthikbtf, line 55: you can keep this as-is.

```
import { ..., useTheme } from '@mui/material';
...
const theme = useTheme();
...
const titleStyle = {
  color: theme.palette.TwClrTxtSecondary,
};
```
i believe `sx` is preferred over `makeStyles/className` where feasible - such as here in the previous version.
something to do with mui v5 etc. Constanza knows more about this preferred usage.

src/components/common/EmptyMessage.tsx
```diff
@@ -73,10 +85,10 @@ export default function EmptyMessage(props: EmptyMessageProps): JSX.Element {
           {rowItems.map((rowItem, index) => (
             <div className={classes.rowItem} key={index}>
               <div className={classes.rowItemInfo}>
-                <Typography fontSize='16px' fontWeight={500} color='#000000' lineHeight='20px'>
+                <Typography className={classes.rowItemInfoTitle}>
                   {rowItem.title}
                 </Typography>
-                <Typography fontSize='14px' fontWeight={500} color='#3A4445' lineHeight='20px'>
```
karthikbtf: same here, leverage `useTheme` and you can access theme directly here

src/theme.ts
```diff
@@ -3,5 +3,6 @@ import { theme } from '@terraware/web-components';
 // This import is needed to get the types for customized Palette, unfortunately this is the easiest fix for now.
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import TerrawareThemeOptions from '@terraware/web-components/style-dictionary-dist/TerrawareTheme';
+import TerrawareThemeOptions2 from '@terraware/web-components/style-dictionary-dist2/TerrawareTheme2';
```
karthikbtf, line 6: a million thanks 🙏 for keeping this in mind

karthikbtf (commented):
@charlie-tf this is super awesome,, exciting to see the plans coming to life!
Thank you so much!
There are a few comments from me about keeping `sx` usage and leveraging `useTheme`.

src/components/accession2/viabilityTesting/ViewViabilityTestModal.tsx
charlie-tf: oh! okay, i felt it was more organized to put everything I could up into `makeStyles`, but if there is a technical reason to prefer `useTheme` and `sx` then I'm happy to change this back and stick with that convention.

src/theme.ts
charlie-tf: I actually wasn't sure about this, as I could not see either `TerrawareThemeOptions` or `TerrawareThemeOptions2` exported from those files, but I figured if it compiled successfully then it must be happening where I could not see.

karthikbtf: Oh it's just importing the default-export from those files, the naming here was intentional (my comments hopefully adds some context).. the main intention of this import is to get the theme options defined in memory - this is purely for typescript to stop complaining. if there's a better way to do this, we should.. maybe something for later.

charlie-tf: Ohh, gotcha 🎉

karthikbtf (approved):
🎉 great!!

PR #658, opened 2022-10-24
SW-1928 Fix attempting to remove the last person in an org generates error on first attempt

constanzauanini:
This error was because the removed row was still "selected" after removing, and modal was still open.
Fix is to clean selectedRows after deleting and close the modal.


charlie-tf (approved):
🎉

PR #659, opened 2022-10-25
SW-1739 Replace Hardcoded Colors (2)

charlie-tf:
Part 2/2


constanzauanini (approved):
Thanks for replacing all this colors!

PR #660, opened 2022-10-25
SW-1882 Support for rendering Withdraw modal

constanzauanini:
No actual withdraw happens here. Just UI

Screen Shot 2022-10-25 at 12 46 31


src/components/Inventory/view/BatchesCellRenderer.tsx
```diff
@@ -20,8 +20,9 @@ export default function BatchesCellRenderer(props: RendererProps<TableRowType>):
   const { column, row, value, index, onRowClick } = props;
 
   const withdraw = () => {
-    // TODO
-    return;
+    if (onRowClick) {
```
karthikbtf, line 23: here and for edit/click... we may need to prevent event propagation (row gets selected otherwise)?

constanzauanini: I tried doing it here but it's not working. Maybe we need to do it on the table component

karthikbtf: we can follow up on this later maybe, someone will surely file a bug :)

constanzauanini: Sounds good!

karthikbtf: I took care of it for now https://github.com/terraware/terraware-web/pull/665

PR #661, opened 2022-10-25
Update database dump with v2 accession migration

sgrimm:
https://github.com/terraware/terraware-server/commit/fe3e5b1276 introduced a
migration step that only existed for long enough to run it in production, and
was removed afterwards because it relied on obsolete code we needed to remove.

Apply this migration to the test database used by the Cypress test suite so
that the suite can be run with more recent server versions.

Also add some documentation to help server developers update the database dump
in the future.


sgrimm:
This dump hadn't been updated since early June and was behind by about 75 database migrations! This PR will probably shave a couple hundred milliseconds off the Cypress test suite's server startup time, not that anyone will notice given how long the suite takes to run.

constanzauanini (approved):

PR #665, opened 2022-10-26
SW-2052 Disable click propagation from withdraw button and batch-number link

karthikbtf:
- allow row selection only if click happens outside of these 2 other calls to action


src/components/Inventory/view/BatchesCellRenderer.tsx
```diff
@@ -19,15 +20,18 @@ export default function BatchesCellRenderer(props: RendererProps<TableRowType>):
   const classes = useStyles();
   const { column, row, value, index, onRowClick } = props;
 
-  const withdraw = () => {
+  const rowClick = (event?: React.SyntheticEvent) => {
```
constanzauanini, line 23: Oh, this was easy! thanks for doing it! I tried with MouseEvent and e.preventDefault and didn't work!

constanzauanini (approved):

PR #670, opened 2022-10-27
SW-1886 Withdraw from seedbank for nursery transfer purposes 

karthikbtf:
- use new API
- tested and verified in inventory seedlings batch
- updated accession id link from batch modal
- TODO: figure out what to do when accession quantity is in weight and not count (this will be a follow up - see https://terraformation.atlassian.net/browse/SW-1886 or https://www.figma.com/file/GuJDioJNGC91EF10YVXuGD?node-id=620%3A82424&fuid=1093672503971536507#287198104 for figma conversation)

Terraware App 2022-10-26 16-41-11

Terraware App 2022-10-27 06-08-05


src/components/Inventory/view/BatchDetailsModal.tsx
```diff
@@ -19,7 +19,7 @@ export interface BatchDetailsModalProps {
   onClose: () => void;
   reload: () => void;
   organization: ServerOrganization;
-  selectedBatch: Batch | undefined;
+  selectedBatch: any;
```
karthikbtf, line 22: selectedBatch is not of type `Batch` - it has search response type data not GET type data

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -41,12 +44,25 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
+    withdrawnByUserId: user.id,
+  };
+
   const newViabilityTesting: ViabilityTestPostRequest = {
     testType: 'Lab',
     seedsTested: 0,
   };
 
   const [record, setRecord, onChange] = useForm(newWithdrawal);
+  const [nurseryTransferRecord, setNurseryTransferRecord, onChangeNurseryTransfer] = useForm(nurseryTransfer);
```
karthikbtf, line 64: we use 2 types of records because we have 2 separate api - we maintain a copy of the common fields in both records because user can switch purpose and not want to see selection lost in other fields

constanzauanini (approved):
I was testing and this looks good. When having an accession in grams we are not creating an inventory batch for now?

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -310,7 +391,9 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
               errorText={fieldsErrors.withdrawnQuantity}
             />
             <Box paddingLeft={1}>
-              {accession.remainingQuantity?.units === 'Seeds' || record.purpose === 'Viability Testing' ? (
+              {accession.remainingQuantity?.units === 'Seeds' ||
+              record.purpose === 'Viability Testing' ||
+              isNurseryTransfer ? (
```
constanzauanini, line 396: If the accession is in grams with purpose nursery I'm still seeing the dropdown for units

karthikbtf: you will see that in old behavior - do you have nursery management enabled?
here's what it should look like 
Screen Shot 2022-10-27 at 7 45 00 AM

karthikbtf: with nursery management disabled 
Screen Shot 2022-10-27 at 7 46 48 AM

constanzauanini:
> > I was testing and this looks good. When having an accession in grams we are not creating an inventory batch for now?
> 
> Accession in grams will work if a subset weight/count is set - BE will error out otherwise. I think this is similar to withdrawal for viability testing? As a follow up, I will add error handling (as per Allison's request) within UI before sending it to BE for withdrawal. I think long term we want to show the modal that allows user to set the subset weight.

Yes, that is how viability testing work!

src/components/accession2/withdraw/WithdrawModal.tsx
constanzauanini: I'm seeing correctly now. I had nursery management enabled but I don't know why it wasn't working. Now it is! Thanks

constanzauanini: I was trying to reproduce it and I couldn't. Maybe I was seeing an old version in Vercel

PR #673, opened 2022-10-27
Go with simplified approach to get last tag

alexTerraformation:
Cleaned this up because previous solution wouldn't have worked for branched releases


karthikbtf (approved):
👍

PR #677, opened 2022-10-27
SW-2062 use 14px font-size in inventory table

karthikbtf:
- also removed some dupe keys
- rest of the web-app seems to use 16px, something we should address with designers' help

Terraware App 2022-10-27 15-36-01


charlie-tf (approved):
thanks for doing this!

PR #679, opened 2022-10-28
SW-2027 allow importing accessions csv during onboarding 

karthikbtf:
- show import option only if seedbanks exist (since import requires selecting a seedbank)
- reused some link text options from other empty state content
- introduced a reload data callback upon successful import

Terraware App 2022-10-28 07-53-54

Terraware App 2022-10-28 07-53-38

Terraware App 2022-10-28 07-53-14

Terraware App 2022-10-28 07-52-55


src/components/common/ImportModal.tsx
```diff
@@ -137,16 +149,9 @@ export default function ImportSpeciesModal(props: ImportSpeciesModalProps): JSX.
...
-
-  const downloadCsvTemplate = async () => {
-    const apiResponse = await templateApi();
-    const csvContent = 'data:text/csv;charset=utf-8,' + apiResponse;
-    const encodedUri = encodeURI(csvContent);
-    const link = document.createElement('a');
-    link.setAttribute('href', encodedUri);
-    link.setAttribute('download', `template.csv`);
-    link.click();
+    if (reloadData) {
```
constanzauanini, line 152: Was this refreshing before this change?

karthikbtf: here we want to refresh the entire state including hasSeedBanks, hasSpecies etc which is passed in from App.tsx
.. wasn't a requirement before this feature

constanzauanini (approved):

PR #680, opened 2022-10-28
SW-1888 Deeplink to Batch Details

charlie-tf:
- extract query parameter
- open batch details modal on batch number if found
- callback to update the query param when select or close


constanzauanini (approved):
😃

PR #683, opened 2022-10-28
SW-1978 Remove script to check for gh-pages deployment

alexTerraformation:
This is now done with a simpler command in the vercel settings


karthikbtf (approved):
👍  thanks for the git log enhancements

PR #686, opened 2022-10-31
SW-2065 Interactive Header

charlie-tf:
- Create wrapper component
- Define class for fixed header
- On scrolling below header fix header to top
- Add padding to immediate sibling below header
- On scroll-down hide the header
- On scroll-up show the header
- Currently only implemented for Accession view

image

Desktop:
Kapture 2022-10-31 at 00 02 24

Mobile:
Kapture 2022-10-31 at 00 03 49

### DIFF
--- a/src/components/common/PageHeaderWrapper.tsx
+++ b/src/components/common/PageHeaderWrapper.tsx
@@ -55,20 +55,17 @@ export default function PageHeaderWrapper({ children, nextElement }: Props): JSX
     };
   }, [children, nextElement, height, sticky]);
 
-  const styles = useMemo<Record<string, any>>(
-    () => ({
-      background: sticky ? theme.palette.TwClrBg : undefined,
-      boxShadow: sticky ? `0px 3px 3px -3px ${theme.palette.TwClrBaseGray200}` : undefined,
-      paddingTop: sticky ? theme.spacing(3) : undefined,
-      paddingRight: sticky ? theme.spacing(3) : undefined,
-      position: sticky ? 'fixed' : undefined,
-      top: sticky ? `${TOP_BAR_HEIGHT}px` : undefined,
-      visibility: scrollDown && sticky ? 'hidden' : 'visible',
-      width: sticky ? '-webkit-fill-available' : undefined,
-      zIndex: sticky ? 100 : undefined,
-    }),
-    [theme, sticky, scrollDown]
-  );
+  const styles: Record<string, any> = {
+    background: sticky ? theme.palette.TwClrBg : undefined,
+    boxShadow: sticky ? `0px 3px 3px -3px ${theme.palette.TwClrBaseGray200}` : undefined,
+    paddingTop: sticky ? theme.spacing(3) : undefined,
+    paddingRight: sticky ? theme.spacing(3) : undefined,
+    position: sticky ? 'fixed' : undefined,
+    top: sticky ? `${TOP_BAR_HEIGHT}px` : undefined,
+    visibility: scrollDown && sticky ? 'hidden' : 'visible',
+    width: sticky ? '-webkit-fill-available' : undefined,
+    zIndex: sticky ? 100 : undefined,
+  };
 
   return (
     <Box ref={ref} sx={styles}>
--- a/src/components/common/PageHeaderWrapper.tsx
+++ b/src/components/common/PageHeaderWrapper.tsx
@@ -1,0 +1,72 @@
+import React, { useEffect, useRef, useState } from 'react';
+import { makeStyles } from '@mui/styles';
+import { Theme } from '@mui/material';
+
+const TOP_BAR_HEIGHT = 64;
+
+const useStyles = makeStyles((theme: Theme) => ({
+  headerFixed: {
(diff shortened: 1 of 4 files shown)


src/index.css
```diff
@@ -9,7 +9,6 @@ body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   min-height: 100%;
-  height: 100%;
```
charlie-tf, line 12: This caused the scroll event to not fire

src/components/common/PageHeaderWrapper.tsx
```diff
@@ -0,0 +1,72 @@
...
+    };
+    window.addEventListener('scroll', handleScroll, { passive: true });
+
+    return () => {
+      window.removeEventListener('scroll', handleScroll);
+    };
+  }, [children, nextElement, height]);
+
+  return (
+    <div ref={ref} className={`${sticky ? classes.headerFixed : ''} ${scrollDown && sticky ? classes.hide : ''}`}>
```
constanzauanini, line 68: Using makeStyles is not correct on MUI5. We could have a Box here instead of a div and add styles in sx={{ }} 
You could have for example sx = {{ visibility: scrollDown && sticky ? 'hidden' : 'visible' }} and for other styles maybe create a variable

src/components/common/PageHeaderWrapper.tsx
```diff
@@ -64,9 +49,22 @@ export default function PageHeaderWrapper({ children, nextElement }: Props): JSX
     };
   }, [children, nextElement, height]);
 
+  const styles = useMemo<Record<string, any>>(
```
constanzauanini, line 58: Is it necessary to put it inside a useMemo?
Could it be ```
const styles = {
      visibility: scrollDown && sticky ? 'hidden' : 'visible',
      position: sticky ? 'fixed' : undefined,
      top: sticky ? `${TOP_BAR_HEIGHT}px` : undefined,
      width: sticky ? '-webkit-fill-available' : undefined,
      background: sticky ? theme.palette.TwClrBg : undefined,
      boxShadow: sticky ? `0px 3px 3px -3px ${theme.palette.TwClrBaseGray200}` : undefined,
      zIndex: sticky ? 100 : undefined,
    }
```

charlie-tf: not necessary, i just thought that might cut down the number of times it had to rebuild the map 🤷

charlie-tf: but it's such a lightweight calc that it probably doesn't matter. I can take it out to simplify

constanzauanini (approved):

karthikbtf (commented):
Thanks for doing this @charlie-tf 
Noticed some behavior that could use tweaking perhaps - as a follow up is fine.
Ofc, my observations could be incorrect, in which case that's fine.
When scrolling, the title bar isn't sticky until scrolling stops (ideally the sticky part should not scroll away from view).
Also, when scrolled all the way to the bottom, the sticky bar never pops back.

charlie-tf:
> Thanks for doing this @charlie-tf Noticed some behavior that could use tweaking perhaps - as a follow up is fine. Ofc, my observations could be incorrect, in which case that's fine. When scrolling, the title bar isn't sticky until scrolling stops (ideally the sticky part should not scroll away from view). Also, when scrolled all the way to the bottom, the sticky bar never pops back.

Thanks for taking a look @karthikbtf !! So I think you are describing the intended behavior: as long as you are scrolling downward the header will disappear. However, when you scroll upward the head should reappear. Perhaps my use of the word "sticky" here was misleading. It's like a "peek-in" header.

karthikbtf:
> > Thanks for doing this @charlie-tf Noticed some behavior that could use tweaking perhaps - as a follow up is fine. Ofc, my observations could be incorrect, in which case that's fine. When scrolling, the title bar isn't sticky until scrolling stops (ideally the sticky part should not scroll away from view). Also, when scrolled all the way to the bottom, the sticky bar never pops back.
> 
> Thanks for taking a look @karthikbtf !! So I think you are describing the intended behavior: as long as you are scrolling downward the header will scroll out of view. However, when you scroll upward the head should reappear. Perhaps my use of the word "sticky" here was misleading. It's like a "peek-in" header.

Ah, ok got it.. I was comparing against confluence ux.
Sounds good if peek-in is the intended behavior.

karthikbtf:
> Thanks for doing this @charlie-tf Noticed some behavior that could use tweaking perhaps - as a follow up is fine. Ofc, my observations could be incorrect, in which case that's fine. When scrolling, the title bar isn't sticky until scrolling stops (ideally the sticky part should not scroll away from view). Also, when scrolled all the way to the bottom, the sticky bar never pops back.

Missed adding this earlier, the call-to-action buttons are in the sticky part as per the design. In case that was missed.

charlie-tf:
> Missed adding this earlier, the call-to-action buttons are in the sticky part as per the design. In case that was missed.

Oh dang, sorry I didn't catch this. I can fix in follow up when I add the wrapper to other pages.

PR #688, opened 2022-10-31
SW-2054 Allow nursery csv import without requiring species creation during onboarding

constanzauanini:
Screen Shot 2022-10-31 at 11 22 18

### DIFF
--- a/src/components/Inventory/index.tsx
+++ b/src/components/Inventory/index.tsx
@@ -216,7 +216,7 @@ export default function Inventory(props: InventoryProps): JSX.Element {
             </div>
           ) : (
             <Container maxWidth={false} className={classes.mainContainer}>
-              <EmptyStatePage pageName={'Inventory'} />
+              <EmptyStatePage pageName={'Inventory'} organization={organization} />
             </Container>
           )
         ) : (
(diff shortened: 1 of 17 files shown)


src/components/Inventory/index.tsx
```diff
@@ -64,15 +68,28 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
     const emptyState = [];
 
     if (!hasSpecies) {
       emptyState.push({
-        title: strings.SPECIES,
+        title: strings.CREATE_SPECIES_LIST,
         text: emptyMessageStrings.INVENTORY_ONBOARDING_SPECIES_MSG,
         buttonText: strings.GO_TO_SPECIES,
         onClick: () => goTo(APP_PATHS.SPECIES),
+        altItem: hasNurseries
```
karthikbtf, line 84: for accessions, importing required user to select a seedbank as the first step.. that's why we had a check.
here, just curious how does nursery get selected for import? is it in the csv?

constanzauanini: There is a dropdown on the import modal and it is send on the api as a queryParameter

karthikbtf: Got it, saw it in vercel. Thanks.

karthikbtf (approved):

PR #691, opened 2022-10-31
SW-2027 updates: Iteration on onboarding for accessions/inventory

karthikbtf:
- show import options disabled if seedbanks/nurseries are not available
- change text as per Ming's suggestion
- disabled button state is weird, captured in bug [SW-2079](https://terraformation.atlassian.net/browse/SW-2079)

Terraware App 2022-10-31 16-10-10

Terraware App 2022-10-31 16-09-24

Terraware App 2022-10-31 16-06-09

Terraware App 2022-10-31 16-05-02


constanzauanini (approved):
Thanks for doing the change in boths!

PR #692, opened 2022-11-01
Refresh inventory table after import

constanzauanini:


src/components/Inventory/index.tsx
```diff
@@ -209,11 +209,16 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
+    onApplyFilters();
+    setImportInventoryModalOpen(false);
+  };
+
   return (
     <TfMain>
       <ImportInventoryModal
         open={importInventoryModalOpen}
-        onClose={() => setImportInventoryModalOpen(false)}
+        onClose={onCloseImportInventoryModal}
```
karthikbtf, line 221: will be cool to avoid the apply-filter if it is a modal cancel.
for later, we should start introducing the concept of onCancel in our modals perhaps.

karthikbtf (approved):
one comment for later if we want to capture it

src/components/Inventory/index.tsx
karthikbtf: nice, thanks @constanzauanini :)

karthikbtf (approved):

PR #697, opened 2022-11-01
SW-2033 Extract inventory search/filter/pills into a reusable component

karthikbtf:
- reuse in inventory table
- reuse in seedlings batches table
- no change in functionality

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.8.7",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.8.7",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
-    "@terraware/web-components": "^1.0.55",
+    "@terraware/web-components": "^1.0.56",
```
karthikbtf, line 19: this was by mistake but guessing it's ok since we need the arrow icon

constanzauanini (approved):

PR #700, opened 2022-11-03
SW-2077 Animate and Delay Header

charlie-tf:
- Add debounce to sticky and scroll state
- Animate in and out
- useLayoutEffect to avoid some flickering
- make TopBar opaque

Kapture 2022-11-02 at 21 20 44


constanzauanini (approved):
I really like how the animation looks!

PR #701, opened 2022-11-03
SW-2093 Use new table version and remove unnecessaries stopPropagations

constanzauanini:
New version of web-components, when rows are not clickable, only clicking on the checkbox is allowed. This is why we are removing all the stopPropagations that are not necessary anymore

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.8.7",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.8.7",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
-    "@terraware/web-components": "^1.0.56",
+    "@terraware/web-components": "^1.0.57",
```
karthikbtf, line 19: 58

karthikbtf (approved):
after fixing version

PR #702, opened 2022-11-03
Create test to select all columns on database

constanzauanini:
I don't think we need to add a value for each filter, because search api is including all the fields, so if api doesn't fail (still returns results) then it means that each field added to the search exists. I tested adding a column with a search value that doesn't exist and search fails after selecting it.

Also, we are redoing filters at some point, so if we test adding a value to each filter then we will have to redo all tests

### DIFF
--- a/cypress/integration/10-database.js
+++ b/cypress/integration/10-database.js
@@ -70,48 +70,11 @@ describe('Database', () => {
       cy.get('#more-options').click();
       cy.get('#edit-columns').click();
 
-      cy.get('#active').click();
-      // cy.get('#state').click(); already selected
-      cy.get('#facility_name').click();
-
-      // cy.get('#speciesName').click(); already selected
-      cy.get('#species_commonName').click();
-      cy.get('#species_familyName').click();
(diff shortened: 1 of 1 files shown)


cypress/integration/10-database.js
```diff
@@ -62,6 +62,66 @@ describe('Database', () => {
       cy.get('#table-header > :nth-child(9)').contains('SEED BANKS');
     });
 
+    it('should select ALL the columns', () => {
+      cy.visit('/accessions');
+      cy.intercept('POST', '/api/v1/search').as('search');
+      cy.intercept('POST', '/api/v1/seedbank/values').as('values');
+
+      cy.get('#more-options').click();
```
karthikbtf, line 70: @constanzauanini can we do this programmatically by search for elements of a certain type? this will be cleaner and won't need any updates.

constanzauanini: Sounds good!

karthikbtf (approved):
Very nice!! 🎉

PR #704, opened 2022-11-04
Use new editable prop on selects

constanzauanini:


src/components/SeedBank/SelectSeedBankModal.tsx
```diff
@@ -77,7 +77,7 @@ export default function SelectSeedBankModal(props: SelectSeedBankProps): JSX.Ele
             label={strings.SEED_BANK}
             aria-label={strings.SELECT_SEED_BANK}
             placeholder={strings.SELECT}
-            readonly={false}
+            allowEditingValue={true}
```
constanzauanini, line 80: Do we intentionally allowed editing in this select?

karthikbtf: looks like a bug

karthikbtf (approved):
can you change the title of the PR/commit before merging?

PR #706, opened 2022-11-04
SW-1815 Fix unnecessary react update in app-version detector

karthikbtf:
- updates happen after component is dismounted, potential memory-leak
- error was seen in js console

### DIFF
--- a/src/components/common/DetectAppVersion.tsx
+++ b/src/components/common/DetectAppVersion.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect } from 'react';
 import { makeStyles } from '@mui/styles';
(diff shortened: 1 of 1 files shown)


constanzauanini (approved):
I've seen that error on console before and wasn't sure why it was happening. Thank you!

karthikbtf:
> I've seen that error on console before and wasn't sure why it was happening. Thank you!

Yes - we have multiple places causing this issue, any component that is awaiting a network response - but is dismounted - and ends up updating state after network response completes. This particular component is on every page and therefore higher potential for memory leaks. The fix is not super elegant I agree.

Next year if/when we slowly integrate with redux, we can get rid of these issues since network layer is separate from components - it's mostly data subscription which gets auto-canceled once component is dismounted.

PR #709, opened 2022-11-04
SW-2162 Fix readonly related dropdown errors

karthikbtf:
- fix usages
- readonly flag was renamed as editable which defaults to false


charlie-tf (approved):
seems good in the vercel deploy, too 👍

karthikbtf:
> seems good in the vercel deploy, too 👍

Thanks @charlie-tf

PR #712, opened 2022-11-04
SW-2076 Add PageHeaderWrapper to Pages

charlie-tf:
- Added the header wrapper to several more pages:
  - species list
  - accessions list
  - check-in accessions
  - monitoring
  - inventory list
- rearrange accessions mobile view to have cta buttons in header

### DIFF
--- a/src/components/Inventory/InventoryTable.tsx
+++ b/src/components/Inventory/InventoryTable.tsx
@@ -1,12 +1,9 @@
 import React from 'react';
 import strings from 'src/strings';
 import { ServerOrganization } from 'src/types/Organization';
-import { Button, Table, TableColumnType } from '@terraware/web-components';
-import { Box, Grid } from '@mui/material';
+import { Table, TableColumnType } from '@terraware/web-components';
+import { Grid } from '@mui/material';
 import { SearchResponseElement } from 'src/api/search';
-import { useHistory } from 'react-router-dom';
-import { useDeviceInfo } from '@terraware/web-components/utils';
-import { APP_PATHS } from 'src/constants';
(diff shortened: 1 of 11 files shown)


karthikbtf (commented):
Spiffy!! The peek-a-boo scrolling works on all the targeted pages, desktop and mobile.
Very cool!

Two observed issues:
1) empty state for monitoring is no longer centered
2) scroll state is preserved across People and SeedBanks (not sure why this is affected as we didn't target these but maybe some unrelated change got picked up?)

charlie-tf:
> Spiffy!! The peek-a-boo scrolling works on all the targeted pages, desktop and mobile. Very cool!
> 
> Two observed issues:
> 
> 1. empty state for monitoring is no longer centered
> 2. scroll state is preserved across People and SeedBanks (not sure why this is affected as we didn't target these but maybe some unrelated change got picked up?)

1. Fixed by wrapping that content in a Container instead of a div
2. This appears to be happening in prod, so not related to these changes, but worth checking out.

constanzauanini (approved):
Lgtm! One question, on accession page, are we sure we want the checkin message to be part of the pageHeader ?

charlie-tf:
> Lgtm! One question, on accession page, are we sure we want the checkin message to be part of the pageHeader ?

I believe so yes, from the figma:
>If there’s a page message, it will be part of the sticky header (below the heading and actions). Page messages will now be full-width.

constanzauanini:
> Member

Ok, I'm not sure if the check in message is a normal page message, but lets keep it like this. If @chudo76 has a comment, we can change it later

PR #713, opened 2022-11-07
SW-2168 Update the Background Gradient

charlie-tf:
- Unfortunately the hex colors are not yet specified as tokens.

e.g.:
image

### DIFF
--- a/src/components/TopBar/TopBar.tsx
+++ b/src/components/TopBar/TopBar.tsx
@@ -5,7 +5,7 @@ import useDeviceInfo from 'src/utils/useDeviceInfo';
 
 const useStyles = makeStyles((theme: Theme) => ({
   appBar: {
-    background: theme.palette.TwClrBg,
+    background: '#FBF9F9',
     color: theme.palette.TwClrTxt,
     boxShadow: 'none',
     minHeight: '64px',
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -56,6 +56,7 @@ import InventoryView from './components/Inventory/InventoryView';
 
 const useStyles = makeStyles((theme: Theme) => ({
   container: {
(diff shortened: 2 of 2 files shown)


src/components/TopBar/TopBar.tsx
```diff
@@ -5,7 +5,7 @@ import useDeviceInfo from 'src/utils/useDeviceInfo';
 
 const useStyles = makeStyles((theme: Theme) => ({
   appBar: {
-    background: theme.palette.TwClrBg,
+    background: '#FBF9F9',
```
constanzauanini, line 8: Is this color going to be used many times? We could ask Chudo to create a token for this

charlie-tf: Good call, I will ask him when he's in today

constanzauanini (approved):

src/components/TopBar/TopBar.tsx
karthikbtf: @charlie-tf this brings up a bigger question - should we have the token values in place for visual updates - in order to develop?
Would be great to keep the tokens as-is in terraware-web - so as not to undo your replace-hardcoded-values effort.
Should we work with @chudo76 to generate new zeroheight links for web-components 2.x.x?
As a hack, we could replace the values in the theme file - in web-components 2.x.x - until we have updated zeroheight files for web-components 2.x.x. That way, terraware-web can remain unaffected.
Thoughts?

charlie-tf: A separate theme for visual-updates in web-components 2.x.x sounds like a good way to go 👍

charlie-tf: Would it feel okay to merge this as is and circle back to update those hex values with tokens when they are in place?

PR #716, opened 2022-11-07
SW-2165 Validate selection in inventory import modal

karthikbtf:
- add component prop (optional) in ImportModal to validate import
- use validate flag in NuseryDropdown to mark selection as required

Terraware App 2022-11-07 10-05-44

Terraware App 2022-11-07 10-06-02

### DIFF
--- a/src/components/Inventory/ImportInventoryModal.tsx
+++ b/src/components/Inventory/ImportInventoryModal.tsx
@@ -29,6 +29,7 @@ export default function ImportInventoryModal(props: ImportInventoryModalProps): 
   const { open, onClose, organization, reloadData } = props;
   const [record, setRecord] = useForm({ facilityId: -1 });
   const [selectedFacility, setSelectedFacility] = useState<Facility>();
+  const [validate, setValidate] = useState<boolean>(false);
 
   useEffect(() => {
     if (record && record.facilityId) {
@@ -36,6 +37,14 @@ export default function ImportInventoryModal(props: ImportInventoryModalProps): 
       setSelectedFacility(found);
     }
   }, [record, organization]);
+
+  const isValid = () => {
+    setValidate(true);
+    if (!selectedFacility) {
+      return false;
+    }
+    return true;
+  };
 
   return (
     <ImportModal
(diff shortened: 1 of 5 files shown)


src/components/Inventory/ImportInventoryModal.tsx
```diff
@@ -29,6 +29,7 @@ export default function ImportInventoryModal(props: ImportInventoryModalProps):
   const { open, onClose, organization, reloadData } = props;
   const [record, setRecord] = useForm({ facilityId: -1 });
   const [selectedFacility, setSelectedFacility] = useState<Facility>();
+  const [validate, setValidate] = useState<boolean>(false);
```
constanzauanini, line 32: We should clean the error on close

src/components/Inventory/NurseryDropdown.tsx
```diff
@@ -35,7 +35,7 @@ export default function NurseryDropdown<T extends AccessionPostRequestBody>(
       selectedValue={record.facilityId?.toString()}
       options={getAllNurseries(organization).map((nursery) => ({ label: nursery.name, value: nursery.id.toString() }))}
       onChange={onChangeHandler}
-      errorText={validate && !record.facilityId ? strings.REQUIRED_FIELD : ''}
+      errorText={validate && (!record.facilityId || record.facilityId === -1) ? strings.REQUIRED_FIELD : ''}
```
charlie-tf, line 38: under what condition is the `record.facilityId` equal to -1?

src/components/Inventory/ImportInventoryModal.tsx
karthikbtf: good catch

src/components/Inventory/NurseryDropdown.tsx
karthikbtf: gripe,, unfortunately it is initiallized as -1 in calling code since facilityId cannot be null/undefined
cc @constanzauanini

karthikbtf: Refactoring this to avoid assumptions

PR #718, opened 2022-11-07
SW-2099 Add opt-in preference for tracking v1

karthikbtf:
- to be used in gating tracking features


karthikbtf:
> have a secret in history

removed from history, phew!

PR #722, opened 2022-11-08
SW-2173 Fix nursery column sorting on inventory table

constanzauanini:


karthikbtf (approved):
cc @charlie-tf in case he has other suggestions

constanzauanini:
> cc @charlie-tf in case he has other suggestions

@charlie-tf I will merge, but if you have a cleaner/better options for this, let me know

PR #724, opened 2022-11-08
SW-2100 Adding planting sites menu item

constanzauanini:
This just adds the item to the menu, but doesn't add the Locations submenu

Screenshot 2022-11-08 at 14 08 32

### DIFF
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -38,6 +38,7 @@ export default function NavBar({ organization, setShowNavBar }: NavBarProps): JS
   const isSeedbanksRoute = useRouteMatch(APP_PATHS.SEED_BANKS + '/');
   const isNurseriesRoute = useRouteMatch(APP_PATHS.NURSERIES + '/');
   const isInventoryRoute = useRouteMatch(APP_PATHS.INVENTORY + '/');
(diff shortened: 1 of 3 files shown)


src/components/NavBar.tsx
```diff
@@ -160,6 +161,16 @@ export default function NavBar({ organization, setShowNavBar }: NavBarProps): JS
               id='nurseries'
             />
           )}
+          {isEnabled('Tracking V1') && (
+            <NavItem
```
karthikbtf, line 165: Guess there's no icon since we will move it under Locations. cc @joshbuchea in case it needs to be rebased into visual_updates branch.

src/constants.ts
```diff
@@ -38,4 +38,5 @@ export enum APP_PATHS {
   WELCOME = '/welcome',
   MY_ACCOUNT = '/myaccount',
   MY_ACCOUNT_EDIT = '/myaccount/edit',
+  PLANTING_SITES = '/planting_sites',
```
karthikbtf, line 41: i think we use '-' instead of '_', see SEEDS_DASHBOARD

karthikbtf (approved):
One comment if you can fix.

src/constants.ts
constanzauanini: Oh, you are right!

PR #730, opened 2022-11-09
SW-2103 Add Planting Sites empty state

constanzauanini:
Screenshot 2022-11-09 at 12 27 24

TODO: Replace with correct icon

### DIFF
--- a/src/components/PlantingSites/index.tsx
+++ b/src/components/PlantingSites/index.tsx
@@ -1,10 +1,32 @@
-import { Box } from '@mui/material';
+import { Box, Container, Grid, Typography } from '@mui/material';
+import { theme } from '@terraware/web-components';
+import { useRef } from 'react';
+import strings from 'src/strings';
+import PageHeaderWrapper from '../common/PageHeaderWrapper';
+import TfMain from '../common/TfMain';
 import EmptyStatePage from '../emptyStatePages/EmptyStatePage';
 
 export default function PlantingSitesList(): JSX.Element {
+  const contentRef = useRef(null);
+
   return (
-    <Box>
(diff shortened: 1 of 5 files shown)


src/components/PlantingSites/index.tsx
```diff
@@ -1,6 +1,10 @@
 import { Box } from '@mui/material';
-import strings from 'src/strings';
+import EmptyStatePage from '../emptyStatePages/EmptyStatePage';
 
 export default function PlantingSitesList(): JSX.Element {
-  return <Box>{strings.PLANTING_SITES}</Box>;
+  return (
+    <Box>
```
karthikbtf, line 6: i think we may need to add the same components like in seedbanks / nurseries etc.. there's some extra padding in the planting sites view.
fwiw, i think the extra padding in planting sites view looks cleaner - it helps fix https://terraformation.atlassian.net/browse/SW-2187 but not sure what the spec is.

karthikbtf (approved):
@constanzauanini two issues I noticed, both unrelated to this PR but just FYI.
1) planting sites menu does not show up for contributors
2) empty state content looks different, padding/positioning for various pages in various scenarios
Not sure how we should proceed regarding 2)
Btw, the search api PR for planting sites in BE was approved - should go in soon.

src/components/PlantingSites/index.tsx
constanzauanini: Mmm but seedbanks and nurseries have the emptyComponent called directly from the app.tsx. I could add the same components as inventory that is the only one calling the empty state inside the component

karthikbtf (approved):
Thanks @constanzauanini

PR #731, opened 2022-11-09
SW-2186 Sensor Monitoring Dashboard

charlie-tf:
image
image

### DIFF
--- a/src/components/common/PageHeaderWrapper.tsx
+++ b/src/components/common/PageHeaderWrapper.tsx
@@ -97,7 +97,7 @@ export default function PageHeaderWrapper({ children, nextElement }: Props): JSX
   }, [debouncedSticky, debouncedScrollDown, height]);
 
   const styles: Record<string, any> = {
-    background: debouncedSticky ? theme.palette.TwClrBg : undefined,
+    background: debouncedSticky ? '#FBF9F9' : undefined,
     boxShadow: debouncedSticky ? `0px 3px 3px -3px ${theme.palette.TwClrBaseGray200}` : undefined,
     paddingRight: debouncedSticky ? theme.spacing(3) : undefined,
     position: debouncedSticky ? 'fixed' : undefined,
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.8.7",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
-    "@terraware/web-components": "^2.0.1",
+    "@terraware/web-components": "^2.0.2",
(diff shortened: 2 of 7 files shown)


src/components/common/PageHeaderWrapper.tsx
```diff
@@ -97,7 +97,7 @@ export default function PageHeaderWrapper({ children, nextElement }: Props): JSX
   }, [debouncedSticky, debouncedScrollDown, height]);
 
   const styles: Record<string, any> = {
-    background: debouncedSticky ? theme.palette.TwClrBg : undefined,
+    background: debouncedSticky ? '#FBF9F9' : undefined,
```
karthikbtf, line 100: if we don't have a ticket to follow up on this hardcoded color, we should probably create one - just so we don't miss it?

karthikbtf (commented):
@charlie-tf post a screenshot?

src/components/common/PageHeaderWrapper.tsx
charlie-tf: Good call, I will create one

charlie-tf: https://terraformation.atlassian.net/browse/SW-2202

hannah-cai:
looks good! vertical alignment of these icons is slightly off with the text:
image

on another note, have you been adding the semantic color tokens with these updates, or is that a separate effort? not sure if I should be calling out color discrepancies I see

charlie-tf:
> looks good! vertical alignment of these icons is slightly off with the text
> 
> on another note, have you been adding the semantic color tokens with these updates, or is that a separate effort? not sure if I should be calling out color discrepancies I see

@hannah-cai I think i've got the alignment better now

On your second question, I have updated the tokens in the code. However, those tokens are still defined as the previous color values. I have chatted with @chudo76 and @karthikbtf a little bit on how we might do the transition to new colors, likely involving generating a new set of zeroheight links

hannah-cai:
@charlie-tf the alignment looks good now! and thanks for clarifying. it sounds like the tokens are being updated, but they will still be linked to the interim semantic colors. in that case, I'd like to call out that all of the icons in the cards should be the secondary icon color (tw-clr-icn-secondary). in addition, I hadn't caught this before, but the icons in the cards on other pages should be the same size as the ones on this page (medium icon size). if you look on the home and seeds dashboard pages, they're using a small icon size.

charlie-tf:
Am I okay to merge this? I created a ticket to update those icon sizes on the Home and Seed dashboards (https://terraformation.atlassian.net/browse/SW-2218)

hannah-cai:
@charlie-tf if you've updated the color tokens I mentioned, yes!

PR #733, opened 2022-11-09
SW-2118 Create a Map component

constanzauanini:


src/components/Map/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+      container: mapContainer.current || '',
+      style: 'mapbox://styles/mapbox/streets-v11',
+      center: [lng, lat],
+      zoom: zoom,
+    });
+  });
+
+  return (
+    <Box>
+      <Box ref={mapContainer} sx={{ height: '400px' }} />
```
charlie-tf, line 31: will this automatically be full-width?

yarn.lock
```diff
@@ -1972,6 +1972,14 @@
     get-stream "^6.0.1"
     minimist "^1.2.5"
 
+"@mapbox/geojson-rewind@^0.5.1":
```
karthikbtf, line 1975: wow mapbox sure pulls in a lot of dependencies.

src/components/Map/index.tsx
karthikbtf: Probably not, this component will need more updates shortly.
Some background if it helps - the current task is mostly to create a placeholder component - which we can enhance over time - as we start developing planting sites and other tracking stuff.

karthikbtf (commented):
looks good for the immediate term, we could leave the ticket open until we add more of the requirements?

yarn.lock
constanzauanini: Yes, I didn't know about a lot of them

constanzauanini:
> looks good for the immediate term, we could leave the ticket open until we add more of the requirements?

You mean leave open this PR, or the jira ticket?

karthikbtf (approved):
@constanzauanini thanks
I meant closing the PR but leaving JIRA open.
However we have to make all those changes anyway so might as well close JIRA too.

PR #734, opened 2022-11-09
SW-2208 handle inventory when species no longer exists

karthikbtf:
- check for species id in renderer
- we can no longer show species info or details by species with a link
- awaiting PM response on this fix

Terraware App 2022-11-09 14-52-08

### DIFF
--- a/src/components/Inventory/InventoryCellRenderer.tsx
+++ b/src/components/Inventory/InventoryCellRenderer.tsx
@@ -53,7 +53,7 @@ export default function InventoryCellRenderer(props: RendererProps<TableRowType>
     );
   }
 
-  if (column.key === 'species_scientificName') {
+  if (column.key === 'species_scientificName' && row.species_id) {
     return (
       <CellRenderer
         index={index}


constanzauanini:
I think Alison wants to still be able to see the species name if it's deleted

karthikbtf:
> I think Alison wants to still be able to see the species name if it's deleted

yes, not sure it is feasible, will check with Steve. In any case, we need this fix -without this inventory page won't load if there is a deleted species in inventory. will show error page. will follow up on jira.

constanzauanini:
> > I think Alison wants to still be able to see the species name if it's deleted
> 
> yes, not sure it is feasible, will check with Steve. In any case, we need this fix -without this inventory page won't load if there is a deleted species in inventory. will show error page. will follow up on jira.

Ok, let's merge this fix and we can follow up later

karthikbtf:
> > > I think Alison wants to still be able to see the species name if it's deleted
> > 
> > 
> > yes, not sure it is feasible, will check with Steve. In any case, we need this fix -without this inventory page won't load if there is a deleted species in inventory. will show error page. will follow up on jira.
> 
> Ok, let's merge this fix and we can follow up later

Thanks, yes, Allison recommended the same. The fixes are a larger enhancement discussion for later.
See JIRA when you get a chance, there will be small changes in species results.

PR #735, opened 2022-11-10
SW-2215 Clear subset weight error in calculate on close

karthikbtf:
- show error only when user clicks 'get total count'
- we don't clear the entered subset weight on close if there is an error, wondering what the behavior should be here


karthikbtf:
I believe there are multiple errors here. Will check with Constanza.

karthikbtf:
reopening since it is still valid

constanzauanini (approved):

PR #737, opened 2022-11-10
SW-2207 Enable Nursery management feature in prod

constanzauanini:


charlie-tf (approved):
exciting!

src/features.ts
```diff
@@ -24,7 +24,7 @@ export const OPT_IN_FEATURES: Feature[] = [
     name: 'Nursery management',
     preferenceName: 'enableUINursery',
     active: true,
-    enabled: false,
+    enabled: true,
```
karthikbtf, line 27: This is enough for now. It will make all the checking code return true.
We can remove the checking code slowly.
You can test what production users will see by enabling production mode in the opt-in page.
If you think removing the checking code is small we can do that too.

karthikbtf (approved):

src/features.ts
karthikbtf: ok just saw that, seemed like more code than i remembered.
maybe we should have just set enabled to true - to avoid risk - since that code path was already validated. my bad.
especially the accession withdrawals code.
can you confirm accession withdrawals is ok?

karthikbtf (approved):
Looks good, Going forward we can agree to use the enabled flag for last minute enabling - to avoid risk.

PR #741, opened 2022-11-10
SW-2209 Render Error Page

charlie-tf:
- fix the main element positioning
- wrap error page in ThemeProvider to get text color

image
image


karthikbtf (approved):
🎉

PR #743, opened 2022-11-11
generate types for tracking api/models

karthikbtf:
- generated from https://github.com/terraware/terraware-server/pull/790
- we can get this in since there will be parallel effort on top of this


karthikbtf:
@constanzauanini we can't generate types for tracking off current BE on main. I have a PR out to enable that.
I generated these types/models off my BE PR.
We can get this in and work on it in parallel - has all models and api coded in.
Let me know.

constanzauanini (approved):
Excellent!

karthikbtf:
> Excellent!

Thanks, I’ll get this in now so you have it.

PR #744, opened 2022-11-11
SW-2255 Fix accessions to nursery transfer withdrawal when selecting-all

karthikbtf:
- use estimatedCount instead of remaining quantity, similar to viability testing purpose
- also fixed notes display on batch details

Terraware App 2022-11-10 16-55-05

Terraware App 2022-11-10 16-49-55

Terraware App 2022-11-10 16-49-22

### DIFF
--- a/src/components/accession2/withdraw/WithdrawModal.tsx
+++ b/src/components/accession2/withdraw/WithdrawModal.tsx
@@ -196,7 +196,7 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element 
         });
         setNurseryTransferRecord({
           ...nurseryTransferRecord,
-          germinatingQuantity: accession.remainingQuantity?.quantity || 0,
+          germinatingQuantity: accession.estimatedCount ? accession.estimatedCount : 0,
         });
       }
     } else {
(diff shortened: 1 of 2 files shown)


src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -196,7 +196,7 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
         });
         setNurseryTransferRecord({
           ...nurseryTransferRecord,
-          germinatingQuantity: accession.remainingQuantity?.quantity || 0,
+          germinatingQuantity: accession.estimatedCount ? accession.estimatedCount : 0,
```
constanzauanini, line 199: Shouldn't this value be the same?

constanzauanini (approved):

src/components/accession2/withdraw/WithdrawModal.tsx
constanzauanini: Oh, yes, thanks for the clarification!

karthikbtf: I’ll do some more testing

PR #745, opened 2022-11-11
SW-2215 Fix different unit error when calculating count

constanzauanini:
If selected units on calculator are different, execute count calculation on backend, and use grams values to see if subset weight is bigger than total weight


karthikbtf (approved):
Tried a qty in gram but subset in mg,  calculation seemed to do the right thing.

PR #746, opened 2022-11-11
SW-2106 Flow to add a planting site

constanzauanini:
Screenshot 2022-11-11 at 13 50 08

Screenshot 2022-11-11 at 13 47 39

### DIFF
--- a/src/components/PlantingSites/PlantingSiteCreate.tsx
+++ b/src/components/PlantingSites/PlantingSiteCreate.tsx
@@ -7,7 +7,7 @@ import FormBottomBar from '../common/FormBottomBar';
 import { useDeviceInfo } from '@terraware/web-components/utils';
 import useForm from 'src/utils/useForm';
 import { PlantingSitePostRequestBody, postPlantingSite } from 'src/api/tracking/tracking';
-import { APP_PATHS } from 'src/constants';
+import { APP_PATHS, TERRAWARE_SUPPORT_LINK } from 'src/constants';
 import { Link, useHistory } from 'react-router-dom';
 import useSnackbar from 'src/utils/useSnackbar';
 import TextField from '@terraware/web-components/components/Textfield/Textfield';
@@ -99,7 +99,7 @@ export default function CreatePlantingSite(props: CreatePlantingSiteProps): JSX.
           <Grid item xs={gridSize()}>
             <TextField
               id='name'
-              label={strings.NAME}
+              label={strings.NAME_REQUIRED}
               type='text'
               onChange={onChange}
               value={record.name}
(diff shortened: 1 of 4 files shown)


src/components/PlantingSites/PlantingSiteCreate.tsx
```diff
@@ -1,10 +1,150 @@
...
+              }}
+            >
+              <Typography fontSize='20px' fontWeight={600} margin={theme.spacing(3, 0)}>
+                {strings.IMPORT_BOUNDARIES_AND_PLOTS}
+              </Typography>
+              <Typography fontSize='16px' fontWeight={400} padding={theme.spacing(1, 0)}>
+                {strings.IMPORT_BOUNDARIES_AND_PLOTS_DESCRIPTION}
+              </Typography>
+              <Box sx={{ paddingY: 2 }}>
+                <Button label={strings.CONTACT_US} onClick={() => true} size='medium' />
```
karthikbtf, line 141: for now maybe we can do the same thing the contact us card does when we click?
i understand there's a jira out for allison on what this should do.

src/components/PlantingSites/index.tsx
```diff
@@ -1,4 +1,4 @@
-import CreatePlantingSite from './PlantingSitesList';
-import PlantingSitesList from './PlantingSiteCreate';
+import CreatePlantingSite from './PlantingSiteCreate';
+import PlantingSitesList from './PlantingSitesList';
```
karthikbtf, line 2: woops :) good catch

src/components/PlantingSites/PlantingSiteCreate.tsx
```diff
@@ -1,10 +1,150 @@
...
+            </Typography>
+            <Typography fontSize='14px' fontWeight={400} margin={theme.spacing(0, 0, 3, 0)}>
+              {strings.ADD_PLANTING_SITE_DESCRIPTION}
+            </Typography>
+          </Grid>
+          <PageSnackbar />
+          <Grid item xs={gridSize()}>
+            <TextField
+              id='name'
+              label={strings.NAME}
```
karthikbtf, line 102: I understand Name and Description don't have the required '*' next to them - but feels like they should? 
We could check with @chudo76 in figma.

src/components/PlantingSites/PlantingSiteCreate.tsx
constanzauanini: Oh, yes, I can add that for now!

src/components/PlantingSites/PlantingSiteCreate.tsx
constanzauanini: Name is required according to the api, and description is optional. I'm validating name when saving

karthikbtf: PRD says name alone is required - which is different from our other pages.
We should at least ask Chudo if we should add the '*' next to 'Name' label for consistency, yes?

karthikbtf (approved):
looks good, we can follow up on name label when chudo responds on figma.
thanks for asking the question there.

constanzauanini:
> looks good, we can follow up on name label when chudo responds on figma. thanks for asking the question there.



> looks good, we can follow up on name label when chudo responds on figma. thanks for asking the question there.

Yes, I added the * for consistency, I can remove it later if we don't need it

PR #747, opened 2022-11-11
SW-2179 Add environment variable for Mapbox token

sgrimm:
To use Mapbox in local dev environments, the server has to be configureed with a
Mapbox API token. Add it to the list of server environment variables in the Docker
configuration so people can put it in their local `.env` files easily.


karthikbtf (approved):
Thanks 👍 
 I set it in my application.yaml and been generating client tokens successfully.
Was going to request this for Constanza/others that use docker.

sgrimm:
I also updated the FE engineer onboarding Confluence doc to include this.

PR #748, opened 2022-11-11
SW-2109 Planting site database table

constanzauanini:
Screenshot 2022-11-11 at 17 37 07

Added search input UI but is not functional yet

### DIFF
--- a/src/components/PlantingSites/PlantingSitesCellRenderer.tsx
+++ b/src/components/PlantingSites/PlantingSitesCellRenderer.tsx
@@ -1,0 +1,24 @@
+import React from 'react';
+import { makeStyles } from '@mui/styles';
+import CellRenderer, { TableRowType } from '../common/table/TableCellRenderer';
+import { RendererProps } from '../common/table/types';
+
+const useStyles = makeStyles(() => ({
+  text: {
+    fontSize: '14px',
+    '& > p': {
+      fontSize: '14px',
+    },
+  },
+}));
+
+export default function PlantingSitesCellRenderer(props: RendererProps<TableRowType>): JSX.Element {
+  const classes = useStyles();
+  const { column, row, value, index } = props;
+
+  if (column.key === 'name') {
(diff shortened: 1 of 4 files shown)


src/components/PlantingSites/PlantingSitesCellRenderer.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export default function PlantingSitesCellRenderer(props: RendererProps<TableRowType>): JSX.Element {
+  const classes = useStyles();
+  const { column, row, value, index } = props;
+
+  if (column.key === 'name') {
+    return <CellRenderer index={index} column={column} value={value} row={row} className={classes.text} />;
+  }
+
+  return <CellRenderer {...props} className={classes.text} />;
```
karthikbtf, line 23: line 23 does the same thing as line 20, we can maybe remove lines 19 - 21

src/components/PlantingSites/PlantingSitesList.tsx
```diff
@@ -39,20 +53,47 @@ export default function PlantingSitesList(props: PlantingSitesListProps): JSX.El
...
       </PageHeaderWrapper>
       <Grid>
         <Box>
-          {searchResults?.length === 0 && (
+          {searchResults?.length === 0 ? (
             <Container sx={{ paddingY: 4 }}>
               <EmptyStatePage pageName={'PlantingSites'} />
             </Container>
+          ) : (
+            <PlantigSitesTable
```
karthikbtf, line 91: this is flashing before search finishes, we may need to hide this while search is happening

karthikbtf (approved):
looks good, some comments for follow up

src/components/PlantingSites/PlantingSitesCellRenderer.tsx
constanzauanini: Yes, I just leave this prepared for adding the link to the name column, but I will remove it and add it back later

karthikbtf (approved):
🎉

PR #749, opened 2022-11-11
SW-2219 More map enhancements

karthikbtf:
- add map component for planting site which provides data to the Map component
- introduced some models and utilities
- Map works around these models with no dependency on actual planting site semantics
- this is still an intermediate step, hoping to get this in and we continue to make improvements
- the calculation from polygon coords (EPSG 3857) to lat/long (EPSG 4326) is done in client for now, this will be handled by BE soon and we can remove that code next
- we can make changes to this code over time, just an initial version of the map for planting site
- on click etc we can continue updating
- Example usage 
```
 <PlantingSiteMap siteId={siteId} />
```

image here is based on testing using contactus page and admin ui to upload shapes to a planting site

Terraware App 2022-11-11 15-02-52

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -32,6 +32,7 @@
       charset="UTF-8"
       src="//cdn.cookie-script.com/s/12c369c418d1c681aa7a0bfe74c63125.js"
     ></script>
+    <link href="https://api.mapbox.com/mapbox-gl-js/v2.10.0/mapbox-gl.css" rel="stylesheet" />
     <!--
       Notice the use of %PUBLIC_URL% in the tags above.
       It will be replaced with the URL of the `public` folder during the build.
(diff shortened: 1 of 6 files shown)


constanzauanini (approved):
This is exciting!

karthikbtf:
> This is exciting!



> This is exciting!

Thanks, this is an intermediate effort.
We have outstanding items to work on.
1) draw boundaries around each polygon
2) render on click needs formatting, and also maybe a callback support (so other components can find which polygon etc was clicked)
3) for now i switched to satellite view (we were using street view earlier) - we can iterate on this. guessing we need stateliite view for planting sites, we can check.
4) points/lines are dropped in the code, only polygons are drawn, we may need to check if lines etc need to be drawn
Will add these in the jira.

Terraware App 2022-11-14 06-28-24

PR #750, opened 2022-11-12
SW-2066: Update Navigation & TopBar

joshbuchea:
This PR updates a couple of elements of the `TopBar` on mobile devices:

- use icon-only Terraformation logo
- update user menu icon: smiley face --> first letter of name

### AFTER

localhost_3000_home(iPhone SE) (2)

localhost_3000_home(iPhone SE) (3)

### BEFORE

localhost_3000_home(iPhone SE) (4)

localhost_3000_home(iPhone SE) (5)

### DIFF
--- a/src/components/SmallDeviceUserMenu.tsx
+++ b/src/components/SmallDeviceUserMenu.tsx
@@ -21,16 +21,22 @@ import strings from 'src/strings';
 import { ServerOrganization } from 'src/types/Organization';
 import { User } from 'src/types/User';
 import AddNewOrganizationModal from './AddNewOrganizationModal';
-import { ReactComponent as AvatarIcon } from './avatar-default.svg';
 import Icon from './common/icon/Icon';
 import useEnvironment from 'src/utils/useEnvironment';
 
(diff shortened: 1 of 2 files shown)


src/components/SmallDeviceUserMenu.tsx
```diff
@@ -109,6 +115,7 @@ export default function SmallDeviceUserMenu({
   const [open, setOpen] = React.useState(false);
   const anchorRef = React.useRef<HTMLButtonElement>(null);
   const { isProduction } = useEnvironment();
+  const iconLetter = user?.firstName?.charAt(0) || user?.lastName?.charAt(0);
```
constanzauanini, line 118: Should we use the first letter email if user doesn't have name/lastname ?

joshbuchea: Great question/suggestion @constanzauanini, I think that's probably a good call — will update.

constanzauanini: Thank you!

constanzauanini (approved):

PR #751, opened 2022-11-14
SW-2110 Planting site search capability

constanzauanini:


src/components/PlantingSites/PlantingSitesList.tsx
```diff
@@ -36,15 +39,29 @@ export default function PlantingSitesList(props: PlantingSitesListProps): JSX.El
...
+            type: 'Fuzzy',
+            values: [debouncedSearchTerm],
+          },
+        ],
       },
       count: 0,
     };
     const apiSearchResults = await search(params);
+    if (!debouncedSearchTerm) {
+      setPlantingSites(apiSearchResults);
```
charlie-tf, line 61: If I'm understanding correctly, is `plantingSites` the full list of planting sites when no search term is supplied?

constanzauanini: Yes, we need to save the full list to show or not the empty state, and in other place save the results when search term is applied, to show results in table

src/components/PlantingSites/PlantingSitesList.tsx
```diff
@@ -36,15 +39,29 @@ export default function PlantingSitesList(props: PlantingSitesListProps): JSX.El
...
+        operation: 'and',
+        children: [
+          {
+            field: 'organization_id',
+            operation: 'field',
+            values: [organization.id],
+          },
+          {
+            operation: 'field',
+            field: 'name',
```
karthikbtf, line 51: suggestion for query:
OR group for name or description

karthikbtf (approved):
Thank you!!!

PR #752, opened 2022-11-14
SW-1405 Use collectionSiteNotes search field

sgrimm:
The field `collectionNotes` was renamed to `collectionSiteNotes` for consistency.
The code uses the new name in its request payloads, but was still using the old
name for searches. Both names are supported; switch to the new one so we can remove
the old one on the server side.


karthikbtf (approved):
Thanks for this change.

PR #753, opened 2022-11-14
SW-1414 Remove strings for unused withdrawal purposes

sgrimm:
We removed some out-of-date accession withdrawal purposes, but their names were
still in the strings list; remove them from there too.


sgrimm:
Seems like TypeScript would complain if these were still used anywhere, but to make sure, I did `git grep` for each of them and got no results.

karthikbtf (approved):
Thanks, again.
We most definitely have a bunch of other unused strings in there. Created a jira to clean up strings files.

PR #754, opened 2022-11-14
Allow refreshing mapbox token

karthikbtf:
- use the on(error) callback
- clean up some assumed style

### DIFF
--- a/src/components/Map/PlantingSiteMap.tsx
+++ b/src/components/Map/PlantingSiteMap.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState, useCallback } from 'react';
-import { Box, CircularProgress } from '@mui/material';
+import { Box, CircularProgress, useTheme } from '@mui/material';
 import { getMapboxToken, getPlantingSite } from 'src/api/tracking/tracking';
 import { Geometry, PlantingSite } from 'src/api/types/tracking';
 import useSnackbar from 'src/utils/useSnackbar';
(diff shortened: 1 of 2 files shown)


src/components/Map/PlantingSiteMap.tsx
```diff
@@ -13,10 +13,15 @@ export type PlantingSiteMapProps = {
 
 export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Element {
   const { siteId } = props;
+  const theme = useTheme();
   const [snackbar] = useState(useSnackbar());
   const [token, setToken] = useState<string>();
   const [mapOptions, setMapOptions] = useState<MapOptions>();
 
+  const getFillColor = useCallback(() => {
+    return theme.palette.TwClrBaseWhite || 'white';
```
karthikbtf, line 22: this is still temporary but looks closer to mocks, previous blue/green/yellow was very wrong. the '||' white is because fillColor for map needs to be string, palette color is string | undefined.

src/components/Map/PlantingSiteMap.tsx
```diff
@@ -48,12 +53,12 @@ export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Elemen
           boundary: getPolygons(boundary),
           name: `zone-${name}`,
           id: `zone-${id}`,
-          fillColor: 'blue',
-          fillOpacity: 0.5,
+          fillColor: getFillColor(),
```
constanzauanini, line 56: Are planting site, planting zones and plots all white?

karthikbtf: this is temporary, added a comment in the PR. looking at the designs they all seem white/transparent. i will check with chudo later.

constanzauanini (approved):

PR #759, opened 2022-11-15
Use react-map-gl for mapbox in web-app

karthikbtf:
- use react-mapbox-gl since it is more react friendly and cleaner
- we used this in our previous implementation
- use new coordinates format returned by BE (generated types and removed coordinate parsing into long/lat)

Terraware App 2022-11-14 17-05-22


karthikbtf:
@constanzauanini I was able to fix the build issue. You can now see the plots and no errors in this vercel link.
Looked up the old implementation you had done, looks like we used react-map-gl then.
After some research we should be using react-map-gl in our react app as well. Did that refactor so it is cleaner now and more 'reactive'. This took me most of the day!
If we need any more changes I'd prefer if we iterate on this later since it is still evolving.

src/components/Map/Map.tsx
```diff
@@ -71,49 +113,55 @@ export default function Map(props: MapProps): JSX.Element {
...
+      >
+        {geoData &&
+          (geoData as any[]).map((geo: any, index) => (
+            <Source type='geojson' key={index} data={geo.data}>
+              <Layer {...geo.layer} />
+            </Source>
+          ))}
+        <NavigationControl showCompass={false} style={navControlStyle} position='bottom-right' />
+        {popupInfo && (
+          <Popup
```
constanzauanini, line 155: I remember this from previous implementation, but do we need it here? I don't see popups on the design

constanzauanini:
> @constanzauanini I was able to fix the build issue. You can now see the plots and no errors in this vercel link. Looked up the old implementation you had done, looks like we used react-map-gl then. After some research we should be using react-map-gl in our react app as well. Did that refactor so it is cleaner now and more 'reactive'. This took me most of the day! If we need any more changes I'd prefer if we iterate on this later since it is still evolving.

Thanks Karthik! I didn't remember there was a react-map-gl, but now that I see it, I think same thing happened, we started with the other library and then we changed it (but not for same error, I think). I see there are no more errors now :) 
Thanks, just one question

constanzauanini (approved):

src/components/Map/Map.tsx
karthikbtf: There is a popup in the withdrawal log. I kept popup code so it can be reused later for withdrawal log.
The data will be different. Will add an option now to enable popup.

karthikbtf:
> > @constanzauanini I was able to fix the build issue. You can now see the plots and no errors in this vercel link. Looked up the old implementation you had done, looks like we used react-map-gl then. After some research we should be using react-map-gl in our react app as well. Did that refactor so it is cleaner now and more 'reactive'. This took me most of the day! If we need any more changes I'd prefer if we iterate on this later since it is still evolving.
> 
> Thanks Karthik! I didn't remember there was a react-map-gl, but now that I see it, I think same thing happened, we started with the other library and then we changed it (but not for same error, I think). I see there are no more errors now :) Thanks, just one question

Thanks, yes added option for popup so it can be used later for withdrawal log. We still need more cleanup in the map. I will break down https://terraformation.atlassian.net/browse/SW-2240 into smaller subtasks soon.

PR #762, opened 2022-11-15
SW-2129 Accession View Updates

charlie-tf:
image
image

#### Content area not bleeding all the way to the screen edge on tablet and mobile:
image

#### Skrim update for dialog boxes:
image


karthikbtf (approved):
wow, that's a lot of work. thanks! looks great.

hannah-cai:
some notes:
- the top-level cards should have 24px padding
- the accession ID and common name should both be 14px and tw-clr-txt
- on sub-pages (pages with a back navigation), the primary button size should be small. the buttons should also be aligned with the main header (the scientific name), instead of the accession id
- there should be 24px padding between the top-level cards and the tabs
- there should be 32px padding between the header and the top-level cards (there should always be 32px bottom padding on the header)
- can you remove the hover states on the edit icons? it's a bit odd since there's already a hover state for the cards
- a new decision we made regarding button patterns is that actions under tabs should use both a label and an icon, except on mobile (refer to [here](https://www.figma.com/file/aLIG1C6uIMyBWiTopSluxe/Visual-Refresh?node-id=1212%3A289143&viewport=3496%2C-2523%2C0.35&t=5SSv7V0zUW4U4FXf-11)). this seems to have been applied for "add test," but not for "edit" yet
- the dates in the history tab should be tw-clr-txt-secondary
- on the viability testing tab, the table header still takes up padding, which makes the spacing look odd. could it be hidden completely?
- the table dividers on the viability testing tab should be tw-clr-brdr-secondary
- some updates were made to the back navigation, which I'd forgotten to call out. I added them to the spec [here](https://www.figma.com/file/aLIG1C6uIMyBWiTopSluxe/Visual-Refresh?node-id=1862%3A288110&viewport=192%2C-10991%2C0.51&t=5SSv7V0zUW4U4FXf-11)

charlie-tf:
>* on the viability testing tab, the table header still takes up padding, which makes the spacing look odd. could it be hidden completely?
>* the table dividers on the viability testing tab should be tw-clr-brdr-secondary

I'll have to dig into these when I refresh the tables component

hannah-cai:
thanks for the changes! some notes:
- missed this earlier but the top-level cards should be 24px border radius, not 16
- the 32px bottom padding on the page header is still missing (commented on this in the other PR too)
image

- the edit icon for the location card is offset to the right for some reason
image

everything else looks great! :)

karthikbtf (approved):
Looks so pretty!!
Feel free to wait for Hannah's feedback to latest changes.

hannah-cai:
everything looks good!

PR #764, opened 2022-11-15
SW-2141 Empty routing/component for single batch withdraw flow

karthikbtf:
- added new route `/batch/<batchId>/withdraw`
- added new empty component with correct cancel callback

Terraware App 2022-11-15 11-23-27


constanzauanini (approved):
👍

PR #766, opened 2022-11-15
SW-2267 reduce # of fetches to BE for accession, in accession view

karthikbtf:
- use react state for snackbar - since it was used as a dependency in useEffect, which was being triggered more often than desired
- reduced fetches from 10 to 2

### DIFF
--- a/src/components/accession2/view/Accession2View.tsx
+++ b/src/components/accession2/view/Accession2View.tsx
@@ -94,11 +94,11 @@ export default function Accession2View(props: Accession2ViewProps): JSX.Element 
   const [openViewViabilityTestModal, setOpenViewViabilityTestModal] = useState(false);
   const [selectedTest, setSelectedTest] = useState<ViabilityTest>();
   const [age, setAge] = useState({ value: '', unit: '' });
+  const [snackbar] = useState(useSnackbar());
   const { organization, user } = props;
   const userCanEdit = !isContributor(organization);
   const { isMobile } = useDeviceInfo();
   const classes = useStyles({ isMobile });
-  const snackbar = useSnackbar();
   const themeObj = useTheme();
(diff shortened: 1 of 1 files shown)


karthikbtf:
probably need to check if other places need similar changes

charlie-tf (approved):
Oh interesting. So `snackbar` was getting mutated multiple times, triggering that `useEffect`?

constanzauanini:
> > Oh interesting. So `snackbar` was getting mutated multiple times, triggering that `useEffect`?
> 
> Since it wasn't in the react state - every re-render was creating a new snackbar object - which is a dependency in a useCallback function that fetches the accession (and uses snackbar to show error messages). The useCallback function is a dependency in a useEffect - which calls the useCallback to initialize the accession. Moved that into a react state and noticed way fewer calls. Why this generated that many calls I don't know, wondering if React calls the useCallback/useEffect functions for every change in the dependencies list instead of an optimized call after all dependencies have changed.

Oh, thanks for fixing this Karthik

karthikbtf:
> > > Oh interesting. So `snackbar` was getting mutated multiple times, triggering that `useEffect`?
> > 
> > 
> > Since it wasn't in the react state - every re-render was creating a new snackbar object - which is a dependency in a useCallback function that fetches the accession (and uses snackbar to show error messages). The useCallback function is a dependency in a useEffect - which calls the useCallback to initialize the accession. Moved that into a react state and noticed way fewer calls. Why this generated that many calls I don't know, wondering if React calls the useCallback/useEffect functions for every change in the dependencies list instead of an optimized call after all dependencies have changed.
> 
> Oh, thanks for fixing this Karthik

I think we may have this happening in other places, I know that I added snackbar in many components. Will do a scan of all components later.

PR #732, opened 2022-11-09
SW-2203: Web-App page content lacks left and top padding

joshbuchea:
This PR adjusts content padding.

### AFTER

Screen Shot 2022-11-09 at 10 17 19 AM

Screen Shot 2022-11-09 at 10 17 25 AM

localhost_3000_home(iPhone SE)

localhost_3000_home(iPhone SE) (1)

### BEFORE

screencast 2022-11-09 09-56-59

PR #767, opened 2022-11-15
SW-2262 Accession Table Updates

charlie-tf:
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.8.7",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
-    "@terraware/web-components": "^2.0.6",
+    "@terraware/web-components": "^2.0.8",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.1.0",
     "@testing-library/user-event": "^12.1.10",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -73,7 +73,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     height: '100%',
     overflow: 'auto',
     '& > div, & > main': {
-      paddingBottom: theme.spacing(25),
       paddingTop: '96px',
     },
   },
--- a/src/components/Home.tsx
+++ b/src/components/Home.tsx
@@ -15,7 +15,7 @@ import useDeviceInfo from 'src/utils/useDeviceInfo';
 
 const useStyles = makeStyles((theme: Theme) => ({
   mainContainer: {
-    padding: '20px 0',
+    padding: 0,
   },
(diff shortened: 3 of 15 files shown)


charlie-tf:
Page message style will be updated separately

hannah-cai:
a few quick fixes:
- page message border radius should be 16px
- page header should have 32px bottom padding (currently doesn't when overlaid)
- 200px extra bottom padding on main
image
- there's a bit of extra padding above the filters, I'm not sure what's causing it
- table dividers should be tw-clr-brdr-secondary
- I randomly noticed that if you switch column order, it distorts the column headers 
https://user-images.githubusercontent.com/45954862/202035874-4046442a-b5b6-4512-8d85-c2bc8f7d1bad.mov
- not specced, but something that was brought up in a user feedback session with forestry—would it be possible for the first fields in filters to be automatically selected/clicked into when that filter is open? right now you have to click to open the filter and then click again on the field before you can enter anything.

charlie-tf:
>table dividers should be tw-clr-brdr-secondary
>
>I randomly noticed that if you switch column order, it distorts the column headers
https://user-images.githubusercontent.com/45954862/202035874-4046442a-b5b6-4512-8d85-c2bc8f7d1bad.mov

I can fix the first one when I do the table updates, as well as investigate what is going on with the second one.

hannah-cai:
> > 200px extra bottom padding on main
> 
> @hannah-cai The additional padding was added here to allow for enough space in the Create Accession Page (see [here](https://github.com/terraware/terraware-web/pull/732#issuecomment-1309489029)). It is actually on most of the pages, and we can revisit if it's a problem.

hmm, once we get to visual updates for that page the contents will be put in a [card](https://www.figma.com/file/aLIG1C6uIMyBWiTopSluxe/Visual-Refresh?node-id=916%3A369505&viewport=4755%2C-6299%2C0.51&t=5SSv7V0zUW4U4FXf-11) and there will be extra padding on the bottom, like this:
image
would it be possible to only have extra padding on pages like that? ideally when you hit the bottom on normal pages, there would be equal padding between the bottom and right.
image

charlie-tf:
>not specced, but something that was brought up in a user feedback session with forestry—would it be possible for the first fields in filters to be automatically selected/clicked into when that filter is open? right now you have to click to open the filter and then click again on the field before you can enter anything.

Figure out how to do this and will add to here once I get a web-components PR merged to support it 😁

charlie-tf:
>would it be possible to only have extra padding on pages like that? ideally when you hit the bottom on normal pages, there would be equal padding between the bottom and right.

Yeah that's totally doable.

hannah-cai:
@charlie-tf the changes look great! I think the only thing that's off is that the page header should still have 32px bottom padding even when it's in an overlaid state. right now I think it has no bottom padding
image

PR #770, opened 2022-11-16
SW-2274 Define a popup renderer prop for Map

karthikbtf:
- needed so we can have different rendering passed into the Map
- needed for withdrawal log at some point
- for now enabled a default popup renderer (which we will remove once withdrawal log rendering works using default as example)
- NO changes to map, just refactored popup rendering to a prop


charlie-tf (approved):
Looks good 👍

constanzauanini (approved):
Looks good!

PR #771, opened 2022-11-16
Single seedling batch withdrawal (nursery transfer, dead, other)

constanzauanini:
Outplant is not ready yet (waiting BE)

Uploading image flow missing

Screenshot 2022-11-16 at 10 56 50
Screenshot 2022-11-16 at 10 56 43


karthikbtf (approved):
@constanzauanini this is good for now. 
The more I look at this, we may benefit from an earlier refactor instead of later.

PR #772, opened 2022-11-16
SW-2270 Make snackbar less mutating

karthikbtf:
- snackbar was returning a copy of the functions every time
- this is not needed, there is no scoped data to protect
- return an initialized value that doesn't change


src/utils/useSnackbar.ts
```diff
@@ -2,52 +2,62 @@ import { useSetRecoilState } from 'recoil';
...
+      toastError,
+      pageInfo,
+      pageWarning,
+      pageSuccess,
+      pageError,
+    };
+  }
+
+  if (!messages) {
+    return (messages = SnackbarMessages());
```
karthikbtf, line 59: this was `auto fixed` by `yarn format`
my original code did not have a return

constanzauanini (approved):
👍

PR #775, opened 2022-11-16
SW-2263 Seed Bank Table

charlie-tf:
image

note: i notice the spacing issue w.r.t. the side navigation bar here. I think I may have addressed it in one of my other outstanding PRs though.


karthikbtf (approved):

hannah-cai:
@charlie-tf it looks good! what are the main changes I should be reviewing? was it moving the table into a card?
side note: the hover state on the rows seems to be missing

charlie-tf:
> @charlie-tf it looks good! what are the main changes I should be reviewing? was it moving the table into a card? side note: the hover state on the rows seems to be missing

@hannah-cai Yes, that's the main thing. Pretty small changeset. I'll look into the hover state—it wasn't originally there

charlie-tf:
> > @charlie-tf it looks good! what are the main changes I should be reviewing? was it moving the table into a card? side note: the hover state on the rows seems to be missing
> 
> @hannah-cai Yes, that's the main thing. Pretty small changeset. I'll look into the hover state—it wasn't originally there

Relating to the hover state: this is enabled when the row is clickable. Do you want the entire row to be clickable (thereby opening the seedbank details), or only the link in the first cell?

karthikbtf:
> > > @charlie-tf it looks good! what are the main changes I should be reviewing? was it moving the table into a card? side note: the hover state on the rows seems to be missing
> > 
> > 
> > @hannah-cai Yes, that's the main thing. Pretty small changeset. I'll look into the hover state—it wasn't originally there
> 
> @hannah-cai Relating to the hover state: this is enabled when the row is clickable. Do you want the entire row to be clickable (thereby opening the seedbank details), or only the link in the first cell?

@charlie-tf some orthogonal reference if it matters. Our hover logic may be a bit too restrictive.
`hover` is enabled if clickable and onSelect are defined on the table component (probably based on earlier requirements).
We now have tables where different cells have different click actions - (seedlings/batches table is an example). onSelect and clickable are disabled for such tables (onSelect is not relevant here, clickable sends double the click events).
Some tables - like species table for contributors, are not clickable (read-only view).
If we want clickable table rows to support hover on the row, we may need to detach hover from onSelect/clickable on the table component. Maybe a simple hover=true/false prop.

charlie-tf:
>@charlie-tf some orthogonal reference if it matters. Our hover logic on the table component may be a bit too restrictive.
hover is enabled if clickable and onSelect are defined on the table component (probably based on earlier requirements).
We now have tables where different cells have different click actions - (seedlings/batches table is an example). onSelect and clickable are disabled for such tables (onSelect is not relevant here, clickable sends double the click events).
Some tables - like species table for contributors, are not clickable (read-only view).
If we want clickable table rows to support hover on the row, we may need to detach hover from onSelect/clickable on the table component. Maybe a simple support-hover=true/false prop.

@karthikbtf thanks for that background. That's helpful to know. Hannah said not to worry about the hover thing for now. She's going to bring it up in their design meeting for discussion. I do like your suggestion of decoupling the hover state from the onSelect/clickable logic though.

PR #781, opened 2022-11-17
SW-2287 Add multiple selection logic to seedlings batches table for withdraw

constanzauanini:


src/components/Inventory/view/InventorySeedlingsTable.tsx
```diff
@@ -193,6 +194,48 @@ export default function InventorySeedslingsTable(props: InventorySeedslingsTable
...
   };
 
+  const areAllFromSameNursery = () => {
+    const initialNurseryId = selectedRows[0].facilityId;
+    const otherNursery = selectedRows.some((row) => row.facilityId.toString() !== initialNurseryId.toString());
+    return !otherNursery;
+  };
+
+  const getSelectedRowsAsQueryParams = () => {
+    const queries = selectedRows.map((row, index) => {
```
karthikbtf, line 204: suggestion:
```
const batchIds = selectedRows.map(row = > `batchId=${row.id}`);
return `?${batchIds.join('&')}`;
```

karthikbtf (approved):
looks good, one suggestion (similar to the code in my PR)

src/components/Inventory/view/InventorySeedlingsTable.tsx
constanzauanini: Like it 👍

PR #782, opened 2022-11-18
Update all batches on select purpose form

constanzauanini:


src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -72,33 +67,41 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
-            },
-          ],
-        });
-      } else {
-        snackbar.toastError(response.error);
-      }
+    const fetchBatches = () => {
+      const withdrawalBatches: NurseryWithdrawal[] = [];
+      batchIds.forEach(async (batchId) => {
+        const response = await getBatch(Number(batchId));
```
karthikbtf, line 82: This will be very slow. Best to do one search API with a values field that is a list of all batch ids.
Should this come from the parent component  as a prop `originalBatches` or here is ok? if other components need batches then should come from parent. I think the grid of selecting quantities form needs batches (filtered down by selected nursery).
We may need to maintain a local state of records that is a 1:1 mapping from original batch to a record.
every time we update this local record, we can update the parent record based on a filtered down list of records (based on from nursery).

Basically:
```
Props {
  batches: any [] // not typed since it comes from search
  records;
  setRecords;
};

const [localRecords, setLocalRecords] = useState<>([]);
useEffect() {
   populate localREcords with NurseryWithdrawal
}

// for every change in the form
onChangeXYZ( {
   setLocalRecords(new changes)
   if (fromNursery) { // set records in parent, filtered by from nursery
      setRecords(localRecords.filter(facilityId === fromNursery)
   }
}
```

src/components/Inventory/withdraw/BatchWithdrawFlow.tsx
```diff
@@ -11,15 +15,34 @@ type BatchWithdrawFlowProps = {
 
 export default function BatchWithdrawFlow(props: BatchWithdrawFlowProps): JSX.Element {
   const { organization, batchIds } = props;
-  const [, setFlowState] = useState<FlowStates>('purpose' as unknown as FlowStates);
+  const [flowState, setFlowState] = useState<FlowStates>('purpose');
+  const [record, setRecord] = useForm<NurseryWithdrawal[]>([]);
```
karthikbtf, line 19: plural records and setRecords?

karthikbtf (commented):
This is really awesome!! The approach is great.
Some comments which I think need follow up.

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -176,71 +170,81 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
                     label={strings.FROM_NURSERY}
-                    selectedValue={batch.facilityId.toString()}
-                    options={[getNurseriesById(organization, batch.facilityId)].map((nursery) => ({
+                    selectedValue={record[0].facilityId.toString()}
+                    options={[getNurseriesById(organization, record[0].facilityId)].map((nursery) => ({
                       label: nursery.name,
                       value: nursery.id.toString(),
                     }))}
-                    onChange={(value) => onChange('destinationFacilityId', value)}
+                    onChange={() => true}
```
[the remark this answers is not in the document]
karthikbtf: ignore this comment ^, my earlier comment addresses it with `onChangeXYZ stuff`

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -176,71 +170,81 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
+              </Grid>
+
+              <Grid display='flex'>
                 <Grid item xs={isNurseryTransfer ? 6 : 12} sx={{ marginTop: theme.spacing(2) }} paddingRight={1}>
                   <Dropdown
                     id='from'
                     label={strings.FROM_NURSERY}
-                    selectedValue={batch.facilityId.toString()}
-                    options={[getNurseriesById(organization, batch.facilityId)].map((nursery) => ({
+                    selectedValue={record[0].facilityId.toString()}
```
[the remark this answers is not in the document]
constanzauanini: But aren't they all from the same nursery? If not we don't show the withdraw button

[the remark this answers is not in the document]
karthikbtf: If they come from inventory they won’t all be from same nursery

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -176,71 +182,81 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
-            </Grid>
-            <Grid item xs={12} sx={{ marginTop: theme.spacing(2) }}>
-              <Textfield id='notes' value={record.notes} onChange={onChange} type='textarea' label={strings.NOTES} />
+                {isNurseryTransfer && (
+                  <Grid item xs={6} sx={{ marginTop: theme.spacing(2) }} paddingLeft={1}>
+                    <Dropdown
+                      id='destinationFacilityId'
+                      label={strings.TO_NURSERY_REQUIRED}
+                      selectedValue={firstBatch.destinationFacilityId?.toString()}
+                      options={getAllNurseries(organization).map((nursery) => ({
```
karthikbtf, line 224: Missed this. The list should only be nurseries that the batches belong to.
This feature is primarily for narrowing down nursery for batches coming from inventory page.

karthikbtf: For batches coming from details page this will end up being a list of one nursery

karthikbtf: Hence auto selected / defaulted to that value

src/components/Inventory/withdraw/BatchWithdrawFlow.tsx
```diff
@@ -32,13 +75,13 @@ export default function BatchWithdrawFlow(props: BatchWithdrawFlowProps): JSX.El
 
   return (
     <>
-      {flowState === 'purpose' && (
+      {flowState === 'purpose' && batches && (
```
karthikbtf, line 78: i think we can move the batches check above?

```
if (!batches) {
  return <CircularProgress />
}
```
so we can avoid the batches check on all the components below

src/components/Inventory/withdraw/BatchWithdrawFlow.tsx
```diff
@@ -16,7 +17,49 @@ type BatchWithdrawFlowProps = {
 export default function BatchWithdrawFlow(props: BatchWithdrawFlowProps): JSX.Element {
   const { organization, batchIds } = props;
   const [flowState, setFlowState] = useState<FlowStates>('purpose');
-  const [record, setRecord] = useForm<NurseryWithdrawal[]>([]);
+  const [records, setRecords] = useForm<NurseryWithdrawal[]>([]);
+  const [batches, setBatches] = useState<any[]>();
```
karthikbtf, line 21: 👍 👍

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -78,41 +70,28 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
   useEffect(() => {
     const fetchBatches = () => {
```
karthikbtf, line 71: `populateWithdrawals`?

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -146,25 +125,55 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
+    if (localRecords) {
       const nurseryTransferInvalid = !validateNurseryTransfer();
-      if (fieldsErrors.withdrawDate || nurseryTransferInvalid) {
+      const selectedNurseryInvalid = !validateSelectedNursery();
+      if (fieldsErrors.withdrawDate || nurseryTransferInvalid || selectedNurseryInvalid) {
         return;
       }
     }
 
+    setRecords(localRecords.filter((rcd) => rcd.facilityId.toString() === selectedNursery));
```
karthikbtf, line 153: nice, this is a good place to do this

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -146,25 +125,55 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
 
+  const validateSelectedNursery = () => {
+    if (!selectedNursery) {
+      setIndividualError('fromFacilityId', strings.REQUIRED_FIELD);
+      return false;
+    }
+    return true;
+  };
+
   const saveWithdrawal = async () => {
```
karthikbtf, line 144: `onNext`?

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -146,25 +125,55 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
     }
 
+    setRecords(localRecords.filter((rcd) => rcd.facilityId.toString() === selectedNursery));
+
     onNext();
   };
 
+  const onChangeFromNursery = (facilityIdSelected: string) => {
+    setSelectedNursery(facilityIdSelected);
+    setRecords(records.filter((localR) => localR.facilityId.toString() === facilityIdSelected));
```
karthikbtf, line 160: we can remove this maybe, since you are handling this on line 153

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
constanzauanini: But if we have selected only one batch and we want to do a "Nursery Transfer", would it be "from" the nursery it belongs "to" the nursery it belongs?

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -146,25 +125,55 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
+  };
+
+  const getNurseriesOptions = () => {
+    const nurseriesOptions: DropdownItem[] = [];
+    batches.forEach((bat) => {
```
karthikbtf, line 165: Optional suggestion (since `find` is a sequential search, we can use assoc array). Current code is just as fine.
```
const nurseries = batches.reduce((acc, batch) => {
  if (!acc[batch.facility_id.toString()]) { // this lookup may be faster
     acc[batch.facility_id.toString()] = { label: batch.facility_id, value: batch.facility_name};
  }
  return acc;
}, {});
const options = Object.values(nurseries); // or Object.keys(nurseries).map(id => nurseries[id]);
if (options.length === 1) {
  setSelectedNursery(...)
}
return options;
```

karthikbtf (approved):
This is very nice, matches the requirements. Some suggestions for followup.

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -72,32 +68,30 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
-      const response = await getBatch(Number(batchId));
-      if (response.requestSucceeded && response.batch) {
-        setBatch(response.batch);
-        setRecord({
+    const fetchBatches = () => {
+      const withdrawalBatches: NurseryWithdrawal[] = [];
+      batches.forEach((batch, index) => {
+        const withdrawalBatch: NurseryWithdrawal = {
+          id: -1,
           purpose: 'Out Plant',
```
karthikbtf, line 76: assuming you will change this to something else and hide outplant for contributors

src/components/Inventory/withdraw/BatchWithdrawFlow.tsx
constanzauanini: Good idea

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
constanzauanini: Yes, sorry, this was the old name

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
constanzauanini: Yes, will do it in separate PR

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
karthikbtf: I ended up commenting on the wrong code, I was referring to the list of 'from nurseries' - which you have already fixed. my bad!!

karthikbtf: btw do we may need to filter out the from nursery from this list? we can do that as a follow up

karthikbtf (approved):
🎉 🎉 🎉  looks great, thanks @constanzauanini

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
constanzauanini: Correct, I will do this in a following PR

PR #783, opened 2022-11-18
SW-2264 Species Table Updates

charlie-tf:
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.8.7",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
-    "@terraware/web-components": "^2.0.14",
+    "@terraware/web-components": "^2.0.15",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.1.0",
     "@testing-library/user-event": "^12.1.10",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.8.7",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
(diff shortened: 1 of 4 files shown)


karthikbtf (approved):

hannah-cai:
look good! my only comment is that the "clear" icon (x) in the search bar should be 16px instead of 24.
image

charlie-tf:
> look good! my only comment is that the "clear" icon (x) in the search bar should be 16px instead of 24. image

@hannah-cai We have set both the `TextField` left and right icons to equal the default line height, 24px. Is that something that should change to 16px app-wide for `TextField`s? They also have a default font-size of 16px.

hannah-cai:
> > look good! my only comment is that the "clear" icon (x) in the search bar should be 16px instead of 24. image
> 
> @hannah-cai We have set both the `TextField` left and right icons to equal the default line height, 24px. Is that something that should change to 16px app-wide for `TextField`s? They also have a default font-size of 16px.

I don't think so, when there's a search icon or other icon it should still be 24px. it's just the clear icon that should be smaller

charlie-tf:
> I don't think so, when there's a search icon or other icon it should still be 24px. it's just the clear icon that should be smaller

Okay, I can work with that

hannah-cai:
> > I don't think so, when there's a search icon or other icon it should still be 24px. it's just the clear icon that should be smaller
> 
> Okay, I can work with that

it looks smaller than 16 right now, I think the padding-right is scaling it to be smaller. the info icons on the table are at 16px for reference

charlie-tf:
> > > I don't think so, when there's a search icon or other icon it should still be 24px. it's just the clear icon that should be smaller
> > 
> > 
> > Okay, I can work with that
> 
> it looks smaller than 16 right now, I think the padding-right is scaling it to be smaller. the info icons on the table are at 16px for reference

dang, you are right. I should have done margin-right instead 😓 I'll fix it and put it up to review again when that's done

charlie-tf:
yayy
image

PR #784, opened 2022-11-18
Use tokens for polygon fill/border on map

karthikbtf:
- mapbox requires rgba(r,g,b,a) format.. using hexrgb package. rgba(hex, a) doesn't work
- we could do fillColor and fillOpacity separately but guessing that's not the same? @chudo76 

Terraware App 2022-11-18 09-47-25

Terraware App 2022-11-18 09-47-10


src/components/Map/PlantingSiteMap.tsx
```diff
@@ -114,21 +113,18 @@ export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Elemen
...
-            fillOpacity,
-            lineColor,
-            lineWidth,
+            ...renderAttributes,
             isInteractive: true,
             annotation: {
               textField: 'fullName',
-              textColor: fillColor,
-              textSize: 10, // no tokens available yet
+              textColor: theme.palette.TwClrBaseWhite as string,
```
karthikbtf, line 120: plot label color

constanzauanini: Why are you using `as string` here and not `|| 'white'` as you did in other places?

constanzauanini (approved):
Looks good, just one question

src/components/Map/PlantingSiteMap.tsx
karthikbtf: theme.palette.TwClrBaseWhite can be a string | undefined
mapbox requires a string
i removed the '|| 'white' because it wasn't as clean

karthikbtf: assumption here is that the theme color exists

charlie-tf (approved):
Looks good to me 👍

PR #785, opened 2022-11-18
Remove selected nursery from destination nurseries

constanzauanini:


src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -100,6 +101,14 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
     }));
   };
 
+  useEffect(() => {
+    const allNurseries = getAllNurseries(organization);
+    const destinantionNurseries = allNurseries.filter((nur) => nur.id.toString() !== selectedNursery);
```
karthikbtf, line 106: typo destinationNurseries
also we could rename `nur` to `nursery`? :) more readable..

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -231,10 +240,7 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
                       id='destinationFacilityId'
                       label={strings.TO_NURSERY_REQUIRED}
                       selectedValue={localRecords[0].destinationFacilityId?.toString()}
-                      options={getAllNurseries(organization).map((nursery) => ({
-                        label: nursery.name,
-                        value: nursery.id.toString(),
-                      }))}
+                      options={destinationNurseriesOptions}
```
karthikbtf, line 243: should this be a state or can it be a computed function like we do on line #230 
either is fine, just looking at consistency

karthikbtf (commented):
looks good, some comments

karthikbtf (approved):
Looks good otherwise, feel free to merge after following up 🎉

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
constanzauanini: The problem here is that we need to change the options according to selectedNursery. I don't think the function will execute automatically every time that value change

PR #786, opened 2022-11-18
SW-2265 People Table Updates

charlie-tf:
image
image
* Adding a searchbar is more feature work that I want to put in for this visual update


constanzauanini (approved):
Looks good to me, but maybe wait for Hannah

PR #788, opened 2022-11-18
SW-2279 Nurseries Table

charlie-tf:
image
image

### DIFF
--- a/src/components/Nurseries/index.tsx
+++ b/src/components/Nurseries/index.tsx
@@ -1,8 +1,8 @@
-import { Box, Grid, Typography } from '@mui/material';
+import { Box, Grid, Typography, useTheme } from '@mui/material';
 import { Button, Table, TableColumnType } from '@terraware/web-components';
 import TextField from '@terraware/web-components/components/Textfield/Textfield';
 import { useDeviceInfo } from '@terraware/web-components/utils';
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


src/components/Nurseries/index.tsx
```diff
@@ -95,22 +98,41 @@ export default function NurseriesList({ organization }: NurseriesListProps): JSX
...
-          </Typography>
+      <PageHeaderWrapper nextElement={contentRef.current}>
+        <Grid container paddingBottom={theme.spacing(4)} paddingLeft={isMobile ? 0 : theme.spacing(3)}>
+          <Grid item xs={9} alignItems='center'>
+            <Typography fontSize='24px' fontWeight={600}>
+              {strings.NURSERIES}
+            </Typography>
+          </Grid>
+          <Grid item xs={3} sx={{ textAlign: 'right' }}>
+            {['Admin', 'Owner'].includes(organization.role) &&
```
karthikbtf, line 109: we could use isAdmin from src/utils/organization (captures admin and owner). unrelated to your change for sure.

karthikbtf (approved):
Looks good to me 👍

hannah-cai:
looks good! not sure if this is already being worked on in the other branch, but the icon size is also off here, and the border stroke is a bit cut off
image

charlie-tf:
added margin to keep the searchBar border from being cut off. the icon size is addressed in the other PR and will carry over to here once this is merged.

PR #790, opened 2022-11-20
Manage NavBar Transparency

charlie-tf:
- should be transparent on pages with a background image
- this includes the Home page only for the time being

image

Note for non-Home page, the horizontally scrolled page is concealed by the navbar 
image


src/App.tsx
```diff
@@ -393,7 +393,11 @@ export default function App() {
                   </div>
                 </Slide>
               ) : (
-                <NavBar organization={selectedOrganization} setShowNavBar={setShowNavBar} />
+                <NavBar
+                  organization={selectedOrganization}
+                  setShowNavBar={setShowNavBar}
+                  backgroundTransparent={location.pathname === APP_PATHS.HOME}
```
karthikbtf, line 399: what to capture both `/home` and `/home/`?

karthikbtf (approved):
Woot!
One comment in there.

joshbuchea:
The `NoOrgLandingPage` also has a transparent page background and the page shows when there are no organizations. May want to add `organizations?.length === 0` to the expression in the `backgroundTransparent` prop also.

charlie-tf:
> The `NoOrgLandingPage` also has a transparent page background and the page shows when there are no organizations. May want to add `organizations?.length === 0` to the expression in the `backgroundTransparent` prop also.

@joshbuchea I think that may not be a relevant place. There isn't actually a navbar on that screen if I'm not mistaken? https://github.com/terraware/terraware-web/blob/main/src/App.tsx#L300

src/App.tsx
charlie-tf: changed it to `location.pathname.startsWith(APP_PATHS.HOME)`

constanzauanini (approved):

src/App.tsx
karthikbtf: Nice, that works 👍

joshbuchea:
@charlie-tf you're technically right (the best kind of right) 😄 Please disregard my comment.

joshbuchea:
Ooh however, there are multiple empty state pages that also need the navbar to be transparent. And unfortunately, these pages only need the navbar to be transparent for the empty state but not when there's data. Examples are the empty states of these pages:

- Species
- Planting Sites
- Inventory
- Seed Banks
- Nurseries

I believe there are a few other pages aside from the ones mentioned above.

Here's [an example of one of the pages from the Figma designs](https://www.figma.com/file/aLIG1C6uIMyBWiTopSluxe/Visual-Refresh?node-id=853%3A112604&t=d5nT2hwXnoXig61i-0):

Screen Shot 2022-11-22 at 6 58 19 PM

charlie-tf:
> ...there are multiple empty state pages that also need the navbar to be transparent...

@joshbuchea ah, okay good call. How about we merge this into `main` and then merge main into `feature/visual-updates`. i can then follow up with a PR into the feature branch to take care of this?

joshbuchea:
@charlie-tf if you have a plan/process in mind for this, that sounds good to me 👍 Please let me know if you could use a hand or another set of eyes on anything related to this.

PR #793, opened 2022-11-21
SW-2159 Skrim

charlie-tf:
- apply skrim to dialogs with destructive actions

example:
image


karthikbtf (approved):
seems like a straightforward change

karthikbtf:
> One question — what's a skrim? 🤔

I believe it's the modal backdrop .. which blocks the content behind, we are making it a tad blurred.

joshbuchea:
Thanks @karthikbtf, TIL. I wasn't familiar with the term and also hadn't ever heard it in the context of UI so I had no idea what it was.

PR #794, opened 2022-11-21
SW-2328 Support 'Outplant' in nursery withdrawals

karthikbtf:
- refactored some models
- added sites/zones/plots mappings
- tested with real withdrawals

Terraware App 2022-11-21 15-22-14

### DIFF
--- a/src/api/tracking/tracking.ts
+++ b/src/api/tracking/tracking.ts
@@ -59,16 +59,13 @@ type ListPlantingSitesResponse = {
   error?: string;
 };
 
-export const listPlantingSites = async (full?: boolean): Promise<ListPlantingSitesResponse> => {
+export const listPlantingSites = async (organizationId: number, full?: boolean): Promise<ListPlantingSitesResponse> => {
   const response: ListPlantingSitesResponse = {
     requestSucceeded: true,
   };
 
   try {
-    let endpoint = PLANTING_SITES_ENDPOINT;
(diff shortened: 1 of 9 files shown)


src/api/tracking/tracking.ts
```diff
@@ -59,16 +59,13 @@ type ListPlantingSitesResponse = {
   error?: string;
 };
 
-export const listPlantingSites = async (full?: boolean): Promise<ListPlantingSitesResponse> => {
+export const listPlantingSites = async (organizationId: number, full?: boolean): Promise<ListPlantingSitesResponse> => {
```
karthikbtf, line 62: missed this when I created this function (not used until now)

src/components/Inventory/withdraw/BatchWithdrawFlow.tsx
```diff
@@ -5,7 +5,7 @@ import React, { useEffect, useState } from 'react';
 import strings from 'src/strings';
 import { APP_PATHS } from 'src/constants';
 import { search } from 'src/api/search';
-import { NurseryWithdrawal } from 'src/api/types/batch';
+import { NurseryWithdrawalRequest } from 'src/api/types/batch';
```
karthikbtf, line 8: we were using the response type as the request, fixed that

constanzauanini (approved):
Looking good! Thanks!

PR #797, opened 2022-11-22
Background Updates

charlie-tf:
- fixed background-attachment (background doesn't scroll)
- changed gradient based on design guidance

image
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,6 +3,7 @@ import { CircularProgress, CssBaseline, Slide, StyledEngineProvider, Theme } fro
 import React, { useCallback, useEffect, useState } from 'react';
 import { useHistory } from 'react-router';
 import { Redirect, Route, Switch } from 'react-router-dom';
+import hexRgb from 'hex-rgb';
 import useQuery from './utils/useQuery';
 import useStateLocation, { getLocation } from './utils/useStateLocation';
 import { getOrganizations } from 'src/api/organization/organization';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -66,7 +67,13 @@ interface StyleProps {
 
 const useStyles = makeStyles((theme: Theme) => ({
   container: {
-    background: 'linear-gradient(180deg, #FBF9F9 0%, #EFF5EF 100%)',
+    backgroundColor: theme.palette.TwClrBaseGray025,
+    backgroundImage:
+      'linear-gradient(180deg,' +
+      `${hexRgb(`${theme.palette.TwClrBaseGreen050}`, { alpha: 0, format: 'css' })} 0%,` +
+      `${hexRgb(`${theme.palette.TwClrBaseGreen050}`, { alpha: 0.4, format: 'css' })} 100%)`,
```
karthikbtf, line 74: for some reason I couldn't get this to work well, I had to use format: 'object' and create a string from that. See https://github.com/terraware/terraware-web/blob/main/src/components/Map/PlantingSiteMap.tsx#L29 let me know if I'm doing something wrong?

karthikbtf (approved):

src/App.tsx
charlie-tf: Hmm, `hexRgb(`${theme.palette.TwClrBaseGreen050}`, { alpha: 0.4, format: 'css' })` generates `rgb(214 253 229 / 40%)` in my code here.

I think I did try using `theme.palette.TwClrBaseGreen050 as string` but the typechecker didn't like that, so I reverted to the `` `${...}` `` format instead. If I recall correctly.

PR #798, opened 2022-11-22
SW-2330 Format success message for batch withdrawals

karthikbtf:
- some singular/plural adjustment which we should ideally get from a 3rd party library

Screen Shot 2022-11-22 at 6 36 40 AM

Screen Shot 2022-11-22 at 6 38 23 AM

Screen Shot 2022-11-22 at 6 37 50 AM

### DIFF
--- a/src/components/Inventory/withdraw/BatchWithdrawFlow.tsx
+++ b/src/components/Inventory/withdraw/BatchWithdrawFlow.tsx
@@ -5,7 +5,7 @@ import React, { useEffect, useState } from 'react';
 import strings from 'src/strings';
 import { APP_PATHS } from 'src/constants';
 import { search } from 'src/api/search';
-import { NurseryWithdrawalRequest } from 'src/api/types/batch';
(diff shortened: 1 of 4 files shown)


src/components/Inventory/withdraw/BatchWithdrawFlow.tsx
```diff
@@ -118,11 +118,36 @@ export default function BatchWithdrawFlow(props: BatchWithdrawFlowProps): JSX.El
...
+      if (readyQuantityWithdrawn) {
+        total += readyQuantityWithdrawn;
+      }
+      return total;
+    }, 0);
+
+    return strings.formatString(
+      strings.BATCH_WITHDRAW_SUCCESS,
+      numBatches,
+      numBatches === 1 ? strings.BATCHES_SINGULAR : (strings.BATCHES_PLURAL as any),
```
constanzauanini, line 145: why is that “as any” needed?

karthikbtf: Typescript was expecting formatted number object

PR #802, opened 2022-11-22
SW-2136 Organization View

charlie-tf:
image
image

### DIFF
--- a/src/components/Organization/index.tsx
+++ b/src/components/Organization/index.tsx
@@ -74,15 +74,15 @@ export default function OrganizationView({ organization }: OrganizationViewProps
 
   return (
     <TfMain>
-      <Box margin={theme.spacing(0, 0, 4, 3)}>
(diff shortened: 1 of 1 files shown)


hannah-cai:
looks good! just two notes:
- this element should have 24px right padding as well
image

- not sure why all secondary buttons seem to have an overlay even in their default state. they should only have an overlay on hover or when pressed. if this is wip then disregard
image

hannah-cai:
another thing I noticed (not directly related) is that the left nav selected states aren't working properly, it's not indicating which tab you're currently on

https://user-images.githubusercontent.com/45954862/203416404-7db52d71-20f3-42f0-93b0-04404a125f1b.mov

joshbuchea (approved):

joshbuchea:
@hannah-cai the nav selected/active state was working previously, but must have stopped working at some point. The changes are unrelated to this PR, but I can take a look at the issue separately 👀

PR #804, opened 2022-11-22
SW-2136 Person View

charlie-tf:
image
image


hannah-cai:
when there's a page message in the header, it should go to the outer margin (to the edges of the white card). also unsure if this is on another branch, but the border radius should be 16px for page messages
image

joshbuchea (approved):
Changes are clean

hannah-cai:
commented on another branch, but padding between fields when they're stacked (eg in mobile view) should be 16px. currently it's 32px here

PR #807, opened 2022-11-23
SW-2351 Fix rendering performance issues on planting site maps

karthikbtf:
- Code was overzealous creating Mapbox Layers
- One layer per plot was being created.. this was not necessary
- One layer encompassing all plots should be the correct approach
- One layer per rendering style is the purpose of a Layer
- With large number of layers, mapbox was slow in rendering and creating tileset data on the fly


constanzauanini (approved):
Tested on Vercel with Alex's organization! Thanks!

PR #813, opened 2022-11-23
SW-2358 Adjust image thumbnail display style (#810)

karthikbtf:
- scale image
- fit image in parent box
- center image
- as per design changes


constanzauanini (approved):

karthikbtf:
this is meant to amend the commit message only.. but looks like this could mess up the commit history if someone had already rebased. closing for now, not worth the correction

PR #814, opened 2022-11-23
SW-2261 Update map polygon types with recent backend changes that removes points/lines, etc.

karthikbtf:
- MultiPolygon alone is supported
- to avoid ambiguity


karthikbtf:
@constanzauanini you will need this change if you end up running `yarn generate-types` since BE made some changes to geometry/polygon types in the API (cleaned up).

constanzauanini (approved):
Thanks 🙏

PR #818, opened 2022-11-24
SW-2314 Add title and dropdown of planting sites to select current planting site

constanzauanini:
Desktop:
Screenshot 2022-11-24 at 11 43 12

Mobile:
Screenshot 2022-11-24 at 11 48 17

### DIFF
--- a/src/components/Plants/index.tsx
+++ b/src/components/Plants/index.tsx
@@ -1,17 +1,55 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { ServerOrganization } from 'src/types/Organization';
 import strings from 'src/strings';
 import TfMain from 'src/components/common/TfMain';
-import Title from 'src/components/common/Title';
+import { Box, Grid, Typography, useTheme } from '@mui/material';
+import { Select } from '@terraware/web-components';
+import { listPlantingSites } from 'src/api/tracking/tracking';
+import { PlantingSite } from 'src/api/types/tracking';
+import { useDeviceInfo } from '@terraware/web-components/utils';
 
 type PlantsDashboardProps = {
   organization: ServerOrganization;
 };
 
 export default function PlantsDashboard(props: PlantsDashboardProps): JSX.Element {
+  const { organization } = props;
+  const [selectedPlantingSite, setSelectedPlantingSite] = useState<PlantingSite>();
+  const [plantingSites, setPlantingSites] = useState<PlantingSite[]>([]);
+  const theme = useTheme();
+  const { isMobile } = useDeviceInfo();
+
+  useEffect(() => {
+    const populatePlantingSites = async () => {
+      const serverResponse = await listPlantingSites(organization.id);
+      if (serverResponse.requestSucceeded) {
+        setPlantingSites(serverResponse.sites ?? []);
+      }
+    };
+    populatePlantingSites();
+  }, [organization]);
+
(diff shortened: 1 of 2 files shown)


src/components/Plants/index.tsx
```diff
@@ -1,17 +1,55 @@
...
+  const [plantingSites, setPlantingSites] = useState<PlantingSite[]>([]);
+  const theme = useTheme();
+  const { isMobile } = useDeviceInfo();
+
+  useEffect(() => {
+    const populatePlantingSites = async () => {
+      const serverResponse = await listPlantingSites(organization.id);
+      if (serverResponse.requestSucceeded) {
+        setPlantingSites(serverResponse.sites ?? []);
+      }
```
karthikbtf, line 27: else toast error

src/components/Plants/index.tsx
```diff
@@ -1,17 +1,55 @@
...
+    };
+    populatePlantingSites();
+  }, [organization]);
```
karthikbtf, line 30: organization.id should be sufficient

src/components/Plants/index.tsx
```diff
@@ -1,17 +1,55 @@
...
+  const [selectedPlantingSite, setSelectedPlantingSite] = useState<PlantingSite>();
+  const [plantingSites, setPlantingSites] = useState<PlantingSite[]>([]);
+  const theme = useTheme();
+  const { isMobile } = useDeviceInfo();
+
+  useEffect(() => {
+    const populatePlantingSites = async () => {
+      const serverResponse = await listPlantingSites(organization.id);
+      if (serverResponse.requestSucceeded) {
+        setPlantingSites(serverResponse.sites ?? []);
```
karthikbtf, line 26: We should default to the first one - if the url does not have a planting site id in the path.
Otherwise match it to the planting site id in the url path if one exists.
We should do the same thing we do for monitoring.

src/components/Plants/index.tsx
```diff
@@ -1,17 +1,55 @@
...
+      }
+    };
+    populatePlantingSites();
+  }, [organization]);
+
+  const onChangePlantingSite = (newValue: string) => {
+    setSelectedPlantingSite(plantingSites.find((ps) => ps.name === newValue));
```
karthikbtf, line 33: ideally set it in the url path, have another useEffect that runs off the url path param to set the selected planting site

karthikbtf (commented):
looks good, a few comments

src/components/Plants/index.tsx
constanzauanini: I didn't add this change because I was getting an infinit loop

karthikbtf: you can do `const snackbar = useState(useSnackbar());`

src/components/Plants/index.tsx
```diff
@@ -1,17 +1,85 @@
...
+    [history]
+  );
+
+  useEffect(() => {
+    const populatePlantingSites = async () => {
+      const serverResponse = await listPlantingSites(organization.id);
+      if (serverResponse.requestSucceeded) {
+        setPlantingSites(serverResponse.sites ?? []);
+      } else {
+        snackbar.toastError();
```
karthikbtf, line 43: you can do `snackbar.toastError(serverResponse.error)` if we want to show more details on the error

PR #820, opened 2022-11-24
SW-2055 Replace usage of React logo in terraware

constanzauanini:


karthikbtf (approved):
Looks good, can you add a description on why this change - just for context?

PR #825, opened 2022-11-28
SW-2278 Show plot labels only after zoomed in

karthikbtf:
- avoid showing during the zoom process to avoid labels dancing around
- to test, view difference with vercel link versus staging

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -49,6 +49,7 @@ export default function Map(props: MapProps): JSX.Element {
     // fit to bounding box
     if (mapRef?.current !== undefined) {
       const mapInstance: any = mapRef.current.getMap();
+      setZoomEnd(false);
       setZoomBegin(true);
       mapInstance.fitBounds([bbox.lowerLeft, bbox.upperRight], { padding: 20 });
     }


src/components/Map/Map.tsx
```diff
@@ -40,13 +40,16 @@ export default function Map(props: MapProps): JSX.Element {
...
+  const [zoomBegin, setZoomBegin] = useState<boolean>(false);
+  const [zoomEnd, setZoomEnd] = useState<boolean>(false);
   const mapRef = useRef<any | null>(null);
 
   const onMapLoad = useCallback(() => {
     const { bbox } = options;
     // fit to bounding box
     if (mapRef?.current !== undefined) {
       const mapInstance: any = mapRef.current.getMap();
+      setZoomBegin(true);
```
charlie-tf, line 52: should we also `setZoomEnd(false)` here?

charlie-tf: Oh I do see now that this should only run on map load and that zoomEnd is initialized to false 🤦 Apologies. I guess it was not clear to me when zoomEnd is ever set to false (other than when initialized). Is there not an "onZoomBegin" handler to mirror the "onZoomEnd" one?

constanzauanini (approved):
Is this change only for the first loading zoom right? Because when zooming in or out with the control, I don't see any change

karthikbtf:
> Is this change only for the first loading zoom right? Because when zooming in or out with the control, I don't see any change

Yes only on the zoom when the labels are dancing around.

src/components/Map/Map.tsx
karthikbtf: Your comment was totally valid, explicitly setting zoomEnd made the code readable and also made the assumptions very clear.

PR #828, opened 2022-11-28
Refactor plants dashboard components to be cleaner

constanzauanini:


src/components/Plants/index.tsx
```diff
@@ -151,69 +60,10 @@ export default function PlantsDashboard(props: PlantsDashboardProps): JSX.Elemen
...
     setActivePlantingSite(plantingSites.find((ps) => ps.name === newValue));
   };
 
-  const onChangePlot = (newValue: string) => {
-    if (plots) {
-      setSelectedPlot(plots.find((p) => p.fullName === newValue));
-    }
-  };
-
   return (
```
karthikbtf, line 67: can we add a spinner while we are listing planting sites?

karthikbtf (approved):
Thank you so much for doing this 🙏 
This is cleaner and better separation of code/responsibility.

src/components/Plants/index.tsx
```diff
@@ -73,12 +73,18 @@ export default function PlantsDashboard(props: PlantsDashboardProps): JSX.Elemen
...
+              options={plantingSites.map((ps) => ps?.name || '')}
+              onChange={onChangePlantingSite}
+              selectedValue={selectedPlantingSite?.name}
+              placeholder={strings.SELECT}
+            />
+          ) : (
+            <Box sx={{ position: 'fixed', top: '50%', left: '50%' }}>
+              <CircularProgress />
+            </Box>
+          )}
```
karthikbtf: would be more readable to have this outside - also  will avoid mistakes like showing planting sites details in this case when it should't show

```
  if (!plantingSites) {
   return (
     <Box>
      spinner
   </box>
   )
  }

return (
   normal content
);

```

karthikbtf (approved):
Thanks 🙏

PR #829, opened 2022-11-29
SW-2318 Add site geographical map to plants dashboard

karthikbtf:
- this does not cover the click/popup changes, separate jira
- renamed some variables/classes
- added a styles override to the map (for border radius)
- show no map object if there's no boundary

Terraware App 2022-11-28 16-37-37

Terraware App 2022-11-28 16-37-13


src/components/Plants/PlantingSiteDashboardMap.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Box } from '@mui/material';
+import { PlantingSite } from 'src/api/types/tracking';
+import { getPlantingSite } from 'src/api/tracking/tracking';
+import useSnackbar from 'src/utils/useSnackbar';
+import { PlantingSitesPlots } from './PlantingSiteDetails';
+import { PlantingSiteMap } from 'src/components/Map';
+
+type PlantingSiteDashboardMapProps = {
+  siteId?: number;
+  plots?: PlantingSitesPlots[];
```
constanzauanini, line 11: Looks like we don't need to pass the plots

src/components/Plants/PlantingSiteDetails.tsx
```diff
@@ -89,12 +95,14 @@ export default function PlantingSiteDetails(props: PlantingSiteDetailsProps): JS
...
     populateResults();
-  }, [selectedPlantingSite]);
+  }, [plantingSite]);
 
   return (
     <Grid container display='flex' marginTop={6}>
-      <PlantingSiteMap plots={plots} />
-      <Grid item xs={isMobile ? 12 : 6} sx={{ paddingLeft: 1 }}>
+      <Grid item xs={isMobile ? 12 : 6} sx={mapCardStyle}>
+        <PlantingSiteDashboardMap plots={plots} siteId={plantingSite?.id} />
```
constanzauanini, line 103: Why don't we pass the plantingSite object instead of the id, so we don't need to fetch the planting site again in the PlantingSiteDashboardMap component?

src/components/Plants/index.tsx
```diff
@@ -68,9 +68,9 @@ export default function PlantsDashboard(props: PlantsDashboardProps): JSX.Elemen
 
   if (!plantingSites) {
     return (
-      <Box sx={{ position: 'fixed', top: '50%', left: '50%' }}>
-        <CircularProgress />
-      </Box>
+      <TfMain>
```
constanzauanini, line 71: Oh, thanks for doing this! I will close mi PR fixing this

src/components/Plants/PlantingSiteDashboardMap.tsx
karthikbtf: we need this for the tooltip which is a followup

src/components/Plants/PlantingSiteDetails.tsx
karthikbtf: yes, good point. not sure why i did it this way.

karthikbtf: oh wait, yes, we don't get the polygon info here

karthikbtf: that's why fetching full details of the planting site

karthikbtf: We need to fetch the planting site here to get the full details of zones and plots with polygons.
The planting site information at the parent level is shallow (no zones/plots) and we should keep it that way for performance.

constanzauanini: Oh, you mean the planting zones?

karthikbtf: Yes the zones and plots with polygon info. To show on map

karthikbtf: When we list all planting sites in the parent we can theoretically fetch all the info for every planting site. That’s one option but it will be expensive.

PR #831, opened 2022-11-29
SW-2322 Pick colors for the bar chart from base color 300s

constanzauanini:
Screenshot 2022-11-29 at 11 16 03

### DIFF
--- a/src/utils/generateRandomColor.tsx
+++ b/src/utils/generateRandomColor.tsx
@@ -1,3 +1,17 @@
+import { Theme } from '@mui/material';
+
+const terrawareColors = (theme: Theme) => [
+  theme.palette.TwClrBaseYellow300,
+  theme.palette.TwClrBaseRed300,
+  theme.palette.TwClrBaseBlue300,
+  theme.palette.TwClrBaseGreen300,
+  theme.palette.TwClrBasePurple300,
+  theme.palette.TwClrBaseOrange300,
+  theme.palette.TwClrBasePink300,
+  theme.palette.TwClrBaseGray300,
+  theme.palette.TwClrBaseTeal300,
+];
+
 const generateRandomColor = () => {
   return `#${Math.floor(Math.random() * 16777215).toString(16)}`;
 };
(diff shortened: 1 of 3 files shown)


src/utils/generateRandomColor.tsx
```diff
@@ -9,4 +23,14 @@ const generateRandomColors = (numberOfColors: number) => {
   return colors;
 };
 
-export { generateRandomColor, generateRandomColors };
+const generateTerrawareRandomColors = (theme: Theme, numberOfColors: number) => {
+  const terrawareColorsList = terrawareColors(theme);
+  const colors = Array.from({ length: numberOfColors }).map((_, i) => {
+    const randomIndex = Math.floor(Math.random() * terrawareColorsList.length);
+    const selectedColor = terrawareColorsList.splice(randomIndex, 1);
+    return selectedColor[0] || generateRandomColor();
```
constanzauanini, line 31: if numberOfColors is bigger than terrawareColors' length, we return a random color

karthikbtf: We need to cycle through colors, no need to use random.
if there are 10 colors.. and 12 bars, we cycle through colors and repeat.
ie.
bar1 - color1
bar2 - color2
bar3 - color3...
bar10 - color10
bar11 - color1
bar 12 - color2

constanzauanini: Oh, ok, but we will have repeated colors. Also between the two chart we will get same colors for the bars

karthikbtf: That was my understanding from Rachel. We can check with her if needed.
In this current PR, is it possible to have 2 bars next to each other with the same color?

karthikbtf (approved):
Looks good otherwise, feel free to follow up with Rachel.

src/utils/generateRandomColor.tsx
constanzauanini: It could be possible if generateRandomColor() returns same color twice, but low probability

karthikbtf: Ok we can get this in maybe and ask Rachel in the Jira, follow up later if needed.

PR #833, opened 2022-11-29
SW-2375 Save last visited planting site and plot in user preferences

constanzauanini:


charlie-tf (approved):
Looks good!

src/components/Plants/SpeciesByPlotChart.tsx
```diff
@@ -26,12 +29,17 @@ export default function SpeciesByPlotChart({ plots }: Props): JSX.Element {
...
     }
   };
 
   React.useEffect(() => {
-    setSelectedPlot(undefined);
-  }, [plots]);
+    if (lastPlot) {
+      setSelectedPlot(plots?.find((p) => p.fullName === lastPlot));
+    } else {
+      setSelectedPlot(undefined);
```
karthikbtf, line 40: default to the first plot here or if we don't find the lastPlot in the 'if' above.
also should we be checking for id instead of fullName?
```
  if (!plots?.length) {
    return;
  }
  const plot = plots.find(p => p.id === lastPlot);
  setSelectedPlot(plot || plots[0]);
```

src/components/Plants/index.tsx
```diff
@@ -48,17 +50,31 @@ export default function PlantsDashboard(props: PlantsDashboardProps): JSX.Elemen
...
-        setActivePlantingSite(plantingSiteToUse);
+    const initializePlantingSite = async () => {
+      if (plantingSites && plantingSites.length) {
+        let lastDashboardPlantingSite: any = {};
+        const response = await getPreferences(organization.id);
+        if (response.requestSucceeded && response.preferences?.lastDashboardPlantingSite) {
+          lastDashboardPlantingSite = response.preferences.lastDashboardPlantingSite;
+        }
+        const plantingSiteIdToUse = plantingSiteId || lastDashboardPlantingSite.plantingSiteId;
+        const requestedPlantingSite = plantingSites.find((ps) => ps?.id === parseInt(plantingSiteIdToUse, 10));
```
karthikbtf, line 61: rename `ps` to `plantingSite`?

src/components/Plants/SpeciesByPlotChart.tsx
```diff
@@ -26,12 +29,17 @@ export default function SpeciesByPlotChart({ plots }: Props): JSX.Element {
   const onChangePlot = (newValue: string) => {
     if (plots) {
       setSelectedPlot(plots.find((p) => p.fullName === newValue));
+      updatePlotPreferences(newValue);
```
karthikbtf, line 32: use the plot id
```
if (plots) {
  const plot = plots.find(p => p.fullName === newValue);
  setSeletedPlot(plot);
  updatePlotPreferences(plot.id)
}
```
also, can we use the id instead of fullname for the onChangePlot value?
We can use the fullname for the label

karthikbtf (commented):
looks good, some comments

karthikbtf (approved):
Thanks 🎉

PR #835, opened 2022-11-29
NavBar Updates

charlie-tf:
- update padding
- update background and scrollbar thumb styles
- add transparent background for empty states:
  - species, seedbanks, nurseries, planting-sites

color and padding around scrollbar:
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -61,6 +61,8 @@ import {
   SpeciesBulkWithdrawWrapperComponent,
 } from './components/Inventory/withdraw';
 import PlantsDashboard from './components/Plants';
+import { listPlantingSites } from './api/tracking/tracking';
+import { PlantingSite } from './api/types/tracking';
 
 interface StyleProps {
   isDesktop?: boolean;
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -206,6 +219,22 @@ export default function App() {
     reloadSpecies();
   }, [reloadSpecies]);
 
+  const reloadTracking = useCallback(() => {
```
karthikbtf, line 222: are these from merging main into visual-updates?

karthikbtf (approved):
Thanks 🎉 finally!!! The navbar scrollbar squeeze has been a big pain.

src/App.tsx
charlie-tf: No, I added those myself. I needed to add set the navbar background to transparent for the planting-sites page empty state, but if one is added I needed to trigger a refresh to set it to not be transparent (I'll need to do an update of the planting-sites list as currently it shows a background image, but shouldn't, when non-empty)

karthikbtf: Got it 👍

PR #837, opened 2022-11-29
SW-2384/SW-2385 Fix fuzzy searches that were using empty strings

karthikbtf:
- in inventory seedlings batches and nurseries tables
- this is expensive and not ideal


constanzauanini (approved):
Looks good! Tested on staging 👍

PR #843, opened 2022-11-30
SW-2383 Show small values in chart when the range of values is wide

constanzauanini:
Add minBarLength to ensure all values (including very small ones) draw a bar in the chart

### DIFF
--- a/src/components/Plants/DashboardChart.tsx
+++ b/src/components/Plants/DashboardChart.tsx
@@ -58,11 +58,7 @@ export default function DashboardChart(props: DashboardChartProps): JSX.Element 
           scales: {
             y: {
               ticks: {
-                callback: (value, index, ticks) => {
-                  if (+value % 1 === 0) {
-                    return value;
-                  }
-                },
+                precision: 0,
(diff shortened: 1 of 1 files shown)


karthikbtf:
Update the PR title and description if you can?

src/components/Plants/DashboardChart.tsx
```diff
@@ -35,6 +35,7 @@ export default function DashboardChart(props: DashboardChartProps): JSX.Element
               data: chartValues,
               barThickness: 50, // number (pixels) or 'flex'
               backgroundColor: colors,
+              minBarLength: 3,
```
karthikbtf, line 38: that was it? nice :)

constanzauanini: Yes :(  after fighting a lot with the code, i realize there was this prop :(

karthikbtf: works in vercel 🎉 👏

karthikbtf (approved):

PR #844, opened 2022-11-30
SW-2401 / SW-2398 / SW-2399 Empty state updates

karthikbtf:
- show sorted plots that only have plants
- show map without polygons if site has no polygons
- show disabled plot dropdown when there are no plots

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -47,9 +47,13 @@ export default function Map(props: MapProps): JSX.Element {
   const mapRef = useRef<any | null>(null);
 
   const onMapLoad = useCallback(() => {
-    const { bbox } = options;
-    // fit to bounding box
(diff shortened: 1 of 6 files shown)


constanzauanini:
I was testing on staging and I see that Number of Plants by Plot and Species chart is not changing when changing planting site

karthikbtf:
> I was testing on staging and I see that Number of Plants by Plot and Species chart is not changing when changing planting site

because of this change?

karthikbtf:
> > I was testing on staging and I see that Number of Plants by Plot and Species chart is not changing when changing planting site
> 
> because of this change?

looking

karthikbtf:
> I was testing on staging and I see that Number of Plants by Plot and Species chart is not changing when changing planting site

fixed now

constanzauanini (approved):
Lgtm!

PR #848, opened 2022-11-30
SW-2397 / SW-2405 More empty state updates for plants dashboard 

karthikbtf:
- set aspect ratio of map and charts
- set height of charts
- don't show plot/species chart if site has no plots
- fix some margin bottom issues

Terraware App 2022-11-30 15-33-00

Terraware App 2022-11-30 15-28-36

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -81,6 +81,10 @@ const useStyles = makeStyles((theme: Theme) => ({
       paddingBottom: theme.spacing(25),
       paddingTop: '96px',
     },
+    '& > .moreScreen': {
+      paddingBottom: theme.spacing(10),
+      paddingTop: '96px',
+    },
   },
   contentWithNavBar: {
     '& > div, & > main': {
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -81,6 +81,10 @@ const useStyles = makeStyles((theme: Theme) => ({
       paddingBottom: theme.spacing(25),
       paddingTop: '96px',
     },
+    '& > .moreScreen': {
```
karthikbtf, line 84: @joshbuchea I added this as a way to override the 200px margin bottom. Hoping this looks ok.

src/components/PlantingSites/PlantingSiteCreate.tsx
```diff
@@ -129,7 +129,7 @@ export default function CreatePlantingSite(props: CreatePlantingSiteProps): JSX.
   };
 
   return (
-    <TfMain>
+    <TfMain moreScreen={true}>
```
constanzauanini, line 132: I think I missed in visual changes what this moreScreen prop does?

karthikbtf: Every page with TfMain has 200px bottom padding which is a lot. This option reduces bottom padding to 80px so we can show a taller map

constanzauanini: Thanks for the explanation!

PR #851, opened 2022-12-01
SW-2300 Add tooltip for disabled quantity field in the accession detail page

constanzauanini:
Screenshot 2022-12-01 at 12 04 20

### DIFF
--- a/src/strings/index.tsx
+++ b/src/strings/index.tsx
@@ -872,6 +872,7 @@ const strings = new LocalizedStrings({
     NO_SEEDLINGS_AVAILABLE_TO_OUTPLANT_DESCRIPTION:
       'There are no batches selected with seedlings that are "ready." Select another withdraw purpose or cancel and select a different batch.',
     OUTPLANTS_REQUIRE_READY_SEEDLINGS: 'Outplants require batches with seedlings that are "ready."',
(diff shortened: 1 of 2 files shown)


src/strings/index.tsx
```diff
@@ -872,6 +872,7 @@ const strings = new LocalizedStrings({
     NO_SEEDLINGS_AVAILABLE_TO_OUTPLANT_DESCRIPTION:
       'There are no batches selected with seedlings that are "ready." Select another withdraw purpose or cancel and select a different batch.',
     OUTPLANTS_REQUIRE_READY_SEEDLINGS: 'Outplants require batches with seedlings that are "ready."',
+    EDIT_QUANTITY_DISABLED: 'Currently, you may only input a quantity when the status is "Drying" or “In Storage”.',
```
karthikbtf, line 875: as per @chudo76 's earlier comments, the "." should be within the quotes.
or "In Storage."

karthikbtf (approved):

src/strings/index.tsx
constanzauanini: 😮

joshbuchea:
This was confusing me to as a new user, so I'm glad to see it get fixed 🙂 🎉

PR #854, opened 2022-12-01
SW-2410 Planting site map should start with the site boundary

karthikbtf:
- use `initialViewState` set to bbox
- remove zoom to boundary code

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { Box } from '@mui/material';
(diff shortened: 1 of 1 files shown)


constanzauanini (commented):
This looks good! Just one thing I noticed when I was testing, is that for a few seconds the map is the generic map and then, when bounds are loaded, it changes to the other map. This is not related to this PR but now that zoom is not going in and out, i can notice it more

constanzauanini (approved):

karthikbtf:
> This looks good! Just one thing I noticed when I was testing, is that for a few seconds the map is the generic map and then, when bounds are loaded, it changes to the other map. This is not related to this PR but now that zoom is not going in and out, i can notice it more

Will follow up on this separately.

karthikbtf:
- use initialViewState set to bbox
- remove zoom to boundary code

PR #855, opened 2022-12-01
SW-2407 Routing and empty components for withdrawal log

constanzauanini:


charlie-tf (approved):

src/components/NavBar.tsx
```diff
@@ -125,6 +126,14 @@ export default function NavBar({
           }}
           id='inventory'
         />
+        <NavItem
```
karthikbtf, line 129: need to gate with `trackingEnabled && `

constanzauanini: Good catch!

karthikbtf (approved):
🎉

PR #857, opened 2022-12-01
Allow access to admin UI without trailing slash

sgrimm:
Currently, Nginx is configured to pass requests whose URL paths start with
`/admin/` through to terraware-server. But people sometimes forget the trailing
slash, and `/admin` doesn't match the passthrough rule so they get redirected to
the web app's home page.

Update the rule so that it matches URL paths that start with `/admin` if there
are no more characters in the path or if the next character is `/` or `?`. That
way if we add, say, `/administration` as a path, it won't match the rule.


karthikbtf (approved):
Thanks!

PR #859, opened 2022-12-02
SW-2412 Initial withdrawal log table view with withdrawal-id link and reassign button

constanzauanini:
Screenshot 2022-12-02 at 11 43 51

### DIFF
--- a/src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
+++ b/src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
@@ -1,10 +1,119 @@
+import { Box, Grid, Theme, Typography, useTheme } from '@mui/material';
+import TextField from '@terraware/web-components/components/Textfield/Textfield';
+import { useEffect, useRef, useState } from 'react';
+import strings from 'src/strings';
 import { ServerOrganization } from 'src/types/Organization';
+import PageHeaderWrapper from '../common/PageHeaderWrapper';
 import TfMain from '../common/TfMain';
+import { makeStyles } from '@mui/styles';
+import { Table, TableColumnType } from '@terraware/web-components';
+import { listNurseryWithdrawals } from 'src/api/tracking/withdrawals';
+import { SearchResponseElement } from 'src/api/search';
+import WithdrawalLogRenderer from './WithdrawalLogRenderer';
+import { APP_PATHS } from 'src/constants';
+import { useHistory } from 'react-router-dom';
 
 type NurseryWithdrawalsProps = {
   organization: ServerOrganization;
 };
 
+const useStyles = makeStyles((theme: Theme) => ({
+  searchField: {
+    width: '300px',
+  },
+}));
+
+const columns: TableColumnType[] = [
+  { key: 'withdrawnDate', name: strings.DATE, type: 'string' },
+  { key: 'purpose', name: strings.PURPOSE, type: 'string' },
+  { key: 'facility_name', name: strings.FROM_NURSERY, type: 'string' },
+  { key: 'destinationName', name: strings.DESTINATION, type: 'string' },
+  { key: 'plotNames', name: strings.TO_PLOT, type: 'string' },
+  { key: 'speciesScientificNames', name: strings.SPECIES, type: 'string' },
+  { key: 'totalWithdrawn', name: strings.TOTAL_QUANTITY, type: 'string' },
+  { key: 'hasReassignments', name: '', type: 'string' },
+];
+
 export default function NurseryWithdrawals(props: NurseryWithdrawalsProps): JSX.Element {
-  return <TfMain>Nursery Withdrawals</TfMain>;
+  const { organization } = props;
+  const theme = useTheme();
+  const contentRef = useRef(null);
+  const classes = useStyles();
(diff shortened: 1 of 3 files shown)


src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
```diff
@@ -1,10 +1,119 @@
...
+                id='search'
+                type='text'
+                className={classes.searchField}
+                iconRight='cancel'
+              />
+            </Grid>
+            <Grid item xs={12}>
+              <div>
+                <Grid container spacing={4}>
+                  <Grid item xs={12}>
```
charlie-tf, line 101: Are all of these layers necessary?

src/components/NurseryWithdrawals/WithdrawalLogRenderer.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { makeStyles } from '@mui/styles';
+import { APP_PATHS } from 'src/constants';
+import CellRenderer, { TableRowType } from '../common/table/TableCellRenderer';
+import { RendererProps } from '../common/table/types';
+import { Theme } from '@mui/material';
+import { Button } from '@terraware/web-components';
```
charlie-tf, line 8: as far as import patterns.. we also re-export this from `src/components/common/button/Button.tsx`. Not sure which is our preferred pattern or the history behind those re-exports. Personally, I like this direct import from `web-components` better.

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
```diff
@@ -1,10 +1,119 @@
+import { Box, Grid, Theme, Typography, useTheme } from '@mui/material';
+import TextField from '@terraware/web-components/components/Textfield/Textfield';
+import { useEffect, useRef, useState } from 'react';
+import strings from 'src/strings';
 import { ServerOrganization } from 'src/types/Organization';
+import PageHeaderWrapper from '../common/PageHeaderWrapper';
 import TfMain from '../common/TfMain';
+import { makeStyles } from '@mui/styles';
+import { Table, TableColumnType } from '@terraware/web-components';
```
charlie-tf, line 9: similarly for `src/components/common/table/index.tsx` and `types.ts` 🤷 Again, I prefer what you have here.

src/components/NurseryWithdrawals/WithdrawalLogRenderer.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const rowClick = (event?: React.SyntheticEvent) => {
+    if (onRowClick) {
+      onRowClick();
+    }
+  };
+
+  const createLinkToNurseryWithdrawalDetail = (iValue: React.ReactNode | unknown[]) => {
+    const nurseryWithdrawalDetailLocation = {
+      pathname: APP_PATHS.NURSERY_WITHDRAWALS_DETAILS.replace(':withdrawalId', row.id.toString()),
+    };
```
charlie-tf, line 36: Why create this object instead of just a const string?

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
```diff
@@ -1,10 +1,119 @@
...
+
+const columns: TableColumnType[] = [
+  { key: 'withdrawnDate', name: strings.DATE, type: 'string' },
+  { key: 'purpose', name: strings.PURPOSE, type: 'string' },
+  { key: 'facility_name', name: strings.FROM_NURSERY, type: 'string' },
+  { key: 'destinationName', name: strings.DESTINATION, type: 'string' },
+  { key: 'plotNames', name: strings.TO_PLOT, type: 'string' },
+  { key: 'speciesScientificNames', name: strings.SPECIES, type: 'string' },
+  { key: 'totalWithdrawn', name: strings.TOTAL_QUANTITY, type: 'string' },
+  { key: 'hasReassignments', name: '', type: 'string' },
```
charlie-tf, line 34: would this be better named simply `'reassignments'`, as `'hasReassignments'` implies a boolean value?

charlie-tf (commented):
Looks good! Just a couple comments

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
constanzauanini: I will try to clean these, I just copied/pasted from other table

src/components/NurseryWithdrawals/WithdrawalLogRenderer.tsx
constanzauanini: Yes, I think we should always import from '@terraware/web-components'; It looks cleaner

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
constanzauanini: The "keys" correspond to the values that come from the search api and the "name" is the title of the column in the table

charlie-tf: gotcha 👍 Thanks!

charlie-tf (approved):
Looks great. Thanks!

PR #861, opened 2022-12-02
SW-2334 Search and filters to withdrawal log table

constanzauanini:
Screenshot 2022-12-05 at 12 20 49

### DIFF
--- a/src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
+++ b/src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
@@ -18,10 +18,14 @@ import { getRequestId, setRequestId } from 'src/utils/requestsId';
 import NurseryWithdrawalsFiltersPopover from './NurseryWithdrawalsFiltersPopover';
 import Pill from '../Pill';
 import { getAllNurseries } from 'src/utils/organization';
+import { getAllSpecies } from 'src/api/species/species';
+import { Species } from 'src/types/Species';
 
 export type NurseryWithdrawalsFiltersType = {
   fromNurseryIds?: string[];
   purposes?: string[];
+  destinationNames?: string[];
+  speciesId?: string[];
 };
 
 type NurseryWithdrawalsProps = {
@@ -50,32 +54,28 @@ export default function NurseryWithdrawals(props: NurseryWithdrawalsProps): JSX.
   const theme = useTheme();
   const contentRef = useRef(null);
   const classes = useStyles();
-  const [nurseryWithdrawals, setNurseryWithdrawals] = useState<SearchResponseElement[]>();
-
   const [searchResults, setSearchResults] = useState<SearchResponseElement[] | null>();
   const history = useHistory();
   const [searchValue, setSearchValue] = useState('');
   const debouncedSearchTerm = useDebounce(searchValue, 250);
   const [filters, setFilters] = useForm<NurseryWithdrawalsFiltersType>({});
-
-  useEffect(() => {
-    const getNurseryWithdrawals = async () => {
-      if (organization) {
-        const withdrawals = await listNurseryWithdrawals(organization.id, []);
-        if (withdrawals) {
-          setNurseryWithdrawals(withdrawals);
-        }
-      }
-    };
-
-    getNurseryWithdrawals();
-  }, [organization]);
+  const [species, setSpecies] = useState<Species[]>();
 
   const onWithdrawalClicked = (withdrawal: any) => {
     history.push({
       pathname: APP_PATHS.NURSERY_REASSIGNMENT.replace(':deliveryId', withdrawal.delivery_id),
     });
   };
+
+  useEffect(() => {
+    const populateSpecies = async () => {
+      const result = await getAllSpecies(organization.id);
+      if (result.requestSucceeded) {
+        setSpecies(result.species);
+      }
+    };
+    populateSpecies();
+  }, [organization]);
 
   const getSearchChildren = useCallback(() => {
     const finalSearchValueChildren: FieldNodePayload[] = [];
@@ -106,6 +106,7 @@ export default function NurseryWithdrawals(props: NurseryWithdrawalsProps): JSX.
       searchValueChildren.push(speciesNameNode);
     }
 
+    const filterValueChildren: FieldNodePayload[] = [];
     let nurseryFilter: FieldNodePayload;
     if (filters.fromNurseryIds && filters.fromNurseryIds.length > 0) {
       nurseryFilter = {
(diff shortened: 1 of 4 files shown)


src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
```diff
@@ -93,13 +212,32 @@ export default function NurseryWithdrawals(props: NurseryWithdrawalsProps): JSX.
...
+                  }
+                )}
+              </Grid>
             </Grid>
             <Grid item xs={12}>
               <Table
                 id='withdrawal-log'
                 columns={columns}
-                rows={nurseryWithdrawals || []}
+                rows={searchResults || nurseryWithdrawals || []}
```
karthikbtf, line 240: we have a mix of latest search results vs cached memory results which can be inconsistent, ok to always use latest search results? which means if the filters are cleared - search for all nursery withdrawals again. Have one state to hold results.

src/components/NurseryWithdrawals/NurseryWithdrawalsFiltersPopover.tsx
```diff
@@ -0,0 +1,207 @@
...
+    fontSize: '20px',
+    fontWeight: 600,
+  },
+  container: {
+    padding: '24px',
+    maxHeight: 'calc(100vh - 150px)',
+    overflow: 'auto',
+  },
+  footer: {
+    background: '#F2F4F5',
```
karthikbtf, line 58: use theme colors everywhere

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
```diff
@@ -50,33 +54,29 @@ export default function NurseryWithdrawals(props: NurseryWithdrawalsProps): JSX.
...
       pathname: APP_PATHS.NURSERY_REASSIGNMENT.replace(':deliveryId', withdrawal.delivery_id),
     });
   };
 
+  useEffect(() => {
+    const populateSpecies = async () => {
+      const result = await getAllSpecies(organization.id);
+      if (result.requestSucceeded) {
+        setSpecies(result.species);
+      }
```
karthikbtf, line 75: should we show an error otherwise?

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
```diff
@@ -115,32 +116,76 @@ export default function NurseryWithdrawals(props: NurseryWithdrawalsProps): JSX.
         values: filters.fromNurseryIds.map((id) => id.toString()),
       };
     }
+    if (nurseryFilter) {
```
karthikbtf, line 119: for all these filters, can we simply push as part of the if (filters.xyz && filters.xyz.length > 0) check?

```
if (filters.fromNurseryIds && filters.fromNurseryIds.length > 0) {
  filterValueChildren.push({
  });
}
```

src/components/NurseryWithdrawals/NurseryWithdrawalsFiltersPopover.tsx
```diff
@@ -194,6 +211,38 @@ export default function NurseryWithdrawalsFiltersPopover({
...
+                      name={plantingSite.name}
+                      label={plantingSite.name}
+                      value={hasFilter('destinationNames', plantingSite.name)}
+                      onChange={(id, value) => onChangeHandler(id, value, 'destinationNames')}
+                    />
+                  </Grid>
+                ))}
+              </Grid>
+              <Grid item xs={6}>
+                <Typography fontSize='16px' paddingLeft={theme.spacing(2)} color='#708284'>
```
karthikbtf, line 231: theme color?

src/components/NurseryWithdrawals/NurseryWithdrawalsFiltersPopover.tsx
```diff
@@ -194,6 +211,38 @@ export default function NurseryWithdrawalsFiltersPopover({
                   </Grid>
                 ))}
               </Grid>
+              <Grid item xs={6}>
+                <Typography fontSize='16px' paddingLeft={theme.spacing(2)} color='#708284'>
```
karthikbtf, line 215: theme color? and in other places

src/components/NurseryWithdrawals/NurseryWithdrawalsFiltersPopover.tsx
```diff
@@ -95,6 +102,16 @@ export default function NurseryWithdrawalsFiltersPopover({
     setTemporalRecord(filters);
   }, [filters, setTemporalRecord]);
 
+  useEffect(() => {
+    const populatePlantingSites = async () => {
+      const result = await listPlantingSites(organization.id);
+      if (result.requestSucceeded) {
+        setPlantingSites(result.sites);
```
karthikbtf, line 109: else error?

karthikbtf (commented):
looks good, some comments for follow up.
after looking more into this.. for next time perhaps.. this would be easier if implemented one filter at a time for review purposes?

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
constanzauanini: Yes, we could show an error, but the thing is that you could still load the withdrawals and search if this fails, just the species filter would not have species to select if this fails. Showing an error explaining that wouldn't be confusing?

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
constanzauanini: Totally! 💯 Much better

src/components/NurseryWithdrawals/NurseryWithdrawalsFiltersPopover.tsx
constanzauanini: Same thing here, what message could we show that would not generate confussion?

karthikbtf: if user searches but the results don’t update because there was a silent error.. that would be confusing as well. If user searches and there’s an error and we show an error .. not sure that would be confusing? A generic error is better than silent errors

karthikbtf: Okay maybe I'm not understanding the flow too well.
Let's just iterate on this.
We don't have to complete the entire filtering in one PR, not easy to review.

karthikbtf (approved):
Looks good, just clean up the theme colors at a minimum.
Error messaging we can leave as-is, no need to show errors - we can iterate later if needed.

src/components/NurseryWithdrawals/NurseryWithdrawals.tsx
```diff
@@ -39,28 +55,201 @@ export default function NurseryWithdrawals(props: NurseryWithdrawalsProps): JSX.
...
-        const withdrawals = await listNurseryWithdrawals(organization.id, []);
-        if (withdrawals) {
-          setNurseryWithdrawals(withdrawals);
-        }
+    const populateSpecies = async () => {
+      const result = await getAllSpecies(organization.id);
+      if (result.requestSucceeded) {
+        setSpecies(result.species);
+      } else {
+        snackbar.toastError();
```
karthikbtf, line 78: Thanks! Can we iterate later on what a good error message might be? I do agree with you that it could be confusing if we don't give context.
Something like `couldn't load species list for search`

src/components/NurseryWithdrawals/NurseryWithdrawalsFiltersPopover.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  </Grid>
+                ))}
+              </Grid>
+              <Grid item xs={6}>
+                <Typography fontSize='16px' paddingLeft={theme.spacing(2)} color={theme.palette.TwClrBaseGray500}>
+                  {strings.SPECIES}
+                </Typography>
+                {species?.map((iSpecies) => (
+                  <Grid item xs={12} key={iSpecies.id}>
+                    <Checkbox
```
karthikbtf, line 240: suggestion - for all the checkboxes, align with top of string?
Terraware App 2022-12-05 08-06-58

karthikbtf (approved):
👍

src/components/NurseryWithdrawals/NurseryWithdrawalsFiltersPopover.tsx
karthikbtf: yes, ofc, was just pointing this out.

PR #881, opened 2022-12-07
SW-2482 Home page cards should redirect user to target page

karthikbtf:
- allow click on box (for links only, not button based cards such as on the contact-us page)
- add wrapper around link so we can prevent bubbling of click - this preserves cmd+click behavior

### DIFF
--- a/src/components/common/PageCard.tsx
+++ b/src/components/common/PageCard.tsx
@@ -2,15 +2,23 @@ import { Box, Link, Theme, Typography, useTheme } from '@mui/material';
 import React from 'react';
 import Icon from './icon/Icon';
 import { IconName } from './icon/icons';
-import { Link as RouterLink } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


src/components/common/PageCard.tsx
```diff
@@ -96,9 +118,11 @@ export default function PageCard(props: PageCardProps): JSX.Element {
         {description}
       </Typography>
       {linkStyle === 'plain' && (
-        <Link className={classes.link} component={RouterLink} to={link}>
-          {linkText}
-        </Link>
+        <Box onClick={stopBubblingEvent}>
```
karthikbtf, line 121: This is needed to prevent event from reaching up to the Box for double processing.
Also this is needed as a wrapper so we can preserver cmd+click behavior on the link.

constanzauanini (approved):

PR #885, opened 2022-12-07
SW-2468 Add messages to empty state generic map

constanzauanini:
Screenshot 2022-12-07 at 12 18 54

### DIFF
--- a/src/components/Map/GenericMap.tsx
+++ b/src/components/Map/GenericMap.tsx
@@ -4,6 +4,9 @@ import { getMapboxToken } from 'src/api/tracking/tracking';
 import useSnackbar from 'src/utils/useSnackbar';
 import Map from './Map';
 import { MapOptions, MapPopupRenderer } from './MapModels';
+import { isContributor } from 'src/utils/organization';
+import { ServerOrganization } from 'src/types/Organization';
+import strings from 'src/strings';
 
 const DUMMY_MAP_OPTIONS: MapOptions = {
   bbox: {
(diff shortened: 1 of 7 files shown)


src/components/Map/GenericMap.tsx
```diff
@@ -17,13 +20,20 @@ type GenericMapProps = {
   contextRenderer?: MapPopupRenderer;
   options?: MapOptions;
   style?: object;
+  organization?: ServerOrganization;
```
karthikbtf, line 23: Looks good.
can we just pass the message in here instead?
the context of showing this message is in PlantingSiteDashboardMap - which is where we should make the decision to set the message

src/components/Map/MapTopMessage.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react';
+import { Box } from '@mui/material';
+import { Message } from '@terraware/web-components';
+
+type MapTopMessageProps = {
+  message: string;
+};
+
+export default function MapTopMessage({ message }: MapTopMessageProps): JSX.Element | null {
```
karthikbtf, line 9: Nice abstraction.
Suggestion, call this MapBanner?

karthikbtf (commented):
Looks good, a few suggestions.

src/components/Map/GenericMap.tsx
karthikbtf: On second thought, leave it this way for now.

src/components/Map/MapTopMessage.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const [showMessage, setShowMessage] = useState(true);
+
+  if (showMessage) {
+    return (
+      <Box
```
karthikbtf, line 14: can you post a mobile screenshot?

karthikbtf (approved):
looks good, rename that one class if you think that's good

src/components/Map/MapTopMessage.tsx
karthikbtf: @constanzauanini we don't show this on mobile view.. ignore my request

karthikbtf: 😂 my bad

PR #886, opened 2022-12-07
Fix message appearing on plot maps

constanzauanini:


src/components/Map/GenericMap.tsx
```diff
@@ -20,20 +17,19 @@ type GenericMapProps = {
   contextRenderer?: MapPopupRenderer;
   options?: MapOptions;
   style?: object;
-  organization?: ServerOrganization;
+  topMessage?: string;
```
karthikbtf, line 20: bannerMessage?

PR #900, opened 2022-12-08
Fix validations on single withdraw flow

constanzauanini:
- Added NaN validation
- Convert strings to number before comparing


karthikbtf (approved):
Looks clear to me!

PR #902, opened 2022-12-09
SW-2514 Update links in accession edit

karthikbtf:
- also use + icon
- introduced an AddLink component

Terraware App 2022-12-09 08-33-07

Terraware App 2022-12-09 10-07-06

Terraware App 2022-12-09 08-22-38

Terraware App 2022-12-09 08-22-08


src/components/common/AddLink.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type AddLinkProps = {
+  text: string;
+  id?: string;
+  large?: boolean;
+  onClick: (args: any) => void;
+  disabled?: boolean;
+};
+
+export default function AddLink(props: AddLinkProps): JSX.Element {
```
charlie-tf, line 27: nice component!

src/components/common/AddLink.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Box, Theme } from '@mui/material';
+import { Icon } from '@terraware/web-components';
+import Link from 'src/components/common/Link';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  addIcon: {
+    fill: theme.palette.TwClrIcnBrand,
+  },
+  disabledIcon: {
+    fill: theme.palette.TwClrBaseGray300,
```
charlie-tf, line 11: base color instead of a semantic one?

hannah-cai: we're not really using disabled icons, what is this for?

hannah-cai (approved):

src/components/common/AddLink.tsx
karthikbtf: @hannah-cai here's the use-case (a completed viability test)
Terraware App 2022-12-09 13-26-18

karthikbtf: @charlie-tf good catch, i copied it without checking for correctness. will update both

karthikbtf: removing the color change, just using opacity 0.5 as per our design

src/components/common/AddLink.tsx
karthikbtf: Thanks @charlie-tf - I also added a few others like Link.tsx, TitleDescription.tsx, Card.tsx .. which seem reusable with specs. I can move the useful ones from that list to web-components at some point.

src/components/common/AddLink.tsx
karthikbtf: Terraware App 2022-12-09 13-31-18

PR #903, opened 2022-12-09
SW-2338 Withdrawal Details

charlie-tf:
- [x] overview card info
- [x] species table
- [ ] photos
  - [ ] request nursery withdrawal photos list from API
image

### DIFF
--- a/src/api/tracking/withdrawals.ts
+++ b/src/api/tracking/withdrawals.ts
@@ -81,13 +81,13 @@ type NurseryWithdrawalResponse = {
   error?: string;
 };
 
-export const getNurseryWithdrawal = async (withdrawalId: number): Promise<NurseryWithdrawalResponse> => {
+export const getNurseryWithdrawal = async (withdrawalId: string): Promise<NurseryWithdrawalResponse> => {
   const response: NurseryWithdrawalResponse = {
     requestSucceeded: true,
   };
 
   try {
-    const endpoint = NURSERY_WITHDRAWAL_ENDPOINT.replace('{withdrawalId}', withdrawalId.toString());
+    const endpoint = NURSERY_WITHDRAWAL_ENDPOINT.replace('{withdrawalId}', withdrawalId);
     const serverResponse: GetNurseryWithdrawalResponsePayload = (await axios.get(endpoint)).data;
     if (serverResponse.status === 'error') {
       response.requestSucceeded = false;
--- a/src/api/tracking/withdrawals.ts
+++ b/src/api/tracking/withdrawals.ts
@@ -81,13 +81,13 @@ type NurseryWithdrawalResponse = {
   error?: string;
 };
 
-export const getNurseryWithdrawal = async (withdrawalId: string): Promise<NurseryWithdrawalResponse> => {
+export const getNurseryWithdrawal = async (withdrawalId: number): Promise<NurseryWithdrawalResponse> => {
   const response: NurseryWithdrawalResponse = {
     requestSucceeded: true,
   };
 
   try {
-    const endpoint = NURSERY_WITHDRAWAL_ENDPOINT.replace('{withdrawalId}', withdrawalId);
+    const endpoint = NURSERY_WITHDRAWAL_ENDPOINT.replace('{withdrawalId}', withdrawalId.toString());
     const serverResponse: GetNurseryWithdrawalResponsePayload = (await axios.get(endpoint)).data;
     if (serverResponse.status === 'error') {
       response.requestSucceeded = false;
--- a/src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
+++ b/src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
@@ -125,8 +125,6 @@ export default function NurseryWithdrawalsDetails({
     updateWithdrawal();
   }, [organization.id, withdrawalId, snackbar]);
 
-  console.log(withdrawalSummary);
-
   useEffect(() => {
     setSelectedTab((query.get('tab') || 'withdrawal') as string);
   }, [query]);
--- a/src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
+++ b/src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
@@ -1,27 +1,28 @@
 import { Link, useHistory, useParams } from 'react-router-dom';
 import { ServerOrganization } from 'src/types/Organization';
-import TfMain from '../common/TfMain';
+import TfMain from 'src/components/common/TfMain';
 import { Box, Tab, Theme, Typography, useTheme } from '@mui/material';
-import { APP_PATHS } from '../../constants';
+import { APP_PATHS } from 'src/constants';
 import { Button, Icon } from '@terraware/web-components';
-import strings from '../../strings';
-import PageSnackbar from '../PageSnackbar';
-import PageHeaderWrapper from '../common/PageHeaderWrapper';
+import strings from 'src/strings';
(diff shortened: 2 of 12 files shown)


src/api/tracking/withdrawals.ts
```diff
@@ -81,13 +81,13 @@ type NurseryWithdrawalResponse = {
   error?: string;
 };
 
-export const getNurseryWithdrawal = async (withdrawalId: number): Promise<NurseryWithdrawalResponse> => {
+export const getNurseryWithdrawal = async (withdrawalId: string): Promise<NurseryWithdrawalResponse> => {
```
karthikbtf, line 84: We have this pattern `(id: number)` in most new API.
Calling code could theoretically pass a Number(stringId) to this API to workaround this change.
Kept the i`d: number` signature because id is defined as a number in the object type from BE.

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
```diff
@@ -1,13 +1,193 @@
...
+    history.push(getLocation(location.pathname, location, query.toString()));
+  };
 
-  return <TfMain>Nursery Withdrawal Details {withdrawalId} </TfMain>;
+  return (
+    <TfMain>
+      <PageHeaderWrapper nextElement={contentRef.current} nextElementInitialMargin={-24}>
+        <Box marginBottom={theme.spacing(4)}>
+          <Box>
+            <Link id='back' to={APP_PATHS.NURSERY_WITHDRAWALS} className={classes.backToWithdrawals}>
```
karthikbtf, line 116: you can use Link component from `components/common/Link` - you may need to rebase your branch to get latest code
speaking of which,, I missed updating the back to links on all views..

karthikbtf (approved):
Looks great on vercel! The code and approach looks great.
This looks good enough to me as an intermediate commit on main.
A few comments...

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
constanzauanini: I didn't update the back links because I thought they were different from other links

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
```diff
@@ -1,13 +1,193 @@
...
+  const [delivery, setDelivery] = useState<Delivery | undefined>(undefined);
+  const [batches, setBatches] = useState<Batch[] | undefined>(undefined);
+  useEffect(() => {
+    const updateWithdrawal = async () => {
+      const withdrawalResponse = await getNurseryWithdrawal(withdrawalId);
+      if (!withdrawalResponse.requestSucceeded || withdrawalResponse.error) {
+        setWithdrawal(undefined);
+        setDelivery(undefined);
+        setBatches(undefined);
+        snackbar.toastError(`An Error Occurred: ${withdrawalResponse.error}`);
```
constanzauanini, line 74: We can use strings.GENERIC_ERROR

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
karthikbtf: good point, @chudo76 do the back links need to be updated as well? (underline + 500 weight)

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
```diff
@@ -1,13 +1,193 @@
...
+            padding={isMobile ? theme.spacing(3, 0, 4, 0) : theme.spacing(3, 3, 4, 3)}
+            display='flex'
+            justifyContent='space-between'
+            alignItems='center'
+          >
+            <Typography color={theme.palette.TwClrTxt} fontSize='20px' fontWeight={600} fontStyle='italic'>
+              {withdrawal?.withdrawnDate}
+            </Typography>
+            {withdrawal?.purpose === 'Out Plant' && (
+              <Button size='medium' priority='secondary' onClick={() => undefined} label={strings.REASSIGN} />
```
constanzauanini, line 131: This button should be disabled if this withdrawal has already had reassignments

src/components/NurseryWithdrawals/WithdrawalDetails/sections/SpeciesTable.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+export default function SpeciesTable({ organization, delivery }: SpeciesTableSectionProps): JSX.Element {
+  const [rowData, setRowData] = useState<{ [p: string]: unknown }[]>([]);
+
+  useEffect(() => {
+    const rows: { [p: string]: unknown }[] = [];
+    delivery?.plantings?.forEach(async (planting) => {
+      rows.push({
+        species: (await getSpecies(planting.speciesId, organization.id.toString()))?.species?.scientificName ?? '',
```
constanzauanini, line 26: We have an array of all the species of the org on the NurseryWithdrawals parent component. We could pass the species array to the withdrawal detail and then to this table and here we could do speciesList.find(species => species.id === planting.speciesId).scientificName, so we don't have to do an api call for each row

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
karthikbtf: I updated the WithdrawalLogRenderer https://github.com/terraware/terraware-web/blob/main/src/components/NurseryWithdrawals/WithdrawalLogRenderer.tsx#L71 to also not show this if the withdrawal did not have a plot to begin with. (cannot reassign if it wasn't belonging to a plot)

src/components/NurseryWithdrawals/WithdrawalDetails/sections/SpeciesTable.tsx
charlie-tf: good call, I'll get that data down to here to avoid those calls 👍

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
```diff
@@ -1,13 +1,244 @@
...
+        setDelivery(undefined);
+        setBatches(undefined);
+        snackbar.toastError(withdrawalResponse.error);
+      } else {
+        setWithdrawal(withdrawalResponse.withdrawal);
+        setDelivery(withdrawalResponse.delivery);
+        setBatches(withdrawalResponse.batches);
+      }
+      // get summary information
+      const apiSearchResults = await listNurseryWithdrawals(organization.id, [
```
charlie-tf, line 100: Added this call to get summary information about the withdrawal, which includes `hasReassignments`

karthikbtf (approved):
Good stopping point, we can iterate / follow up separately if there are any more questions.

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
chudo76: > good point, @chudo76 do the back links need to be updated as well? (underline + 500 weight)

Yes, all links styles going forward will be:

- Links are `tw-clr-txt-brand`, no text decoration.
- On hover, `text-decoration: underline;` 
- No additional styles for `:active` or `:visited`.

karthikbtf: Thanks @chudo76 i will open up a new ticket out-of-band of withdrawal log effort

charlie-tf: +1 thanks for responding!

PR #906, opened 2022-12-12
SW-2491 New menu component with dropdown styles

constanzauanini:
- Create a new PopoverMenu component and use it in all menus (organization menu, user menu, and kabob menus) 
- This new menu has similar styles to a dropdown component

### DIFF
--- a/src/components/Inventory/index.tsx
+++ b/src/components/Inventory/index.tsx
@@ -1,15 +1,4 @@
-import {
-  Box,
-  CircularProgress,
-  Container,
-  Grid,
-  MenuItem,
-  MenuList,
-  Popover,
-  Theme,
-  Typography,
-  useTheme,
-} from '@mui/material';
+import { Box, CircularProgress, Container, Grid, Theme, Typography, useTheme } from '@mui/material';
 import { makeStyles } from '@mui/styles';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { useHistory } from 'react-router-dom';
@@ -34,7 +23,8 @@ import { downloadInventoryTemplate } from 'src/api/inventory/inventory';
 import ImportInventoryModal from './ImportInventoryModal';
 import { Button } from '@terraware/web-components';
 import PageHeaderWrapper from 'src/components/common/PageHeaderWrapper';
-import Icon from '../common/icon/Icon';
+import { DropdownItem } from '@terraware/web-components/components/Dropdown';
+import PopoverMenu from '../common/PopoverMenu';
 
 interface StyleProps {
   isMobile: boolean;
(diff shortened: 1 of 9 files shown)


constanzauanini:
@hannah-cai can you take a look at the menus? I can update any style as necessary

src/components/Inventory/index.tsx
```diff
@@ -34,7 +23,8 @@ import { downloadInventoryTemplate } from 'src/api/inventory/inventory';
 import ImportInventoryModal from './ImportInventoryModal';
 import { Button } from '@terraware/web-components';
 import PageHeaderWrapper from 'src/components/common/PageHeaderWrapper';
-import Icon from '../common/icon/Icon';
+import { DropdownItem } from '@terraware/web-components/components/Dropdown';
```
karthikbtf, line 26: as much as possible we should import from '@terraware/web-components' only. can we add a task to export DropdownItem from web-components index.ts?

src/components/common/PopoverMenu.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import { ListItem, MenuItem, MenuList, Popover, useTheme } from '@mui/material';
+import { DropdownItem } from '@terraware/web-components/components/Dropdown';
+
+type PopoverMenuProps = {
+  items: DropdownItem[] | [];
```
karthikbtf, line 6: why the ` | []`;

src/components/common/PopoverMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+  handleClick: (item: DropdownItem) => void;
+  otherItems?: DropdownItem[];
```
karthikbtf, line 8: can we make items and otherItems just sections, each section is a list of menu itmes and a handeclick.
sections are separated by '---'
this way it's not very specifically coded for the org menu create.
example:
```
type MenuSection = {
  handleClick: (item: DropdownItem) => void;
  items: DropdownItem[];
}
```
PopoverMenuProps can then be:
```
  menuSections: MenuSection[];
  anchorElement: ....
```

constanzauanini:
> @hannah-cai can you take a look at the menus? I can update any style as necessary

@chudo76 here is the menu updates to look like dropdowns

src/components/common/PopoverMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+                onClick={() => otherItemClick(item)}
+                key={`other-option-${index}`}
+                sx={itemStyles}
+                disableRipple={true}
+              >
+                {item.label}
+              </MenuItem>
+            );
+          })}
+        </>
```
karthikbtf, line 80: this can be
```
   sections.map((section, index) => (
      if (index > 0) 
          ListITem>---</ListItem>
      section.menItems.map(item, menuIndex => (
          <MenuItem onClick={() => handleClick(item)} key={`option-${index}-${menuIndex}`} sx={itemStyles} disableRipple={true}>
              {item.label}
            </MenuItem>
      )
   )
```

src/components/common/PopoverMenu.tsx
karthikbtf: The other option is to have a single list of menuitems and support a 'divider' item in the list

karthikbtf (commented):
Looks great in vercel... a few comments

src/components/common/PopoverMenu.tsx
karthikbtf: Actually - a single list of menuitems, and if item is null - we could assume a divider (with some comments in the file explaining this behavior).

chudo76:
> @chudo76 here is the menu updates to look like dropdowns

Styling looks right. 👍

src/components/common/PopoverMenu.tsx
constanzauanini: Because in the org dropdown orgs could be empty

karthikbtf: DropdownItem[] also captures empty arrays.. maybe I'm missing some TS error you are seeing

karthikbtf (approved):

PR #909, opened 2022-12-13
Derive hasReassignments from Deliveries

charlie-tf:
- information is in the deliveries, instead of needing an additional query
- hasReassignments is true if any of the delivery types is
  "Reassignment To" or "Reassignment From"
- also add back in the query for the destination planting site name

### DIFF
--- a/src/components/NurseryWithdrawals/WithdrawalDetails/WithdrawalTabPanelContent.tsx
+++ b/src/components/NurseryWithdrawals/WithdrawalDetails/WithdrawalTabPanelContent.tsx
@@ -7,14 +7,14 @@ import { Delivery } from 'src/api/types/tracking';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
 import OverviewItemCard from 'src/components/common/OverviewItemCard';
 import { ServerOrganization } from 'src/types/Organization';
+import { useEffect, useState } from 'react';
+import { getPlantingSite } from 'src/api/tracking/tracking';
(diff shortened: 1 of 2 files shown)


src/components/NurseryWithdrawals/WithdrawalDetails/WithdrawalTabPanelContent.tsx
```diff
@@ -41,19 +53,23 @@ export default function WithdrawalTabPanelContent({
...
     },
     {
       title: strings.DESTINATION,
-      data: withdrawalSummary?.destinationName ?? '',
+      data: siteName,
     },
     {
       title: strings.TO_PLOT,
-      data: withdrawalSummary?.plotNames ?? '',
+      data:
```
constanzauanini, line 68: We were showing the plot name before and now just the id, is that intentional ?

charlie-tf: Yeah, the search result was nice in that it had a lot of this information populated in the summary. I will take a look at the `getPlantingSiteData()` method though as well.

charlie-tf:
Going to leave in the search query for now. I'll create a new PR to show/hide the Reassign button.

karthikbtf:
@charlie-tf you would still need to hide the reassign button for deliveries that don't involve a plot.
You can use the existence of plotNames as the criterion.

PR #910, opened 2022-12-13
SW-2447 Header Spacing Fix

charlie-tf:
- fix header padding on monitoring page

image

### DIFF
--- a/src/components/Monitoring/index.tsx
+++ b/src/components/Monitoring/index.tsx
@@ -42,7 +42,6 @@ const useStyles = makeStyles((theme: Theme) => ({
   titleContainer: {
     display: 'flex',
     alignItems: 'center',
-    // paddingBottom: theme.spacing(2),
     paddingLeft: (props: StyleProps) => theme.spacing(props.isDesktop ? 3 : 0),
     paddingTop: (props: StyleProps) => theme.spacing(props.isMobile ? 2 : 0),
   },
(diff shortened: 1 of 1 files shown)


src/components/Monitoring/index.tsx
```diff
@@ -152,7 +152,7 @@ export default function Monitoring(props: MonitoringProps): JSX.Element {
       <Grid container>
         {hasSeedBanks ? (
           <>
-            <PageHeaderWrapper nextElement={contentRef.current}>
+            <PageHeaderWrapper nextElement={contentRef.current} nextElementInitialMargin={-24}>
```
karthikbtf, line 155: may want to skip the negative margin on mobile, it seems to be blending into the seedbank picker

src/components/Monitoring/index.tsx
```diff
@@ -37,11 +37,11 @@ const useStyles = makeStyles((theme: Theme) => ({
   titleRootContainer: {
     display: 'flex',
     alignItems: 'center',
+    paddingBottom: theme.spacing(4),
```
karthikbtf, line 40: ideally margin to allow for space between components, padding if we want space inside the component so we can draw borders etc..

karthikbtf (approved):
looks good, a few comments

src/components/Monitoring/index.tsx
charlie-tf: Left it in, and added the margin to the header for mobile. It looks correct now.

karthikbtf (approved):

src/components/Monitoring/index.tsx
charlie-tf: made the change. Thanks for the feedback!

PR #911, opened 2022-12-13
SW-2519 Topbar Dropdown Hover States

charlie-tf:
- add radius and padding to dropdown buttons

image
image


karthikbtf (approved):
Nice change,, looks nicer.. 👍

constanzauanini (approved):

PR #913, opened 2022-12-13
SW-2510 My Account Page Visual Updates

charlie-tf:
Per spec:
- put the body content in a card to match the rest of the pages
- change the subtitles (general, notifications, organizations) to be 20px, so they aren’t the same size as the header
- remove the horizontal rules (we’re no longer using that component)
- update the header and description text sizes to match the rest of the pages (eg seeds dashboard, accessions)
- the edit button should be a primary button, and it should include the edit icon.
image
image

### DIFF
--- a/src/components/MyAccount/index.tsx
+++ b/src/components/MyAccount/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import { updateOrganizationUser, updateUserProfile } from 'src/api/user/user';
 import Button from 'src/components/common/button/Button';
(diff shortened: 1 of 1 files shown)


karthikbtf (approved):
looks good to me 👍

src/components/MyAccount/index.tsx
```diff
@@ -246,25 +218,46 @@ export default function MyAccount({ user, organizations, edit, reloadUser, reloa
...
+          marginBottom={theme.spacing(4)}
+          paddingLeft={theme.spacing(3)}
+        >
+          <Box>
+            <Typography marginBottom={theme.spacing(1.5)} fontSize='24px' fontWeight={600}>
+              {strings.MY_ACCOUNT}
+            </Typography>
+            <Typography fontSize='14px' fontWeight={400}>
+              {strings.MY_ACCOUNT_DESC}
+            </Typography>
```
karthikbtf, line 234: as an aside.. I created a `TitleDescription` component in case it can be reused..
wondering if we should extend that to support these right hand side buttons as well...
we can remove a bunch of boilerplate code across views after that..

src/components/MyAccount/index.tsx
```diff
@@ -246,25 +218,46 @@ export default function MyAccount({ user, organizations, edit, reloadUser, reloa
...
+          </Box>
+          <Button
+            id='edit-account'
+            icon='iconEdit'
+            label={isMobile ? strings.EDIT : dictionary.EDIT_ACCOUNT}
```
constanzauanini, line 239: I think for mobile we are not showing any label, just the icon. But maybe @chudo76 can confirm that

charlie-tf: I think you are correct. I'll make that change now and we can always change it back if needed.

src/components/MyAccount/index.tsx
charlie-tf: works great!

karthikbtf (approved):

PR #914, opened 2022-12-13
SW-2493 / SW-2495 Plot selector component with search/autocomplete

karthikbtf:
- extracted out from withdrawal purpose form
- swapped out dropdowns with autocomplete
- added a horizontal layout feature to be used with Plot charts next in plants dashboard

Terraware App 2022-12-13 13-27-27

### DIFF
--- a/src/components/PlotSelector/index.tsx
+++ b/src/components/PlotSelector/index.tsx
@@ -28,19 +28,19 @@ export type PlotSelectorProps = {
   zones: ZoneInfo[];
   onPlotSelected: (plot?: PlotInfo) => void;
   onZoneSelected?: (zone?: ZoneInfo) => void;
-  initialSelectedPlotId?: number;
   zoneError?: string;
   plotError?: string;
   horizontalLayout?: boolean;
+  siteId?: string | number | undefined;
 };
 
 export default function PlotSelector(props: PlotSelectorProps): JSX.Element {
-  const { zones, onZoneSelected, onPlotSelected, initialSelectedPlotId, zoneError, plotError, horizontalLayout } =
-    props;
+  const { zones, onZoneSelected, onPlotSelected, zoneError, plotError, horizontalLayout, siteId } = props;
   const { isMobile } = useDeviceInfo();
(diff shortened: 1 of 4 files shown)


src/components/Plants/PlantingSiteDetails.tsx
```diff
@@ -67,9 +67,19 @@ export default function PlantingSiteDetails(props: PlantingSiteDetailsProps): JS
...
-            values: [plantingSite.id],
+            operation: 'and',
+            children: [
+              {
+                operation: 'field',
+                field: 'plantingSite_id',
+                values: [plantingSite.id],
+              },
+              {
+                operation: 'field',
```
constanzauanini, line 78: Were we showing results from other orgs previously?  😮

src/components/PlotSelector/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  plots?: PlotInfo[];
+};
+
+export type PlotSelectorProps = {
+  zones: ZoneInfo[];
+  onPlotSelected: (plot?: PlotInfo) => void;
+  onZoneSelected?: (zone?: ZoneInfo) => void;
+  zoneError?: string;
+  plotError?: string;
+  horizontalLayout?: boolean;
```
constanzauanini, line 33: Where is this coming from?

karthikbtf: This will be used in the plots chart . Added this in the PR description.
Have asked Rachel if we can keep existing autocomplete layout let’s see. For now we need this for next task.

src/components/Plants/PlantingSiteDetails.tsx
karthikbtf: I don’t think so but added if.

constanzauanini (approved):

src/components/Plants/PlantingSiteDetails.tsx
karthikbtf: i will remove this, we can add it later if needed

PR #917, opened 2022-12-14
SW-2534 Clean up notifications context

karthikbtf:
- notifications state was being held in the root level App
- probably an artifact of previous notifications design
- every time notifications were fetched (every minute), the app would re-render unnecessarily (due to notifications state change)
- moved notifications state into the notifications dropdown
- removed all the plumbing to App

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,7 +30,6 @@ import TopBar from 'src/components/TopBar/TopBar';
 import TopBarContent from 'src/components/TopBar/TopBarContent';
 import { APP_PATHS } from 'src/constants';
 import ErrorBoundary from 'src/ErrorBoundary';
-import { Notifications } from 'src/types/Notifications';
 import { ServerOrganization } from 'src/types/Organization';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -429,8 +425,6 @@ export default function App() {
       <ToastSnackbar />
       <TopBar>
         <TopBarContent
-          notifications={notifications}
```
constanzauanini, line 432: Oh, were we passing an empty object?

constanzauanini (approved):
Thanks for cleaning this!

src/App.tsx
karthikbtf: we were holding the notifications state in the App (i think because previously the information was shared across components). we don't share this info anymore.

karthikbtf:
> Thanks for cleaning this!

there is going to be a side-effect though.. now the app won't re-render every minutes, maybe data will not refresh as much as it did before.

PR #922, opened 2022-12-15
SW-2470 Add support for hover and select state fill colors on plots 

karthikbtf:
- as per specs in the jira
- this PR does not cover selected plots from the chart
- that will be a follow up PR

Terraware App 2022-12-14 16-29-30


src/components/Map/Map.tsx
```diff
@@ -123,7 +192,23 @@ export default function Map(props: MapProps): JSX.Element {
             id: `${source.id}-fill`,
             type: 'fill',
             paint: {
-              'fill-color': source.fillColor,
+              'fill-color': [
+                // Use a case-expression to decide which color to use for fill
```
constanzauanini, line 196: Thanks for adding this clarification here

constanzauanini (approved):
Looks much better now!

PR #923, opened 2022-12-15
SW-2338, SW-2339 Withdrawal/Reassignment Table

charlie-tf:
Table calculations:
On the withdrawal tab: only show quantities for the original withdrawal,
and do not take into account reassignments. This is because the user
wants to know what was originally withdrawn from the nursery and expects
to go to the reassignment tab to find more information about whether
plants ended up in different plots than expected.

On the reassignment tab: show a row on the table for each plot that a
species is delivered to. The Original Qty column is essentially identical to
the quantities on the withdrawal tab and shows how many plants are
originally intended for the plot. The Final Qty tab shows how many ended
up in the plot at the end of the day, accounting for changes due to
reassignments.

image
image


src/components/NurseryWithdrawals/WithdrawalDetails/sections/OutplantReassignmentTable.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { Delivery } from 'src/api/types/tracking';
+
+type ReassignmentTableSectionProps = {
+  species: Species[];
+  plotNames: Record<number, string>;
+  delivery?: Delivery;
+};
+
+export default function OutplantReassignmentTable(props: ReassignmentTableSectionProps): JSX.Element {
+  return <Box>Reassignment Table</Box>;
```
karthikbtf, line 12: use strings for i18n

karthikbtf (commented):
looks good so far

src/components/NurseryWithdrawals/WithdrawalDetails/sections/OutplantReassignmentTable.tsx
charlie-tf: yup! this is just a placeholder

src/components/NurseryWithdrawals/WithdrawalDetails/sections/OutplantReassignmentTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+      // if reassignment plantings are found, create table rows
+      if (deliveryPlanting && reassignmentFromPlanting && reassignmentToPlanting) {
+        rows.push({
+          species: speciesName,
+          from_plot: '',
+          to_plot: deliveryPlanting.plotId ? plotNames[deliveryPlanting.plotId] : '',
+          original_qty: deliveryPlanting.numPlants.toString(),
+          final_qty: (deliveryPlanting.numPlants + reassignmentFromPlanting.numPlants).toString(),
+          notes: reassignmentToPlanting.notes ?? '',
```
karthikbtf, line 52: this is interesting, probably need to follow up on figma if we need to show the notes twice. for now we could leave it in for sure.

src/components/NurseryWithdrawals/WithdrawalDetails/sections/OutplantReassignmentTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+          final_qty: reassignmentToPlanting.numPlants.toString(),
+          notes: reassignmentToPlanting.notes ?? '',
+        });
+      }
+    }
+
+    setRowData(rows);
+  }, [delivery, species, plotNames]);
+
+  return <Table id='outplant-reassignment-table' columns={columns} rows={rowData} orderBy={'name'} />;
```
karthikbtf, line 68: here and in the outplant withdrawal table, thought on renaming rowData to reassignments and withdrawals respectively?

karthikbtf (approved):
Looks great!

charlie-tf:
> Side comment, on the withdrawals page, we may be missing 'notes' and the to-Plot font seems to be lighter weight in figma.

@karthikbtf I have it hide the card if there are no withdrawal notes. Would it be better to show it empty? I asked about the to-Plot font-weight in the figma.

src/components/NurseryWithdrawals/WithdrawalDetails/sections/OutplantReassignmentTable.tsx
charlie-tf: I don't see a great amount of value in renaming, tbh. But I can certainly do so if you think it makes things clearer 🙂

karthikbtf:
> > Side comment, on the withdrawals page, we may be missing 'notes' and the to-Plot font seems to be lighter weight in figma.
> 
> @karthikbtf I have it hide the card if there are no withdrawal notes. Would it be better to show it empty? I asked about the to-Plot font-weight in the figma.

In viability tests, we show Notes regardless (screenshot). just for consistency maybe. T
My opinion is to show it regardless.. if QA decides to remove it - we can. If we don't show it, QA might miss the behavior.

Terraware App 2022-12-16 07-27-49

Re: font-size, yes we can iterate based on QA for sure.

src/components/NurseryWithdrawals/WithdrawalDetails/sections/OutplantReassignmentTable.tsx
karthikbtf: rowData is just as fine. We don't have any guidelines at this point, was mostly projecting my preferences for readability. In this case it is quite obvious anyways.

src/components/NurseryWithdrawals/WithdrawalDetails/sections/OutplantReassignmentTable.tsx
charlie-tf: from @rachel-tf-design:
> The notes for the first row is the withdrawal notes and the notes for the second row is the reassignment notes so they're different.

I will go ahead and make this change.

charlie-tf:
>My opinion is to show it regardless.. if QA decides to remove it - we can. If we don't show it, QA might miss the behavior.

That's a good call, I'll add it back in.

karthikbtf (approved):
🎉 thanks for the changes

PR #924, opened 2022-12-15
SW-2537 Support preselection in zone/plot selector

constanzauanini:


src/components/Plants/index.tsx
```diff
@@ -141,13 +142,11 @@ export default function PlantsDashboard(props: PlantsDashboardProps): JSX.Elemen
       <Grid item xs={12}>
         <PageSnackbar />
       </Grid>
-      <Box ref={contentRef} display='flex' flexGrow={1}>
+      <Box ref={contentRef}>
```
karthikbtf, line 145: This change undoes my last fix.. add it back?

karthikbtf (approved):
Looks good, thanks for this enhancement.
One comment about a change

src/components/Plants/index.tsx
constanzauanini: Oh, good catch, I must have been during the merge

karthikbtf (approved):
Thanks 👍 looks good on vercel!
I do notice the 'x' icon showing up even if there is no selection - something we can clean up later maybe

PR #926, opened 2022-12-15
SW-2470 Highlight the selected plot in maps dashboard

karthikbtf:
- plumb the active object through

Terraware App 2022-12-15 10-30-15


constanzauanini (approved):
Lgtm! Should we zoom to the selected plot in the map?

karthikbtf:
> Lgtm! Should we zoom to the selected plot in the map?

The requirement is to pan to the zone when the zone is selected. That's my next PR. They don't want zoom to plot (as of now).

karthikbtf:
Ok, I believe I fixed outstanding issues. Primary issue was initializing autocomplete to undefined value.
We need this web-components version to go in first https://github.com/terraware/web-components/pull/205
I will have a follow up PR after that.

PR #931, opened 2022-12-16
SW-2341 / SW-2552 Dead/Other withdrawal overview

karthikbtf:
- extracted out the overview item card from outplant
- combined dead and other into DeadOtherContent (no distinction for now)
- I can keep them separate if we want that separation
- if this approach is fine, I can follow up with a table PR

Terraware App 2022-12-16 09-27-29

Terraware App 2022-12-16 09-26-59


charlie-tf (approved):
I like it! Nice refactor.

karthikbtf:
> I like it! Nice refactor.

Thank you!

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
```diff
@@ -234,14 +233,13 @@ export default function NurseryWithdrawalsDetails({
             <NurseryTransferContent />
           </Box>
         )}
-        {withdrawal?.purpose === 'Dead' && (
+        {(withdrawal?.purpose === 'Dead' || withdrawal?.purpose === 'Other') && (
```
charlie-tf, line 236: just looked at the Figma again, and it doesn't appear that there are differences between these and the Nursery Transfer details either. Would it make sense to include that here as well?

karthikbtf: good point, will do that

charlie-tf: Thanks, Karthik

PR #932, opened 2022-12-16
Set Cursor for Withdrawal Details Photos

charlie-tf:
- set `cursor: pointer` for the `Box` containing the `img`


karthikbtf (approved):
thanks for following up

PR #936, opened 2022-12-19
SW-2441 Fix Error focus active input field on withdrawal form batches cut off on left

constanzauanini:
When focusing on input field, the border increases to 2px and takes more space, so the fix was remove overflow: hidden to avoid cutting off the border

Screenshot 2022-12-19 at 13 30 26


karthikbtf (approved):
PR description and screenshot helps.

PR #937, opened 2022-12-19
SW-2443 Disable non-numeric and negative entries on withdrawal flow

constanzauanini:
Screenshot 2022-12-19 at 15 11 30

### DIFF
--- a/src/components/Inventory/withdraw/flow/SelectBatchesWithdrawnQuantity.tsx
+++ b/src/components/Inventory/withdraw/flow/SelectBatchesWithdrawnQuantity.tsx
@@ -145,6 +145,8 @@ export default function SelectBatches(props: SelectBatchesWithdrawnQuantityProps
     });
   };
 
+  const isInvalidQuantity = (val: any) => isNaN(val) || +val < 0;
+
   const validateQuantities = () => {
     let noErrors = true;
     let newRecords: BatchWithdrawalForTable[] = [];
@@ -153,7 +155,7 @@ export default function SelectBatches(props: SelectBatchesWithdrawnQuantityProps
       newRecords = record.map((rec) => {
         let readyQuantityWithdrawnError = '';
         if (rec.readyQuantityWithdrawn) {
-          if (isNaN(rec.readyQuantityWithdrawn) || +rec.readyQuantityWithdrawn < 0) {
+          if (isInvalidQuantity(rec.readyQuantityWithdrawn)) {
             readyQuantityWithdrawnError = strings.INVALID_VALUE;
             noErrors = false;
           } else {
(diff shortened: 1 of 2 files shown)


src/components/Inventory/withdraw/flow/WithdrawalBatchesCellRenderer.tsx
```diff
@@ -41,12 +41,13 @@ export default function WithdrawalBatchesCellRenderer(props: RendererProps<Table
         <Box display='flex' alignItems={row.error[id] ? 'start' : 'center'}>
           <Textfield
             id={id}
-            type='text'
+            type='number'
```
karthikbtf, line 44: our number input is currently broken
you can't enter 0

src/components/Inventory/withdraw/flow/SelectBatchesWithdrawnQuantity.tsx
```diff
@@ -178,21 +183,31 @@ export default function SelectBatches(props: SelectBatchesWithdrawnQuantityProps
...
+              readyQuantityWithdrawnError = '';
+            }
           }
         } else {
           unsetValues++;
         }
         if (rec.notReadyQuantityWithdrawn) {
-          if (+rec.notReadyQuantityWithdrawn > rec.notReadyQuantity) {
-            notReadyQuantityWithdrawnError = strings.WITHDRAWN_QUANTITY_ERROR;
+          if (isNaN(rec.notReadyQuantityWithdrawn) || +rec.notReadyQuantityWithdrawn < 0) {
```
karthikbtf, line 201: could extract this into a function maybe

const isInvalidQuantity = (val: any) => isNaN(val) || (+val < 0);

src/components/Inventory/withdraw/flow/WithdrawalBatchesCellRenderer.tsx
constanzauanini: I tested entering 0 and it worked as expected.  In which case it doesn't work?

karthikbtf: interesting, it never worked for me.. this line https://github.com/terraware/web-components/blob/main/src/components/Textfield/Textfield.tsx#L124 changes the 0 to '' (i may have added it because we did not want to display a default 0, this change is still wrong anyways).
will try in the vercel link.. in any case, if it works good

karthikbtf (approved):
looks good

src/components/Inventory/withdraw/flow/WithdrawalBatchesCellRenderer.tsx
karthikbtf: it works in vercel.. i should probably switch to this on the reassignments page

constanzauanini: Should we remove the `|| ''` from the value? Because if we do that, we maybe start showing zeros in some places we don't wan't

karthikbtf: yes we should probably remove it and fix the places where we were initializing to 0

PR #939, opened 2022-12-19
Use number texfield for reassignment flow

karthikbtf:
- number did not work previously but seems to work now... 

Screen Shot 2022-12-19 at 12 36 37 PM


src/components/NurseryWithdrawals/ReassignmentRenderer.tsx
```diff
@@ -94,7 +94,7 @@ export default function ReassignmentRenderer({ plots, setReassignment }: Reassig
         <Box display='flex' alignItems={error.quantity ? 'start' : 'center'}>
           <Textfield
             id={`quantity_${plantingId}`}
-            type='text'
+            type='number'
```
constanzauanini, line 97: Can we also add the min=0 ?

constanzauanini (approved):
Thanks!

PR #940, opened 2022-12-19
SW-2467 Use consistent spacing between page and bottom cancel/save bar

karthikbtf:
- created a PageForm components that instantiates children + FormBottomBar with consistent spacing (height of bottom bar + 32px)
- refactored code to use PageForm
- removed margin/padding bottom from various components that were using form bottom bar
- kept a theme spacing(8) in some components that had a dropdown as last element to allow for the dropdown
- ideally would have preferred to remove the fixed position in formbottombar and use some Box magic between the bottom bar and children
- will save that for another review, for now the code is managed in one-place which will make it easier
- tested all those pages in desktop and mobile view

### DIFF
--- a/src/components/EditOrganization/index.tsx
+++ b/src/components/EditOrganization/index.tsx
@@ -7,7 +7,7 @@ import { ServerOrganization } from 'src/types/Organization';
 import TextField from '../common/Textfield/Textfield';
 import useForm from 'src/utils/useForm';
 import Select from '../common/Select/Select';
-import FormBottomBar from '../common/FormBottomBar';
+import PageForm from '../common/PageForm';
 import { Country, Subdivision } from 'src/types/Country';
 import { searchCountries } from 'src/api/country/country';
 import { updateOrganization } from 'src/api/organization/organization';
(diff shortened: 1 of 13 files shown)


src/components/EditOrganization/index.tsx
```diff
@@ -117,66 +117,67 @@ export default function OrganizationView({ organization, reloadOrganizationData
 
   return (
     <TfMain>
-      <Box margin={theme.spacing(0, 3, 4, 3)}>
-        <Box display='flex' flexDirection='column' justifyContent='space-between' marginBottom={theme.spacing(1)}>
-          <Typography margin={0} fontSize='24px' fontWeight={600}>
-            {strings.ORGANIZATION}
-          </Typography>
-          <PageSnackbar />
+      <PageForm onCancel={goToOrganization} onSave={saveOrganization}>
```
karthikbtf, line 120: most of the changes below are indentation, similarly in other places

joshbuchea (approved):
I like that this PR makes the code more DRY 💚

constanzauanini (approved):

PR #941, opened 2022-12-19
Fix sort order of zone and plot names

karthikbtf:
- use localeCompare with numeric = true

Screen Shot 2022-12-19 at 2 20 01 PM

Screen Shot 2022-12-19 at 2 19 32 PM


joshbuchea (approved):
`localeCompare` is pretty cool 🎉

PR #943, opened 2022-12-20
SW-2470 Use zoom instead of pan - to zone, in plants dashboard

karthikbtf:
- Rachel chose zoom over pan, both were functional
- removed the feature based if/else and default to zoom
- Pan-to will center the zone but won’t change the zoom level. If you are zoomed in or out that won’t change.
- With new change the map will fit the sleeted zone boundary on screen


constanzauanini (approved):
I was trying to see the difference between Vercel and staging but I'm not sure which is, could you add in the description the before and after?

karthikbtf:
> I was trying to see the difference between Vercel and staging but I'm not sure which is, could you add in the description the before and after?

Updated the description.

constanzauanini:
> > I was trying to see the difference between Vercel and staging but I'm not sure which is, could you add in the description the before and after?
> 
> Updated the description.

Thanks! I see the difference now!

karthikbtf:
> > > I was trying to see the difference between Vercel and staging but I'm not sure which is, could you add in the description the before and after?
> > 
> > 
> > Updated the description.
> 
> Thanks! I see the difference now!

Thanks 🙏🏻

PR #944, opened 2022-12-20
SW-2558 Change font size on withdrawals details pages

karthikbtf:
- match figma design
- also updated a padding on the map (unrelated change)

Terraware App 2022-12-20 07-27-21

Terraware App 2022-12-20 07-27-03

### DIFF
--- a/src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
+++ b/src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
@@ -174,7 +174,7 @@ export default function NurseryWithdrawalsDetails({
             justifyContent='space-between'
             alignItems='center'
           >
-            <Typography color={theme.palette.TwClrTxt} fontSize='20px' fontWeight={600} fontStyle='italic'>
+            <Typography color={theme.palette.ClrTextFill} fontSize='24px' lineHeight='32px' fontWeight={600}>
               {withdrawal?.withdrawnDate}
             </Typography>
             {withdrawal?.purpose === OUTPLANT && hasPlots && (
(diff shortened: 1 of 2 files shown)


src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
```diff
@@ -174,7 +174,7 @@ export default function NurseryWithdrawalsDetails({
             justifyContent='space-between'
             alignItems='center'
           >
-            <Typography color={theme.palette.TwClrTxt} fontSize='20px' fontWeight={600} fontStyle='italic'>
+            <Typography color={theme.palette.ClrTextFill} fontSize='24px' lineHeight='32px' fontWeight={600}>
```
charlie-tf, line 177: This is a token from the old theme. Do we want to use it here?

karthikbtf: We don't have this color in the new theme. I added a comment in figma asking the new theme to be updated. We have a few other such usages in our codebase.

karthikbtf: see https://www.figma.com/file/yRSfer1xtWKAJr4IETPml1?node-id=2074:504340#330623402

charlie-tf: Thanks!

charlie-tf (approved):

src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
karthikbtf: I will create a ticket to track removal of old theme.

PR #947, opened 2022-12-20
SW-2584 Show withdrawal log menu after a withdrawal is created

karthikbtf:
- previously was checking on org change only
- also check once a withdrawal is created
- also fixed an infinite planting sites fetch error


src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -90,7 +90,7 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
   };
 
   const fetchPlantingSites = useCallback(async () => {
-    if (plantingSites.length) {
+    if (plantingSites) {
```
karthikbtf, line 93: this was being called infinitely because we update plantingsites object.
updated the type to be PlantingSite[] or undefined

constanzauanini (approved):
Worked in Vercel!

PR #950, opened 2022-12-21
SW-2586 Fix date column on withdrawal log not sortable

constanzauanini:
Add sortOrder to search nursery withdrawals api to allow sorting by date.
MUI front end default sorter doesn't work for dates.


karthikbtf (approved):
looks good on vercel, changes look good
confirmed that other tables sort works the same

PR #951, opened 2022-12-21
SW-2575 Hierarchical nav updates from web-components

karthikbtf:
- allow opening nav root without selection
- fix strings
- fix selected route missing batch withdrawals and reassignment
- test in vercel


constanzauanini (approved):
Thanks for replacing with the strings!

karthikbtf:
> Thanks for replacing with the strings!

Thanks 🙏

PR #953, opened 2022-12-21
SW-2595 Hide Reassignment Panel if no Reassignments

charlie-tf:
No Reassignments:
image

Reassignments Present:
image

### DIFF
--- a/src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
+++ b/src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
@@ -191,7 +191,7 @@ export default function NurseryWithdrawalsDetails({
         </Box>
       </PageHeaderWrapper>
       <div ref={contentRef}>
-        {withdrawal?.purpose === OUTPLANT && (
+        {withdrawal?.purpose === OUTPLANT && withdrawalSummary?.hasReassignments && (
           <TabContext value={selectedTab}>
             <Box
               sx={{
@@ -238,6 +238,19 @@ export default function NurseryWithdrawalsDetails({
               />
             </TabPanel>
           </TabContext>
+        )}
+        {withdrawal?.purpose === OUTPLANT && !withdrawalSummary?.hasReassignments && (
+          <Box sx={contentPanelProps}>
+            <WithdrawalTabPanelContent
+              organization={organization}
+              species={species}
+              plotNames={plotNames}
+              withdrawal={withdrawal}
+              withdrawalSummary={withdrawalSummary}
+              delivery={delivery}
+              batches={batches}
+            />
+          </Box>
         )}
         {withdrawal?.purpose !== OUTPLANT && (
           <Box sx={contentPanelProps}>


src/components/NurseryWithdrawals/NurseryWithdrawalsDetails.tsx
```diff
@@ -239,6 +239,19 @@ export default function NurseryWithdrawalsDetails({
             </TabPanel>
           </TabContext>
         )}
+        {withdrawal?.purpose === OUTPLANT && !withdrawalSummary?.hasReassignments && (
+          <Box sx={contentPanelProps}>
```
karthikbtf, line 243: we should probably title this section as 'Details' instead of 'Withdrawal' to stay consistent with the other views that don't have tabs?
Also.. seems like a disabled Reassigment tab would be most helpful here instead of hiding the tab - which would confuse the users. But.. that would be a PM decision.

charlie-tf: Thanks Karthik. @rachel-tf-design what are your thoughts here?

charlie-tf: Looks good as in, you're happy with it as is? Or Karthik's suggestions look good? Or?

rachel-tf-design: Looks good as in I'm happy with it as is. I'd prefer not disabling the reassignments tab because we don't expect reassignment to happen much nor are we really encouraging it, so to me, it makes more sense to have it hidden until it's necessary to show.

charlie-tf: 👍 alright. Thank you for the clarification!

karthikbtf: > Looks good as in I'm happy with it as is. I'd prefer not disabling the reassignments tab because we don't expect reassignment to happen much nor are we really encouraging it, so to me, it makes more sense to have it hidden until it's necessary to show.
If we aren't encouraging it - the dedicated reassign column in the withdrawal log sends a mixed message.

rachel-tf-design: Yeah, the design and product team discussed this a while back and debated where to put that action (i.e. reassign). Since we're not encouraging it, we didn't want to make it a primary page level button but also, since it's the only real action a user can take on that page, we didn't want to hide it in a kebab icon. As a result, we settled on make it a secondary row level button.

karthikbtf: > Yeah, the design and product team discussed this a while back and debated where to put that action (i.e. reassign). Since we're not encouraging it, we didn't want to make it a primary page level button but also, since it's the only real action a user can take on that page, we didn't want to hide it in a kebab icon. As a result, we settled on make it a secondary row level button.

Thanks @rachel-tf-design this context helps. I should read the design notes more often, my bad.

rachel-tf-design: No worries! This discussion was during the designing stage, so before we did the first handoff. I'll try to do a better job of documenting context/reasoning for design decisions in the future :)

karthikbtf (approved):
Oops, I missed approving this earlier.

PR #954, opened 2022-12-21
SW-2583 / SW-2574 Show planting site data for sites that don't have zones/plots

karthikbtf:
- code was only processing zones/plots driven info
- using a separate query now for totals and by species
- also show map on mobile for planting sites with plots even if there are no plants
- also show the plot selector for empty state in species chart

Terraware App 2022-12-21 09-19-00

### DIFF
--- a/src/components/Plants/PlantingSiteDetails.tsx
+++ b/src/components/Plants/PlantingSiteDetails.tsx
@@ -11,7 +11,6 @@ import { ServerOrganization } from 'src/types/Organization';
 
 export type Population = {
   species_scientificName: string;
-  species_commonName: string;
   totalPlants: number;
 };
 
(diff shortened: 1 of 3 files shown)


src/components/Plants/PlantingSiteDetails.tsx
```diff
@@ -27,6 +26,11 @@ export type PlantingSiteZone = {
   plots: PlantingSitePlot[];
 };
 
+export type PlantingSiteSpecies = {
```
constanzauanini, line 29: This has the same attributes than Population, could we use that type?

karthikbtf: good catch 👍

karthikbtf: fixed

karthikbtf:
@constanzauanini i also added fix for SW-2574
- also show map on mobile for planting sites with plots even if there are no plants
- also show the plot selector for empty state in species chart

constanzauanini (approved):
Looks good! Tested mobile on vercel

karthikbtf:
> Looks good! Tested mobile on vercel

Thank you 🙏

PR #958, opened 2022-12-21
SW-2601 Withdrawal flow: filter nurseries by ready quantity (if outplant)

karthikbtf:
- same for batches so we don't show batches with 0 ready quantity on the batch-quantity page

Terraware App 2022-12-21 15-48-55

Terraware App 2022-12-21 15-47-45

Terraware App 2022-12-21 15-53-27

### DIFF
--- a/src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
+++ b/src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
@@ -297,7 +297,9 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
       plotId: isOutplant ? localRecord.plotId : undefined,
       facilityId: Number(selectedNursery as string),
       batchWithdrawals: batches
-        .filter((batch) => batch.facility_id.toString() === selectedNursery)
+        .filter((batch) => {
+          return batch.facility_id.toString() === selectedNursery && (!isOutplant || +batch.readyQuantity > 0);
+        })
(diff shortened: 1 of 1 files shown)


src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -315,12 +317,19 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
...
-    }, {});
+    const nurseries = batches
+      .filter((batch) => {
+        if (isOutplant) {
+          return +batch.readyQuantity > 0;
+        }
+        return +batch.totalQuantity > 0;
+      })
+      .reduce((acc, batch) => {
+        if (!acc[batch.facility_id.toString()]) {
```
constanzauanini, line 328: I think we could include the filter condition above in this same reduce:

`  if ((!isOutplant || +batch.readyQuantity > 0) && !acc[batch.facility_id.toString()]) {`

(didn't tested, just thinking)

constanzauanini (approved):
Lgtm! Just one suggestion, not necessary

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
karthikbtf: That will also work. Being functional is easier to read, we should leverage functional programming more if we can. Unless performance is a concern.

PR #959, opened 2022-12-22
SW-2587 Avoid purpose label turn blue on selected

constanzauanini:


src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -450,7 +450,15 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
             </Grid>
             <Grid xs={12} padding={theme.spacing(4, 0, 0)}>
               <FormControl>
-                <FormLabel sx={{ color: theme.palette.TwClrTxtSecondary, fontSize: '14px' }}>
+                <FormLabel
```
karthikbtf, line 453: We could make this a group component at some point.

PR #960, opened 2022-12-22
SW-2594 Fix Zone field's width changing when selecting planting site

constanzauanini:
When zone dropdown has values, the width of the field was increasing. Set flex={1} to both fields (zone and plot) to ensure width doesn't change

### DIFF
--- a/src/components/PlotSelector/index.tsx
+++ b/src/components/PlotSelector/index.tsx
@@ -104,11 +104,10 @@ export default function PlotSelector(props: PlotSelectorProps): JSX.Element {
       display='flex'
       flexWrap='wrap'
       flexDirection={isMobile ? 'column' : 'row'}
-      flexGrow={1}
       className={horizontalLayout ? `${classes.horizontal}` : ''}
     >
       <Box
-        flexGrow={1}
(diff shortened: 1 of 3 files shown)


src/components/PlotSelector/index.tsx
```diff
@@ -104,11 +104,10 @@ export default function PlotSelector(props: PlotSelectorProps): JSX.Element {
       display='flex'
       flexWrap='wrap'
       flexDirection={isMobile ? 'column' : 'row'}
-      flexGrow={1}
       className={horizontalLayout ? `${classes.horizontal}` : ''}
     >
       <Box
-        flexGrow={1}
+        flex={1}
```
karthikbtf, line 110: Can you test that resizing window works in the chart and the purpose form?

karthikbtf (approved):

src/components/PlotSelector/index.tsx
karthikbtf: looks good in vercel

constanzauanini: Yes, I tested it and it works!

PR #962, opened 2022-12-27
SW-2582 Validate withdrawal quantity for single/outplant withdrawal

constanzauanini:
Add quantity validation for single withdrawal with outplant purpose. Previously, the validation was happening on the backend in the last step of the flow (after photo upload)

### DIFF
--- a/src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
+++ b/src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
@@ -192,6 +192,11 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
         if (isNaN(withdrawnQuantity)) {
           setIndividualError('withdrawnQuantity', strings.INVALID_VALUE);
           return false;
+        } else {
+          if (withdrawnQuantity > +batches[0].readyQuantity) {
+            setIndividualError('withdrawnQuantity', strings.WITHDRAWN_QUANTITY_ERROR);
+            return false;
+          }
         }
       } else {
         setIndividualError('withdrawnQuantity', strings.REQUIRED_FIELD);


src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -192,6 +192,11 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
         if (isNaN(withdrawnQuantity)) {
           setIndividualError('withdrawnQuantity', strings.INVALID_VALUE);
           return false;
+        } else {
+          if (withdrawnQuantity > +batches[0].readyQuantity) {
```
charlie-tf, line 196: Will validation still be deferred to the backend if there is more than one batch? Is it possible to validate multiple batches in the frontend?

charlie-tf (approved):
Looks good, just a question.

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
constanzauanini: Yes, multiple batches are already validated in a different view (the next step)

karthikbtf: Charlie, the bulk flow validation happens here https://github.com/terraware/terraware-web/blob/main/src/components/Inventory/withdraw/flow/SelectBatchesWithdrawnQuantity.tsx#L162

Fyi, here's a jira https://terraformation.atlassian.net/browse/SW-2605 I created recently - the single batch flow does not inform user of remaining quantities (except in outplant withdrawal).  We may follow the [input]/remaining pattern soon for single batches (just like we do in bulk and reassignment). My recommendation to design is to also consider using the same flow for both single and bulk. More pros than cons to that.

PR #963, opened 2022-12-27
SW-2569 Add missing validations to dates when creating viability test

constanzauanini:
Screenshot 2022-12-27 at 16 34 06

### DIFF
--- a/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
+++ b/src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
@@ -192,6 +192,34 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
     return true;
   };
 
+  const validateStartDate = () => {
+    if (!record?.startDate) {
+      setIndividualError('startDate', strings.REQUIRED_FIELD);
+      return false;
+    } else {
+      const startDateMs = new Date(record.startDate).getTime();
+      if (isNaN(startDateMs)) {
+        setIndividualError('startDate', strings.INVALID_DATE);
+        return false;
+      } else {
+        if (isInTheFuture(startDateMs)) {
+          setIndividualError('startDate', strings.NO_FUTURE_DATES);
+          return false;
+        } else {
+          if (accession.collectedDate) {
+            const collectedDateMs = new Date(accession.collectedDate).getTime();
+            if (startDateMs < collectedDateMs) {
+              setIndividualError('startDate', strings.VIABILITY_TEST_START_DATE_ERROR);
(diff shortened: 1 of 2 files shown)


src/strings/index.tsx
```diff
@@ -642,6 +642,7 @@ const strings = new LocalizedStrings({
     TOTAL_SEEDS_GERMINATED_ERROR: 'Germinated amount should not be more than tested amount.',
     INVALID_VALUE: 'Invalid Value',
     RECORDING_DATE_ERROR: 'This should be later than Start Date',
+    VIABILITY_TEST_START_DATE_ERROR: 'This should be later than the date seeds were collected',
```
karthikbtf, line 645: This may be technically incorrect.. can it be the same date as seed collection date (we may need to check with whomever gave you this text)? 
If no, we need to update the check on line 211 to <= in viability test modal.
If yes, we need to update this string.

src/components/accession2/viabilityTesting/NewViabilityTestModal.tsx
```diff
@@ -192,6 +192,34 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
     return true;
   };
 
+  const validateStartDate = () => {
```
karthikbtf, line 195: Unrelated to this PR.. we have a lot of such date validate code..
We could start thinking about a date validator utility that accepts some rules for validation. Maybe as part of the timezone effort.

karthikbtf (commented):
looks good, i have one comment for follow up

src/strings/index.tsx
constanzauanini: I created the text but the requirement is: "viability test start dates should not be allowed to start before the date that the seeds were collected", so maybe the text is incorrect

constanzauanini: Updated

src/strings/index.tsx
```diff
@@ -642,7 +642,7 @@ const strings = new LocalizedStrings({
     TOTAL_SEEDS_GERMINATED_ERROR: 'Germinated amount should not be more than tested amount.',
     INVALID_VALUE: 'Invalid Value',
     RECORDING_DATE_ERROR: 'This should be later than Start Date',
-    VIABILITY_TEST_START_DATE_ERROR: 'This should be later than the date seeds were collected',
+    VIABILITY_TEST_START_DATE_ERROR: 'This should not be before the date seeds were collected',
```
karthikbtf, line 645: thanks..
add 'when'...
..before the date when seeds were...

karthikbtf (approved):

PR #964, opened 2022-12-28
SW-659 Add search functionality to users list

constanzauanini:
Screenshot 2022-12-28 at 15 31 19

### DIFF
--- a/src/components/People/index.tsx
+++ b/src/components/People/index.tsx
@@ -2,7 +2,6 @@ import React, { useEffect, useRef, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import {
   deleteOrganization,
-  getOrganizationUsers,
   leaveOrganization,
   listOrganizationRoles,
   UpdateOrganizationResponse,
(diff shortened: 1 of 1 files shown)


src/components/People/index.tsx
```diff
@@ -93,6 +109,56 @@ export default function PeopleList({ organization, reloadData, user }: PeopleLis
...
+            role: result.roleName as AllOrganizationRoles,
+            addedTime: result.createdTime as string,
+          });
+        });
+        if (getRequestId('searchUsers') === requestId) {
+          setResults(usersResults);
+        }
+      } else {
+        setResults(people);
+      }
```
karthikbtf, line 157: we are using a combination of real-time search results and cached results.
can we switch to using real-time results every time? just like what we discussed in other search related PRs

constanzauanini: Yes, I forgot about that. Will update it

karthikbtf (approved):
Thank you 🙏

PR #967, opened 2022-12-28
SW-2036 Combine terraware-web strings into a single file

constanzauanini:


charlie-tf (approved):
Great work! Thanks so much for doing these!

PR #975, opened 2022-12-30
SW-2614 Avoid zooming into zone on plot selection

karthikbtf:
- zoom into zone on new zone selection only
- bug was from useEffect triggerring even when there were no changes to selected zone
- behavior was due to parent prop being reconstructed from scratch even if a plot was changed
- fixed that by using separate memoization for plot and zone objects


constanzauanini:
In which planting site the zone was being zoomed when changing the plot? I was testing but couldn't find an example

karthikbtf:
> In which planting site the zone was being zoomed when changing the plot? I was testing but couldn't find an example

Thanks for testing, Constanza. If you zoom into the map and change the plot, the zone is zoomed back to again. This was not part of the requirement (and it was a bug tbh).

constanzauanini (approved):
I was able to test it now! Good fix!

PR #976, opened 2022-12-30
SW-2056 Remove hardcoded colors from index.css

karthikbtf:
- we don't need this style, probably an artifact from initial project setup
- all our pages are styled by themes and override this index.css background


charlie-tf (approved):
Thanks for doing this, @karthikbtf

PR #979, opened 2022-12-30
SW-2608 Add IDs to some components to allow for easier tracking

alexTerraformation:
In GTM we have some triggers on appearance of error text labels.  When reporting which label appeared it is useful to have a value in the "for" attribute.

### DIFF
--- a/src/components/accession2/edit/EditLocationModal.tsx
+++ b/src/components/accession2/edit/EditLocationModal.tsx
@@ -89,7 +89,7 @@ export default function EditLocationModal(props: EditLocationModalProps): JSX.El
       <Grid item xs={12} textAlign='left'>
         <Grid item xs={12}>
           <StorageLocationSelector
-            id='location'
+            id='edit-location'
             label={strings.LOCATION}
             selectedStorageLocation={seedBanks.find((sb) => sb.id === record.facilityId)}
             storageLocations={seedBanks}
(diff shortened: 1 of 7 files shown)


src/components/accession2/properties/StorageSubLocationSelector.tsx
```diff
@@ -3,6 +3,7 @@ import strings from 'src/strings';
 import { Select } from '@terraware/web-components';
 
 type StorageSubLocationSelectorProps = {
+  id?: string;
```
karthikbtf, line 6: use the id below?

src/components/accession2/edit/EditLocationModal.tsx
```diff
@@ -89,6 +89,7 @@ export default function EditLocationModal(props: EditLocationModalProps): JSX.El
       <Grid item xs={12} textAlign='left'>
         <Grid item xs={12}>
           <StorageLocationSelector
+            id='location'
```
karthikbtf, line 92: not sure if you want to distinguish between this use-case and the other.. if so, maybe call it `edit-location` .. etc.

karthikbtf (approved):
🎉 thanks.. some comments

src/components/accession2/edit/EditLocationModal.tsx
karthikbtf: i guess the page url helps distinguish here,, since these flows aren't modals

alexTerraformation: Good practice anyways!

karthikbtf (approved):

PR #986, opened 2023-01-03
SW-2628 Planting Site Creation Map vs No-Map

charlie-tf:
- add modal to choose between having map data or not when creating
  a new planting site.
- when 'with map' is selected, show a modal with a link to a contact 
  form page saying that they will need help from Terraformation
- when 'without map' is selected, proceed to the planting site creation
  flow
- don't show the Boundaries and Plot section for planting sites without
  map data during creating, viewing, or editing

image
image
image
image

### DIFF
--- a/src/api/search/index.ts
+++ b/src/api/search/index.ts
@@ -21,7 +21,7 @@ export function convertToSearchNodePayload(
   criteria: SearchCriteria,
   organizationId?: number
 ): SearchNodePayload | undefined {
-  if (criteria === {}) {
+  if (Object.keys(criteria).length === 0) {
     return undefined;
   }
   let newCriteria = criteria;
--- a/src/components/PlantingSites/PlantingSiteCreate.tsx
+++ b/src/components/PlantingSites/PlantingSiteCreate.tsx
@@ -173,7 +173,7 @@ export default function CreatePlantingSite(props: CreatePlantingSiteProps): JSX.
                   </Grid>
                 </Box>
               </Grid>
-              <BoundariesAndPlots plantingSite={record} />
+              {record?.boundary && <BoundariesAndPlots plantingSite={record} />}
             </>
           )}
         </Container>
--- a/src/components/PlantingSites/PlantingSiteSelectTypeModal.tsx
+++ b/src/components/PlantingSites/PlantingSiteSelectTypeModal.tsx
@@ -1,0 +1,83 @@
+import strings from 'src/strings';
+import Button from 'src/components/common/button/Button';
+import React, { useState } from 'react';
+import DialogBox from 'src/components/common/DialogBox/DialogBox';
+import { FormControlLabel, Radio, RadioGroup, Theme, Typography, useTheme } from '@mui/material';
+import { makeStyles } from '@mui/styles';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  buttonSpacing: {
+    marginRight: theme.spacing(2),
+  },
+}));
+
+export type PlantingSiteSelectTypeModalProps = {
+  open: boolean;
+  onNext: (goToCreate: boolean) => void;
+  onClose: () => void;
+};
+
+export default function PlantingSiteSelectTypeModal(props: PlantingSiteSelectTypeModalProps): JSX.Element {
+  const { open, onNext, onClose } = props;
+  const classes = useStyles();
+  const theme = useTheme();
+  const [withMap, setWithMap] = useState<boolean | null>(null);
+
+  const handleClose = () => {
+    setWithMap(null);
+    onClose();
+  };
+
+  const handleNext = () => {
+    onNext(withMap === false);
+    handleClose();
+  };
+
+  const handleTypeChange = (_: React.ChangeEvent<HTMLInputElement>, value: string) => {
+    if (value === 'withMap') {
+      setWithMap(true);
+    } else {
+      setWithMap(false);
+    }
+  };
+
+  return (
+    <DialogBox
+      scrolled
+      onClose={handleClose}
+      open={open}
+      title={strings.SELECT_PLANTING_SITE_TYPE}
+      size={'medium'}
+      middleButtons={[
+        <Button
+          onClick={handleClose}
+          id='cancel'
+          label={strings.CANCEL}
+          priority='secondary'
+          type='passive'
+          className={classes.buttonSpacing}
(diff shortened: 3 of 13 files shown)


src/components/PlantingSites/PlantingSiteSelectTypeModal.tsx
```diff
@@ -0,0 +1,83 @@
+import strings from 'src/strings';
+import Button from 'src/components/common/button/Button';
```
karthikbtf, line 2: you could
```
import { Button, DialogBox } from '@terraware/web-components;
```

src/components/PlantingSites/PlantingSiteSelectTypeModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+      </Typography>
+      <Typography marginBottom={theme.spacing(3)} justifyContent='center'>
+        {strings.SELECT_PLANTING_SITE_TYPE_DESCRIPTION_3}
+      </Typography>
+
+      <Typography color={theme.palette.TwClrTxtSecondary} display='flex' fontSize={14}>
+        {strings.PLANTING_SITE_TYPE}
+      </Typography>
+      <RadioGroup defaultValue={null} name='radio-buttons-group' onChange={handleTypeChange}>
+        <FormControlLabel value='withMap' control={<Radio />} label={strings.WITH_MAP} />
```
karthikbtf, line 78: we have a radio button in web-components.. i see now we aren't using it in other places either.. maybe constanza knows. fine for now.

src/components/PlantingSites/PlantingSiteWithMapHelpModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <DialogBox
+      scrolled
+      onClose={onClose}
+      open={open}
+      title={strings.SELECT_PLANTING_SITE_TYPE}
+      size={'medium'}
+      middleButtons={[<Button onClick={onClose} id='done' label={strings.DONE} key='button-1' />]}
+    >
+      <Typography>
+        {parseHtml(
```
karthikbtf, line 27: this is cool!
fwiw, see https://terraformation.atlassian.net/wiki/spaces/PM/pages/708280327/PRD+Support+Spanish+localization#Stop-mixing-link-and-non-link-words-in-phrases-(preferred-approach) for preferred approach.. not sure if it's less of an issue here but just to check

src/components/emptyStatePages/EmptyStatePage.tsx
```diff
@@ -296,6 +309,23 @@ export default function EmptyStatePage({
...
+              setPlantingSiteTypeHelpModalOpen(true);
+            }
+          }}
+          onClose={() => setPlantingSiteTypeModalOpen(false)}
+        />
+        <PlantingSiteWithMapHelpModal
+          open={plantingSiteTypeHelpModalOpen}
+          onClose={() => setPlantingSiteTypeHelpModalOpen(false)}
+        />
+      </>
```
karthikbtf, line 328: extract out as a component for reuse in both places?

src/components/PlantingSites/PlantingSiteCreate.tsx
```diff
@@ -173,7 +173,7 @@ export default function CreatePlantingSite(props: CreatePlantingSiteProps): JSX.
                   </Grid>
                 </Box>
               </Grid>
-              <BoundariesAndPlots plantingSite={record} />
+              {record?.boundary && <BoundariesAndPlots plantingSite={record} />}
```
karthikbtf, line 176: the boundaries and plots component has a rendering for sites without boundaries.. we should clean that up maybe?

karthikbtf (commented):
Looks great, a few comments for your consideration.
This looks good to go otherwise...

src/api/search/index.ts
```diff
@@ -21,7 +21,7 @@ export function convertToSearchNodePayload(
   criteria: SearchCriteria,
   organizationId?: number
 ): SearchNodePayload | undefined {
-  if (criteria === {}) {
+  if (Object.keys(criteria).length === 0) {
```
charlie-tf, line 24: changed because of 
```
src/api/search/index.ts:24:7 - error TS2839: This condition will always return 'false' since JavaScript compares objects by reference, not value.
```

src/components/PlantingSites/PlantingSiteWithMapHelpModal.tsx
charlie-tf: This was the cleanest way to add the link inline. The preferred approach would involve adding a button or something

src/components/PlantingSites/PlantingSiteCreate.tsx
charlie-tf: I'll remove that. thanks!

src/components/emptyStatePages/EmptyStatePage.tsx
charlie-tf: That'd be less repetition, so I'll make that change. Thanks

src/components/PlantingSites/PlantingSiteWithMapHelpModal.tsx
charlie-tf: from a translation standpoint we are looking at, "Please _{blank}_ to create a site..." where the {blank} is some call-to-action ("contact us" in this case). Maybe this doesn't translate well, or lend itself to other grammars, I'm not sure.

karthikbtf: Got it.. the preferred approach is to link the entire sentence. We can revisit this when we get to i18n for sure.
Also would be good to have a 'contact us' string that is already lower cased.. ideally we use translated strings as-is.

charlie-tf: Oh gotcha. thanks

karthikbtf: My bad.. ignore my previous comment. 
The string replace approach may break in general - depending on the language. As an example - “contact us” can theoretically be contracted when used in a full sentence.. in some language. 
Let’s defer all this for i18n discussions.
Code looks great as-is for now.

karthikbtf (approved):
Nice 🎉

PR #987, opened 2023-01-04
SW-2640 Fix theme initialization

constanzauanini:
- Use web-components version with mui/styles version updated (previous to revert)
- Upgrade mui/styles version in terraware-web also, to be the same as web-components
- Regenerate yarn.lock file to avoid having @mui/* dependencies messed up


karthikbtf (approved):
code changes look good

karthikbtf:
terraware/web-components version 2.0.58 seems to work just fine for me.
I'm very confused why this PR or what is actually broken.

karthikbtf:
we can close this?

constanzauanini:
> we can close this?

Can we wait until I rebase it with main after charlie's changes, and generating a new web-components version with mui/styles updated?

karthikbtf:
> > we can close this?
> 
> Can we wait until I rebase it with main after charlie's changes, and generating a new web-components version with mui/styles updated?

Just to confirm, the plan is to pick up new mui-styles changes? Undo the revert of mui-styles update?

karthikbtf:
> > > > we can close this?
> > > 
> > > 
> > > Can we wait until I rebase it with main after charlie's changes, and generating a new web-components version with mui/styles updated?
> > 
> > 
> > Just to confirm, the plan is to pick up new mui-styles changes? Undo the revert of mui-styles update?
> 
> Yes, use the new version of mui-styles in web components, but I don't want to revert the revert. Could we re-do the update and generate a new version?

we can apply this patch https://github.com/terraware/web-components/pull/225
i already generated a branch for the revert of the revert (which is totally fine too and may accurately express our historical intention - but if we aren't confident yes make changes from scratch)

constanzauanini:
> > > > > we can close this?
> > > > 
> > > > 
> > > > Can we wait until I rebase it with main after charlie's changes, and generating a new web-components version with mui/styles updated?
> > > 
> > > 
> > > Just to confirm, the plan is to pick up new mui-styles changes? Undo the revert of mui-styles update?
> > 
> > 
> > Yes, use the new version of mui-styles in web components, but I don't want to revert the revert. Could we re-do the update and generate a new version?
> 
> we can apply this patch [terraware/web-components#225](https://github.com/terraware/web-components/pull/225) i already generated a branch for the revert of the revert (which is totally fine too and may accurately express our historical intention - but if we aren't confident yes make changes from scratch)

Could you create a PR in web-components with the revert of the revert?

karthikbtf:
> > > > > > we can close this?
> > > > > 
> > > > > 
> > > > > Can we wait until I rebase it with main after charlie's changes, and generating a new web-components version with mui/styles updated?
> > > > 
> > > > 
> > > > Just to confirm, the plan is to pick up new mui-styles changes? Undo the revert of mui-styles update?
> > > 
> > > 
> > > Yes, use the new version of mui-styles in web components, but I don't want to revert the revert. Could we re-do the update and generate a new version?
> > 
> > 
> > we can apply this patch [terraware/web-components#225](https://github.com/terraware/web-components/pull/225) i already generated a branch for the revert of the revert (which is totally fine too and may accurately express our historical intention - but if we aren't confident yes make changes from scratch)
> 
> Could you create a PR in web-components with the revert of the revert?

https://github.com/terraware/web-components/pull/237

src/components/Inventory/InventoryCellRenderer.tsx
```diff
@@ -37,7 +37,9 @@ export default function InventoryCellRenderer(props: RendererProps<TableRowType>
   };
 
   const createLinkToInventoryDetail = (iValue: React.ReactNode | unknown[]) => {
-    return <Link to={APP_PATHS.INVENTORY_ITEM.replace(':speciesId', row.species_id.toString())}>{iValue}</Link>;
+    if (!Array.isArray(iValue)) {
```
karthikbtf, line 40: sorry to suggest this so late.. can we make this check inside the Link component instead? That could avoid touching many files

constanzauanini: To do this we should allow children prop of Link component, to be a ReactNode or unknown[]; and then check that if its an array don't pass it, do we want that? It doesn't seem correct to me that children can be of type unknown[]

karthikbtf: Adding the isArray check everywhere is not going to be fun and something we will have to do everytime we use Link. None of the use cases are arrays. This is very weird.
Not sure what the real problem is here but handling this in Link would be a better thing to do. Or maybe cast every use case to a ReactNode instead of doing an isarray check.
Or we could do the ReactNode or unknown[] for now with a comment in the code.
I don't know - will leave it to you.

karthikbtf: we can revisit this later for sure if you want to get this in as-is

constanzauanini: I casted every use case for now and we can revisit later

karthikbtf: awesome, thanks @constanzauanini this is much better from a usage standpoint.. looks good 🎉

karthikbtf (approved):

PR #988, opened 2023-01-04
SW-2561 Remove Several Old Theme Values

charlie-tf:


src/App.tsx
```diff
@@ -125,7 +125,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     margin: 'auto',
     minHeight: '100vh',
     '& .MuiCircularProgress-svg': {
-      color: theme.palette.ClrIconFillProductive,
+      color: theme.palette.TwClrIcnBrand,
```
charlie-tf, line 128: @hannah-cai @rachel-tf-design Does this seem like a suitable replacement? It is `#007DF2` (old) vs `#2C8658` (new) and is for the initial large app loading spinner.

rachel-tf-design: @charlie-tf Yes, that's the correct replacement

karthikbtf (approved):
🙏

PR #989, opened 2023-01-04
SW-2535 Pass user_id and internal_user semantics to google tag manager configuration

karthikbtf:
This is an iterative attempt to figure out how to use google analytics + reporting for our purposes.
Alex has already created/configured the user_id and internal_user properties in google tag manager.
Here, we wait until we have a user and then initialize google tag manager with these property values pushed onto the data layer.
Downside to this approach is we won't have analytics for unauthenticated views at this point.
Something we will figure out down the road.


public/index.html
```diff
@@ -13,27 +13,44 @@
...
+            internalUser,
+            'GTM-TNX8NWM',
+            '&gtm_auth=oxl4mxgvnvET8ov_75EuYA&gtm_preview=env-1&gtm_cookies_win=x'
+          );
+        };
       } else {
-        // Uncomment this line to track non-staging/prod environments
-        // gtm(window, document, 'script', 'dataLayer', 'GTM-TNX8NWM', '&gtm_auth=1591mjZEVL_pulZkOX4gWg&gtm_preview=env-11&gtm_cookies_win=x');
+        // Uncomment this function to track non-staging/prod environments
+        window.INIT_GTAG = function (userId, internalUser) {
```
karthikbtf, line 46: this will be commented out prior to checking-in, for new we need this to test in vercel

src/App.tsx
```diff
@@ -312,10 +315,17 @@ export default function App() {
       const response = await getUser();
       if (response.requestSucceeded) {
         setUser(response.user ?? undefined);
+        if (response.user && !userState?.gtmInstrumented && (window as any).INIT_GTAG) {
```
karthikbtf, line 318: using recoil to keep state of whether this was instrumented.. we are loading the user multiple times depending on the view.. we don't want to re-init gtag

charlie-tf (approved):
Pretty easy to follow. Nice work! 🎉

constanzauanini (approved):

karthikbtf:
@alexTerraformation I will get this into staging for now

PR #985, opened 2023-01-03
SW-2629 Set id to all onChange() functions

constanzauanini:
This is because new version of web-component doesn't return the id from the component, so we need to explicitly set it. After this change any component's id can be changed to whatever id we need in the HTML, because data model id is set in the onChange function

PR #991, opened 2023-01-05
SW-2565 Use New Theme Only

charlie-tf:
Consume web-components 2.0.61


karthikbtf:
Some of the errors will be fixed in Constanza's PR https://github.com/terraware/terraware-web/pull/985
All the web-components' onChange signatures were updated from `(id, value)` to just `(value)`
Seems like we may need one PR -  containing both = Constanza's changes and other fixes that resolve remaining build failures.
Fwiw, we were going to hold off on merging the onChange related changes until after tracking-v1 is deployed to production.

karthikbtf:
> Some of the errors will be addressed in Constanza's PR #985 All the web-components' onChange signatures were updated from `(id, value)` to just `(value)` Seems like we may need one PR - containing both = Constanza's changes and other fixes that resolve remaining build failures. Fwiw, we were going to hold off on merging the onChange related changes until after tracking-v1 is deployed to production.

oops.. we won't need one PR. maybe get the other PR in first followed up this (rebased with changes from other PR). my bad.

constanzauanini:
> > Some of the errors will be addressed in Constanza's PR #985 All the web-components' onChange signatures were updated from `(id, value)` to just `(value)` Seems like we may need one PR - containing both = Constanza's changes and other fixes that resolve remaining build failures. Fwiw, we were going to hold off on merging the onChange related changes until after tracking-v1 is deployed to production.
> 
> oops.. we won't need one PR. maybe get the other PR in first followed by this (rebased with changes from other PR). my bad.

We may need to merge this one https://github.com/terraware/terraware-web/pull/987, since the other one is breaking with the theme initialization (the web/styles update in web-components is reverted in the version that that PR uses, and it stills break)

constanzauanini (approved):

PR #996, opened 2023-01-05
SW-2667 Add a user-provider in the web-app react tree

karthikbtf:
- this PR primarily introduces how to use a provider
- introduced a directory and export structure, this is totally changeable and not set in stone
- application of user provider is a bit forced at this point since we don't have that many uses
- we can re-evaluate this later, good start for now
- tested all occurrences of user/reloadUser
- cleaned up props

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -62,6 +62,7 @@ import PlantsDashboard from './components/Plants';
 import { NurseryWithdrawals, NurseryWithdrawalsDetails, NurseryReassignment } from './components/NurseryWithdrawals';
 import { listPlantingSites } from './api/tracking/tracking';
 import { PlantingSite } from './api/types/tracking';
+import { UserProvider } from './providers';
 
 interface StyleProps {
   isDesktop?: boolean;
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -750,6 +742,13 @@ export default function App() {
           </ErrorBoundary>
         </div>
       </div>
+    </>
+  );
+
+  return (
+    <StyledEngineProvider injectFirst>
+      <CssBaseline />
+      <UserProvider data={{ user, reloadUser }}>{getContent()}</UserProvider>
```
karthikbtf, line 751: ideally we fetch the user within the provider layer, for now i'm keeping the existing fetching sequence just not to break anything. will revisit this later.

src/components/NavBar.tsx
```diff
@@ -94,6 +94,7 @@ export default function NavBar({
           closeAndNavigateTo(APP_PATHS.INVENTORY);
         }}
         id='inventory'
+        key='inventory'
```
constanzauanini, line 97: What is this key for?

constanzauanini (approved):
Looks good!

src/components/NavBar.tsx
karthikbtf: This is an array of nav items, react was complaining the elements needed keys.

constanzauanini: Thanks!

src/components/MyAccount/index.tsx
```diff
@@ -33,14 +34,36 @@ const columns: TableColumnType[] = [
...
+type MyAccountContentProps = {
   user: User;
   organizations?: ServerOrganization[];
   edit: boolean;
   reloadUser: () => void;
   reloadData?: () => void;
 };
 
-export default function MyAccount({ user, organizations, edit, reloadUser, reloadData }: MyAccountProps): JSX.Element {
+const MyAccountContent = ({
```
charlie-tf, line 60: I like your refactor strategy here 👍

charlie-tf (approved):
Nice!

src/providers/hooks.ts
```diff
@@ -0,0 +1,4 @@
+import { useContext } from 'react';
+import { UserContext } from './contexts';
+
+export const useUser = () => useContext(UserContext);
```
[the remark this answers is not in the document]
charlie-tf: I like it!

PR #997, opened 2023-01-06
SW-2666 Add opt-in flag for timezones

constanzauanini:


src/features.ts
```diff
@@ -34,6 +34,15 @@ export const OPT_IN_FEATURES: Feature[] = [
     get: env().isForcedProductionView,
     set: env().forceProductionView,
   },
+  {
+    name: 'Timezones',
+    preferenceName: 'enableTimezones',
+    active: true,
+    enabled: false,
+    allowInternalProduction: false,
+    description: ['Allow choosing timezone'],
```
karthikbtf, line 43: time zone (that's our preference as per the PRD)

karthikbtf (approved):
Yay! 🎉

src/features.ts
constanzauanini: Good catch! Thanks

PR #998, opened 2023-01-06
SW-2674 Add timezones BE api

constanzauanini:


src/api/timezones/timezones.ts
```diff
@@ -0,0 +1,14 @@
...
+const TIMEZONES_ENDPOINT = '/api/v1/i18n/timeZones';
+
+type ListTimeZoneNamesResponsePayload =
+  paths[typeof TIMEZONES_ENDPOINT]['get']['responses'][200]['content']['application/json'];
+
+export type TimeZone = ListTimeZoneNamesResponsePayload['timeZones'][0];
+
+export const getTimeZones = async (): Promise<TimeZone[]> => {
+  const response: ListTimeZoneNamesResponsePayload = (await axios.get(TIMEZONES_ENDPOINT)).data;
+  return response.timeZones;
```
karthikbtf, line 13: should we do the try catch and populate with requestSucceeded etc.. just like we do other API?
like here in updateAccession2 https://github.com/terraware/terraware-web/blob/main/src/api/accessions2/accession.ts
fwiw, looks like we aren't doing try/catch for getAccession2 but unrelated to this PR

src/api/types/generated-schema.ts
```diff
@@ -1599,6 +1638,8 @@ export interface components {
       emailNotificationsEnabled: boolean;
       firstName?: string;
       lastName?: string;
+      /** Time zone name in IANA tz database format */
+      timeZone?: string;
```
karthikbtf, line 1642: i think we need to update the types under `/src/types` and `src/api/types` etc..

karthikbtf (commented):
looks good.. a few comments

src/api/timezones/timezones.ts
constanzauanini: Right, I just used getAccession2 as guide and forgot about try/catch

src/api/types/generated-schema.ts
constanzauanini: Planting Sites and facilities types are defined referencing this file, so no need to update. I just updated Organization and User type

karthikbtf (approved):
🙌

src/api/timezones/timezones.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type TimeZone = ListTimeZoneNamesResponsePayload['timeZones'][0];
+
+type GetTimeZonesResponse = {
+  requestSucceeded: boolean;
+  timeZones: TimeZone[] | undefined;
+  error?: string;
+};
+
+export const getTimeZones = async (): Promise<GetTimeZonesResponse> => {
```
karthikbtf, line 18: oops, missed this.. it supports an optional locale
locale?: string
this is important

PR #1005, opened 2023-01-09
SW-2679 Initialize user timezone

constanzauanini:
- Set user time zone to browser time zone only when user has no time zone. 
- @karthikbtf user here is being populated with getUser api. Should we refactor to use the UserProvider? 
- Update web components version needed to allow JSX.Elements on the Message component

Screenshot 2023-01-09 at 10 55 11


src/strings/index.tsx
```diff
@@ -940,6 +940,8 @@ const strings = new LocalizedStrings({
     UNSPECIFIED: 'Unspecified',
     UNSURE: 'Unsure',
     UPDATE_STATUS_WARNING: 'You‘re about to update the status',
+    UPDATED_TIMEZONE_TITLE: 'Time zone was updated to {0} zone.',
```
karthikbtf, line 943: `Your time zone..`?
In case we need to differentiate between user and org at some point

src/components/Home.tsx
```diff
@@ -75,7 +78,22 @@ export default function Home({ organizations, selectedOrganization, setSelectedO
     const populateUser = async () => {
       const response = await getUser();
       if (response.requestSucceeded && !cancel) {
-        setUser(response.user ?? undefined);
+        if (response.user) {
```
karthikbtf, line 81: I don't really know where a good place for this might be.
One way to to do this inside the UserProvider, have a useEffect that depends on user..
and check there.

src/components/Home.tsx
```diff
@@ -75,7 +78,22 @@ export default function Home({ organizations, selectedOrganization, setSelectedO
...
+          setUser(response.user);
+          if (!response.user.timeZone) {
+            const browserTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+            const updateResponse = await updateUserProfile({ ...response.user, timeZone: browserTimeZone });
```
karthikbtf, line 85: the only tricky part here is that the timezone needs to be among the support timezones returned by the BE.
We can keep this as-is for now but add a comment to check against BE timezones.

karthikbtf (commented):
looks good, some comments

src/components/Home.tsx
constanzauanini: Oh, ok. I thought the IANA database was only one

src/providers/UserProvider.tsx
```diff
@@ -8,5 +13,30 @@ export type UserProviderProps = {
...
+
+  useEffect(() => {
+    const populateTimeZone = async () => {
+      if (data.user) {
+        if (!data.user.timeZone) {
+          // Should validate bowser time zone with supported time-zones returned by backend
+          const browserTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+          const updateResponse = await updateUserProfile({ ...data.user, timeZone: browserTimeZone });
+          if (updateResponse.requestSucceeded) {
+            data.reloadUser();
```
karthikbtf, line 26: nice!

src/providers/UserProvider.tsx
```diff
@@ -8,5 +13,30 @@ export type UserProviderProps = {
...
+                strings.UPDATED_TIMEZONE_MSG,
+                <Link to={APP_PATHS.MY_ACCOUNT}>{strings.MY_ACCOUNT}</Link>
+              ),
+              strings.formatString(strings.UPDATED_TIMEZONE_TITLE, browserTimeZone)
+            );
+          }
+        }
+      }
+    };
+    populateTimeZone();
```
karthikbtf, line 38: can we call this only if the time zone opt-in feature is enabled?

karthikbtf (approved):
thanks for the changes... one comment

src/providers/UserProvider.tsx
constanzauanini: Right!

karthikbtf (approved):

src/components/Home.tsx
karthikbtf: The BE will provide the map of time zones like so:
{
  <time zone id>: <locale display name>
}
So 'America/Los_Angeles' can be displayed in user's preferred locale.
If the browser time zone is not in the BE provided mapping, we will use UTC.
We decided to use this approach since browsers have different supported lists.. we picked the common ones across browsers.
Since localization provider will be a layer below user provider, we can move this code into the localization provider once we have it.
My planning hasn't been too good so far..
the code is good as-is, we just need to move it later and add one extra check.

PR #1006, opened 2023-01-09
SW-2668 Organization Provider

charlie-tf:
- Add provider layer for Organization.
- useOrganization hook to get organization
- Remove prop drilling for selected organization.


src/providers/hooks.ts
```diff
@@ -1,4 +1,6 @@
 import { useContext } from 'react';
-import { UserContext } from './contexts';
+import { OrganizationContext, UserContext } from './contexts';
 
 export const useUser = () => useContext(UserContext);
+
+export const useOrganization = () => useContext(OrganizationContext);
```
karthikbtf, line 6: One way to remove the `!!` is to maybe introduce 2 hooks.
```
export const useOrganization = () : ServerOrganization => useContext(OrganizationContext) as ServerOrganization;
export const useRawOrganization = () => useContext(OrganizationContext);
```
The second one can be used in TopBarContent since it is used in the orphaned user path.
Could be hacky but perhaps ok?

karthikbtf: On rethinking.. since the major use-case is fetching the selectedOrganizationOnly.. 
ignore the above comment.
Add a new hook:
```
export const useOrganization = () : ServerOrganization => useContext(OrganizationContext).selectedOrganization as ServerOrganization;
```
Maybe rename the original `useOrganization` to something else for the smaller use-cases that need all the data.
In all usecases:
```
  const organization = useOrganization();
```
This would be a sweeping change.. but just some feedback.

karthikbtf: The last option would be to initialize the default org as non-null with some dummy values:
{
  name: '',
  id: -1,
 ...
}

and if id is -1, assume no org (only place this matters is TopBarContent)

charlie-tf: I think I understand what you're getting at. However, might the first one need to look like this, perhaps?
```typescript
export const useSelectedOrg = () => useContext(OrganizationContext).selectedOrganization as ServerOrganization;
```

charlie-tf: oop, didn't see your later message

charlie-tf: okay, that's making sense. Thanks so much for your suggestions! I'll get something in there

charlie-tf:
I went with the route of providing a default dummy org (with -1 id). This seemed like a clean approach. Thanks @karthikbtf for your discussion!

src/providers/DataTypes.ts
```diff
@@ -1,6 +1,15 @@
...
+import React from 'react';
 
 export type ProvidedUserData = {
   user?: User;
   reloadUser: () => void;
 };
+
+export type ProvidedOrganizationData = {
+  selectedOrganization: ServerOrganization;
+  setSelectedOrganization: React.Dispatch<React.SetStateAction<ServerOrganization | undefined>>;
```
karthikbtf, line 12: my 2cents.. i know we have been using this pattern of exposing setState from a parent.. that's usually an internal detail in parent.
would be cleaner to maybe:
setSelectedOrganization: (org?: ServerOrganization) => void;
and in parent:
setSelectedOrganization={org => setSelectOrganziation(org)}

also for another time.. thoughts on this just being
organization: 
setOrganization: 
...
in the provider (the selected is implied..) we could just have
const { organization } = useOrganization();
across the board.

just a question. this is great for now.

karthikbtf (approved):
Thanks @charlie-tf this is a ton of cleanup 🎉

src/App.tsx
```diff
@@ -352,17 +353,13 @@ export default function App() {
...
-                organizations={organizations}
-                selectedOrganization={selectedOrganization}
-                setSelectedOrganization={setSelectedOrganization}
-                reloadOrganizationData={reloadData}
-                setShowNavBar={setShowNavBar}
-              />
-            </TopBar>
-            <ToastSnackbar />
-            <NoOrgLandingPage reloadOrganizationData={reloadData} />
+            <OrganizationProvider data={{ selectedOrganization, setSelectedOrganization, organizations, reloadData }}>
```
karthikbtf, line 356: re: the typescript errors, my suggestion:
expose that default value in a constant and use that here (since we won't have a selectedOrganization in this path anyways)

src/App.tsx
```diff
@@ -748,7 +703,11 @@ export default function App() {
   return (
     <StyledEngineProvider injectFirst>
       <CssBaseline />
-      <UserProvider data={{ user, reloadUser }}>{getContent()}</UserProvider>
+      <UserProvider data={{ user, reloadUser }}>
+        <OrganizationProvider data={{ selectedOrganization, setSelectedOrganization, organizations, reloadData }}>
```
karthikbtf, line 707: for the typescript error, we could do a 
```
(selectedOrganization && (
   ...
)
```
on line 704.

Feels like our App.tsx could use some cleanup at some point. We added a lot of patch up style code recently.

charlie-tf: ooh, that's nice. it'd allow me to get rid of a whole bunch of `selectedOrganization &&`

karthikbtf (approved):
Nice,, and thanks!! 🎉

PR #1007, opened 2023-01-09
SW-2679 Validate that browser time zone is in backend supported time zones

constanzauanini:


src/providers/UserProvider.tsx
```diff
@@ -23,16 +24,25 @@ export default function UserProvider({ children, data }: UserProviderProps): JSX
...
-          const updateResponse = await updateUserProfile({ ...data.user, timeZone: browserTimeZone });
-          if (updateResponse.requestSucceeded) {
-            data.reloadUser();
-            snackbar.pageSuccess(
-              strings.formatString(
-                strings.UPDATED_TIMEZONE_MSG,
-                <Link to={APP_PATHS.MY_ACCOUNT}>{strings.MY_ACCOUNT}</Link>
-              ),
-              strings.formatString(strings.UPDATED_TIMEZONE_TITLE, browserTimeZone)
+          const backendTimeZonesResponse = await getTimeZones();
```
karthikbtf, line 27: The localization provider will be fetching and providing this data.
Can we wait until that is implemented - we can move this check there later?

src/providers/UserProvider.tsx
```diff
@@ -23,16 +24,25 @@ export default function UserProvider({ children, data }: UserProviderProps): JSX
...
+          if (backendTimeZonesResponse.requestSucceeded) {
+            const foundTimeZone = backendTimeZonesResponse.timeZones?.find(
+              (timeZone) => timeZone.id === browserTimeZone
             );
+            const updateResponse = await updateUserProfile({
+              ...data.user,
+              timeZone: foundTimeZone ? browserTimeZone : strings.UTC,
```
karthikbtf, line 34: UTC will be a key in the map, we can use that value.

src/providers/UserProvider.tsx
```diff
@@ -23,16 +24,25 @@ export default function UserProvider({ children, data }: UserProviderProps): JSX
...
-            data.reloadUser();
-            snackbar.pageSuccess(
-              strings.formatString(
-                strings.UPDATED_TIMEZONE_MSG,
-                <Link to={APP_PATHS.MY_ACCOUNT}>{strings.MY_ACCOUNT}</Link>
-              ),
-              strings.formatString(strings.UPDATED_TIMEZONE_TITLE, browserTimeZone)
+          const backendTimeZonesResponse = await getTimeZones();
+          if (backendTimeZonesResponse.requestSucceeded) {
+            const foundTimeZone = backendTimeZonesResponse.timeZones?.find(
```
karthikbtf, line 29: is timeZones possibly undefined in BE response type?.. that's interesting...

src/providers/UserProvider.tsx
karthikbtf: actually this is fine, not sure what i was thinking.. the key is still 'UTC'

karthikbtf: actually no, this is not fine .. Mondays are hard..
The key will always be 'UTC', it will not be translated.
basically:
```
timeZone: foundTimeZone?.id || 'UTC'
```

karthikbtf: @constanzauanini i think i'm wrong again.. give me a minute let me get back.

karthikbtf: Ok, 
```
 {
      "id": "Etc/UTC",
      "longName": "Coordinated Universal Time"
    }
```
The code should be:
```
timeZone: foundTimeZone?.id || 'Etc/UTC',
```

src/providers/UserProvider.tsx
```diff
@@ -23,16 +24,25 @@ export default function UserProvider({ children, data }: UserProviderProps): JSX
...
+              timeZone: foundTimeZone ? browserTimeZone : strings.UTC,
+            });
+            if (updateResponse.requestSucceeded) {
+              data.reloadUser();
+              snackbar.pageSuccess(
+                strings.formatString(
+                  strings.UPDATED_TIMEZONE_MSG,
+                  <Link to={APP_PATHS.MY_ACCOUNT}>{strings.MY_ACCOUNT}</Link>
+                ),
+                strings.formatString(strings.UPDATED_TIMEZONE_TITLE, foundTimeZone ? browserTimeZone : strings.UTC)
```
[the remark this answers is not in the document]
karthikbtf: we have to use the `longName` for zone whose `id === 'Etc/UTC'`

src/providers/UserProvider.tsx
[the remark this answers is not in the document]
karthikbtf: We should maybe add a comment to move the code to localization provider.. when it is ready?

src/providers/UserProvider.tsx
karthikbtf: We can get this in since you have already coded it up. No need to redo this.. we can just move the code later. Thoughts?

constanzauanini: Yes, let's move it later

src/providers/UserProvider.tsx
```diff
@@ -40,7 +43,10 @@ export default function UserProvider({ children, data }: UserProviderProps): JSX
                   strings.UPDATED_TIMEZONE_MSG,
                   <Link to={APP_PATHS.MY_ACCOUNT}>{strings.MY_ACCOUNT}</Link>
                 ),
-                strings.formatString(strings.UPDATED_TIMEZONE_TITLE, foundTimeZone ? browserTimeZone : strings.UTC)
+                strings.formatString(
+                  strings.UPDATED_TIMEZONE_TITLE,
+                  foundTimeZone.id === 'Etc/UTC' ? foundTimeZone.longName : foundTimeZone.id
```
karthikbtf, line 48: this should be foundTimeZone.longName for all zones (this is the display name for the time zone id)

karthikbtf (approved):
One change is needed but otherwise good 👍

karthikbtf (approved):
thanks!!

PR #1008, opened 2023-01-09
SW-2687 Temporary time zones hook to unblock selector component and other page views

karthikbtf:
- this is temporary
- localization provider will replace this with it's time zones

### DIFF
--- a/src/providers/hooks.ts
+++ b/src/providers/hooks.ts
@@ -6,8 +6,8 @@ export const useUser = () => useContext(UserContext);
 /**
  * Default time zones until we have a provider.
  */
-//const DEFAULT_TIME_ZONES = (Intl as any).supportedValuesOf('timeZone').map((tz: any) => ({ id: tz, longName: tz }));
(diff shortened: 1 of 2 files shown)


charlie-tf (approved):
Looks good. Thanks for doing this

constanzauanini (approved):
Tests are failing 😕

karthikbtf:
> Tests are failing 😕

I saw that, what does it have to do with my change though 🤯

src/providers/hooks.ts
```diff
@@ -1,6 +1,16 @@
...
 import { OrganizationContext, UserContext } from './contexts';
 
 export const useUser = () => useContext(UserContext);
 
+/**
+ * Default time zones until we have a provider.
+ */
+export const useTimeZones = () =>
+  useMemo(() => {
+    const defaultTimeZones = (Intl as any).supportedValuesOf('timeZone').map((tz: any) => ({ id: tz, longName: tz }));
```
karthikbtf, line 11: moving this into the function seems to fix the tests error.. not sure how related.
will get this in for now and see how it works.

PR #1009, opened 2023-01-09
SW-2688 Fix search criteria transform function in FE to consider the org id

karthikbtf:


src/api/search/index.ts
```diff
@@ -21,7 +21,7 @@ export function convertToSearchNodePayload(
   criteria: SearchCriteria,
   organizationId?: number
 ): SearchNodePayload | undefined {
-  if (Object.keys(criteria).length === 0) {
+  if (Object.keys(criteria).length === 0 && !organizationId) {
```
karthikbtf, line 24: this code was previously
```
if (criteria === {}) {
  return undefined;
}
```
That check was incorrect as well for not considering the org id...  we got lucky since it never tested true.

constanzauanini (approved):

karthikbtf:
For hotfixes, we won't be merging the PR, see https://terraformation.atlassian.net/wiki/spaces/SW/pages/482017308/Deploying+code#Fix-a-bug-in-production-without-pulling-in-other-changes-from-main

charlie-tf (approved):
Testing out in staging looks good

PR #1010, opened 2023-01-09
SW-2689 Fix redundant search on accessions page

constanzauanini:
Get unfiltered results and pending accessions only one time (when changing selected org) to avoid executing them in every search

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -266,6 +266,27 @@ export default function Database(props: DatabaseProps): JSX.Element {
   }, [query, location, history, setSearchCriteria, selectedOrganization, searchCriteria]);
 
   useEffect(() => {
+    if (selectedOrganization) {
+      const populateUnfilteredResults = async () => {
+        const apiResponse = await search({
+          prefix: 'facilities.accessions',
+          fields: ['id'],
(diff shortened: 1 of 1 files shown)


src/components/seeds/database/index.tsx
```diff
@@ -265,6 +265,27 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
+        setUnfilteredResults(apiResponse);
+      };
+      const populatePendingAccessions = async () => {
+        const data = await getPendingAccessions(selectedOrganization.id);
+        setPendingAccessions(data);
+      };
+      populateUnfilteredResults();
+      populatePendingAccessions();
+    }
+  }, [selectedOrganization]);
```
karthikbtf, line 287: [selectedOrganization.id]?

karthikbtf (approved):
looks good on vercel

src/components/seeds/database/index.tsx
constanzauanini: I'm getting the linter "React Hook useEffect has a missing dependency: 'selectedOrganization'. Either include it or remove the dependency array  react-hooks/exhaustive-deps" if I use selectedOrganization.id :(

PR #1011, opened 2023-01-09
remove unused strings

karthikbtf:
- the banner is no longer shown on maps (we won't show the map if there's no shapefile)
- the banner property is still useful if we want to show something there later? keeping it for now.. unless anyone feels otherwise

### DIFF
--- a/src/strings/index.tsx
+++ b/src/strings/index.tsx
@@ -314,9 +314,6 @@ const strings = new LocalizedStrings({
     GA3: 'GA3',
     GENERAL: 'General',
     GENERIC_ERROR: 'An error occurred',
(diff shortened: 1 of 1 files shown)


joshbuchea (approved):

constanzauanini:
We are still showing the generic map in some cases right? Don't we need the message for those cases?

karthikbtf:
> We are still showing the generic map in some cases right? Don't we need the message for those cases?

The empty state - when there is a planting site but no shapefiles - was the only case. We removed the map from that empty state.
 We were also showing it when there were no planting sites - that was actually incorrect. Design did not want to show it then - the message was specific to adding shapefiles to the empty planting site. That's why removed it.
The banner props are still there if we want to add any additional info later.

PR #1012, opened 2023-01-09
SW-2660 Create Localization Provider Layer

charlie-tf:
- create the data types, context, provider, and modify the useTimeZones hook
- at the App layer, wrap the component tree in the provider
- add a useEffect to fetch the time zone data from the backend
- default the locale to "en"

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -62,9 +62,8 @@ import PlantsDashboard from './components/Plants';
 import { NurseryWithdrawals, NurseryWithdrawalsDetails, NurseryReassignment } from './components/NurseryWithdrawals';
 import { listPlantingSites } from './api/tracking/tracking';
 import { PlantingSite } from './api/types/tracking';
-import { LocalizationProvider, OrganizationProvider, TimeZoneDescription, UserProvider } from './providers';
+import { LocalizationProvider, OrganizationProvider, UserProvider } from './providers';
 import { defaultSelectedOrg } from './providers/contexts';
-import { getTimeZones } from './api/timezones/timezones';
 
 interface StyleProps {
   isDesktop?: boolean;
@@ -341,18 +340,6 @@ export default function App() {
   }, [type]);
 
   const [locale] = useState('en');
-  const [timeZones, setTimeZones] = useState<TimeZoneDescription[]>([]);
-
-  useEffect(() => {
-    const fetchTimeZones = async () => {
-      const timeZoneResponse = await getTimeZones(locale);
-      if (!timeZoneResponse.error && timeZoneResponse.timeZones) {
-        setTimeZones(timeZoneResponse.timeZones);
-      }
-    };
-
-    fetchTimeZones();
-  }, [locale]);
 
   if (orgAPIRequestStatus === APIRequestStatus.AWAITING || orgAPIRequestStatus === APIRequestStatus.FAILED_NO_AUTH) {
     return (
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -340,6 +340,20 @@ export default function App() {
     }
   }, [type]);
 
+  const [locale] = useState('en');
```
charlie-tf, line 343: Default the locale to 'en' for now. This will be changed later when we get to the languages portion of i18n.

karthikbtf: Default sounds good 👍 
Thoughts on moving the timezones fetching etc useEffect into the LocalizationProvider - separate the responsibility?
The user and org stuff we left them in the App because it was already tightly coupled with the rest of the code.

src/providers/contexts.ts
```diff
@@ -27,3 +27,7 @@ export const OrganizationContext = createContext<ProvidedOrganizationData>({
   },
   selectedOrganization: defaultSelectedOrg,
 });
+
+export const LocalizationContext = createContext<ProvidedTimeZones>({
```
karthikbtf, line 31: Ideally this should provide the time zones and the strings.
```
{
  timeZones: TimeZoneDescription[],
  strings: { [key: string]: string },
```
For now the provider can load the strings from src/strings.. when we have i18n we will get the strings from the BE.

Also, if you prefer we do this separately as part of the i18n work that's fine too.

src/providers/hooks.ts
```diff
@@ -1,16 +1,8 @@
...
-/**
- * Default time zones until we have a provider.
- */
-export const useTimeZones = () =>
-  useMemo(() => {
-    const defaultTimeZones = (Intl as any).supportedValuesOf('timeZone').map((tz: any) => ({ id: tz, longName: tz }));
-    const supportedTimeZones = [...defaultTimeZones, { id: 'Etc/UTC', longName: 'Coordinated Universal Time' }];
-    return supportedTimeZones;
-  }, []);
+export const useTimeZones = () => useContext(LocalizationContext);
```
karthikbtf, line 6: thoughts on returning the array here, for usage-friendly purposes:
useTimeZones = () => useContext(LocalizationContext).supportedTimeZones;
we will have.. at some point..
useStrings = () => useContext(LocalizationContext).strings;

karthikbtf (commented):
Looks good, a few comments/thoughts..

src/providers/DataTypes.ts
```diff
@@ -13,3 +13,12 @@ export type ProvidedOrganizationData = {
   organizations?: ServerOrganization[];
   reloadData: () => void;
 };
+
+export type TimeZoneDescription = {
```
constanzauanini, line 17: Should we add this type to a file in src/types?

src/providers/contexts.ts
charlie-tf: good call, I misunderstood that part of the requirement

src/providers/DataTypes.ts
charlie-tf: that would make sense as it is not specifically tied to the Provider

src/providers/contexts.ts
karthikbtf: i created tickets to move user logic from App into UserProvider, same for org. I can take a crack at those after this lands.

karthikbtf: Btw - if you expose strings as well, we can share the load of updating the rest of the codebase as a separate effort if you prefer.

karthikbtf (approved):

karthikbtf:
neat!! 🙌

PR #1013, opened 2023-01-10
SW-2670 Time zone selector - initial revision with alphabetical sort

constanzauanini:
- Create an initial time zone selector
- Have to update this changes when localization provider is merged
- Used the selector in my account page for testing

### DIFF
--- a/src/components/TimeZoneSelector.tsx
+++ b/src/components/TimeZoneSelector.tsx
@@ -1,0 +1,76 @@
+import React, { useEffect, useMemo, useState } from 'react';
+import { Autocomplete } from '@terraware/web-components';
+import strings from 'src/strings';
+import { getTimeZones } from 'src/api/timezones/timezones';
+
+export type TimeZoneDescription = {
+  id: string;
+  longName: string;
+};
+
+export type TimeZoneSelectorProps = {
+  onTimeZoneSelected: (tzSelected: TimeZoneDescription) => void;
+};
+
+export default function TimeZoneSelector(props: TimeZoneSelectorProps): JSX.Element {
+  const { onTimeZoneSelected } = props;
+  const [selectedTimeZone, setSelectedTimeZone] = useState<TimeZoneDescription>();
+  const [timeZones, setTimeZones] = useState<TimeZoneDescription[]>([]);
+
+  useEffect(() => {
+    const fetchTimeZones = async () => {
+      const timeZoneResponse = await getTimeZones();
+      if (!timeZoneResponse.error && timeZoneResponse.timeZones) {
+        setTimeZones(
+          timeZoneResponse.timeZones.sort((a, b) => {
+            if (a.longName > b.longName) {
+              return 1;
+            }
+            if (b.longName > a.longName) {
+              return -1;
+            }
+            return 0;
+          })
+        );
+      }
+    };
+
+    fetchTimeZones();
+  }, []);
+
+  const tzToDropdownItem = (tz?: TimeZoneDescription) =>
+    tz ? { label: tz.longName, value: tz.id } : { label: '', value: '' };
+
+  const tzOptions: any[] = useMemo(() => {
+    return timeZones.map((tz) => tzToDropdownItem(tz));
+  }, [timeZones]);
+
+  const onChangeTimeZone = (timeZone: any) => {
+    const foundTimeZone = timeZones.find((iTtimeZone) => iTtimeZone.id.toString() === timeZone.value?.toString());
+    if (foundTimeZone) {
+      setSelectedTimeZone(foundTimeZone);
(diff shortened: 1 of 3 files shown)


src/components/TimeZoneSelector.tsx
```diff
@@ -0,0 +1,80 @@
...
+      if (!timeZoneResponse.error && timeZoneResponse.timeZones) {
+        setTimeZones(
+          timeZoneResponse.timeZones.sort((a, b) => {
+            if (a.longName > b.longName) {
+              return 1;
+            }
+            if (b.longName > a.longName) {
+              return -1;
+            }
+            return 0;
```
charlie-tf, line 32: You may be able to replace this with simply
```typescript
timeZoneResponse.timeZones.sort((a, b) => a.longName > b.longName)
```

charlie-tf: nope sorry, that's wrong, please ignore 😅

charlie-tf: this is what i was thinking of: `a.longName.compareTo(b.longName)`

constanzauanini: Oh, this looks nicer. Will test. Thanks @charlie-tf !

charlie-tf: actually it's `a.longName.localeCompare(b.longName)`

src/components/TimeZoneSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }, [timeZones]);
+
+  const onChangeTimeZone = (timeZone: any) => {
+    const foundTimeZone = timeZones.find((iTtimeZone) => iTtimeZone.id.toString() === timeZone.value?.toString());
+    if (foundTimeZone) {
+      onTimeZoneSelected(foundTimeZone);
+    }
+  };
+
+  const isEqualAut = (optionA: any, optionB: any) => {
```
charlie-tf, line 61: #naming What is "Aut"?

src/components/TimeZoneSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+  disabled?: boolean;
+};
+
+export default function TimeZoneSelector(props: TimeZoneSelectorProps): JSX.Element {
+  const { onTimeZoneSelected, selectedTimeZone, disabled } = props;
+  const [timeZones, setTimeZones] = useState<TimeZoneDescription[]>([]);
+
+  useEffect(() => {
+    const fetchTimeZones = async () => {
+      const timeZoneResponse = await getTimeZones();
```
karthikbtf, line 23: use the `useTimeZones` hook?

src/components/TimeZoneSelector.tsx
constanzauanini: Oh, sorry, It came from "autocomplete" because initially I did a <Select> component and an <Autocomplete> component and each one has it owns _isEqual_ function. But finally, I chose the autocomplete an forgot to rename

src/components/TimeZoneSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const onChangeTimeZone = (timeZone: any) => {
+    const foundTimeZone = timeZones.find((iTtimeZone) => iTtimeZone.id.toString() === timeZone.value?.toString());
+    if (foundTimeZone) {
+      onTimeZoneSelected(foundTimeZone);
+    }
+  };
+
+  const isEqualAut = (optionA: any, optionB: any) => {
+    if (optionA?.value && optionB?.value) {
```
karthikbtf, line 62: could return optionA?.value === optionB?.value;

karthikbtf (approved):
Looks good 🎉 Thanks!!
One comment... also we need to inform Ming (see Jira https://terraformation.atlassian.net/browse/SW-2671) on how to test.. she will give feedback on the selector

karthikbtf:
@constanzauanini I just tested it in vercel.. does the selector need to be full-width in desktop view?

karthikbtf:
> @constanzauanini I just tested it in vercel.. does the selector need to be full-width in desktop view?

Nvm, we can let Ming give feedback. This works great!

charlie-tf (approved):
Looks great!

constanzauanini:
> https://terraformation.atlassian.net/browse/SW-2671

I don't think it has to be full-width, but yes, we can fix it later according to Ming's review

karthikbtf:
I wanted to bring this up on slack but maybe here is fine.

We have a mixed usage of
`src/components/<component-name>/index.ts`
and
`src/components/<component-name>.tsx`

Should we decide on one pattern?
Do we need more than one file to warrant a directory structure? Or always use directory structures?

PR #1014, opened 2023-01-10
Move user time zone logic to localization provider

constanzauanini:


src/providers/LocalizationProvider.tsx
```diff
@@ -23,6 +32,38 @@ export default function LocalizationProvider({ children, locale }: LocalizationP
     fetchTimeZones();
   }, [locale]);
 
+  useEffect(() => {
+    const populateTimeZone = async () => {
+      if (user && !user.timeZone) {
```
karthikbtf, line 37: just occurred to me that if BE time zones change and user already has a time zone set.. but it is no longer in the BE list.. we may need to update again.
for now i am probably over thinking this.. no change required

karthikbtf (approved):
🙏  thanks @constanzauanini

src/providers/LocalizationProvider.tsx
constanzauanini: Yes, that could happen but I don't think it would be a common case. We could improve this for that use case later

PR #1015, opened 2023-01-10
Make useSnackbar return a constant object so it doesn't trigger useEffects

karthikbtf:
- leverage useMemo
 - tested in all views


charlie-tf (approved):
🎉 Nice!

karthikbtf:
> 🎉 Nice!

Thanks 🙏 
Fwiw, my testing showed a lot of views making redundant requests (but unrelated to snackbar). There's another ticket open to track that.

PR #1016, opened 2023-01-10
SW-2683 Time zones utils

karthikbtf:
- for now just one, to get the default time zone
- this will be used across the board

### DIFF
--- a/src/utils/useTimeZoneUtils.ts
+++ b/src/utils/useTimeZoneUtils.ts
@@ -1,0 +1,34 @@
+import { useOrganization, useTimeZones, useUser } from 'src/providers';
+
+import { TimeZoneDescription } from 'src/types/TimeZones';
+
+const DEFAULT_UTC = { id: 'Etc/UTC', longName: 'Coordinated Universal Time' };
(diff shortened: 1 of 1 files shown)


src/utils/useTimeZoneUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Helper function to return UTC as per the supported time zones list
+ */
+export const getUTC = (timeZones: TimeZoneDescription[]): TimeZoneDescription => {
+  return getTimeZone(timeZones, DEFAULT_UTC.id) ?? DEFAULT_UTC;
+};
+
+/**
+ * Get a fallback time zone (based on org, user, etc.)
+ */
+export const useDefaultTimeZone = () => {
```
karthikbtf, line 24: this is needed in places where we initialize an entity to a time zone (if it doesn't already have one)

karthikbtf:
I missed one util, getTodaysDateFormatted (which is in web-components), will fix that

charlie-tf (approved):
Nice 👍

PR #1017, opened 2023-01-11
SW-2648 Add time zone display and selection to the Nursery view/edit page

constanzauanini:
- Tooltip is missing because Autocomplete component doesn't accept tooltipTitle as Textfield.

Screenshot 2023-01-11 at 11 39 33
Screenshot 2023-01-11 at 11 39 59
Screenshot 2023-01-11 at 11 40 18

### DIFF
--- a/src/components/NewNursery/index.tsx
+++ b/src/components/NewNursery/index.tsx
@@ -14,6 +14,10 @@ import useSnackbar from 'src/utils/useSnackbar';
 import { getAllNurseries } from 'src/utils/organization';
 import TfMain from 'src/components/common/TfMain';
 import { useOrganization } from 'src/providers/hooks';
+import TimeZoneSelector from '../TimeZoneSelector';
+import { TimeZoneDescription } from 'src/types/TimeZones';
+import { Checkbox } from '@terraware/web-components';
+import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
 
 export default function NurseryView(): JSX.Element {
   const { selectedOrganization, reloadData } = useOrganization();
@@ -21,6 +25,8 @@ export default function NurseryView(): JSX.Element {
   const [descriptionError, setDescriptionError] = useState('');
   const snackbar = useSnackbar();
   const theme = useTheme();
+  const [orgTZChecked, setOrgTZChecked] = useState<boolean>(false);
+  const defaultTimeZone = useDefaultTimeZone();
 
   const [record, setRecord, onChange] = useForm<Facility>({
     name: '',
@@ -53,7 +59,14 @@ export default function NurseryView(): JSX.Element {
       organizationId: selectedOrganization.id,
       type: 'Nursery',
       connectionState: 'Not Connected',
+      timeZone: selectedNursery?.timeZone,
     });
+
+    if (!selectedNursery?.timeZone) {
+      setOrgTZChecked(true);
+    } else {
+      setOrgTZChecked(false);
+    }
   }, [selectedNursery, setRecord, selectedOrganization]);
 
   const goToNurseries = () => {
@@ -82,6 +95,29 @@ export default function NurseryView(): JSX.Element {
       goToNurseries();
     } else {
       snackbar.toastError();
+    }
+  };
+
+  const onChangeTimeZone = (newTimeZone: TimeZoneDescription) => {
+    setRecord((previousRecord: Facility): Facility => {
+      return {
+        ...previousRecord,
+        timeZone: newTimeZone.id,
+      };
+    });
+  };
+
+  const onOrgTimeZoneChecked = (checked: boolean) => {
+    if (checked) {
+      setOrgTZChecked(true);
+      setRecord((previousRecord: Facility): Facility => {
+        return {
+          ...previousRecord,
+          timeZone: undefined,
+        };
+      });
+    } else {
+      setOrgTZChecked(false);
     }
   };
 
@@ -120,6 +156,21 @@ export default function NurseryView(): JSX.Element {
                 onChange={(value) => onChange('description', value)}
                 value={record.description}
                 errorText={record.description ? '' : descriptionError}
+              />
+            </Grid>
+            <Grid item xs={gridSize()}>
+              <TimeZoneSelector
+                selectedTimeZone={record?.timeZone || defaultTimeZone.id}
+                onTimeZoneSelected={onChangeTimeZone}
+                label={strings.TIME_ZONE}
+                disabled={orgTZChecked}
+              />
+              <Checkbox
+                label={strings.FALLBACK_TO_ORG_TZ}
+                onChange={(value) => onOrgTimeZoneChecked(value)}
+                id='orgTZ'
+                name='orgTZ'
+                value={orgTZChecked}
               />
             </Grid>
           </Grid>
--- a/src/components/NewNursery/index.tsx
+++ b/src/components/NewNursery/index.tsx
@@ -18,6 +18,7 @@ import TimeZoneSelector from '../TimeZoneSelector';
 import { TimeZoneDescription } from 'src/types/TimeZones';
 import { Checkbox } from '@terraware/web-components';
 import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
+import isEnabled from 'src/features';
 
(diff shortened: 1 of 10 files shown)


src/components/NewNursery/index.tsx
```diff
@@ -53,7 +59,14 @@ export default function NurseryView(): JSX.Element {
...
       type: 'Nursery',
       connectionState: 'Not Connected',
+      timeZone: selectedNursery?.timeZone,
     });
+
+    if (!selectedNursery?.timeZone) {
+      setOrgTZChecked(true);
+    } else {
+      setOrgTZChecked(false);
+    }
```
karthikbtf, line 71: nice

src/components/NewNursery/index.tsx
```diff
@@ -85,6 +98,29 @@ export default function NurseryView(): JSX.Element {
...
+    if (checked) {
+      setOrgTZChecked(true);
+      setRecord((previousRecord: Facility): Facility => {
+        return {
+          ...previousRecord,
+          timeZone: undefined,
+        };
+      });
+    } else {
+      setOrgTZChecked(false);
```
karthikbtf, line 122: set timeZone to default value?

src/components/NewNursery/index.tsx
```diff
@@ -14,13 +14,19 @@ import useSnackbar from 'src/utils/useSnackbar';
...
+import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
 
 export default function NurseryView(): JSX.Element {
   const { selectedOrganization, reloadData } = useOrganization();
   const [nameError, setNameError] = useState('');
   const [descriptionError, setDescriptionError] = useState('');
   const snackbar = useSnackbar();
   const theme = useTheme();
+  const [orgTZChecked, setOrgTZChecked] = useState<boolean>(false);
+  const defaultTimeZone = useDefaultTimeZone();
```
karthikbtf, line 30: i messed up, i need to update the utility to consider if editing or read-only.
For read-only we won't use the user's time zone. For edit we will.
Can you add an optional 'forEdit?: boolean' in the utility.
If forEdit, then also use user's time zone, otherwise don't use the user's time zone.

src/components/NewNursery/index.tsx
```diff
@@ -122,6 +158,21 @@ export default function NurseryView(): JSX.Element {
...
+                disabled={orgTZChecked}
+              />
+              <Checkbox
+                label={strings.FALLBACK_TO_ORG_TZ}
+                onChange={(value) => onOrgTimeZoneChecked(value)}
+                id='orgTZ'
+                name='orgTZ'
+                value={orgTZChecked}
+              />
+            </Grid>
```
karthikbtf: extract into a component since we will reuse?

src/components/NewNursery/index.tsx
```diff
@@ -122,6 +158,21 @@ export default function NurseryView(): JSX.Element {
                 errorText={record.description ? '' : descriptionError}
               />
             </Grid>
+            <Grid item xs={gridSize()}>
```
karthikbtf: render if opt-in enabled

src/components/Nurseries/index.tsx
```diff
@@ -23,6 +23,7 @@ type NurseriesListProps = {
 const columns: TableColumnType[] = [
   { key: 'name', name: 'Name', type: 'string' },
   { key: 'description', name: 'Description', type: 'string' },
+  { key: 'timeZone', name: strings.TIME_ZONE, type: 'string' },
```
karthikbtf: only if opt-in enabled, we can do this dynamically in the table props maybe for now

src/components/Nursery/index.tsx
```diff
@@ -102,6 +105,14 @@ export default function NurseryDetails(): JSX.Element {
             display={true}
           />
         </Grid>
+        <Grid item xs={gridSize()}>
```
karthikbtf: opt-in enabled

src/strings/index.tsx
```diff
@@ -283,6 +283,7 @@ export const stringsMap: ILocalizedStringsMap = {
     EXPORT_RECORDS: 'Export Records',
     EXPORT: 'Export',
     FACILITY: 'Facility',
+    FALLBACK_TO_ORG_TZ: 'Fallback to organization’s time zone',
```
karthikbtf, line 286: can we call this 'Use organization's time zone'?

karthikbtf (commented):
looks good, some comments

src/components/NewNursery/index.tsx
karthikbtf: ignore, i see the latest updates, thanks

src/components/Nursery/index.tsx
```diff
@@ -102,6 +107,16 @@ export default function NurseryDetails(): JSX.Element {
             display={true}
           />
         </Grid>
+        {timeZoneFeatureEnabled && (
+          <Grid item xs={gridSize()}>
```
karthikbtf, line 111: it will be cool to inform user if the tz is being inherited from org.

src/components/NewNursery/index.tsx
```diff
@@ -122,6 +160,23 @@ export default function NurseryView(): JSX.Element {
...
             </Grid>
+            {timeZoneFeatureEnabled && (
+              <Grid item xs={gridSize()}>
+                <TimeZoneSelector
+                  selectedTimeZone={record?.timeZone || defaultTimeZone.id}
+                  onTimeZoneSelected={onChangeTimeZone}
+                  label={strings.TIME_ZONE}
+                  disabled={orgTZChecked}
+                />
+                <Checkbox
```
karthikbtf, line 171: since the dropdown blocks the checkbox.. should we move this to the right (or i wish between label and dropdown but i guess that's not feasible)... just a thought. or we can let design make a decision

src/components/NewNursery/index.tsx
karthikbtf: actually let me send a PR out for this, since it's my mess up

karthikbtf: I have this out https://github.com/terraware/terraware-web/pull/1018 also added a `useFacilityTimeZone` which handles the fallback - so client code finds it easier to use

constanzauanini: I don't get it. If editing, we default the user's time zone, but if it is read-only and it doesn't have any timezone value, what are we going to show?

src/components/Nursery/index.tsx
constanzauanini: I thought we could add that later as an info tooltip in the label (autocomplete does not support tooltip yet). Thoughts?

src/components/NewNursery/index.tsx
constanzauanini: I think we can update it later with design decision

src/components/NewNursery/index.tsx
constanzauanini: Looking at your PR, it seems for ready-only we show UTC time zone, and not the user timezone. Why is that?

karthikbtf: 1) in display mode: the nursery time zone shows what the system/BE uses. nusery-timezone || org-timezone || UTC.
2) in edit mode, we are trying to suggest a time zone to use. If org doesn't have a time zone set, we are suggesting user to use his/her time zone value (we will almost never end up suggesting UTC fallback since user time zone is set by us).
Basically - in edit mode, we are 'suggesting' a value to use.
In read-only mode, we are showing what is actually being used.
Not sure if this clarifies?

src/components/Nursery/index.tsx
karthikbtf: Oh yes, tooltip works great.

src/components/NewNursery/index.tsx
karthikbtf: sounds good

src/components/NewNursery/index.tsx
karthikbtf: Actually, this is still technically valid - since setting the org or user time zone may fail. And data model has time zone as optional. I can take care of this later if you wish.

karthikbtf (approved):
looks good for now 👍

PR #1019, opened 2023-01-11
SW-2701 Introduce a bootstrapping layer in FE, move data initialization into Provider layers

karthikbtf:
- introduce bootstrapping layers
- move provider data initialization into providers


karthikbtf:
This will break until we clean up the App some more.
Move org initialization into org provider.
There's a whole bunch of bootstrapping data for the app.. maybe that should be another layer?
@charlie-tf seems like maybe we should take this on (user, org, bootstrap initialization) after time zones is completed? The clean up seems a bit complex at this point.

charlie-tf:
> This will break until we clean up the App some more. Move org initialization into org provider. There's a whole bunch of bootstrapping data for the app.. maybe that should be another layer? @charlie-tf seems like maybe we should take this on (user, org, bootstrap initialization) after time zones is completed? The clean up seems a bit complex at this point.

@karthikbtf, yeah I started to look at moving the org bootstrapping into the provider and there's a lot to it.

charlie-tf:
> > This will break until we clean up the App some more. Move org initialization into org provider. There's a whole bunch of bootstrapping data for the app.. maybe that should be another layer? @charlie-tf seems like maybe we should take this on (user, org, bootstrap initialization) after time zones is completed? The clean up seems a bit complex at this point.
> 
> @karthikbtf, yeah I started to look at moving the org bootstrapping into the provider and there's a lot to it.

I was wondering whether it might make sense to move Organization and User Providers up to the index.tsx level, because the provided values are used quite a lot throughout App.tsx. What do you think of that idea?

karthikbtf:
> > This will break until we clean up the App some more. Move org initialization into org provider. There's a whole bunch of bootstrapping data for the app.. maybe that should be another layer? @charlie-tf seems like maybe we should take this on (user, org, bootstrap initialization) after time zones is completed? The clean up seems a bit complex at this point.
> 
> @karthikbtf, yeah I started to look at moving the org bootstrapping into the provider and there's a lot to it.

@charlie-tf yes, I agree 🤣 .. we could keep the tasks open and revisit later.

karthikbtf:
> > > This will break until we clean up the App some more. Move org initialization into org provider. There's a whole bunch of bootstrapping data for the app.. maybe that should be another layer? @charlie-tf seems like maybe we should take this on (user, org, bootstrap initialization) after time zones is completed? The clean up seems a bit complex at this point.
> > 
> > 
> > @karthikbtf, yeah I started to look at moving the org bootstrapping into the provider and there's a lot to it.
> 
> I was wondering whether it might make sense to move Organization and User Providers up to the index.tsx level, because the provided values are used quite a lot throughout App.tsx. What do you think of that idea?

That's a good idea. 
Might be good to keep index.tsx clean and split App into your vision of the index.tsx change.. ?
Basically, rename App to boostrapped App, and have another App that's basically the various providers with bootstrapped app as the child? Same design as you vision.

karthikbtf:
@charlie-tf I leveraged your idea to lift up UserProvider outside the App monolith. It is much cleaner now.
We can technically bring in other bootstrapping into this layer if needed. Reopened this PR. Tests great.

src/App.tsx
```diff
@@ -680,13 +659,19 @@ export default function App() {
...
+          <LocalizationProvider locale={locale}>{getContent()}</LocalizationProvider>
+        </OrganizationProvider>
+      )}
     </StyledEngineProvider>
   );
 }
+
+export default function App(): JSX.Element {
+  return (
+    <UserProvider>
```
karthikbtf, line 673: We can add the OrgProvider below this.. it doesn't have to be a blocking provider.
AppBase can continue to do it's thing based on the selectedOrg,, we can clean that up later.
We could theoretically load strings as part of the UserProvider and expose the data. Localization provider could just re-expose it for semantics purposes, maybe 🤷 
Just some thoughts.

constanzauanini (approved):
Refactor looks good to me

karthikbtf:
> Refactor looks good to me

Thanks @constanzauanini  - btw, I tried updating the time zone but the x-axis range stayed the same. Can you take a look - it's possible I didn't test it right since I have no data.

charlie-tf (approved):
This is great, Karthik! Makes sense for UserProvider to block.

src/App.tsx
karthikbtf: We could refactor this some more by introducing a 'Bootstrap' layer - this would be blocking since it's bootstrapping the app. What happens inside that is up to us. This way we can separate out providers and blocking semantics.
```
<App>
  <Bootstrap>
      <AppBase />
  </Bootstrap>

Boostrap:
   initialize the various providers maybe.
   block on required info but not on optional info
```

Fwiw, I noticed that we - today - block the app until we have data back from the orgs fetch. App requires knowledge of whether user has orgs no no-orgs - so it can render the correct path (no-orgs-landing-page versus full-fledged-app).

charlie-tf: Ah, makes sense

karthikbtf:
@charlie-tf I did the refactor here, had to be done for all otherwise was going to break.

src/AppBootstrap.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+export default function AppBootstrap({ children }: AppBootstrapProps): JSX.Element {
+  const [locale] = useState('en');
+
+  return (
+    <UserProvider>
+      <OrganizationProvider>
+        <LocalizationProvider locale={locale}>
+          <BlockingBootstrap>{children}</BlockingBootstrap>
```
karthikbtf, line 70: this layer decides how to block

src/AppBootstrap.tsx
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+  }, [bootstrapped, userBootstrapped, organizationBootstrapped, localizationBootstrapped]);
+
+  if (!bootstrapped) {
+    return (
+      <StyledEngineProvider injectFirst>
+        <CircularProgress className={classes.spinner} size='193' />
+      </StyledEngineProvider>
+    );
```
karthikbtf, line 53: block while not bootstrapped yet

src/AppBootstrap.tsx
```diff
@@ -0,0 +1,75 @@
...
+function BlockingBootstrap({ children }: BlockingBootstrapProps): JSX.Element {
+  const classes = useStyles();
+  const [bootstrapped, setBootstrapped] = useState<boolean>(false);
+  const { bootstrapped: userBootstrapped } = useUser();
+  const { bootstrapped: organizationBootstrapped } = useOrganization();
+  const { bootstrapped: localizationBootstrapped } = useLocalization();
+
+  useEffect(() => {
+    if (!bootstrapped) {
+      const isBootstrapped = userBootstrapped && organizationBootstrapped && localizationBootstrapped;
```
karthikbtf, line 41: infer bootstrapped from provider bootstrapped status

src/providers/DataTypes.ts
```diff
@@ -6,16 +6,21 @@ import { TimeZoneDescription } from '../types/TimeZones';
 export type ProvidedUserData = {
   user?: User;
   reloadUser: () => void;
+  bootstrapped: boolean;
```
karthikbtf, line 9: all provider data has a bootstrapped boolean, up to provider to flag if data is sufficient

src/providers/OrganizationProvider.tsx
```diff
@@ -1,12 +1,142 @@
...
+import { defaultSelectedOrg } from './contexts';
 
 export type OrganizationProviderProps = {
   children?: React.ReactNode;
-  data: ProvidedOrganizationData;
 };
 
-export default function OrganizationProvider({ children, data }: OrganizationProviderProps): JSX.Element {
-  return <OrganizationContext.Provider value={data}>{children}</OrganizationContext.Provider>;
+enum APIRequestStatus {
```
karthikbtf, line 17: moved all the org based initialization into org provider

src/App.tsx
```diff
@@ -256,87 +211,22 @@ export default function App() {
     setPlotNames(plots);
   }, [plantingSites]);
 
-  const reloadPreferences = useCallback(() => {
```
karthikbtf, line 259: removed all the org based logic

charlie-tf:
> @charlie-tf I did the refactor here, had to be done for all otherwise was going to break.

Oh! I was going to tackle the Org layer after the User layer merged—thanks for doing it! It looks good at first blush, and I'll take a deeper look to review.

src/App.tsx
```diff
@@ -461,232 +286,253 @@ export default function App() {
...
-        )}
-      </UserProvider>
+      <ToastSnackbar />
+      <TopBar>
+        <TopBarContent setShowNavBar={setShowNavBar} />
+      </TopBar>
+      <div className={classes.container}>{organizations.length === 0 ? getOrphanedUserContent() : getContent()}</div>
     </StyledEngineProvider>
   );
 }
```
karthikbtf, line 530: cleaned up rendering of different paths in one place instead of scattered across the app

src/1
```diff
@@ -0,0 +1,545 @@
+/* eslint-disable import/no-webpack-loader-syntax */
```
charlie-tf, line 1: Why is this file named "`1`"?

karthikbtf: accidental check-in, i removed it

karthikbtf:
> > @charlie-tf I did the refactor here, had to be done for all otherwise was going to break.
> 
> Oh! I was going to tackle the Org layer after the User layer merged—thanks for doing it! It looks good at first blush, and I'll take a deeper look to review.

seemed better to do in one shot, hope that's ok

charlie-tf:
> > > @charlie-tf I did the refactor here, had to be done for all otherwise was going to break.
> > 
> > 
> > Oh! I was going to tackle the Org layer after the User layer merged—thanks for doing it! It looks good at first blush, and I'll take a deeper look to review.
> 
> seemed better to do in one shot, hope that's ok

Works for me!

charlie-tf (approved):
This all looks pretty good to me. Just a few questions.

src/AppBootstrap.tsx
```diff
@@ -0,0 +1,75 @@
...
+      if (isBootstrapped) {
+        setBootstrapped(isBootstrapped);
```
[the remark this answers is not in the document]
karthikbtf: I had it that way.. but got a big anal about avoiding unnecessary set-states. Will change it back - definitely more readable here.

src/providers/UserProvider.tsx
```diff
@@ -1,12 +1,36 @@
...
-  return <UserContext.Provider value={data}>{children}</UserContext.Provider>;
+export default function UserProvider({ children }: UserProviderProps): JSX.Element {
+  const reloadUser = useCallback(() => {
+    const populateUser = async () => {
+      const response = await getUser();
+      if (response.requestSucceeded) {
+        setUserData((previous: ProvidedUserData) => {
+          return {
+            ...previous,
+            user: response.user ?? undefined,
```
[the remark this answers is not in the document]
karthikbtf: response.user can't be null or undefined here, it either succeeds or we get a 401 status code and redirected to login.
I suppose we could just use force it as valid here

src/providers/OrganizationProvider.tsx
```diff
@@ -1,12 +1,142 @@
...
+    if (organizations && preferencesOrg) {
+      const organizationId = query.get('organizationId');
+      const querySelectionOrg = organizationId && organizations.find((org) => org.id === parseInt(organizationId, 10));
+      setSelectedOrganization((previouslySelectedOrg: ServerOrganization | undefined) => {
+        let orgToUse = querySelectionOrg || organizations.find((org) => org.id === previouslySelectedOrg?.id);
+        if (!orgToUse && preferencesOrg.lastVisitedOrg) {
+          orgToUse = organizations.find((org) => org.id === preferencesOrg.lastVisitedOrg);
+        }
+        if (!orgToUse) {
+          orgToUse = organizations[0];
```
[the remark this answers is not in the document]
karthikbtf: correct, undefined.. js is forgiving.
fwiw, this was relocated code

charlie-tf (approved):
🎉

karthikbtf:
> 🎉

Thanks @charlie-tf - if anything breaks I can follow up. Initial testing looks good.

karthikbtf:
@constanzauanini if you have any thoughts/suggestions lmk - i will follow up separately.

PR #1020, opened 2023-01-11
Extract TimeZone Selector With Checkbox into new component

constanzauanini:


src/components/NewNursery/index.tsx
```diff
@@ -14,20 +14,16 @@ import useSnackbar from 'src/utils/useSnackbar';
 import { getAllNurseries } from 'src/utils/organization';
 import TfMain from 'src/components/common/TfMain';
 import { useOrganization } from 'src/providers/hooks';
-import TimeZoneSelector from '../TimeZoneSelector';
 import { TimeZoneDescription } from 'src/types/TimeZones';
-import { Checkbox } from '@terraware/web-components';
-import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
 import isEnabled from 'src/features';
+import TimeZoneSelectorWithCheckbox from '../common/TimeZoneSelectorWithCheckbox';
```
karthikbtf, line 19: `src/components/common`?

src/components/NewNursery/index.tsx
```diff
@@ -162,18 +138,10 @@ export default function NurseryView(): JSX.Element {
...
-                  label={strings.TIME_ZONE}
-                  disabled={orgTZChecked}
-                />
-                <Checkbox
-                  label={strings.USE_ORG_TZ}
-                  onChange={(value) => onOrgTimeZoneChecked(value)}
-                  id='orgTZ'
-                  name='orgTZ'
-                  value={orgTZChecked}
+                <TimeZoneSelectorWithCheckbox
```
karthikbtf, line 141: Call it `LocationTimeZoneSelector` since it is use for locations (seed bank, nursery, planting site)

src/components/NewNursery/index.tsx
```diff
@@ -162,18 +138,10 @@ export default function NurseryView(): JSX.Element {
...
+                  record={record}
+                  setRecord={setRecord}
```
karthikbtf, line 143: i know we have been using record/setRecord as props in many places.
These are semantics local to the component.
Ideally the props should be in the time zone selector component's semantics.
example:
```
  location={record}
```
we don't need to expose setState semantics etc.
remove setRecord?

src/components/common/TimeZoneSelectorWithCheckbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
+import TimeZoneSelector from '../TimeZoneSelector';
+
+type UpdateTimeZone = {
+  timeZone?: string;
+};
+
+interface TimeZoneSelectorWithCheckboxProps<T extends UpdateTimeZone> {
+  onChangeTimeZone: (tzSelected: TimeZoneDescription) => void;
+  record: T;
```
karthikbtf, line 14:
```
location: TimeZoneEntity
```
we can remove the generic T

src/components/common/TimeZoneSelectorWithCheckbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+  setRecord: React.Dispatch<React.SetStateAction<T>>;
```
karthikbtf, line 15: ideally don't expose other components' state, this is a very internal semantic

src/components/common/TimeZoneSelectorWithCheckbox.tsx
```diff
@@ -0,0 +1,64 @@
+import { Checkbox } from '@terraware/web-components';
+import { useEffect, useState } from 'react';
+import strings from 'src/strings';
+import { TimeZoneDescription } from 'src/types/TimeZones';
+import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
+import TimeZoneSelector from '../TimeZoneSelector';
+
+type UpdateTimeZone = {
```
karthikbtf, line 8: call this TimeZoneEntity?

src/components/common/TimeZoneSelectorWithCheckbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+  onChangeTimeZone: (tzSelected: TimeZoneDescription) => void;
+  record: T;
+  setRecord: React.Dispatch<React.SetStateAction<T>>;
+}
+
+export default function TimeZoneSelectorWithCheckbox<T extends UpdateTimeZone>(
+  props: TimeZoneSelectorWithCheckboxProps<T>
+): JSX.Element {
+  const [orgTZChecked, setOrgTZChecked] = useState<boolean>(false);
+  const { onChangeTimeZone, record, setRecord } = props;
```
karthikbtf, line 22: we don't need setRecord,, we can just maintain a local state.. also might end up modifying the parent record if user cancels
```
  const [timeZoneEntity, setTimeZoneEntity] = useState<TimeZoneEntity>(location);
```

src/components/common/TimeZoneSelectorWithCheckbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+      setOrgTZChecked(true);
+    } else {
+      setOrgTZChecked(false);
+    }
+  }, [record]);
+
+  const onOrgTimeZoneChecked = (checked: boolean) => {
+    if (checked) {
+      setOrgTZChecked(true);
+      setRecord((previousRecord: T): T => {
```
karthikbtf, line 36: local state

src/components/common/TimeZoneSelectorWithCheckbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+        return {
+          ...previousRecord,
+          timeZone: undefined,
+        };
+      });
+    } else {
+      setOrgTZChecked(false);
```
karthikbtf, line 43: we need to default to a value here for time zone..

karthikbtf (commented):
a few comments

src/components/common/TimeZoneSelectorWithCheckbox.tsx
constanzauanini: Thanks!

karthikbtf (approved):
Looks great 🎉  thank you so much!

src/components/LocationTimeZoneSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+  timeZone?: string;
+};
+
+type LocationTimeZoneSelectorProps = {
+  onChangeTimeZone: (tzSelected: TimeZoneDescription | undefined) => void;
+  location: TimeZoneEntity;
+};
+
+export default function LocationTimeZoneSelector(props: LocationTimeZoneSelectorProps): JSX.Element {
+  const [orgTZChecked, setOrgTZChecked] = useState<boolean>(false);
```
karthikbtf, line 18: instead of false can we use `!!location.timeZone`

src/components/LocationTimeZoneSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [timeZoneEntity, setTimeZoneEntity] = useState<TimeZoneEntity>();
+
+  useEffect(() => {
+    if (!location?.timeZone) {
+      setOrgTZChecked(true);
+    } else {
+      setOrgTZChecked(false);
+      setTimeZoneEntity(location);
+    }
+  }, [location]);
```
karthikbtf, line 30: maybe we don't need this useEffect

src/components/LocationTimeZoneSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+type LocationTimeZoneSelectorProps = {
+  onChangeTimeZone: (tzSelected: TimeZoneDescription | undefined) => void;
+  location: TimeZoneEntity;
+};
+
+export default function LocationTimeZoneSelector(props: LocationTimeZoneSelectorProps): JSX.Element {
+  const [orgTZChecked, setOrgTZChecked] = useState<boolean>(false);
+  const { onChangeTimeZone, location } = props;
+  const defaultTimeZone = useDefaultTimeZone();
+  const [timeZoneEntity, setTimeZoneEntity] = useState<TimeZoneEntity>();
```
karthikbtf, line 21: we could get rid of this and only use location maybe (since onchange will update the location)

src/components/LocationTimeZoneSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+      setTimeZoneEntity(location);
+    }
+  }, [location]);
+
+  const onOrgTimeZoneChecked = (checked: boolean) => {
+    if (checked) {
+      setOrgTZChecked(true);
+      setTimeZoneEntity({
+        timeZone: undefined,
+      });
```
karthikbtf, line 37: can remove this

src/components/LocationTimeZoneSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+      onChangeTimeZone(undefined);
+    } else {
+      setOrgTZChecked(false);
+      setTimeZoneEntity({
+        timeZone: defaultTimeZone.id,
+      });
```
karthikbtf, line 43: can remove this

src/components/LocationTimeZoneSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+      onChangeTimeZone(defaultTimeZone);
+    }
+  };
+
+  return (
+    <>
+      <TimeZoneSelector
+        selectedTimeZone={timeZoneEntity?.timeZone || defaultTimeZone.id}
```
karthikbtf, line 51: location.timeZone || defaultTimeZone.id

karthikbtf (approved):
Thanks for the changes. Sorry, a few more comments...

src/components/LocationTimeZoneSelector.tsx
constanzauanini: This doesn't work, that's why I created the useEffect. I think location might be undefined at the beginning until the location object (the nursery in this case) is fetched

PR #1023, opened 2023-01-12
SW-2652 - Add time zone and display to the Monitoring charts

constanzauanini:
- Adding time zone selector to monitoring
- Default value to seedbank value

### DIFF
--- a/src/components/Monitoring/dashboard/SeedBankDashboard.tsx
+++ b/src/components/Monitoring/dashboard/SeedBankDashboard.tsx
@@ -70,19 +70,17 @@ export default function SeedBankDashboard(props: SeedBankDashboardProps): JSX.El
   const [defaultSensor, setDefaultSensor] = useState<Device>();
   const timeZoneFeatureEnabled = isEnabled('Timezones');
   const tz = useLocationTimeZone().get(seedBank);
+  const [tzSelected, setTzSelected] = useState<string>();
 
   const onChangeTimeZone = (newTimeZone: TimeZoneDescription | undefined) => {
-    updatePreferences({
-      ...monitoringPreferences,
-      timeZone: newTimeZone?.id,
-    });
+    setTzSelected(newTimeZone?.id);
   };
 
   const gridSize = () => {
     if (isMobile) {
       return 12;
     }
-    return 6;
+    return timeZoneFeatureEnabled ? 4 : 6;
   };
 
   useEffect(() => {
@@ -219,15 +217,16 @@ export default function SeedBankDashboard(props: SeedBankDashboardProps): JSX.El
 
   return (
     <Grid container spacing={3} marginTop={0}>
-      <Grid item xs={12}>
-        {timeZoneFeatureEnabled && (
-          <TimeZoneSelector
-            selectedTimeZone={(monitoringPreferences.timeZone as string) || tz.id}
-            onTimeZoneSelected={onChangeTimeZone}
-            label={strings.TIME_ZONE}
-          />
-        )}
-      </Grid>
+      {timeZoneFeatureEnabled && (
+        <Grid item xs={gridSize()}>
+          <div className={classes.graphContainer}>
+            <div className={classes.panelTitle}>
+              <p>{strings.TIME_ZONE}</p>
+            </div>
+            <TimeZoneSelector selectedTimeZone={tzSelected || tz.id} onTimeZoneSelected={onChangeTimeZone} />
+          </div>
+        </Grid>
+      )}
       <Grid item xs={gridSize()}>
         <div className={classes.graphContainer}>
           <div className={classes.panelTitle}>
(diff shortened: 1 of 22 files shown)


src/components/Monitoring/dashboard/SeedBankDashboard.tsx
```diff
@@ -64,6 +68,15 @@ export default function SeedBankDashboard(props: SeedBankDashboardProps): JSX.El
   const [defaultSensorTimePeriod, setDefaultSensorTimePeriod] = useState<string>();
   const [defaultPvTimePeriod, setDefaultPvTimePeriod] = useState<string>();
   const [defaultSensor, setDefaultSensor] = useState<Device>();
+  const timeZoneFeatureEnabled = isEnabled('Timezones');
+  const tz = useLocationTimeZone().get(seedBank);
+
+  const onChangeTimeZone = (newTimeZone: TimeZoneDescription | undefined) => {
+    updatePreferences({
+      ...monitoringPreferences,
+      timeZone: newTimeZone?.id,
```
karthikbtf, line 77: i don't know if we need to update the user preference here, might be good to always use seed bank time zone on load. Otherwise we have to give additional info and allow user to reset back to seed bank time zone.

karthikbtf (commented):
@constanzauanini the timestamps are unix timestamps, we just need to transform value with selected time zone. I'm pretty sure the chart library has some time zone option. BE data is already complete.

src/components/Monitoring/dashboard/SeedBankDashboard.tsx
karthikbtf: maybe we need a reset button always?

karthikbtf:
Thoughts on moving the timepicker into a white card and layout horizontally (desktop)? The current version looks a bit off..
I know design will give feedback but we can make our best judgement as well.

Terraware App 2023-01-12 07-12-40

Basically another card just like battery and connected status.

constanzauanini:
> @constanzauanini the timestamps are unix timestamps, we just need to transform value with selected time zone. I'm pretty sure the chart library has some time zone option. BE data is already complete.

I will take a look on chart library. Thanks

constanzauanini:
> Thoughts on moving the timepicker into a white card and layout horizontally (desktop)? The current version looks a bit off.. I know design will give feedback but we can make our best judgement as well.
> 
> Terraware App 2023-01-12 07-12-40
> 
> Basically another card just like battery and connected status.

Would that be a card that takes the full width?

src/components/Monitoring/dashboard/SeedBankDashboard.tsx
constanzauanini: Ok, I will not save it in preferences for now

karthikbtf:
> > Thoughts on moving the timepicker into a white card and layout horizontally (desktop)? The current version looks a bit off.. I know design will give feedback but we can make our best judgement as well.
> > Terraware App 2023-01-12 07-12-40
> > Basically another card just like battery and connected status.
> 
> Would that be a card that takes the full width?

Same row as other cards (share width) .. it’s only a suggestion - but design may not consider it as same category of information. The charts are already pushed down below the fold and with another row they get pushed down further. If you have no preference we can make it another card sharing same row as other cards.

constanzauanini:
> > > Thoughts on moving the timepicker into a white card and layout horizontally (desktop)? The current version looks a bit off.. I know design will give feedback but we can make our best judgement as well.
> > > Terraware App 2023-01-12 07-12-40
> > > Basically another card just like battery and connected status.
> > 
> > 
> > Would that be a card that takes the full width?
> 
> Same row as other cards (share width) .. it’s only a suggestion - but design may not consider it as same category of information. The charts are already pushed down below the fold and with another row they get pushed down further. If you have no preference we can make it another card sharing same row as other cards.

Yes, same row sounds good 👌

karthikbtf:
> > > > Thoughts on moving the timepicker into a white card and layout horizontally (desktop)? The current version looks a bit off.. I know design will give feedback but we can make our best judgement as well.
> > > > Terraware App 2023-01-12 07-12-40
> > > > Basically another card just like battery and connected status.
> > > 
> > > 
> > > Would that be a card that takes the full width?
> > 
> > 
> > Same row as other cards (share width) .. it’s only a suggestion - but design may not consider it as same category of information. The charts are already pushed down below the fold and with another row they get pushed down further. If you have no preference we can make it another card sharing same row as other cards.
> 
> Yes, same row sounds good 👌

Thanks, this looks good... we could move it to the right so the other 2 (more important) show first?

karthikbtf:
@constanzauanini you may need to rebase... my recent changes are showing up for review.

src/utils/useTimeZoneUtils.ts
```diff
@@ -73,3 +73,19 @@ export const useLocationTimeZone = () => {
 };
 
 // TODO - add more utilities as we see fit
+
+export const changeTimezone = (date: Date, ianatz: string) => {
```
karthikbtf, line 77: call it timeZone instead of ianatz?

karthikbtf: this might work 🎉

karthikbtf (approved):
🎉

karthikbtf (approved):
👍

PR #1024, opened 2023-01-12
SW-2646 Time Zone display and Selection for Organization

charlie-tf:
In organizations for which the time zone has not been set, it will not appear in the organization view page:
image

Editing the organization will cause the time zone selector to be populated with the user's time zone (falling back to UTC if the user time zone is not set):
image

If Cancel is selected, no change will be made to the organization.

If Save is selected, the organization time zone will be updated, and the toast message will indicate that it was set to the user time zone (the time zone will now be displayed in the organization view page):
image

If the organization is edited, and a different time zone is selected, choosing Save will cause the time zone to be updated and the toast will not show that it was set to the user's time zone:
image
image


src/components/Organization/index.tsx
```diff
@@ -149,6 +153,11 @@ export default function OrganizationView(): JSX.Element {
             display={true}
           />
         </Grid>
+        {timeZoneFeatureEnabled && currentTimeZone && (
```
karthikbtf, line 156: The org will default to UTC in the BE.
We could still show the default here with a tooltip that it's a default?
The locations (seedbank, nursery, etc.) show the time zone value - explicit or inferred.
Not sure what's best.. we could either remove it from the other views or add it here.

karthikbtf (commented):
Looks good 🙌 
A few comments - would like to get your thoughts.
Also, rebase if you can.. the components version is updated on main.

src/components/EditOrganization/index.tsx
```diff
@@ -16,6 +16,11 @@ import useDeviceInfo from 'src/utils/useDeviceInfo';
 import PageSnackbar from 'src/components/PageSnackbar';
 import useSnackbar from 'src/utils/useSnackbar';
 import TfMain from 'src/components/common/TfMain';
+import isEnabled from '../../features';
+import { getUTC, useUserTimeZone } from '../../utils/useTimeZoneUtils';
+import TimeZoneSelector from '../TimeZoneSelector';
+import { TimeZoneDescription } from '../../types/TimeZones';
+import { useTimeZones } from '../../providers';
```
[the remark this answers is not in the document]
charlie-tf: oops, thank you

src/components/EditOrganization/index.tsx
```diff
@@ -175,6 +201,17 @@ export default function OrganizationView({ organization, reloadOrganizationData
                 fullWidth
               />
             </Grid>
+          ) : (
+            !isMobile && <Grid item xs={gridSize()} paddingLeft={isMobile ? 0 : theme.spacing(2)} />
+          )}
+          {timeZoneFeatureEnabled && (
+            <Grid item xs={gridSize()}>
+              <TimeZoneSelector
```
karthikbtf, line 209: i think this has a tooltip option, we could pass one with placeholder text.. objective is to show users the implications of setting a time zone

src/components/EditOrganization/index.tsx
```diff
@@ -97,9 +118,14 @@ export default function OrganizationView({ organization, reloadOrganizationData
     if (organizationRecord.name === '') {
       setNameError('Required field.');
     } else {
+      if (!tzChangedFromDefault) {
+        organizationRecord.timeZone = defaultTimeZone;
+      }
       const response = await updateOrganization(organizationRecord);
       if (response.requestSucceeded) {
-        snackbar.toastSuccess(strings.CHANGES_SAVED);
+        snackbar.toastSuccess(
```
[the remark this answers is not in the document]
charlie-tf: Hm, I wonder whether it is too out of the ordinary though. Time zone isn't necessarily something that needs to be treated special. Will we need to have this kind of messaging for language settings as well? I may take this out depending on your thoughts here.

src/components/EditOrganization/index.tsx
```diff
@@ -97,9 +118,14 @@ export default function OrganizationView({ organization, reloadOrganizationData
     if (organizationRecord.name === '') {
       setNameError('Required field.');
     } else {
+      if (!tzChangedFromDefault) {
+        organizationRecord.timeZone = defaultTimeZone;
+      }
```
karthikbtf, line 123: actually we may not need to do this..
there's another big ticket to initialize user and org time zones.. https://terraformation.atlassian.net/browse/SW-2708
for now, might be good do assume the org retains it's original time zone (if not set in the selector) - even if it is undefined.

src/components/EditOrganization/index.tsx
charlie-tf: The tooltip can also explain the default as well.

src/components/Organization/index.tsx
charlie-tf: I can add a tooltip 👍 I like that pattern.

src/components/EditOrganization/index.tsx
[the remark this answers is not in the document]
constanzauanini: When we work on the ticket of setting the org time zone, we will show a page message. So modifying the value here, it would be as modifying the user time zone in my account page, and we are not sowing anything special there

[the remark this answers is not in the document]
karthikbtf: I agree, just a generic 'saved changes' might suffice. anyways, since we will be handling the org tz initialization separately.. this is a moot point :)

karthikbtf (approved):
With agreed upon changes,,, looks great to me 🎉

karthikbtf (approved):
🎉

PR #1025, opened 2023-01-12
SW-2714 Simplify "has withdrawals?" search request

sgrimm:
At page load, we query the backend to find out if there are any nursery
withdrawals; if not, we don't show the "Withdrawal Log" item in the left nav.
Previously, this was searching for all the withdrawal details, but really we
only need to know if a withdrawal exists. Change it to just search for the ID.

### DIFF
--- a/src/api/tracking/withdrawals.ts
+++ b/src/api/tracking/withdrawals.ts
@@ -25,7 +25,6 @@ const DELETED_SPECIES = [
 export async function listNurseryWithdrawals(
   organizationId: number,
   searchCriteria: SearchCriteria,
-  count = 1000,
   sortOrder?: SearchSortOrder
 ): Promise<SearchResponseElement[] | null> {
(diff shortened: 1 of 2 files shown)


karthikbtf (approved):
Looks good to me, thanks very much!

src/api/tracking/withdrawals.ts
```diff
@@ -44,7 +43,7 @@ export async function listNurseryWithdrawals(
     ],
     search: convertToSearchNodePayload(searchCriteria, organizationId),
     sortOrder: sortOrder ? [sortOrder] : [{ field: 'id', direction: 'Ascending' }],
-    count,
+    count: 1000,
```
sgrimm, line 46: `hasNurseryWithdrawals` was the only place this had a value other than 1000, so I figured I'd simplify the function signature. Happy to undo this part of the change if we want to keep the parameter.

PR #1029, opened 2023-01-13
SW-2655 Default date/time in seed bank time zone for add accession

constanzauanini:
- When selecting seedbank, set time zone to receive/collected date pickers
- Change maxDate according to time zone


src/components/accession2/properties/CollectedReceivedDate2.tsx
```diff
@@ -4,20 +4,22 @@ import { DatePicker } from '@terraware/web-components';
...
 import { isInTheFuture } from '@terraware/web-components/utils';
 import strings from 'src/strings';
+import { changeTimezone } from 'src/utils/useTimeZoneUtils';
 
 interface Props {
   onChange: (id: string, value: string) => void;
   record: Accession2 | AccessionPostRequestBody;
   type: 'collected' | 'received';
   validate?: boolean;
+  timeZone?: string;
```
karthikbtf, line 14: this will always be valid since the utils function defaults to UTC

src/components/accession2/properties/CollectedReceivedDate2.tsx
```diff
@@ -84,7 +86,8 @@ export default function CollectedReceivedDate2({ onChange, record, type, validat
           value={dates.collectedDate}
           onChange={(value) => changeDate('collectedDate', value)}
           errorText={dateErrors.collectedDate}
-          maxDate={Date.now()}
+          maxDate={timeZone ? changeTimezone(new Date(), timeZone) : new Date()}
```
karthikbtf, line 89: we can assume timeZone will always be set

src/components/accession2/create/Accession2Create.tsx
```diff
@@ -45,6 +49,21 @@ export default function CreateAccession(): JSX.Element {
   const snackbar = useSnackbar();
   const [validateFields, setValidateFields] = useState<boolean>(false);
   const [record, setRecord, onChange] = useForm<AccessionPostRequestBody>(defaultAccession());
+  const [selectedSeedBank, setSelectedSeedBank] = useState<Facility>();
```
karthikbtf, line 52: on line # 38 we have to replace the default receivedDate date to use the time zone, we have to pass the timezone into the function maybe.. some logic we have add.
There's a new util in web-components/utils/date.ts getTodaysDateFormatted(timeZone) we can use that

src/components/accession2/properties/CollectedReceivedDate2.tsx
constanzauanini: I just added this because I didn't know if this component was used in other place. WIll remove

karthikbtf: i think it is used in the edit view as well

karthikbtf: oops, don't need to remove.. make it non-optional.

src/components/accession2/properties/CollectedReceivedDate2.tsx
karthikbtf: this can just be `maxDate={changeTimezone(new Date(), timeZone}`
we should rename changeTimezone to changeTimeZone since we use 'time zone' and not 'timezone'. can we do that? i missed it in the earlier review.

src/components/accession2/create/Accession2Create.tsx
```diff
@@ -44,8 +42,30 @@ export default function CreateAccession(): JSX.Element {
...
+  const [timeZone, setTimeZone] = useState<string>();
+  const { selectedOrganization } = useOrganization();
+
+  const defaultAccession = (): AccessionPostRequestBody =>
+    ({
+      state: 'Awaiting Check-In',
+      receivedDate: getTodaysDateFormatted(timeZone),
+    } as AccessionPostRequestBody);
+
   const [record, setRecord, onChange] = useForm<AccessionPostRequestBody>(defaultAccession());
```
karthikbtf, line 56: i don't know if this will auto-update for every time zone change, can we test?

src/components/accession2/properties/CollectedReceivedDate2.tsx
constanzauanini: Yes, I meant remove the optional, but I have to modify the edit view to do that

src/components/accession2/properties/CollectedReceivedDate2.tsx
karthikbtf, line 89: leave maxDate as-is, @charlie-tf added a timeZone property to datepicker, use that instead

karthikbtf: @constanzauanini i made this change it is working for me (i checked out your branch)
keep maxDate={new Date()} and use defaultTimeZone
You have to select the seed-bank 'location dropdown' to see the change.
Select a seedbank that has a time zone you want to test.
If you don't use a seed bank, the default datepicker will show UTC.

constanzauanini:
This needs newer version from web-components to work as expected

src/components/accession2/properties/CollectedReceivedDate2.tsx
```diff
@@ -42,22 +51,22 @@ export default function CollectedReceivedDate2({ onChange, record, type, validat
 
   const validateDate = useCallback(
     (id: string, value?: any) => {
-      const date = new Date(value);
+      const date = DateTime.fromISO(value, { zone: timeZone });
```
karthikbtf, line 54: there is a fromJS in luxon, if you want to keep web-components returning the JS object like it did previously.. it has more info than the string

karthikbtf (commented):
looks good, a few comments

src/components/accession2/properties/CollectedReceivedDate2.tsx
karthikbtf: fwiw, we can keep original code as-is, here.
the only thing we need to do is pass the timeZone on line 62
there's no need to convert to luxon in this use-case

karthikbtf: i take that back, this change is a good thing if we are going to move to luxon throughout anyways. 
we need to update Datepicker to return luxon date as well, and move away from momentjs. that's a breaking/bigger change.. for another time.

constanzauanini: Changed it to receive a string when validating

src/components/accession2/create/Accession2Create.tsx
```diff
@@ -44,8 +42,39 @@ export default function CreateAccession(): JSX.Element {
   const history = useHistory();
   const snackbar = useSnackbar();
   const [validateFields, setValidateFields] = useState<boolean>(false);
+  const [selectedSeedBank, setSelectedSeedBank] = useState<Facility>();
+  const tz = useLocationTimeZone().get(selectedSeedBank);
```
karthikbtf, line 46: do you think we should use the isEnabled here and pass undefined if not enabled (instead of selected seedbank)? 
i think if someone has never set the timezone we should not be impacted .. so maybe we could leave it as-is here.. just calling this out if we missed it

src/components/accession2/edit/Accession2EditModal.tsx
```diff
@@ -30,6 +33,10 @@ export default function Accession2EditModal(props: Accession2EditModalProps): JS
   const [record, setRecord, onChange] = useForm(accession);
   const [validateFields, setValidateFields] = useState<boolean>(false);
   const snackbar = useSnackbar();
+  const { selectedOrganization } = useOrganization();
+  const selectedSeedBank = getSeedBank(selectedOrganization, record.facilityId);
+  const tz = useLocationTimeZone().get(selectedSeedBank);
```
karthikbtf, line 38: use isEnabled flag here.. or maybe in the line below and initialize timeZone to undefined..
i think maybe it's ok to leave it as-is. just calling it out

karthikbtf (approved):
thank you!

src/components/accession2/create/Accession2Create.tsx
constanzauanini: I will add it just in case

src/components/accession2/edit/Accession2EditModal.tsx
constanzauanini: Yes, I will add it just in case

src/components/accession2/create/Accession2Create.tsx
```diff
@@ -44,8 +43,40 @@ export default function CreateAccession(): JSX.Element {
...
+  useEffect(() => {
+    if (record.facilityId) {
+      const accessionSeedBank = getSeedBank(selectedOrganization, record.facilityId);
+      setSelectedSeedBank(accessionSeedBank);
+    }
+  }, [record.facilityId, selectedOrganization]);
+
+  useEffect(() => {
+    setTimeZone(tz.id);
+  }, [tz]);
```
karthikbtf, line 69: @constanzauanini can we do a test to make sure this isn't triggering a lot because tz is not in the state? maybe a console.log will show

karthikbtf (approved):
thanks for the changes

src/components/accession2/create/Accession2Create.tsx
constanzauanini: I see it just one time

PR #1031, opened 2023-01-13
WIP load strings dynamically

sgrimm:


src/providers/LocalizationProvider.tsx
```diff
@@ -32,6 +41,32 @@ export default function LocalizationProvider({ children, locale }: LocalizationP
...
+        // Default to English
+        stringsModule = import('../strings/strings-en');
+      }
+
+      const stringsTable = (await stringsModule).strings;
+
+      const localeMap: ILocalizedStringsMap = {};
+      localeMap[locale] = stringsTable;
+      strings.setContent(localeMap);
+      strings.setLanguage(locale);
```
karthikbtf, line 62: this is great!

karthikbtf (commented):
Exciting!!

src/components/seeds/database/columns.ts
```diff
@@ -1,266 +1,268 @@
-import strings from 'src/strings';
+import strings, { ILocalizedStrings } from 'src/strings';
 import { DatabaseColumn } from '@terraware/web-components/components/table/types';
 
-const COLUMNS: DatabaseColumn[] = [
+export type LocalizableDatabaseColumn = Omit<DatabaseColumn, 'name'> & { code: keyof ILocalizedStrings };
+
+const COLUMNS: LocalizableDatabaseColumn[] = [
```
karthikbtf, line 6: Optionally,, move this into the component scope so it's evaluated on render time.
Another option, convert this into a function that returns the list of columns, so strings are evaluate when the function is called.

sgrimm: Converting to a function seems like a good way to go, yeah. That'll also make it easy to memoize.

sgrimm: Updated this to use a function, which definitely seems cleaner. Thanks for that suggestion. It's https://github.com/terraware/terraware-web/pull/1031/commits/6dcd78b75b83c1674fca7ebee0350935c30dcc23 if you want to see it in isolation.

karthikbtf: This looks good, thanks @sgrimm

sgrimm: Updated all the other tables too. Their column lists are all way shorter than the accessions table so it seemed cleaner to inline them into the render functions. (In some cases, they were already inlined and I didn't have to change anything.)

src/components/Species/index.tsx
```diff
@@ -152,8 +153,13 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
       return '';
     }
   };
-  const columns: TableColumnType[] = React.useMemo(
-    () => [
+  const columns: TableColumnType[] = React.useMemo(() => {
```
karthikbtf, line 156: We could remove the useMemo and see if it really affects performance.
Another option is to maybe pass the strings into the useMemo function - expose the strings in useLocalization, that perhaps negates the benefit of loading strings from `/src/strings`.
For now this is harmless and works.

sgrimm: Removing `useMemo` makes the compiler complain that the `useEffect` call that declares `columns` as a dependency will be flagged as changed on every render.

karthikbtf: Oops, yes that won't work.

sgrimm:
The changes here have either been merged or are in non-draft PRs, so no more need for this.

PR #1039, opened 2023-01-17
Move English strings to a separate source file

sgrimm:
In preparation for dynamically loading localized strings, move the English strings
to a separate source file and tighten up the type declarations of the strings
table. For now, the new file is loaded statically, so there's no change in
behavior here; it will still be included in the Webpack bundle as before.


sgrimm:
This one is a real PR that I'd like to merge so as to avoid having to constantly resolve merge conflicts on the strings table as I work on #1031 and pull in updates.

karthikbtf (approved):
Thanks!

PR #1040, opened 2023-01-17
Use strings table for Cancel and Save button labels

sgrimm:


charlie-tf (approved):
Nice find!

karthikbtf (approved):
👍 i think we have similar hardcoding in the monitoring charts (legends, etc.)

PR #1042, opened 2023-01-17
SW-2721 Look up column labels at render time

sgrimm:
Update the table views so they look up the display names of the columns at the
time the render function is called rather than at the time the module is imported.

For most of the tables, this is a simple matter of moving the column list from the
top level of the module into the body of the render function.

The list of columns on the accessions table, though, is referenced in multiple
places and is also pretty large. Change it from a constant to a function and call
the function in the places that used to reference the constant.

No user-visible behavior change here; the English strings table is still imported
statically as before.

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -204,6 +204,7 @@ export default function Database(props: DatabaseProps): JSX.Element {
   const updateSearchColumns = useCallback(
     (columnNames?: string[]) => {
       if (columnNames) {
+        const columns = columnsIndexed();
         const searchSelectedColumns = columnNames.reduce((acum, value) => {
           acum.push(value);
           const additionalColumns = columns[value].additionalKeys;
@@ -218,7 +219,7 @@ export default function Database(props: DatabaseProps): JSX.Element {
         setDisplayColumnNames(columnNames);
       }
     },
-    [columns, setSearchColumns, setDisplayColumnNames]
+    [setSearchColumns, setDisplayColumnNames]
   );
 
   useEffect(() => {
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -218,7 +218,7 @@ export default function Database(props: DatabaseProps): JSX.Element {
         setDisplayColumnNames(columnNames);
       }
     },
-    [setSearchColumns, setDisplayColumnNames]
+    [columns, setSearchColumns, setDisplayColumnNames]
   );
 
   useEffect(() => {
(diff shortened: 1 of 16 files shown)


karthikbtf (approved):
This is very clean, thanks!

sgrimm:
Cypress tests are failing, maybe for legit reasons or maybe not. Looks like it's time for me to figure out how to get Cypress running on my M1 Mac.

charlie-tf (approved):
Nice changes! fwiw, the cypress tests are known to be flaky. You can try re-running them and see if they pass.

sgrimm:
Cypress failure turned out to be legit. https://github.com/terraware/terraware-web/pull/1042/commits/63a66c78758730094cb082a526c3653d7cc121d1 was me making the edit suggested by a lint error message, but it ended up confusing React.

src/components/seeds/database/index.tsx
```diff
@@ -203,9 +204,10 @@ export default function Database(props: DatabaseProps): JSX.Element {
   const updateSearchColumns = useCallback(
     (columnNames?: string[]) => {
       if (columnNames) {
+        const columnInfo = columnsIndexed();
```
karthikbtf, line 207: This works just fine. 
Another option is to keep columns in the component state, the value won't change on every render. More of a hack.
Or [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo) - which returns a memoized value (even though this is not really compute intensive).
Line 169 could be `const columns = useMemo(() => columnsIndexed(), []);`
I don't believe we have a particular preferred way of doing such things today - so anything goes.

sgrimm: The reason I didn't just use the `columns` from the parent scope was that the linter complained that I was referencing `columns` in the `useEffect` callback but wasn't declaring it as a dependency. So I added it as a dependency, but then React seemed to get confused and go into an infinite recursive loop (which it detected and aborted). I guess maybe it uses reference equality, which would fail because `columnsIndexed()` returns a new array instance each time.

If I used `useMemo`, the reference equality problem would go away but I guess I'd have to pass the locale as a dependency to `useMemo` or it'd keep reusing the strings from the initial locale after you switched to a new one. That does seem a bit cleaner than what I'm doing here.

karthikbtf: That is correct, good call out - the locale would need to be a dependency.
I'm afraid the linter might complain that locale is not being directly referenced in the useMemo function. If that happens, the current version would be the best!!

sgrimm: Tried `useMemo` and it has a similar lint problem where if I tell it the locale is a dependency, it complains that I have an unnecessary dependency because the actual function doesn't reference the locale. I could add a dummy reference to the locale to make the lint error go away but at that point I think it's no longer cleaner than the way I have it now.

sgrimm: Ha, glad we reached the same conclusion!

PR #1043, opened 2023-01-17
Upgrade to Cypress 12.3.0

sgrimm:
Older versions of Cypress don't run well on M1 Mac systems, so upgrade to the
latest release.

One of the changes they've made is that each test runs in a fresh browser rather
than sitting on whatever page happened to be shown at the end of the previous
test. Update the tests of the accession table filter presets to navigate to the
accessions page for each preset.


cypress/e2e/10-database.cy.js
```diff
@@ -86,13 +86,15 @@ describe('Database', () => {
     });
 
     context('Presets', () => {
-      it('General Inventory', () => {
+      beforeEach(() => {
```
karthikbtf, line 89: This is good, makes sense.

karthikbtf (approved):
Thanks so much, sorry about the repeated effort from last year.

PR #1045, opened 2023-01-18
SW-2658 Default date/time in nursery time zone for inventory add batch

constanzauanini:
- Added timezone to create inventory and add batch.
- Noticed that updating the "addedDate" when editing a batch doesn't update it in the backend, the api doesn't accept an addedDate. Is it correct that "addedDate" is editable in the frontend?

### DIFF
--- a/src/components/Inventory/InventoryCreate.tsx
+++ b/src/components/Inventory/InventoryCreate.tsx
@@ -124,7 +124,7 @@ export default function CreateInventory(): JSX.Element {
   const paddingSeparator = () => (isMobile ? 0 : 1.5);
 
   const changeDate = (id: string, value?: any) => {
-    setAddedDateChanged(true);
+    setAddedDateChanged(id === 'addedDate');
     const date = value ? getDateDisplayValue(value.getTime(), timeZone) : null;
     onChange(id, date);
   };
--- a/src/components/Inventory/InventoryCreate.tsx
+++ b/src/components/Inventory/InventoryCreate.tsx
@@ -43,6 +43,7 @@ export default function CreateInventory(): JSX.Element {
   const [selectedNursery, setSelectedNursery] = useState<Facility>();
   const tz = useLocationTimeZone().get(timeZoneFeatureEnabled ? selectedNursery : undefined);
   const [timeZone, setTimeZone] = useState(tz.id);
+  const [addedDateChanged, setAddedDateChanged] = useState(false);
 
   const defaultBatch = (): CreateBatchRequestPayload =>
     ({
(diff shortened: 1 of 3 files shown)


karthikbtf:
> * Added timezone to create inventory and add batch.
> * Noticed that updating the "addedDate" when editing a batch doesn't update it in the backend, the api doesn't accept an addedDate. Is it correct that "addedDate" is editable in the frontend?

I see that even if we edit addedDate, we don't pass the edited value to the BE, we pass the initial value. The BE does accept it. Could be a UI bug.

src/components/Inventory/InventoryCreate.tsx
```diff
@@ -142,6 +176,7 @@ export default function CreateInventory(): JSX.Element {
                   value={record.addedDate}
                   onChange={(value) => changeDate('dateAdded', value)}
```
karthikbtf, line 177: this should be `addedDate` .. oh noes :(
that's the bug

src/components/Inventory/view/BatchDetailsModal.tsx
```diff
@@ -54,6 +63,30 @@ export default function BatchDetailsModal(props: BatchDetailsModalProps): JSX.El
...
+    if (timeZone !== tz.id) {
+      setTimeZone(tz.id);
+    }
+  }, [tz.id, timeZone]);
+
+  useEffect(() => {
+    setRecord((previousRecord: CreateBatchRequestPayload): CreateBatchRequestPayload => {
+      return {
+        ...previousRecord,
+        addedDate: getTodaysDateFormatted(timeZone),
```
karthikbtf, line 85: we should be fetching the date from the record yes?

src/components/Inventory/InventoryCreate.tsx
```diff
@@ -42,8 +37,46 @@ export default function CreateInventory(): JSX.Element {
...
+    if (timeZone !== tz.id) {
+      setTimeZone(tz.id);
+    }
+  }, [tz.id, timeZone]);
+
+  useEffect(() => {
+    setRecord((previousRecord: CreateBatchRequestPayload): CreateBatchRequestPayload => {
+      return {
+        ...previousRecord,
+        addedDate: getTodaysDateFormatted(timeZone),
```
karthikbtf, line 76: if the record has a date set by user, we should use that.. not sure how we can detect it here.. maybe this will be called before user edits?

constanzauanini: This is called everytime the user changes the "receiving nursery", so if the user selected the date before the nursery, when the nursery is changed, this date will be changed to today's date (in the time zone of the nursery)

src/components/Inventory/InventoryCreate.tsx
constanzauanini: Oh, yes, for inventory create this is a bug. Will fix it! For editing a batch, the value is not saved because the api doesn't accept that value

src/components/Inventory/view/BatchDetailsModal.tsx
constanzauanini: Yes, the only problem of doing that is that if you have the default today's date selected and then you change to another nursery that has other time zone, the date will remain with the old "today's date". That's why I added this here. I can remove it, and it will keep the record value

karthikbtf: If user edited the value, we should leave it as-is, otherwise make this change. We could maybe use a boolean state to track it?

karthikbtf (approved):
one comment on latest change

PR #1046, opened 2023-01-18
SW-2721 Look up accession enum values at render time

sgrimm:
Look up human-readable values in the strings table for:

- Labels on the "Accessions by Status" panel on the dashboard (this is refactored
  to use a loop).
- Value of the "Status" box on the accession details page.
- Menu items in the "Edit Status" dialog on the accession details page.
- Test method, seed type, substrate, and treatment menu items on the Add Viability
  Test dialog.
- Purpose and test method menu items on the Withdraw dialog.

Also remove a couple unused lists of enum values from `src/types/Accession.ts`.


karthikbtf (approved):
Looks good to me, thanks for the cleanup/refactor as well 👍

PR #1047, opened 2023-01-18
SW-2721 Look up species dropdown values at render time

sgrimm:
Use the strings table instead of the English enum values for dropdowns on the
species page:

- Growth forms, storage behaviors, and conservation statuses in the filter popover.
- Growth forms and storage behaviors in the Add Species modal.


constanzauanini (approved):
Looks good! Thanks!

PR #1049, opened 2023-01-18
SW-2721 Look up accession quantity strings at render time

sgrimm:
Look up strings from the strings table at render time for:

- The field labels on count/weight range and number range filter popovers
- The list of weight units in filter popovers, the "edit quantity" popover,
  and the withdrawal modal


charlie-tf (approved):
Looks great. Thanks for doing these.

PR #1050, opened 2023-01-18
SW-2721 Look up monitoring chart labels at render time

sgrimm:


karthikbtf (approved):
Thanks! 👍 
@charlie-tf thoughts on `yarn format` also sorting the keys?

sgrimm:
Oh, oops, you're right, I should have put these above `MONITORING`. Making `yarn format` sort these seems like a great idea to me.

karthikbtf:
> Oh, oops, you're right, I should have put these above `MONITORING`. Making `yarn format` sort these seems like a great idea to me.

My mistake, I didn't notice anything wrong here, lol. I thought it looked fine below MONITORING, I am probably wrong.
I will most likely get manual sorting wrong, will take a crack at `yarn format` soon.

karthikbtf (approved):

PR #1054, opened 2023-01-19
SW-2721 Look up deleted-species placeholder at render time

sgrimm:
On the nursery withdrawal log, we show a client-rendered placeholder instead of
a species name if the species was deleted. Look it up at render time so it uses
the correct translation, and move the angle brackets to the translatable string
so they can be replaced with other symbols as needed in other languages.


karthikbtf (approved):
looks good to me

PR #1055, opened 2023-01-19
SW-2721 Look up People headers, role names at render time

sgrimm:
Look up the column headers for the People page at render time.

Replace the enum values of user roles with names from the strings table
on the My Account page, the person details page, and the add/edit person
page. They remain in English for now on the People page because they are
search results and we'll cover localizing them as part of the updates to
the search API.


karthikbtf (approved):
👍

PR #1056, opened 2023-01-19
SW-2738 Show correct time zone when editing a seed bank or nursery

karthikbtf:
- code was not updating the record once it was fetched from url param id

### DIFF
--- a/src/components/NewNursery/index.tsx
+++ b/src/components/NewNursery/index.tsx
@@ -53,7 +53,7 @@ export default function NurseryView(): JSX.Element {
     setRecord({
       name: selectedNursery?.name || '',
       description: selectedNursery?.description,
-      id: -1,
+      id: selectedNursery?.id ?? -1,
       organizationId: selectedOrganization.id,
       type: 'Nursery',
       connectionState: 'Not Connected',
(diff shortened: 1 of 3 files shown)


src/components/NewNursery/index.tsx
```diff
@@ -53,7 +53,7 @@ export default function NurseryView(): JSX.Element {
     setRecord({
       name: selectedNursery?.name || '',
       description: selectedNursery?.description,
-      id: -1,
+      id: selectedNursery?.id ?? -1,
```
constanzauanini, line 56: With this change, at line 81 is not necessary to add the id separately

src/components/NewSeedBank/index.tsx
```diff
@@ -53,7 +53,7 @@ export default function SeedBankView(): JSX.Element {
     setRecord({
       name: selectedSeedBank?.name || '',
       description: selectedSeedBank?.description,
-      id: -1,
+      id: selectedSeedBank?.id ?? -1,
```
constanzauanini, line 56: Same here, we can remove the id from line 81

constanzauanini (approved):

src/components/NewSeedBank/index.tsx
constanzauanini: Yes, that would be a good option, they have a lot in common

PR #1057, opened 2023-01-19
SW-2741 Fix web-app url based org navigation

karthikbtf:
- previously the organizations list was undefined as initial state
- url based org navigation waited until orgs was populated
- with provider refactor, the orgs list is an empty list as initial state
- the org based navigation check did not account or this new initial state
- fixed that
- also removed some unnecessary checks
- also updated org create to switch user to new org
- i believe this may be new behavior and never worked before

### DIFF
--- a/src/components/AddNewOrganizationModal.tsx
+++ b/src/components/AddNewOrganizationModal.tsx
@@ -86,7 +86,7 @@ export default function AddNewOrganizationModal(props: AddNewOrganizationModalPr
         strings.ORGANIZATION_CREATED_MSG,
         strings.formatString(strings.ORGANIZATION_CREATED_TITLE, response.organization.name)
       );
-      reloadData();
+      reloadData(response.organization.id);
       history.push({ pathname: APP_PATHS.HOME });
     } else {
       snackbar.toastError(strings.GENERIC_ERROR, strings.ORGANIZATION_CREATE_FAILED);
--- a/src/providers/OrganizationProvider.tsx
+++ b/src/providers/OrganizationProvider.tsx
@@ -58,12 +58,12 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
   const reloadPreferences = useCallback(() => {
     const getUserPreferences = async () => {
       const response = await getPreferences();
-      if (organizations && response.requestSucceeded) {
+      if (response.requestSucceeded) {
         setPreferencesOrg(response.preferences);
       }
     };
     getUserPreferences();
-  }, [organizations, setPreferencesOrg]);
+  }, [setPreferencesOrg]);
 
   const [organizationData, setOrganizationData] = useState<ProvidedOrganizationData>({
     selectedOrganization: selectedOrganization || defaultSelectedOrg,
@@ -110,7 +110,7 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
   }, [reloadOrgPreferences, selectedOrganization]);
 
   useEffect(() => {
-    if (organizations && preferencesOrg) {
+    if (organizations?.length && preferencesOrg) {
       const organizationId = query.get('organizationId');
       const querySelectionOrg = organizationId && organizations.find((org) => org.id === parseInt(organizationId, 10));
       setSelectedOrganization((previouslySelectedOrg: ServerOrganization | undefined) => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -404,7 +404,7 @@ function AppBase() {
               <PersonDetails />
             </Route>
             <Route exact path={APP_PATHS.PEOPLE}>
-              <People reloadData={reloadData} />
+              <People />
             </Route>
             <Route exact path={APP_PATHS.SEED_BANKS_NEW}>
               <NewSeedBank />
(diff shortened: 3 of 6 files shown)


src/components/AddNewOrganizationModal.tsx
```diff
@@ -86,7 +86,7 @@ export default function AddNewOrganizationModal(props: AddNewOrganizationModalPr
         strings.ORGANIZATION_CREATED_MSG,
         strings.formatString(strings.ORGANIZATION_CREATED_TITLE, response.organization.name)
       );
-      reloadData();
+      reloadData(response.organization.id);
```
karthikbtf, line 89: this could be new behavior

src/providers/OrganizationProvider.tsx
```diff
@@ -58,12 +58,12 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
   const reloadPreferences = useCallback(() => {
     const getUserPreferences = async () => {
       const response = await getPreferences();
-      if (organizations && response.requestSucceeded) {
+      if (response.requestSucceeded) {
```
karthikbtf, line 61: organizations is always valid, either empty or not empty

src/providers/OrganizationProvider.tsx
```diff
@@ -110,7 +110,7 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
   }, [reloadOrgPreferences, selectedOrganization]);
 
   useEffect(() => {
-    if (organizations && preferencesOrg) {
+    if (organizations?.length && preferencesOrg) {
```
karthikbtf, line 113: this was the bug..

constanzauanini (approved):
Looks good. Tested and working. Also tested deleting an org and is not working (in staging either), we can fix that after this

karthikbtf:
> > Looks good. Tested and working. Also tested deleting an org and is not working (in staging either), we can fix that after this
> 
> Yes, I think this has been broken for a while - the button does nothing. Will take a look.

Seems to work in MyAccount page though

karthikbtf:
> > > Looks good. Tested and working. Also tested deleting an org and is not working (in staging either), we can fix that after this
> > 
> > 
> > Yes, I think this has been broken for a while - the button does nothing. Will take a look.
> 
> Seems to work in MyAccount page though

The if(user) check is failing on People page during deleteOrgHandler.. will take a look

karthikbtf:
> > > > Looks good. Tested and working. Also tested deleting an org and is not working (in staging either), we can fix that after this
> > > 
> > > 
> > > Yes, I think this has been broken for a while - the button does nothing. Will take a look.
> > 
> > 
> > Seems to work in MyAccount page though
> 
> The if(user) check is failing on People page during deleteOrgHandler.. will take a look

Even after fixing that, I get this error:
```
{error: {message: "Cannot remove last owner of organization 57"}, status: "error"}
error
: 
{message: "Cannot remove last owner of organization 57"}
status
: 
"error"
```
Based on code, looks like we skip deleting the active user anyways. This should be fixed outside the scope of current bug.

karthikbtf:
> > > > > Looks good. Tested and working. Also tested deleting an org and is not working (in staging either), we can fix that after this
> > > > 
> > > > 
> > > > Yes, I think this has been broken for a while - the button does nothing. Will take a look.
> > > 
> > > 
> > > Seems to work in MyAccount page though
> > 
> > 
> > The if(user) check is failing on People page during deleteOrgHandler.. will take a look
> 
> Even after fixing that, I get this error:
> 
> ```
> {error: {message: "Cannot remove last owner of organization 57"}, status: "error"}
> error
> : 
> {message: "Cannot remove last owner of organization 57"}
> status
> : 
> "error"
> ```
> 
> Based on code, looks like we skip deleting the active user anyways. This should be fixed outside the scope of current bug.

The org ends up getting deleted behind the scenes.. eventually. confusing.
@constanzauanini I will fix the user not being set issue.. but everything else seems unrelated to provider related changes.
We can fix separately.

constanzauanini:
> > > > > > Looks good. Tested and working. Also tested deleting an org and is not working (in staging either), we can fix that after this
> > > > > 
> > > > > 
> > > > > Yes, I think this has been broken for a while - the button does nothing. Will take a look.
> > > > 
> > > > 
> > > > Seems to work in MyAccount page though
> > > 
> > > 
> > > The if(user) check is failing on People page during deleteOrgHandler.. will take a look
> > 
> > 
> > Even after fixing that, I get this error:
> > ```
> > {error: {message: "Cannot remove last owner of organization 57"}, status: "error"}
> > error
> > : 
> > {message: "Cannot remove last owner of organization 57"}
> > status
> > : 
> > "error"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Based on code, looks like we skip deleting the active user anyways. This should be fixed outside the scope of current bug.
> 
> The org ends up getting deleted behind the scenes.. eventually. confusing. @constanzauanini I will fix the user not being set issue.. but everything else seems unrelated to provider related changes. We can fix separately.

I tested and the org is deleted at the end (despite the error), but the user is not changed from that org. But yes, let's fix that separately

constanzauanini (approved):

PR #1060, opened 2023-01-19
SW-2620 Filter Pill List Component

charlie-tf:
Create a PillList component that renders a list of pills. Also make the
Pill component more generic. If the pill is not rendered with an
onRemovePill handler it will not have an 'X' icon to delete it.

Refactor existing components that display pills (table filters) to use
the new PillList component.


src/components/common/Pill.tsx
```diff
@@ -0,0 +1,66 @@
...
+    width: theme.spacing(1),
+    height: theme.spacing(1.5),
+  },
+}));
+
+type PillProps<IdType> = {
+  id: IdType;
+  label: string;
+  value: string;
+  onRemovePill?: (pillId: IdType) => void;
```
karthikbtf, line 48: since the props/component already qualify this to be a Pill.. thoughts on renaming this to `onRemove?: (id: IdType) => void;`
same in the PillList

src/components/common/PillList.tsx
```diff
@@ -0,0 +1,33 @@
...
+export type PillListItem<IdType> = {
+  pillId: IdType;
+  label: string;
+  value: string;
+  onRemovePill?: (pillId: IdType) => void;
+};
+
+type PillListProps<IdType> = {
+  pillData: PillListItem<IdType>[];
+  onRemovePillFromList?: (pillId: IdType) => void;
```
karthikbtf, line 14: `data` and `onRemove`?

src/components/common/PillList.tsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export default function PillList<IdType>(props: PillListProps<IdType>): JSX.Element {
+  const { pillData, onRemovePillFromList } = props;
+
+  return (
+    <Box display='flex' flexWrap='wrap'>
+      {pillData.map((data) => (
+        <Pill
+          key={String(data.pillId)}
```
karthikbtf, line 24: unless we can guarantee pillId to be unique, I'd recommend using the array index to be the key `pillData.map((data, index) => {`

src/components/common/PillList.tsx
```diff
@@ -0,0 +1,33 @@
...
+          id={data.pillId}
```
karthikbtf, line 25: for analytics/readability - we could make this `{`${data.pillId}-${data.label}`}`?

karthikbtf (approved):
Looks good to me.
Some suggestions if you think they make sense.
Moving to web-components at some point will be good.. we actually have a jira to move Pill to web-components.. i'll try to find that.

charlie-tf:
@karthikbtf all great suggestions. Thanks

src/components/common/PillList.tsx
charlie-tf: Actually, i don't think that would work. The `id` is passed to the `onRemove` handler, so the caller expects it to be the form and type of the `data.pillId` that was passed in.

karthikbtf: ugh ofc.. i misunderstood this id to be the html id..

karthikbtf (approved):

PR #1062, opened 2023-01-20
Use Craco to override Webpack caching config

sgrimm:
To make it easier to test web-components changes in the web app, start using a
react-scripts wrapper that allows overriding the Webpack configuration, and
configure Webpack to stop assuming that the web-components package never changes
and that it's thus safe to minify it once and then ignore it thereafter.

Instead, `yarn start:dev` will now watch web-components for changes and recompile
the app as needed.

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -1,0 +1,12 @@
+const { whenDev } = require('@craco/craco');
+
+module.exports = {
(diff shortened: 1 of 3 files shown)


karthikbtf (approved):
Nice, thanks!

PR #1063, opened 2023-01-20
SW-2762 Add label to allow showing tooltip

constanzauanini:
Tooltip was not showing on time selector because tooltip shows next to the label. Remove Title and add label to allow showing tooltip

Screenshot 2023-01-20 at 11 08 07


karthikbtf (approved):
i should have just done that.. my bad.
thanks

PR #1064, opened 2023-01-20
SW-2732 Fix textfield for seeds remaning filter doesn't allow input

constanzauanini:
onChange method on fields of filter was incorrectly used, so the input value was never set. Fix the onChange method to set the value correctly

### DIFF
--- a/src/components/seeds/database/filters/FilterCountWeight.tsx
+++ b/src/components/seeds/database/filters/FilterCountWeight.tsx
@@ -177,32 +177,16 @@ export default function FilterCountWeight(props: Props): JSX.Element {
           />
         </Grid>
         <Grid item xs={5}>
-          <TextField
-            id='countMinValue'
-            value={countMinValue}
-            onChange={(value) => onChange('countMinValue', value)}
-            label={strings.MIN}
-          />
+          <TextField id='countMinValue' value={countMinValue} onChange={onChange} label={strings.MIN} />
         </Grid>
         <Grid item xs={1} className={classes.flexContainer}>
           <ArrowForward />
         </Grid>
         <Grid item xs={5}>
-          <TextField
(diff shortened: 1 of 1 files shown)


src/components/seeds/database/filters/FilterCountWeight.tsx
```diff
@@ -177,32 +177,16 @@ export default function FilterCountWeight(props: Props): JSX.Element {
           />
         </Grid>
         <Grid item xs={5}>
-          <TextField
-            id='countMinValue'
-            value={countMinValue}
-            onChange={(value) => onChange('countMinValue', value)}
-            label={strings.MIN}
-          />
+          <TextField id='countMinValue' value={countMinValue} onChange={onChange} label={strings.MIN} />
```
karthikbtf, line 180: did we miss updating this signature in the onChange(id,value) cleanup?

karthikbtf: nvm, i think we should use the Textfield from web-components.. why this one left over?

karthikbtf: @charlie-tf just confirming.. the filters epic will be redoing these filters,, yes?

karthikbtf (approved):

src/components/seeds/database/filters/FilterCountWeight.tsx
charlie-tf: Hey Karthik. Yes, the seeds database filters are to be largely revamped

constanzauanini: I don't think this bug was introduced with the onChange pr, because these textfields, as you said, are not from web-components. We can delete all this after the new implementation

PR #1065, opened 2023-01-20
SW-2761 default all time zone tooltips to lorem ipusum, in preparation for design copy

karthikbtf:
- renamed the string constants to keep with the `TOOLTIP_XYZ_ABC` pattern used elsewhere
- switched to using default placeholder content for all time zone tool tips
- updated use-cases in the JIRA instead https://terraformation.atlassian.net/browse/SW-2761
- will get Ming to supply the copy
- happy to revert any changes here if anyone has suggestions
- I will assign the JIRA to Ming after this gets in, otherwise it will be super confusing for Ming to see the tooltips in the various scenarios

### DIFF
--- a/src/components/SeedBank/index.tsx
+++ b/src/components/SeedBank/index.tsx
@@ -112,7 +112,7 @@ export default function SeedBankDetails(): JSX.Element {
               id='timezone'
               type='text'
               value={tz.longName}
-              tooltipTitle={seedBank?.timeZone ? undefined : strings.ORG_TZ_TOOLTIP_INFO}
+              tooltipTitle={strings.TOOLTIP_TIME_ZONE_SEEDBANK}
               display={true}
             />
           </Grid>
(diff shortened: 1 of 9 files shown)


charlie-tf (approved):
lgtm!

src/components/SeedBank/index.tsx
```diff
@@ -112,7 +112,7 @@ export default function SeedBankDetails(): JSX.Element {
               id='timezone'
               type='text'
               value={tz.longName}
-              tooltipTitle={seedBank?.timeZone ? undefined : strings.ORG_TZ_TOOLTIP_INFO}
+              tooltipTitle={strings.TOOLTIP_TIME_ZONE_SEEDBANK}
```
constanzauanini, line 115: With this change, if we manually set a timezone for the seedbank, we will still see this tooltip that says it's inherited from org, and in this case it would not be inherited

constanzauanini: Oh, nevermind, i see it's lorem ipsum now

karthikbtf: Yes, I switched them all to use the placeholder and assigned the jira to Ming - to provide the values.

PR #1067, opened 2023-01-20
SW-2721 Load strings dynamically

sgrimm:
Use dynamic import instead of a static import to load the strings table for the
user's locale.

Currently, only the default locale of English is supported, but this change is
a prerequisite for loading other locales.


karthikbtf (approved):
Thanks so much.. doing the fixes in small chunks certainly helped the review process.
👍

PR #1068, opened 2023-01-20
SW-2733 Pass localization parameters to components

sgrimm:
Some of the components in web-components can display localizable strings.
Update the application to pass the necessary strings and render functions
to the components so they will render in the user's locale.

The general approach here is to put the localization parameters in wrapper
functions that pass through to the actual components. Most of the changes
are thus replacing imports of the underlying components with imports of
the wrappers.


sgrimm:
Haven't merged the web-components side of this yet, so expect the build to fail because it can't find the dependency. Will update this PR once that one is merged, but I want to give people a chance to review both sides of this change before either one goes in.

package.json
```diff
@@ -14,7 +14,7 @@
     "@mui/styled-engine-sc": "^5.8.0",
     "@mui/styles": "^5.11.2",
     "@mui/x-date-pickers": "^5.0.0-alpha.7",
-    "@terraware/web-components": "^2.0.72",
+    "@terraware/web-components": "^2.1.0",
```
sgrimm, line 17: I think this will be the version number after I merge https://github.com/terraware/web-components/pull/261 but I will update it if I'm wrong.

src/components/Inventory/InventoryTable.tsx
```diff
@@ -1,14 +1,15 @@
...
-import { Table, TableColumnType } from '@terraware/web-components';
+import { TableColumnType } from '@terraware/web-components';
 import { Box, Grid, useTheme } from '@mui/material';
 import { SearchResponseElement } from 'src/api/search';
 import InventoryCellRenderer from './InventoryCellRenderer';
 import { InventoryFiltersType } from './InventoryFiltersPopover';
 import PageSnackbar from 'src/components/PageSnackbar';
 import { APP_PATHS } from 'src/constants';
 import Search from './Search';
+import Table from 'src/components/common/table';
```
constanzauanini, line 12: Why are this all import changes (from web-components to common) ?

sgrimm: Because the versions in `common` now pass extra parameters to the components. The other option here would have been to add the parameters individually to every `<Table>` in the application, but it seemed cleaner to put them in one place.

https://github.com/terraware/terraware-web/pull/1068/files#diff-387ed938b72221d7f368435bd9c9e197815d14e9774e37bf7829892820c51def may make this clearer.

constanzauanini: Oh, excellent! Thank you. Looks good!

constanzauanini (approved):
Once web-component version is updated looks good

PR #1069, opened 2023-01-20
SW-2743 Add opt-in flag for weight system

constanzauanini:


src/features.ts
```diff
@@ -43,6 +43,15 @@ export const OPT_IN_FEATURES: Feature[] = [
     description: ['Allow choosing time zone'],
     disclosure: ['This is WIP.'],
   },
+  {
+    name: 'Weight system',
```
karthikbtf, line 47: weight units?
preference can be `enablePreferredWeightUnits`

karthikbtf (approved):
a suggestion if you prefer

karthikbtf (approved):

PR #1072, opened 2023-01-20
SW-2737 Add locale selector and gibberish locale

sgrimm:
Add an opt-in feature to show a locale selector in the left nav footer.

A selector is no good without things to select, so add a programmatically
generated "gibberish" locale that transforms the English strings.


karthikbtf (approved):
looks good to me 👍

sgrimm:
Looks like a GitHub Actions glitch; rerunning the CI build.

PR #1074, opened 2023-01-20
SW-2763 Add 'got it' button to time zone initialization notification

karthikbtf:
- add button which closes the notification and marks user preference to not show notification again
- removed some earlier fixes that were not fixes

Terraware App 2023-01-20 11-20-29


karthikbtf:
still seeing some funkiness on my local session.. may need to redo this further

src/components/AddNewOrganizationModal.tsx
```diff
@@ -110,7 +110,7 @@ export default function AddNewOrganizationModal(props: AddNewOrganizationModalPr
         strings.ORGANIZATION_CREATED_MSG,
         strings.formatString(strings.ORGANIZATION_CREATED_TITLE, response.organization.name)
       );
-      reloadData(response.organization.id);
```
constanzauanini, line 113: Why are we reverting this change? I just tested adding a new org in Vercel and I'm not getting changed into that org

karthikbtf: we never changed to new org.. we only displayed a page message saying you can access your orgs in the list above.
only time we change org if user did not have an org to start with. that still works.
not sure why i assumed we changed orgs.. it was wrong.

PR #1081, opened 2023-01-23
SW-2744 Create weight system selector

constanzauanini:
Just adding the dropdown on my account page.
Saving to user preferences will be done separately.


src/components/MyAccount/index.tsx
```diff
@@ -369,6 +373,26 @@ const MyAccountContent = ({
                 </Grid>
               </>
             )}
+            {weightUnitsEnabled && (
+              <Grid item xs={isMobile ? 12 : 4} sx={{ '&.MuiGrid-item': { paddingTop: theme.spacing(2) } }}>
```
karthikbtf, line 377: may want a little bit of marginTop in mobile view

karthikbtf (approved):

src/components/MyAccount/index.tsx
```diff
@@ -369,6 +373,24 @@ const MyAccountContent = ({
                 </Grid>
               </>
             )}
+            {weightUnitsEnabled && (
+              <Grid item xs={isMobile ? 12 : 4} sx={{ '&.MuiGrid-item': { paddingTop: theme.spacing(3) } }}>
```
karthikbtf, line 377: i think you want margin top 3 in mobile only.. in desktop view it shares same row as time zones and the margin top is not consistent

src/components/MyAccount/WeightSystemSelector.tsx
```diff
@@ -0,0 +1,20 @@
+import { Dropdown } from '@terraware/web-components';
```
karthikbtf, line 1: move this under `/components`?

src/components/MyAccount/index.tsx
constanzauanini: Good catch!

karthikbtf (approved):

karthikbtf (approved):
🎉

PR #1082, opened 2023-01-23
Constanza/sw 2749

constanzauanini:
Save selected value into user preferences


src/components/MyAccount/index.tsx
```diff
@@ -112,6 +116,12 @@ const MyAccountContent = ({
     }
   }, [organizations, loadedStringsForLocale]);
 
+  useEffect(() => {
+    if (userPreferences?.preferredWeightSystem) {
```
karthikbtf, line 120: might be easier to use https://github.com/terraware/terraware-web/blob/main/src/api/preferences/preferences.ts#L71
i still don't know what's better yet.. 
i have been using ^^
same for update

karthikbtf (commented):
one comment..
also update the PR title and description?

karthikbtf (approved):

src/components/MyAccount/index.tsx
karthikbtf: ignore this ^^ i will revisit the pattern later

PR #1083, opened 2023-01-23
SW-2779 Edit org improvements

constanzauanini:
Make country, state and time zone labels mandatory.
Use autocomplete for country and state.


src/components/EditOrganization/index.tsx
```diff
@@ -132,6 +133,27 @@ export default function OrganizationView({ organization, reloadOrganizationData
     return 6;
   };
 
+  const countryToDropdownItem = (country: Country) => {
+    return { label: country.name, value: country.code };
+  };
```
karthikbtf, line 138: could merge the country and state ...
```
const toDropdownItem = (entity: Country | Subdivision) => ({ label: entity.name, value: entity.code });
```

src/components/EditOrganization/index.tsx
```diff
@@ -132,6 +133,27 @@ export default function OrganizationView({ organization, reloadOrganizationData
...
+
+  const stateToDropdownItem = (subdivision: Subdivision) => {
+    return { label: subdivision.name, value: subdivision.code };
+  };
+
+  const subdivisionOptions = () => {
+    const country = getSelectedCountry();
+    if (country) {
+      return country.subdivisions.map((subd) => stateToDropdownItem(subd));
+    }
```
karthikbtf, line 154: return empty array otherwise

src/components/EditOrganization/index.tsx
```diff
@@ -132,6 +133,27 @@ export default function OrganizationView({ organization, reloadOrganizationData
...
 
+  const countryToDropdownItem = (country: Country) => {
+    return { label: country.name, value: country.code };
+  };
+
+  const countriesOptions = () => {
+    if (countries) {
+      return countries.map((country) => countryToDropdownItem(country));
+    }
+  };
```
karthikbtf, line 144: return empty array otherwise.. we can remove the || [] from the autocomplete..
might be good for these functions to return the options based on current state of the component.. not only if countries or country is set?

src/components/EditOrganization/index.tsx
```diff
@@ -172,24 +194,34 @@ export default function OrganizationView({ organization, reloadOrganizationData
             />
           </Grid>
           <Grid item xs={gridSize()} paddingBottom={theme.spacing(4)}>
-            <Select
-              label={strings.COUNTRY}
+            <Autocomplete
```
karthikbtf, line 197: wondering if we should create a dropdown component (like we do for select) that handles {label,value} semantics... avoid some repetition

karthikbtf (commented):
looks good, some thoughts

src/components/EditOrganization/index.tsx
constanzauanini: Yes, that would be useful

karthikbtf: ok, it was a suggestion..  
will do this as part of 2781

karthikbtf (approved):
thanks

PR #1084, opened 2023-01-23
SW-2781 Create org flow, time zone workflow changes

karthikbtf:
- move below country
- make country/state required
- TODO: allow country/state to be searched with input (to be merged with changes from other PR by Constanza)
- we need 2 components, country and state selection components maybe

Terraware App 2023-01-23 09-12-51


src/components/AddNewOrganizationModal.tsx
```diff
@@ -186,39 +200,41 @@ export default function AddNewOrganizationModal(props: AddNewOrganizationModalPr
...
-              label={strings.TIME_ZONE_REQUIRED}
-              onTimeZoneSelected={onTimeZoneChange}
-              selectedTimeZone={newOrganization.timeZone}
-              tooltip={strings.TOOLTIP_TIME_ZONE_ORGANIZATION}
-              errorText={timeZoneError}
-            />
-          </Grid>
-        )}
         <Grid item xs={12}>
           <Select
```
karthikbtf, line 204: @constanzauanini i will change these to autocomplete as well.. i'm wondering if we should just have a component that does the country/state selection since it's redundant.. i can do that after you merge your PR

constanzauanini (approved):

src/components/RegionSelector/index.tsx
```diff
@@ -0,0 +1,143 @@
+import { Grid, useTheme } from '@mui/material';
```
karthikbtf, line 1: @constanzauanini i created a shared component for region selection - used in org create and edit.
not using the DropdownAutocomplete yet.. something is not working there.
will follow up separately

constanzauanini (approved):
Great! Lot of code was re-use with this :)

PR #1085, opened 2023-01-23
SW-2780 Replace autocomplete of TimeSelector with textfield when it's state is disabled

constanzauanini:
Screenshot 2023-01-23 at 15 37 05

### DIFF
--- a/src/components/TimeZoneSelector.tsx
+++ b/src/components/TimeZoneSelector.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo } from 'react';
-import { Autocomplete } from '@terraware/web-components';
+import { Autocomplete, Textfield } from '@terraware/web-components';
 import strings from 'src/strings';
 import { useTimeZones } from 'src/providers';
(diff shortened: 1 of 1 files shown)


src/components/TimeZoneSelector.tsx
```diff
@@ -40,7 +40,16 @@ export default function TimeZoneSelector(props: TimeZoneSelectorProps): JSX.Elem
     return optionA?.value === optionB?.value;
   };
 
-  return (
+  return disabled ? (
+    <Textfield
+      id='timezones2'
```
karthikbtf, line 45: missed this earlier.. can we call it 'time-zone-selector-disabled' and below 'time-zone-selector'? or anything else that is more clarifying?

karthikbtf (approved):
looks good..

src/components/TimeZoneSelector.tsx
constanzauanini: Yes. Not sure why this was the id before

karthikbtf (approved):

PR #1086, opened 2023-01-23
SW-2736 Fix can't edit newly-invited user

constanzauanini:
In people table in people page:
- Make email clikable instead of Name and Last Name to allow editing users with no name/lastname
- Sort rows on table by email


karthikbtf (approved):
Here we can add a screenshot next time since there's a visual change

PR #1089, opened 2023-01-24
SW-2785 Make country and state required if available in org edit

constanzauanini:
Screenshot 2023-01-24 at 11 54 43
Screenshot 2023-01-24 at 11 55 13

### DIFF
--- a/src/components/EditOrganization/index.tsx
+++ b/src/components/EditOrganization/index.tsx
@@ -55,18 +55,23 @@ export default function OrganizationView({ organization, reloadOrganizationData 
   };
 
   const saveOrganization = async () => {
+    let hasErrors = false;
     if (organizationRecord.name === '') {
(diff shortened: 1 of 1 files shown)


src/components/EditOrganization/index.tsx
```diff
@@ -53,17 +56,28 @@ export default function OrganizationView({ organization, reloadOrganizationData
...
+
+    if (!organizationRecord.countryCode) {
+      setCountryError(strings.REQUIRED_FIELD);
+      return;
+    }
+
+    if (requireSubdivision && !organizationRecord.countrySubdivisionCode) {
+      setSubdivisionError(strings.REQUIRED_FIELD);
+      return;
+    }
```
karthikbtf, line 71: can we return after all the checks? maybe use a flag to capture hasErrors or something

karthikbtf (approved):
one suggestion

PR #1091, opened 2023-01-24
Set Accept-Language header to current locale

sgrimm:
The server looks at the `Accept-Language` header to figure out which locale to use
in its API responses. Update the client to set that header on outgoing requests.

### DIFF
--- a/src/providers/LocalizationProvider.tsx
+++ b/src/providers/LocalizationProvider.tsx
@@ -5,6 +5,7 @@ import { TimeZoneDescription } from 'src/types/TimeZones';
 import strings, { ILocalizedStringsMap } from 'src/strings';
 import { ProvidedLocalizationData } from '.';
 import { supportedLocales } from '../strings/locales';
+import axios from 'axios';
 
 export type LocalizationProviderProps = {
   children?: React.ReactNode;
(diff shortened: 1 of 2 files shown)


src/providers/LocalizationProvider.tsx
```diff
@@ -23,9 +24,14 @@ export default function LocalizationProvider({
 }: LocalizationProviderProps): JSX.Element | null {
   const [timeZones, setTimeZones] = useState<TimeZoneDescription[]>([]);
 
+  useEffect(() => {
+    axios.defaults.headers = { ...axios.defaults.headers, 'Accept-Language': locale };
```
sgrimm, line 28: Not sure if there's a more idiomatic way to do this.

karthikbtf: This is fine. Should we clear the header value if locale is not set?
Another way is using an [interceptor](https://axios-http.com/docs/interceptors).
No preference here, just fyi.
I should think about abstracting the network functionality into an Http service at some point... just an aside.

sgrimm: There's no way to unset the locale -- it defaults to English (in `AppBootstrap`) and can only be updated to other nonempty values.

karthikbtf: Got it, at some point when we read the locale from the user (to override the default locale).. guessing we will ignore unset locales.

sgrimm: Yep, and in fact I just submitted a PR to do exactly that: #1092 .

PR #1092, opened 2023-01-24
SW-2776 Load and save user locale selection

sgrimm:
At initialization time, use the locale from the user's profile, if any.

When the user changes locales, save the new selection to their profile.

When the user edits their profile, save the currently-selected locale along with
the other profile data.

### DIFF
--- a/src/components/MyAccount/index.tsx
+++ b/src/components/MyAccount/index.tsx
@@ -97,6 +97,7 @@ const MyAccountContent = ({
   const contentRef = useRef(null);
   const timeZonesEnabled = isEnabled('Timezones');
   const weightUnitsEnabled = isEnabled('Weight units');
+  const locale = useLocalization().locale;
   const timeZones = useTimeZones();
   const tz = timeZones.find((timeZone) => timeZone.id === record.timeZone) || getUTC(timeZones);
   const [preferredWeightSystemSelected, setPreferredWeightSystemSelected] = useState(
@@ -197,7 +198,9 @@ const MyAccountContent = ({
   };
 
   const saveProfileChanges = async () => {
-    const updateUserResponse = await updateUserProfile(record);
+    // Save the currently-selected locale, even if it differs from the locale in the profile data we
+    // fetched from the server.
+    const updateUserResponse = await updateUserProfile({ ...record, locale });
     return updateUserResponse;
   };
 
--- a/src/components/MyAccount/index.tsx
+++ b/src/components/MyAccount/index.tsx
@@ -97,7 +97,7 @@ const MyAccountContent = ({
   const contentRef = useRef(null);
   const timeZonesEnabled = isEnabled('Timezones');
   const weightUnitsEnabled = isEnabled('Weight units');
-  const locale = useLocalization().locale;
+  const { locale } = useLocalization();
   const timeZones = useTimeZones();
   const tz = timeZones.find((timeZone) => timeZone.id === record.timeZone) || getUTC(timeZones);
   const [preferredWeightSystemSelected, setPreferredWeightSystemSelected] = useState(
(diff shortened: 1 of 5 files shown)


src/components/MyAccount/index.tsx
```diff
@@ -197,7 +198,9 @@ const MyAccountContent = ({
   };
 
   const saveProfileChanges = async () => {
-    const updateUserResponse = await updateUserProfile(record);
+    // Save the currently-selected locale, even if it differs from the locale in the profile data we
+    // fetched from the server.
+    const updateUserResponse = await updateUserProfile({ ...record, locale });
```
sgrimm, line 203: My solution to what we were talking about in Slack. This seemed like the simplest approach. Previously I'd been trying to figure out what sequence of hooks to use to get `record` to always include the correct locale setting, but then I realized the only thing that matters is what locale we submit to the server.

src/components/MyAccount/index.tsx
```diff
@@ -97,6 +97,7 @@ const MyAccountContent = ({
   const contentRef = useRef(null);
   const timeZonesEnabled = isEnabled('Timezones');
   const weightUnitsEnabled = isEnabled('Weight units');
+  const locale = useLocalization().locale;
```
karthikbtf, line 100: preferred way is to deconstruct
`const { locale } = useLocalization();`

src/components/MyAccount/index.tsx
karthikbtf: This makes sense and totally works. This was my first hunch/suggestion last night on slack.

src/providers/LocalizationProvider.tsx
```diff
@@ -59,6 +67,19 @@ export default function LocalizationProvider({
     fetchStrings();
   }, [locale, setLoadedStringsForLocale]);
 
+  useEffect(() => {
+    const updateUserLocale = async () => {
+      const oldProfile = (await getUser())?.user;
+
+      if (oldProfile) {
+        const newProfile = { ...oldProfile, locale };
+        await updateUserProfile(newProfile, true);
```
karthikbtf, line 76: may need to call `reloadData` to update the user data in user provider.
reloadData is available in useUser();

karthikbtf (commented):
This approach works. A few comments/suggestions.

src/providers/LocalizationProvider.tsx
```diff
@@ -59,6 +67,19 @@ export default function LocalizationProvider({
     fetchStrings();
   }, [locale, setLoadedStringsForLocale]);
 
+  useEffect(() => {
+    const updateUserLocale = async () => {
+      const oldProfile = (await getUser())?.user;
```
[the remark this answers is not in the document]
sgrimm: No, you're right, I should be using that instead. This was left over from me playing with different ways to get the locale into the right place in MyAccount.

karthikbtf (approved):
Looks great to me!
Thanks so much!!

karthikbtf:
Missed seeing this earlier. App seems to switch between en and gx forever after reloading the app with locale saved as gx. Not sure if my testing is correct.

Switch could be because of the hardcoded 'en' in the app bootstrap.
The following change seems to fix it... not sure if this change is the most elegant.

```
export default function AppBootstrap({ children }: AppBootstrapProps): JSX.Element {
  const [locale, setLocale] = useState('');       <---- changed
  const [loadedStringsForLocale, setLoadedStringsForLocale] = useState<string | null>(null);

  return (
    <UserProvider>
      <UserContext.Consumer>.                    <--- introduced
        {({ user }) => (
          <OrganizationProvider>
            <LocalizationProvider
              locale={locale || user?.locale || 'en'}                  <--- changed
              setLocale={setLocale}
              loadedStringsForLocale={loadedStringsForLocale}
              setLoadedStringsForLocale={setLoadedStringsForLocale}
            >
              <BlockingBootstrap>{children}</BlockingBootstrap>
            </LocalizationProvider>
          </OrganizationProvider>
        )}
      </UserContext.Consumer>
    </UserProvider>
  );
}
```

sgrimm:
I’m actually in the middle of fixing this right now; sorry about that.

sgrimm:
#1096 is out to fix this.

PR #1093, opened 2023-01-24
SW-2745 Initialize weight system preference upon login

constanzauanini:
Screenshot 2023-01-24 at 15 45 55

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -153,7 +153,8 @@ function AppContent() {
   const { isDesktop, type } = useDeviceInfo();
   const classes = useStyles({ isDesktop });
   const location = useStateLocation();
-  const { organizations, selectedOrganization, reloadData, reloadPreferences, orgPreferences } = useOrganization();
+  const { organizations, selectedOrganization, reloadData, reloadPreferences, orgPreferences, userPreferences } =
+    useOrganization();
   const [withdrawalCreated, setWithdrawalCreated] = useState<boolean>(false);
   const [timeZoneInitializedForOrg, setTimeZoneInitializedForOrg] = useState<number>(defaultSelectedOrg.id);
   const { isProduction } = useEnvironment();
@@ -161,6 +162,7 @@ function AppContent() {
   const snackbar = useSnackbar();
   const timeZones = useTimeZones();
   const timeZoneFeatureEnabled = isEnabled('Timezones');
+  const weightUnitsEnabled = isEnabled('Weight units');
 
   // seedSearchCriteria describes which criteria to apply when searching accession data.
   const [seedSearchCriteria, setSeedSearchCriteria] = useState<SearchCriteria>(DEFAULT_SEED_SEARCH_FILTERS);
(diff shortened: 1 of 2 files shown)


karthikbtf:
@constanzauanini this interferes with the other time zone initialized message..
we only show one message today..
we need design to advise on how to show both.. should they be merged? do we need to show more than one message? etc etc.

karthikbtf:
> @constanzauanini this interferes with the other time zone initialized message.. we only show one message today.. we need design to advise on how to show both.. should they be merged? do we need to show more than one message? etc etc.

if we need to show both we will need to redo our page message framework

constanzauanini:
> @constanzauanini this interferes with the other time zone initialized message.. we only show one message today.. we need design to advise on how to show both.. should they be merged? do we need to show more than one message? etc etc.

That's true. I will ask design and come back to this

constanzauanini:
> > @constanzauanini this interferes with the other time zone initialized message.. we only show one message today.. we need design to advise on how to show both.. should they be merged? do we need to show more than one message? etc etc.
> 
> if we need to show both we will need to redo our page message framework

@karthikbtf I updated the page snackbar components with your changes, and now both messages are shown. I think we can leave both messages until we get an answer from design

src/App.tsx
```diff
@@ -262,6 +272,50 @@ function AppContent() {
     }
   }, [type]);
 
+  useEffect(() => {
+    if (showUnitSnackbar) {
+      snackbar.pageInfo(
+        strings.formatString<any>(
+          strings.UNITS_INITIALIZED_MESSAGE,
+          userPreferences.preferredWeightSystem ?? 'metric',
```
karthikbtf, line 280: the strings should be localized some how, perhaps check with Steve how

src/App.tsx
```diff
@@ -262,6 +272,50 @@ function AppContent() {
...
+        const response = await updatePreferences('preferredWeightSystem', 'metric');
+        if (response.requestSucceeded) {
+          setShowUnitSnackbar(true);
+        }
+      }
+
+      if (!userPreferences.unitsAcknowledgedOnMs) {
+        setShowUnitSnackbar(true);
+      }
+    };
```
karthikbtf, line 312: i have been moving logic like this into the api/user (or org depending on context) (which will be rewritten as a service soon).
fwiw - we have 2 ways of initializing .. time zone is done differently.. units done differently.. not sure if this can be done the same way for consistency.. if not no worries

src/App.tsx
```diff
@@ -262,6 +272,50 @@ function AppContent() {
...
+          },
+        },
+        'user'
+      );
+    }
+  }, [showUnitSnackbar, snackbar, userPreferences.preferredWeightSystem]);
+
+  useEffect(() => {
+    const initializaUnit = async () => {
+      if (!user || !userPreferences || !pathname) {
```
karthikbtf, line 298: why pathname? we just want this to happen once on first visit yes?

src/strings/strings-en.ts
```diff
@@ -974,6 +974,8 @@ export const strings = {
   UNDO_SEND_TO_NURSERY: 'Undo Send to Nursery',
   UNEXPECTED_ERROR: 'An unexpected error occurred.',
   UNITS: 'Units',
+  UNITS_INITIALIZED_MESSAGE: 'Your Weight System was updated to {0}. You can edit this setting from {1}.',
+  UNITS_INITIALIZED_TITLE: 'Preferred Weight System set',
```
karthikbtf, line 978: may not need to capitalize weight system

karthikbtf:
> > > @constanzauanini this interferes with the other time zone initialized message.. we only show one message today.. we need design to advise on how to show both.. should they be merged? do we need to show more than one message? etc etc.
> > 
> > 
> > if we need to show both we will need to redo our page message framework
> 
> @karthikbtf I updated the page snackbar components with your changes, and now both messages are shown. I think we can leave both messages until we get an answer from design

@constanzauanini yes i see that.. looks funny but yea design can follow up

src/App.tsx
```diff
@@ -262,6 +272,50 @@ function AppContent() {
...
+        return;
+      }
+
+      if (!userPreferences.preferredWeightSystem) {
```
karthikbtf, line 302: seems like we need to update the userPreferences in the provider after the update on line 303, or we may repeatedly set the preference on every useEffect? may need to confirm

karthikbtf (approved):
some comments

src/App.tsx
constanzauanini: Ok! I tried to follow your implementation for consistency, but notification part is a bit different. Let me know what you think

karthikbtf (approved):
looks good to me

karthikbtf (approved):
Thanks 🎉

PR #1097, opened 2023-01-25
DetectAppVersion Returns null if App Version is Stale

charlie-tf:
Tested locally for both stale and non-stale versions.


karthikbtf (approved):
🎉

PR #1103, opened 2023-01-26
SW-2755 Add weight units columns

constanzauanini:
Screenshot 2023-01-26 at 15 26 26
Screenshot 2023-01-26 at 15 27 30


src/components/seeds/database/EditColumns.tsx
```diff
@@ -155,51 +157,59 @@ function sections(): Section[] {
...
           columns.collectionSiteLandowner,
           columns.collectionSiteNotes,
-          columns.estimatedWeightGrams,
+          columns.collectionSiteName,
         ],
       ],
     },
     {
-      name: 'Processing and Drying',
+      name: strings.PROCESSING_AND_DRYING,
```
karthikbtf, line 196: thanks for catching all these

src/components/seeds/database/EditColumns.tsx
```diff
@@ -217,4 +227,23 @@ function sections(): Section[] {
       ],
     },
   ];
+
+  if (weightUnitsEnabled) {
+    sections
+      .splice(
+        2,
+        0,
+        weightUnitsEnabled && {
```
karthikbtf, line 236: don't need this check, it's under an `if (weightUnitsEnabled)` check

src/components/seeds/database/EditColumns.tsx
```diff
@@ -217,4 +227,23 @@ function sections(): Section[] {
       ],
     },
   ];
+
+  if (weightUnitsEnabled) {
```
karthikbtf, line 231: Can you add a note: when we undo the feature check - we should move this code to insert columns in the right place.. without the splice.

karthikbtf (approved):
looks good, a few suggestions

PR #1104, opened 2023-01-26
rename ServerOrganization type to just Organization

karthikbtf:
- **no** other changes but this touches many files

### DIFF
--- a/src/api/organization/organization.ts
+++ b/src/api/organization/organization.ts
@@ -1,5 +1,5 @@
 import axios from 'src/api/index';
-import { ServerOrganization } from 'src/types/Organization';
+import { Organization } from 'src/types/Organization';
 import { paths } from 'src/api/types/generated-schema';
 import { Facility } from '../types/facilities';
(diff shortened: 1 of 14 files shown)


karthikbtf:
Why the rename? Not sure why we had 'Server' in the name - we don't use this pattern in other types.

charlie-tf (approved):
Thanks!

constanzauanini:
> Why the rename? Not sure why we had 'Server' in the name - we don't use this pattern in other types.

I think it was because we had another Organization type in the FE, but it may not exist anymore

constanzauanini (approved):

karthikbtf:
> > Why the rename? Not sure why we had 'Server' in the name - we don't use this pattern in other types.
> 
> I think it was because we had another Organization type in the FE, but it may not exist anymore

Thanks @constanzauanini - yes I noticed we don't have another one. This rename seems consistent with other types we have.. and just seems simpler/cleaner. hope that's ok.

PR #1105, opened 2023-01-26
SW-2756 Change default preset for imperial preference

constanzauanini:
Change default preset, to include estimatedWeightOunces column instead of estimatedWeightGrams.

Also order columns according to preferred system

### DIFF
--- a/src/components/seeds/database/columns.ts
+++ b/src/components/seeds/database/columns.ts
@@ -1,5 +1,6 @@
 import strings from 'src/strings';
 import { DatabaseColumn } from '@terraware/web-components/components/table/types';
+import isEnabled from 'src/features';
 
 function columns(): DatabaseColumn[] {
   return [
@@ -319,6 +320,20 @@ export const defaultPreset: Preset = {
   ],
 };
 
+export const defaultImperialPreset: Preset = {
+  name: 'Default',
+  fields: [
+    'accessionNumber',
+    'speciesName',
+    'state',
+    'collectionSiteName',
+    'collectedDate',
+    'ageMonths',
+    'estimatedWeightOunces',
+    'estimatedCount',
+  ],
+};
+
 const generalInventoryPreset: Preset = {
   name: 'General Inventory',
   fields: [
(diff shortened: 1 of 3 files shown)


src/components/seeds/database/columns.ts
```diff
@@ -319,6 +320,20 @@ export const defaultPreset: Preset = {
   ],
 };
 
+export const defaultImperialPreset: Preset = {
```
karthikbtf, line 323: suggestion, convert `defaultPreset` to a function and pass in boolean imperialSystem.. and on line 318 (imperialSystem ? 'estimatedWeightOunces' : 'estimatedWeightGrams') .. since the only difference between the two is the weight system.. otherwise we have to keep 2 default lists in sync.. just extra maintenance.

src/components/seeds/database/columns.ts
```diff
@@ -389,12 +404,19 @@ const germinationTestingPreset: Preset = {
...
-  defaultPreset,
-  generalInventoryPreset,
-  seedStoragePreset,
-  viabilitySummaryPreset,
-  germinationTestingPreset,
-];
+export const searchPresets = (preferredWeightSystem: string) => {
+  const weightUnitsEnabled = isEnabled('Weight units');
+
+  if (weightUnitsEnabled && preferredWeightSystem === 'imperial') {
```
karthikbtf, line 410: if you use my suggestion above, this gets simpler.. and only one list to maintain

karthikbtf (approved):
one suggestion, i hope you may consider it

src/components/seeds/database/columns.ts
```diff
@@ -305,18 +306,20 @@ export interface Preset {
...
+  return {
+    name: 'Default',
+    fields: [
+      'accessionNumber',
+      'speciesName',
+      'state',
+      'collectionSiteName',
+      'collectedDate',
+      'ageMonths',
+      system === 'imperial' ? 'estimatedWeightOunces' : 'estimatedWeightGrams',
```
karthikbtf, line 319: guessing 'metric' is the default today?

karthikbtf (approved):
thanks for the refactor

src/components/seeds/database/columns.ts
constanzauanini: Yes, metric is default

PR #1106, opened 2023-01-26
SW-2794 Added Organization*Services

karthikbtf:
- OrganizationService for organization CRUD
- OrganizationUserService for organization user CRUD
- renamed AllOrganizationRoles as OrganizationRole
- updated/tested all changes to calls


src/services/index.ts
```diff
@@ -1,6 +1,17 @@
...
 import PreferencesService from './PreferencesService';
 import UserService from './UserService';
 
-export { CachedUserService, HttpService, PreferencesService, UserService };
+export type { Response } from './HttpService';
+
+export {
+  CachedUserService,
+  HttpService,
+  OrganizationUserService,
```
constanzauanini, line 13: nitpick: if these are in alphabetical order, OrganizationService should be before OrganizationUserService

constanzauanini (approved):
Looks good to me!

src/services/index.ts
karthikbtf: thanks, that's how i would usually sort..  in the strings file we are sorting differently.. applied that here. will fix here.

karthikbtf:
```
node -e 'console.log(["OrganizationService", "OrganizationUserService"].sort())'
[ 'OrganizationService', 'OrganizationUserService' ]
```
oh well...

PR #1107, opened 2023-01-27
SW-2757 Show message first time user select imperial unit column

constanzauanini:
Screenshot 2023-01-27 at 10 26 08


src/components/seeds/database/index.tsx
```diff
@@ -200,10 +202,45 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
+          strings.CHANGE_DEFAULT_WEIGHT_SYSTEM,
+          <Link to={APP_PATHS.MY_ACCOUNT}>{strings.MY_ACCOUNT}</Link>
+        ),
+        '',
+        {
+          label: strings.GOT_IT,
+          apply: async () => {
+            await PreferencesService.updateUserPreferences({ defaultWeightSystemAcknowledgedOnMs: Date.now() });
+            reloadPreferences();
+          },
```
karthikbtf, line 223: unrelated to this PR but a request if possible.. if we could move this as a utility hook somewhere and use it in all those places that have a got-it button to set a preference and reload preference (time zone, unit, etc)...
example:
```
useGetAcknowledgedPreferenceContent(preferenceNames: string[]) {
  const { reloadPreferences } = useOrganization();
return {
  label: strings. GOT_IT,
  apply: async () => {
    await Promise.allSettled(preferenceNames.map(name => PreferencesService.upateUserPreference(name, Date.now()));
    reloadPreferences();
  }
}
```
np if we want to do this later...

karthikbtf (approved):

src/components/seeds/database/index.tsx
constanzauanini: I just tried doing this, but not all the setting preferences have the reload after the update. Some of them have a validation to update or not. For example: 
```
(      if (userUnit.updated) {
        reloadPreferences();
      } ) 
```
And another problem is that I can't call a hook from a useEffect, and we are updating froma useEffect

karthikbtf: Ok got it. No worries.

karthikbtf: Thanks for trying this out,

PR #1108, opened 2023-01-28
SW-2797 Update country names on locale change

sgrimm:
When the user switches locales, we need to show the new locale's country names.


karthikbtf (approved):
Thanks

PR #1109, opened 2023-01-28
SW-2808 move app/server specific calls into a system service

karthikbtf:
- get server clock time (currently unused)
- get app version
- some typing refactor (for GET calls only, to test - others will be followed up on later if the current test is not hideous)


karthikbtf:
My main objective was to remove as much boilerplate  code from all the api calls in services.
Sadly we have a ton of that.
Wrestled with typescript all day, eventually learned that I can't figure it out.. or it can't be supported.
Came up with a mediocre solution,, seems more like overhead but avoids some ugly post-processing.
will probably live with some boilerplate for a bit

karthikbtf:
This one is ready to go, I was unable to make any progress with typescript/generics (to remove redundant patterns) and don't want to waste any more time.

constanzauanini (approved):

PR #1115, opened 2023-01-30
SW-2799 Localize "check species" problem names

sgrimm:
The problem names in the "check species" tooltips need to use localizable strings
rather than the English enum names.


karthikbtf (approved):
👍  looks good
just curious what our preference is re: when to localize server side vs client side.

sgrimm:
My approach is more or less that if the client treats something as an enum, it gets translated client-side, but if the client treats it as an arbitrary string, it gets translated server-side.

I suspect we'll end up revisiting that approach down the road since it does lead to duplication of strings, but I think it's the shortest path to Spanish for now.

PR #1116, opened 2023-01-30
SW-2820 Look up title of Notifications dropdown

sgrimm:
The Notifications dropdown's title was using a hardwired English string.


karthikbtf (approved):
Thanks!

PR #1117, opened 2023-01-30
SW-2821 Show localized enums on species list

sgrimm:
Update the species list page so it always shows localized strings for the growth
form and seed storage behavior columns.

On locale change, reset all the filters and reload the search results because
simply updating the translations in place might cause the list to be shown in the
wrong order.

This requires a minor change to the way the "edit" button works; it now fetches
the species data explicitly rather than relying on search results being
interchangeable with species API payloads.

### DIFF
--- a/src/components/Species/index.tsx
+++ b/src/components/Species/index.tsx
@@ -2,14 +2,14 @@ import { Box, Container, Grid, IconButton, Theme } from '@mui/material';
 import { makeStyles } from '@mui/styles';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { useRecoilState } from 'recoil';
-import { deleteSpecies } from 'src/api/species/species';
+import { deleteSpecies, getSpecies } from 'src/api/species/species';
 import Button from 'src/components/common/button/Button';
 import EmptyMessage from 'src/components/common/EmptyMessage';
 import Table from 'src/components/common/table';
 import { TableColumnType } from 'src/components/common/table/types';
 import speciesAtom from 'src/state/species';
 import strings from 'src/strings';
-import { Species, SpeciesProblemElement } from 'src/types/Species';
+import { getGrowthFormString, getSeedStorageBehaviorString, Species, SpeciesProblemElement } from 'src/types/Species';
 import TfMain from 'src/components/common/TfMain';
 import PageSnackbar from 'src/components/PageSnackbar';
 import AddSpeciesModal from './AddSpeciesModal';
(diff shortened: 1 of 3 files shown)


src/types/Species.ts
```diff
@@ -20,22 +20,22 @@ export type SpeciesProblemElement = {
...
 
-export function growthForms() {
+export function growthForms(useLocalizedValues = false) {
   return [
-    { label: strings.TREE, value: 'Tree' },
-    { label: strings.SHRUB, value: 'Shrub' },
-    { label: strings.FORB, value: 'Forb' },
-    { label: strings.GRAMINOID, value: 'Graminoid' },
-    { label: strings.FERN, value: 'Fern' },
+    { label: strings.TREE, value: useLocalizedValues ? strings.TREE : 'Tree' },
```
constanzauanini, line 25: Don't we need to translate the label instead of the value? Because I tried editing a species and then saving (in gibberish mode) and i get an api error {"error":{"message":"Field value invalid: growthForm"},"status":"error"}

sgrimm: You mean you tried with this version of the code, or with main? That bug should be fixed by this PR.

constanzauanini: With this version, in Vercel

sgrimm: Ah, oops, the fix was actually a backend thing and I neglected to send out a PR for it.

sgrimm: PR is out now: https://github.com/terraware/terraware-server/pull/930

src/components/Species/index.tsx
```diff
@@ -422,14 +434,18 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
     snackbar.toastError(snackbarMessage);
   };
 
-  const OnEditSpecies = () => {
-    setSelectedSpecies(selectedSpeciesRows[0]);
-    setEditSpeciesModalOpen(true);
+  const openEditSpeciesModal = async (speciesId: number) => {
+    const speciesResponse = await getSpecies(speciesId, selectedOrganization.id.toString());
```
karthikbtf, line 438: do we have enough info to reinterpret the search result payload as an api payload? just curious about this fetch..

karthikbtf (approved):
one question... otherwise lgtm

src/components/Species/index.tsx
sgrimm: It's theoretically possible but would be awkward with the way the code is set up right now. You'd have to map from the previous locale's display names for the enum fields to the new locale's names, but the way it's structured right now, by the time you know the new names, the old ones have already been discarded from the `strings` object, so you'd have to keep a cache around.

If we care about this, a better approach might be to stop having the search API expect localized enum values as input. I thought about doing that but I didn't like that it would mean you'd do "search for the exact value `Foo`" and get back results where the value is `Bar` (or whatever the translation of `Foo` is). The way it is now is inconsistent too, though, just in a different way.

karthikbtf: Ok, got it, makes total sense.
Avoiding a network request between user click and viewing the modal would be ideal but in this case correctness is more important.

karthikbtf (approved):
👍

PR #1118, opened 2023-01-30
SW-2814 Facilty service to search/get/create/update facilities - reused in Nursery and Seedbank views

karthikbtf:
- moved existing code into a search service
- re-exporting search types and functions in original file to avoid breaking other use-cases that rely on a direct search.. until refactored
- introduced a facility service which leverages search service, updated seedbanks and nurseries views to use this service and rely less on how search works
- refactored seed banks view to always search (avoid using cached seed banks) similar to nurseries

### DIFF
--- a/src/components/Nursery/index.tsx
+++ b/src/components/Nursery/index.tsx
@@ -3,7 +3,7 @@ import { useEffect, useState } from 'react';
 import { useHistory, useParams } from 'react-router-dom';
 import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
-import { getAllNurseries } from 'src/utils/organization';
+import { FacilityService } from 'src/services';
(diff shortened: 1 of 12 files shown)


src/components/Nursery/index.tsx
```diff
@@ -36,11 +36,15 @@ export default function NurseryDetails(): JSX.Element {
...
+      const selectedNursery = FacilityService.getFacility({
+        organization: selectedOrganization,
+        facilityId: nurseryId,
+        type: 'Nursery',
+      });
       if (selectedNursery) {
         setNursery(selectedNursery);
       } else {
-        history.push(APP_PATHS.SEED_BANKS);
+        history.push(APP_PATHS.NURSERIES);
```
constanzauanini, line 47: 😮 good fix!

PR #1120, opened 2023-01-31
SW-2283 Add Ecosystem Type to Species

charlie-tf:
- display a column for ecosystem type
- allow ecosystem type to be filterable
- when adding or editing a species, multiple ecosystem types may be chosen

image
image
image
image


src/components/Species/AddSpeciesModal.tsx
```diff
@@ -339,6 +346,39 @@ export default function AddSpeciesModal(props: AddSpeciesModalProps): JSX.Elemen
...
+            onAdd={(type: EcosystemType) => {
+              const selectedTypes = record.ecosystemTypes ?? [];
+              selectedTypes.push(type);
+              onChange('ecosystemTypes', selectedTypes);
+            }}
+            onRemove={(type: EcosystemType) => {
+              const selectedTypes = record.ecosystemTypes ?? [];
+              const index = selectedTypes.findIndex((x) => x === type);
+              selectedTypes.splice(index, 1);
+              onChange('ecosystemTypes', selectedTypes);
```
karthikbtf, line 374: Above code works.. in case it helps  functional constructs for readability/brevity ..
‘’’
onChange(‘ecosystemType’, record.ecosystemTypes?.filter(ecosystemType => ecosystemType !== type) ?? []);
‘’’
Can’t find the back tick on my phone

src/types/Species.ts
```diff
@@ -10,8 +10,25 @@ export type Species = {
   rare?: boolean;
   seedStorageBehavior?: 'Orthodox' | 'Recalcitrant' | 'Intermediate' | 'Unknown';
   problems?: SpeciesProblemElement[];
+  ecosystemTypes?: EcosystemType[];
 };
```
karthikbtf, line 15: A thought.. if this was an enumeration, we could refer to them by enum key in the map below.

karthikbtf (approved):
Looks good to me.. some comments.

src/components/Species/AddSpeciesModal.tsx
```diff
@@ -339,6 +346,39 @@ export default function AddSpeciesModal(props: AddSpeciesModalProps): JSX.Elemen
             }
           />
         </Grid>
+        <Grid item xs={12}>
+          <MultiSelect
+            fullWidth={true}
+            label={strings.ECOSYSTEM_TYPE}
+            tooltipTitle={
+              <>
+                {`${strings.TOOLTIP_ECOSYSTEM_TYPE} `}
```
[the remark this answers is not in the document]
charlie-tf: I see what you're saying. In this case it's low risk, as the space denotes a break between separate sentences, but I'll create a ticket to see how we might clean this up.

src/types/Species.ts
charlie-tf: Not sure I understand what you're asking here. `EcosystemType` is an enumeration (below).

src/components/Species/AddSpeciesModal.tsx
charlie-tf: I can do that (I also see that I forgot to check whether index === -1)

src/components/Species/AddSpeciesModal.tsx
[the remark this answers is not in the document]
karthikbtf: Fair enough. Fwiw you had introduced an html parser package.. perhaps an option for cleanup

src/types/Species.ts
karthikbtf: enum EcosystemType {
  BOREAL = ‘Boreal forests/Taiga’,
 …
}
etc but you are right this is overkill.
Objective was to help catch incorrect strings..
Another option I can think of is inferring this type from the generated schema.
type EcosystemType = paths[‘components’].. 
or at a minimum add return type to ecosystemTypes() to be Record<string, EcosystemType> so compiler can catch string errors. Maybe that’s the best..

karthikbtf: Oops return type should be
 { label: string; value: EcosystemType }[]

charlie-tf: I'll explore this a bit. Thanks!

karthikbtf: Thanks, adding a return type to ecosystemTypes() should be enough

charlie-tf: Got it, yeah that does make things safer 👍

karthikbtf (approved):
thanks for the changes

PR #1122, opened 2023-01-31
SW-2727 Show total accession withdrawal quantities

sgrimm:
Remove the per-withdrawal columns from the accession table's customization dialog,
and replace them with a set of columns to show the total withdrawal count/weight
for the accession. This will reduce the confusion caused by the fact that the
per-withdrawal fields returned one table row for each withdrawal, not one per
accession.

We can still get multiple rows for a single accession with the viability test
fields, but users want to be able to see individul viability tests. So add a
tooltip to warn them that they might get multiple rows per accession.


sgrimm:
This depends on https://github.com/terraware/terraware-server/pull/931 which adds the new search columns.

constanzauanini (approved):
lgtm! Thanks for adding the fe changes too!

sgrimm:
Rerunning CI now that the server side of this is live.

PR #1123, opened 2023-01-31
SW-2809 Create i18n service to get timezones

constanzauanini:
- Moved timezones api code into i18n service


src/services/InternationalizationService.ts
```diff
@@ -0,0 +1,37 @@
...
+    timeZones: data?.timeZones ?? [],
+  }));
+
+  return response;
+};
+
+/**
+ * Exported functions
+ */
+const InternationalizationService = {
```
karthikbtf, line 33: I18nService?

karthikbtf (approved):
Thanks

src/services/InternationalizationService.ts
constanzauanini: Ok! I wasn't sure about the "I" in I18n. But will change it

karthikbtf: Oh.. the name came like this:
Internationalization has 18 letters between first letter "i" and last letter "n".. => i18n

constanzauanini: Yes, I know. But since I always see it like "i18n" with the "i" in lowercase, I wasn't sure about using "I18n" with an uppercase "I"

karthikbtf (approved):

PR #1124, opened 2023-01-31
SW-2851 Allow setting accession QTY to 0 manually

constanzauanini:


src/components/accession2/edit/QuantityModal.tsx
```diff
@@ -172,7 +172,9 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
               onChange={(value) => onChangeRemainingQuantity('seedsQuantity', Number(value))}
               type='text'
               value={
-                record.remainingQuantity?.units === 'Seeds' ? record.remainingQuantity?.quantity : record.estimatedCount
+                record.remainingQuantity?.units === 'Seeds'
+                  ? record.remainingQuantity?.quantity.toString()
+                  : record.estimatedCount
```
karthikbtf, line 177: why toString for quantity but not estimatedCount? they are both 'numbers'

src/components/accession2/edit/QuantityModal.tsx
```diff
@@ -48,7 +48,7 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
 
   const validate = () => {
     const quantity = parseFloat(record.remainingQuantity?.quantity as unknown as string);
-    if (isNaN(quantity) || quantity <= 0) {
+    if (isNaN(quantity) || quantity < 0) {
```
karthikbtf, line 51: good catch

karthikbtf (approved):
one comment if you can follow up

src/components/accession2/edit/QuantityModal.tsx
constanzauanini: True, I was thinking in the input value (that is remaining quantity), but estimated value can also be 0

karthikbtf: Oops, I meant why the toString conversion? Can we use them as numbers?

constanzauanini: The problem is that the the Textfield component has an optional "value", and 0 as number is consider false, so it doesn't render. That's why I'm converting the 0 to "0", so that the textfield shows the value

karthikbtf: Yes, I remember - in web-components.. we should probably fix that at some point for input type='number'?
And maybe use number as type in this text field.

karthikbtf (approved):

PR #1126, opened 2023-02-01
SW-2825 Allow viability % to be inputted as 0

constanzauanini:


src/components/accession2/view/Accession2View.tsx
```diff
@@ -452,7 +452,7 @@ export default function Accession2View(): JSX.Element {
               reload={reloadData}
               setNewViabilityTestOpened={setOpenNewViabilityTest}
               changeTab={handleChange}
-              title={accession?.viabilityPercent ? strings.EDIT_VIABILITY : strings.ADD_VIABILITY}
+              title={accession?.viabilityPercent?.toString() ? strings.EDIT_VIABILITY : strings.ADD_VIABILITY}
```
karthikbtf, line 455: Suggestion: use IsNaN instead of toString? And use numbers instead of toString conversions..?

constanzauanini: Yes, I can use NaN in some places, but not in the value of the Textfield. I need to convert it to "0" so that the textfield shows the value

karthikbtf: Got it, we should update textfield component to do the right thing. Can we create a ticket for that - the validation should distinguish between text and number? Also we should be using type=number in these text fields (i can't remember why we aren't..).
Looks good for now

karthikbtf (approved):

src/components/accession2/view/Accession2View.tsx
constanzauanini: Yes, I will create the ticket

PR #1127, opened 2023-02-01
SW-2778 Reset searches when locale changes

sgrimm:
The People page and the Accessions page include localizable search results.
When the user changes locales, we need to show the results in the new locale.

In the case of the Accessions page, we also need to reset the search criteria
because they might include localized values in the old locale, which the
search API won't recognize in the new locale. The People page doesn't allow
filtering on the localizable column (role) so doesn't need this logic.

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -6,6 +6,7 @@ import { Link, useHistory } from 'react-router-dom';
 import useQuery from '../../../utils/useQuery';
 import {
   AllFieldValuesMap,
+  DEFAULT_SEED_SEARCH_FILTERS,
   FieldValuesMap,
   filterSelectFields,
   getAllFieldValues,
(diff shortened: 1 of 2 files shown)


src/components/seeds/database/index.tsx
```diff
@@ -375,18 +378,33 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
+        if (searchCriteria) {
+          populateSearchResults();
+        }
+        populateAvailableFieldOptions();
+
+        populateFieldOptions();
+      } else {
+        searchedLocaleRef.current = null;
+        setSearchCriteria({ ...DEFAULT_SEED_SEARCH_FILTERS });
+      }
```
sgrimm, line 401: This feels a little inelegant to me, but I couldn't come up with a nicer-looking way to do it that didn't end up causing duplicate API requests on locale change.

constanzauanini: This lookg good to me, but maybe @karthikbtf or @charlie-tf have other suggestion

constanzauanini (approved):

src/components/seeds/database/index.tsx
```diff
@@ -375,18 +378,33 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
+      if (searchedLocaleRef.current === loadedStringsForLocale) {
+        if (searchCriteria) {
+          populateSearchResults();
+        }
+        populateAvailableFieldOptions();
+
+        populateFieldOptions();
+      } else {
+        searchedLocaleRef.current = null;
+        setSearchCriteria({ ...DEFAULT_SEED_SEARCH_FILTERS });
```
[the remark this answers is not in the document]
sgrimm: Good call. I'd tried that earlier and had trouble with it, but that was before I realized I needed to use `{...DEFAULT_SEED_SEARCH_FILTERS}` instead of `DEFAULT_SEED_SEARCH_FILTERS ` to force the refresh if the user was looking at the default view. Definitely cleaner this way.

karthikbtf (approved):
Thank you!!

PR #1128, opened 2023-02-01
SW-2822 Save user inputted values for all countries outside the us

constanzauanini:
Show dropdown for state only when country has subdivisions, and show textfield for other countries.
Show value on detail page.

Screenshot 2023-02-01 at 16 25 07

Screenshot 2023-02-01 at 16 22 22

### DIFF
--- a/src/components/accession2/view/DetailPanel.tsx
+++ b/src/components/accession2/view/DetailPanel.tsx
@@ -99,7 +99,8 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
     const subdivision = accession?.collectionSiteCountrySubdivision;
     const data = [
       city,
-      countries && country && subdivision ? getSubdivisionByCode(countries, country, subdivision)?.name : '',
+      (countries && country && subdivision && getSubdivisionByCode(countries, country, subdivision)?.name) ??
+        subdivision,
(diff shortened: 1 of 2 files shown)


src/components/accession2/view/DetailPanel.tsx
```diff
@@ -99,7 +99,8 @@ export default function DetailPanel(props: DetailPanelProps): JSX.Element {
     const subdivision = accession?.collectionSiteCountrySubdivision;
     const data = [
       city,
-      countries && country && subdivision ? getSubdivisionByCode(countries, country, subdivision)?.name : '',
+      (countries && country && subdivision && getSubdivisionByCode(countries, country, subdivision)?.name) ??
+        subdivision,
```
karthikbtf, line 103: do we need another `?? ''` after subdivision on line 103 (in case we already have accessions with an undefined subdivision)

karthikbtf (approved):
one suggestion, feel free to merge otherwise

src/components/accession2/view/DetailPanel.tsx
constanzauanini: Not necessary, just tested and if subdivision is undefined, the location shows the country only

PR #1130, opened 2023-02-02
Improve accession weight quantity styles

constanzauanini:
According to Ming's feedback:

- Add parenthesis for converted weight
- Remove info icon because user is already informed in the quantity form
- Increase vertical spacing to separate weight and count display

Screenshot 2023-02-02 at 12 03 53


charlie-tf (approved):
lgtm!

charlie-tf:
Looks nice in Vercel 👍

PR #1131, opened 2023-02-02
SW-2806 Localize notification timestamps

sgrimm:
Currently, the timestamps on notifications are rendered using MomentJS with a
fixed format string. Replace that with a component that renders a timestamp
using the browser's built-in `Intl.DateTimeFormat` class. Since it's not possible
to add custom locales to `Intl`, the component has some special-case code for
the gibberish locale that can be used to test that dates are in fact localized.

Timestamps are rendered in the user's time zone, if one is set.

The default US English date format is similar to the one we were using before,
but not identical.

- Old: `January 21st 2023 at 12:14:54 am`
- New: `January 21, 2023 at 12:14:54 AM`

### DIFF
--- a/src/components/common/Timestamp.tsx
+++ b/src/components/common/Timestamp.tsx
@@ -8,29 +8,28 @@ export interface TimestampProps {
   isoString: string;
 }
 
-/**
- * Renders a timestamp using the current locale and the user's time zone.
- */
+/** Returns a DateTimeFormat object for a locale and time zone. */
+function newDateTimeFormat(locale?: string, timeZone?: string) {
+  return new Intl.DateTimeFormat(locale, {
+    dateStyle: 'long',
+    timeStyle: 'medium',
+    timeZone,
+    hourCycle: locale === 'gx' ? 'h24' : undefined,
+  });
+}
+
+/** Renders a timestamp using the current locale and the user's time zone. */
 export default function Timestamp({ className, isoString }: TimestampProps): JSX.Element | null {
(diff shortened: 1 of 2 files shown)


sgrimm:
Adding a new component for this seemed like the best way to encapsulate it, but maybe it's overkill?

karthikbtf:
> Adding a new component for this seemed like the best way to encapsulate it, but maybe it's overkill?

New component sounds great here!

src/components/common/Timestamp.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const format = () => {
+    const formatted = dateTimeFormat?.format(new Date(isoString));
+    if (locale === 'gx') {
+      return formatted?.replace(/[A-Za-z]/g, 'XY');
+    } else {
+      return formatted;
+    }
+  };
+
+  return <span className={className}>{format()}</span>;
```
karthikbtf, line 41: should we default to some internal representation if formatted is undefined? This is pretty much unlikely but theoretically possible.

karthikbtf (approved):
Looks great! One question but maybe not an issue.

src/components/common/Timestamp.tsx
sgrimm: Updated this so there will always be a formatter even if the locale is undefined.

karthikbtf: Thanks!

karthikbtf (approved):

PR #1132, opened 2023-02-02
SW-2811 seed bank service + refactor

karthikbtf:
- moved facility types to src/types
- NOTE most of the changes here are just changing the import directory from moving facility types
- removed storage locations and summary api, moved them under seedbank service
- helped catch 2 redundant storage locations api  - updates both uses-cases to use service api
- this service will now host the upcoming customizable storage locations api


src/components/accession2/properties/SeedBank2Selector.tsx
```diff
@@ -34,11 +34,13 @@ export default function SeedBank2Selector(props: SeedBank2SelectorProps): JSX.El
   useEffect(() => {
     const setLocation = async () => {
       if (record.facilityId) {
-        const response = await listStorageLocations(record.facilityId);
```
constanzauanini, line 37: Thanks for fixing this!

constanzauanini (approved):

PR #1133, opened 2023-02-02
SW-2869 Add component for text with embedded link

sgrimm:
To make it easier to translate strings with embedded links, add a new component
that replaces delimiter characters with the HTML link tags. This will allow
translators to use the grammatically correct text for the link based on its
surrounding context.

This initial implementation is pretty limited; it only supports a single link
per string.

We no longer need the `html-react-parser` library, which was only being used
to construct an embedded link.

### DIFF
--- a/src/components/common/TextWithLink.tsx
+++ b/src/components/common/TextWithLink.tsx
@@ -10,14 +10,26 @@ export interface TextWithLinkProps {
  *
  * @param className
  * @param text The text to render. The link should be surrounded by square brackets. For
- * example, `'See [Jane] run'` would be rendered as `See <a href=...>Jane</a> run`.
+ * example, `'See [Jane] run'` would be rendered as `See <a href=...>Jane</a> run`. If there are
+ * no square brackets, [text] is rendered as a link. For example, `'See Jane run'` would be rendered
+ * as, `<a href=...>See Jane run</a>`.
  */
 export default function TextWithLink({ className, href, onClick, text }: TextWithLinkProps): JSX.Element {
   const linkStart = text.indexOf('[');
   const linkEnd = text.indexOf(']');
-  const prefix = text.substring(0, linkStart);
-  const linkText = text.substring(linkStart + 1, linkEnd);
-  const suffix = text.substring(linkEnd + 1);
+  let prefix: string;
+  let linkText: string;
+  let suffix: string;
+
+  if (linkStart >= 0 && linkEnd >= 0) {
+    prefix = text.substring(0, linkStart);
+    linkText = text.substring(linkStart + 1, linkEnd);
+    suffix = text.substring(linkEnd + 1);
+  } else {
+    prefix = '';
+    linkText = text;
+    suffix = '';
+  }
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


src/strings/strings-en.ts
```diff
@@ -622,7 +622,7 @@ export const strings = {
   PLANT: 'plant',
   PLANTING_SITE_TYPE: 'Planting Site Type',
   PLANTING_SITE_WITH_MAP_HELP:
-    'If you are a Terraformation partner or accelerator participant, please {0} to create a site with a map.',
+    'If you are a Terraformation partner or accelerator participant, please [contact us] to create a site with a map.',
```
karthikbtf, line 625: if we need to support valid square brackets in such a string, guessing we'll go with `\[` and `\]` if/when we need to..?

src/components/common/TextWithLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * @param text The text to render. The link should be surrounded by square brackets. For
+ * example, `'See [Jane] run'` would be rendered as `See <a href=...>Jane</a> run`.
+ */
+export default function TextWithLink({ className, href, onClick, text }: TextWithLinkProps): JSX.Element {
+  const linkStart = text.indexOf('[');
+  const linkEnd = text.indexOf(']');
+  const prefix = text.substring(0, linkStart);
+  const linkText = text.substring(linkStart + 1, linkEnd);
+  const suffix = text.substring(linkEnd + 1);
+
```
karthikbtf, line 21: if there's no linkText, return input text as-is?
guessing visually the current code won't show any link if there's no link text .. so maybe it's ok..

karthikbtf (approved):
Looks good to me, thanks 🙏

src/strings/strings-en.ts
sgrimm: Yeah, or we'll change the delimiter to something else. I'm not aware of any language that uses those characters in regular text, but you never know.

src/components/common/TextWithLink.tsx
sgrimm: Good catch. I think in that case it makes more sense to turn the whole string into a link. Otherwise a bad translation could make it impossible to follow the link. Updated the PR to do that.

karthikbtf: That's a good idea!! This is definitely better.

src/components/common/TextWithLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * as, `<a href=...>See Jane run</a>`.
+ */
+export default function TextWithLink({ className, href, onClick, text }: TextWithLinkProps): JSX.Element {
+  const linkStart = text.indexOf('[');
+  const linkEnd = text.indexOf(']');
+  let prefix: string;
+  let linkText: string;
+  let suffix: string;
+
+  if (linkStart >= 0 && linkEnd >= 0) {
```
karthikbtf, line 24: add one more check for linkEnd > linkStart to handle "lorem]ipsum[dolor.."

karthikbtf (approved):
lgtm,, one suggestion

karthikbtf (approved):
Thanks!!

charlie-tf (approved):
This is pretty nice!

PR #1135, opened 2023-02-03
SW-2858 Right-align non-English monitoring chart labels

sgrimm:
The chart rendering code was checking for English dataset names to determine
whether or not the labels in the legend should be right-aligned.


karthikbtf (approved):

constanzauanini (approved):
Thanks!

PR #1136, opened 2023-02-03
SW-2872 avoid redundant preference write when switching orgs

karthikbtf:
- the setSelectedOrg within callback function was being executed twice every time, this is absurd
- figured we don't need this to be in a callback, moved it out and we don't have the duplicate writes

### DIFF
--- a/src/providers/OrganizationProvider.tsx
+++ b/src/providers/OrganizationProvider.tsx
@@ -108,11 +108,11 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
       const organizationId = query.get('organizationId');
       const querySelectionOrg = organizationId && organizations.find((org) => org.id === parseInt(organizationId, 10));
       let orgToUse = querySelectionOrg || organizations.find((org) => org.id === selectedOrganization?.id);
-      if (!orgToUse) {
-        orgToUse = organizations[0];
-      }
       if (!orgToUse && userPreferences.lastVisitedOrg) {
         orgToUse = organizations.find((org) => org.id === userPreferences.lastVisitedOrg);
+      }
+      if (!orgToUse) {
+        orgToUse = organizations[0];
       }
       if (orgToUse) {
         setSelectedOrganization(orgToUse);
--- a/src/providers/OrganizationProvider.tsx
+++ b/src/providers/OrganizationProvider.tsx
@@ -116,9 +116,7 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
       }
       if (orgToUse) {
         setSelectedOrganization(orgToUse);
-        if (userPreferences?.lastVisitedOrg !== orgToUse.id) {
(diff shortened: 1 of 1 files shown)


karthikbtf:
This was failing server-side but that has been fixed and upgrade in prod recently.

constanzauanini (approved):

karthikbtf:
@constanzauanini I will hotfix this in production.. Becky is saying that the web-app is switching user to another org while the app is already running. strange.. will try to repro that problem locally and see if i can fix that too..  if you have any idea lmk.

constanzauanini:
> @constanzauanini I will hotfix this in production.. Becky is saying that the web-app is switching user to another org while the app is already running. strange.. will try to repro that problem locally and see if i can fix that too.. if you have any idea lmk.

But isn't the server-side fixing that already? I will try to reproduce it, but never happened that to me

karthikbtf:
> > @constanzauanini I will hotfix this in production.. Becky is saying that the web-app is switching user to another org while the app is already running. strange.. will try to repro that problem locally and see if i can fix that too.. if you have any idea lmk.
> 
> But isn't the server-side fixing that already? I will try to reproduce it, but never happened that to me

The server fix is not directly related to the auto switching, that was just another issue I observed.

karthikbtf:
Found a case where if user switches from org A to B and back to A - the last preference only stores B, we don't update the preference back to A - since user preference's last org matches A (from app load).
Fixing that.
Also - user preferences are cached in a service, I will refactor code to use the cached preferences, as a separate effort. That will clean all this up.

src/providers/OrganizationProvider.tsx
```diff
@@ -107,19 +107,17 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
...
       const querySelectionOrg = organizationId && organizations.find((org) => org.id === parseInt(organizationId, 10));
-      setSelectedOrganization((previouslySelectedOrg: Organization | undefined) => {
-        let orgToUse = querySelectionOrg || organizations.find((org) => org.id === previouslySelectedOrg?.id);
-        if (!orgToUse && userPreferences.lastVisitedOrg) {
-          orgToUse = organizations.find((org) => org.id === userPreferences.lastVisitedOrg);
-        }
-        if (!orgToUse) {
-          orgToUse = organizations[0];
-        }
-        if (orgToUse && userPreferences?.lastVisitedOrg !== orgToUse.id) {
```
karthikbtf, line 118: @constanzauanini i removed the check here and we always set the preference now, this was a bug - either that or we reload user preferences after every update. either way, this code is getting messy, i will clean it up to use cached preferences later. for now removing this check fixes a bug.. will get this.

constanzauanini: Ok, sounds good. Does setting the preference always triggers a lot of data fetches also?

karthikbtf: Not here because we aren't refreshing the preferences.

PR #1137, opened 2023-02-03
SW-2866 Use popover component in locale selector

constanzauanini:
Screenshot 2023-02-03 at 10 47 24
Screenshot 2023-02-03 at 10 47 30

### DIFF
--- a/src/components/LocaleSelector.tsx
+++ b/src/components/LocaleSelector.tsx
@@ -1,9 +1,33 @@
-import { DropdownItem } from '@terraware/web-components';
+import { DropdownItem, Icon } from '@terraware/web-components';
 import { supportedLocales } from '../strings/locales';
 import { LocalizationContext } from '../providers/contexts';
-import Dropdown from './common/Dropdown';
 import { UserService } from 'src/services';
 import { useUser } from '../providers';
+import { makeStyles } from '@mui/styles';
+import { IconButton, Theme } from '@mui/material';
+import { useState } from 'react';
+import PopoverMenu from './common/PopoverMenu';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  iconContainer: {
+    height: '48px',
+    borderRadius: '16px',
+    padding: theme.spacing(1.5, 2),
+  },
+  icon: {
+    width: '32px',
+    height: '32px',
+  },
+  chevronDown: {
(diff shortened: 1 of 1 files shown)


src/components/LocaleSelector.tsx
```diff
@@ -1,21 +1,55 @@
...
+    paddingLeft: '8px',
+    color: theme.palette.TwClrTxt,
+  },
+}));
 
 export default function LocaleSelector(): JSX.Element {
   const { user, reloadUser } = useUser();
   const localeItems: DropdownItem[] = supportedLocales.map((supportedLocale) => ({
     value: supportedLocale.id,
     label: supportedLocale.name,
```
karthikbtf, line 36: unrelated to your PR but this should probably come from `strings`

karthikbtf (approved):
looks good in vercel

src/components/LocaleSelector.tsx
karthikbtf: actually no, i am wrong here.. i guess each language should show up in it's respective recognizable script. Nothing to do here, but maybe gibberish should be in gibberish. When we show spanish menu item maybe it will show `española` if that's the correct spelling :)
ignore my comments in this thread

src/components/LocaleSelector.tsx
```diff
@@ -28,9 +62,23 @@ export default function LocaleSelector(): JSX.Element {
...
+              <span className={classes.selected}>{localeItems.find((iLocale) => iLocale.value === locale)?.label}</span>
+              <Icon name='chevronDown' size='medium' className={classes.chevronDown} />
+            </IconButton>
+            <PopoverMenu
+              sections={[localeItems]}
+              handleClick={onChange}
+              anchorElement={anchorEl}
+              setAnchorElement={setAnchorEl}
+            />
+          </div>
```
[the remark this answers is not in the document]
constanzauanini: I will create a ticket to move the PopoverMenu to web-components, and when moving we should include the chevron and the label in same component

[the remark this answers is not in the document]
karthikbtf: Thanks.

src/components/LocaleSelector.tsx
constanzauanini: It would be `español` haha

PR #1139, opened 2023-02-03
SW-2823 Add option “none” to substrate viability fields

constanzauanini:
Screenshot 2023-02-03 at 14 58 10


sgrimm:
Should we alphabetize these (using the sorting rules for the current locale, of course)?

constanzauanini:
> Should we alphabetize these (using the sorting rules for the current locale, of course)?

Not sure. I think other and none should be always the last options, but maybe alphabetize the others? We can do that separately if we need it

PR #1140, opened 2023-02-03
SW-2871 Fix race condition that allowed org selection to occur before…

karthikbtf:
… user preferences were loaded

- user is considered bootstrapped after preferences are fetched (regardless of success/failure in fetch)
- org selection happens only after the user preferences are loaded
- orgs data loading happens concurrently as usual
- moved the setting of last visited org out into a separate useEffect, cleaner this way

### DIFF
--- a/src/providers/OrganizationProvider.tsx
+++ b/src/providers/OrganizationProvider.tsx
@@ -31,7 +31,7 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
   const history = useHistory();
   const query = useQuery();
   const location = useStateLocation();
-  const { userPreferences } = useUser();
+  const { userPreferences, updateUserPreferences, bootstrapped: userBootstrapped } = useUser();
 
   const reloadOrganizations = useCallback(async (selectedOrgId?: number) => {
     const populateOrganizations = async () => {
@@ -43,7 +43,6 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
           const orgToSelect = response.organizations.find((org) => org.id === selectedOrgId);
           if (orgToSelect) {
             setSelectedOrganization(orgToSelect);
-            PreferencesService.updateUserPreferences({ lastVisitedOrg: orgToSelect.id });
           }
         }
         if (response.organizations.length === 0) {
@@ -56,6 +55,7 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
         setOrgAPIRequestStatus(APIRequestStatus.FAILED);
       }
     };
+
     await populateOrganizations();
   }, []);
 
@@ -96,6 +96,7 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
         setBootstrapped(true);
       }
     };
+
     getOrgPreferences();
   }, [selectedOrganization]);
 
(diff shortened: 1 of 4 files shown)


src/providers/OrganizationProvider.tsx
```diff
@@ -104,7 +105,7 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
   }, [reloadOrgPreferences, selectedOrganization]);
 
   useEffect(() => {
-    if (organizations.length && userPreferences) {
+    if (userBootstrapped && organizations.length && userPreferences) {
```
karthikbtf, line 108: select the org only after user is bootstrapped - so we have the preferences data. this was the race condition - preferences being available before getting to this code.. which doesn't always happen, esp in prod

src/providers/OrganizationProvider.tsx
```diff
@@ -116,15 +117,20 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
...
       }
     }
-  }, [organizations, selectedOrganization, query, location, history, userPreferences]);
+  }, [organizations, selectedOrganization, query, location, history, userPreferences, userBootstrapped]);
+
+  useEffect(() => {
+    if (selectedOrganization?.id && userPreferences.lastVisitedOrg !== selectedOrganization.id) {
+      updateUserPreferences({ lastVisitedOrg: selectedOrganization.id });
+    }
+  }, [selectedOrganization?.id, updateUserPreferences, userPreferences.lastVisitedOrg]);
```
karthikbtf, line 133: separate useEffect to keep track of last selected org in user preference. this is cleaner and avoids redundant updates

src/providers/UserProvider.tsx
```diff
@@ -11,37 +12,34 @@ export type UserProviderProps = {
...
+  const [userPreferences, setUserPreferences] = useState<PreferencesType>();
+
+  const updateUserPreferences = useCallback(async (preferences: PreferencesType) => {
+    const response = await PreferencesService.updateUserPreferences(preferences);
+    if (response.requestSucceeded) {
+      setUserPreferences(response.preferences ?? {});
+    }
+
+    return Promise.resolve(response.requestSucceeded);
+  }, []);
```
karthikbtf, line 26: exposing this so user preference can be updated.. and patched back into the provided data for downstream consumption. currently only used by the organization provider but will expand the usage across code after this fix is deployed

src/providers/UserProvider.tsx
```diff
@@ -51,26 +49,37 @@ export default function UserProvider({ children }: UserProviderProps): JSX.Eleme
...
+
+  useEffect(() => {
+    setUserData((prev) => ({
+      ...prev,
+      user,
+      userPreferences: userPreferences ?? {},
+      reloadUserPreferences,
+      bootstrapped: Boolean(userPreferences && user),
+    }));
+  }, [user, userPreferences, reloadUserPreferences]);
```
karthikbtf, line 82: moved all useEffects towards the bottom, the primary change is line 80 here

charlie-tf (commented):
Looks good. Thanks for the helpful explanations.

charlie-tf (approved):

karthikbtf:
> Looks good. Thanks for the helpful explanations.

Thanks @charlie-tf

karthikbtf:
this is already merged as part of hotfix workflow

PR #1141, opened 2023-02-04
SW-2873 Localize axis labels on charts

sgrimm:
Update the configuration of the app's charts to use localized labels. This
involved a few interrelated changes.

* Wrap the Chart.js constructor in a helper function that dynamically imports the
  `date-fns` module for the user's locale if there is a time axis on the chart.
  For Gibberish, render dates in Korean so they're visually distinct from English;
  it's a nontrivial project to add a custom locale to `date-fns` and all we need
  is a way to tell whether or not the labels are hardwired English.

  Since module loading is asynchronous, this requires making the constructor
  wrapper async as well.

  The wrapper also sets the top-level `locale` option on the chart as a whole,
  which will cause Chart.js to use locale-specific number formatting.

* Extract the suffixes for temperature and voltage into the strings table.

* Use a locale-aware date format string for the monitoring X axis labels. The
  previous format string was incorrect even in English since it didn't distinguish
  AM and PM times.

* Revamp the viability test observations chart. The current version in production
  has an invalid configuration (and it logs errors to that effect to the console)
  which was impossible to tweak to use localized dates. Most likely it was written
  for an older Chart.js version and not revised when we upgraded.


src/components/accession2/viabilityTesting/ObservationsChart.tsx
```diff
@@ -16,73 +19,103 @@ export interface Props {
...
+  useEffect(() => {
+    const createChart = async () => {
+      const ctx = canvasRef?.current?.getContext('2d');
+      if (ctx && loadedStringsForLocale) {
+        if (chartRef.current) {
+          chartRef.current?.destroy();
+          chartRef.current = null;
+        }
+
+        const localTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
```
constanzauanini, line 36: Should we use the seedbank time zone here?

src/components/Monitoring/dashboard/PVBatteryChart.tsx
```diff
@@ -147,7 +148,7 @@ export default function PVBatteryChart(props: PVBatteryChartProps): JSX.Element
                 time: {
                   unit: getUnit(selectedPVBatteryPeriod),
                   displayFormats: {
-                    hour: 'MMM d h:mm',
+                    hour: 'PPp',
```
constanzauanini, line 151: With this format we are including the year that I think we didn't want to include it (because it takes more space)

constanzauanini: Also, when changing to gibberish, to the format that include hours doesn't look like japanese

src/components/accession2/viabilityTesting/ObservationsChart.tsx
sgrimm: Good point. This should indeed use the seed bank's time zone.

sgrimm: Oh, actually, this is fine as-is, but I will add a comment to describe what it's doing since it's not very obvious (it took me a while to figure out this was needed).

constanzauanini: Excellent. Understood now. Thanks for the explanation!

constanzauanini (approved):

PR #1146, opened 2023-02-06
SW-2865 Move timezone and units init message under notifications popup

constanzauanini:
Screenshot 2023-02-06 at 17 02 48


src/strings/strings-en.ts
```diff
@@ -1027,6 +1025,7 @@ export const strings = {
   UPLOAD_PHOTO: 'Upload Photo...',
   UPLOAD_PHOTOS: 'Upload Photo(s)',
   USED_UP: 'Used Up',
+  USER_NOTIFICATION_ACTION: 'To change the preference, go to “Settings > Organization”',
```
karthikbtf, line 1028: can we make the last part an href?

src/strings/strings-en.ts
```diff
@@ -579,6 +580,7 @@ export const strings = {
   OPT_IN: 'Opt-In Features',
   OR_SEED_WEIGHT: 'or Seed Weight',
   OR: 'OR',
+  ORG_NOTIFICATION_ACTION: 'To change your preferences, go to “My Account”',
```
karthikbtf, line 583: make the last part an href?

src/components/NotificationsDropdown.tsx
```diff
@@ -164,8 +269,73 @@ export default function NotificationsDropdown(props: NotificationsDropdownProps)
       const dateB = new Date(b.createdTime);
       return dateB.getTime() - dateA.getTime();
     });
+
+    if (timeZoneOrgNotification) {
```
karthikbtf, line 273: we need to think about how to move this outside of this file.. since the number of features with this type of notification will increase over time.. maybe need to have some sort of 'FeatureNotifications' component which is embedded into this component?

src/components/NotificationsDropdown.tsx
```diff
@@ -217,14 +387,41 @@ export default function NotificationsDropdown(props: NotificationsDropdownProps)
...
-    }
-    await MarkNotificationRead(read, id);
+  const markUserNotificationAsRead = async () => {
+    await PreferencesService.updateUserPreferences({
+      unitsAcknowledgedOnMs: Date.now(),
+      timeZoneAcknowledgedOnMs: Date.now(),
+    });
+
+    await reloadPreferences();
+    onPopoverClose();
```
karthikbtf, line 397: why are we closing here? we should move this to the top maybe so user is not waiting for the api to finish

karthikbtf (commented):
some comments

src/components/NotificationsDropdown.tsx
karthikbtf: may need to check with Ming on where this notification should disappear or always be present

karthikbtf (approved):
Feel free to merge if we want to get this in and then iterate.. it will be nice to think of a more scalable design since # of features will increase over time.

src/strings/strings-en.ts
constanzauanini: Any part of the notification is clickable and takes to organization page, but I could make it and href for the styles

src/components/NotificationsDropdown.tsx
constanzauanini: Yes, I thought about it, this is not scalable for the future. I can iterate after

karthikbtf: Just so we don't lose track, can we create a ticket for this. I think we may have missed tickets for previous PRs as well (componentize popover menu, etc.)

karthikbtf (approved):
Some suggestions... but feel free to merge

src/components/NotificationsDropdown.tsx
constanzauanini: Yes! We have that previous ticket created also

karthikbtf: Awesome, thanks @constanzauanini

karthikbtf (approved):
🎉

PR #1149, opened 2023-02-07
SW-2897 Pass locale to cookie consent UI

sgrimm:
We want the cookie consent UI to show in Spanish for Spanish-speaking users. It
gets its initial locale from the `lang` attribute of the `<html>` tag, so update
the code to set that attribute before initializing Google Analytics (which is
what triggers loading the cookie consent UI).

### DIFF
--- a/src/providers/UserProvider.tsx
+++ b/src/providers/UserProvider.tsx
@@ -42,6 +42,16 @@ export default function UserProvider({ children }: UserProviderProps): JSX.Eleme
         setUser(response.user!);
         if (response.user && !userState?.gtmInstrumented && (window as any).INIT_GTAG) {
           setUserState({ gtmInstrumented: true });
+
+          // Put the language in the "lang" attribute of the <html> tag before initializing Google
+          // Analytics because the cookie consent UI code will look there to determine which
+          // language to use for the consent UI. This needs to happen before the call to INIT_GTAG,
+          // so it lives here instead of in LocalizationProvider.
+          const locale = response.user.locale;
+          if (locale) {
+            document.documentElement.setAttribute('lang', locale);
+          }
+
           (window as any).INIT_GTAG(
             response.user.id.toString(),
             response.user.email?.toLowerCase()?.endsWith('@terraformation.com') ? 'true' : 'false'
--- a/public/index.html
+++ b/public/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html>
   <head>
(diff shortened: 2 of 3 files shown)


sgrimm:
I created a `gx` custom locale on cookie-scripts.com so we can test with gibberish.

sgrimm:
You may need to do a hard reload or clear your browser cache to see the updated cookie-script code with the translations.

src/providers/UserProvider.tsx
```diff
@@ -42,6 +42,16 @@ export default function UserProvider({ children }: UserProviderProps): JSX.Eleme
...
           setUserState({ gtmInstrumented: true });
+
+          // Put the language in the "lang" attribute of the <html> tag before initializing Google
+          // Analytics because the cookie consent UI code will look there to determine which
+          // language to use for the consent UI. This needs to happen before the call to INIT_GTAG,
+          // so it lives here instead of in LocalizationProvider.
+          const locale = response.user.locale;
+          if (locale) {
+            document.documentElement.setAttribute('lang', locale);
+          }
```
karthikbtf, line 53: else `en`?

karthikbtf (approved):
lgtm, you may need to uncomment out the dev fragment in index.html for us to test in vercel.. (and re-comment prior to checking in)... otherwise np we  can wait until it is in staging.. code looks good to me

karthikbtf:
> lgtm, you may need to uncomment out the dev fragment in index.html for us to test in vercel.. (and re-comment prior to checking in)... otherwise np we can wait until it is in staging.. code looks good to me

ooh.. nvm, i can check out the branch locally to test

karthikbtf:
> > lgtm, you may need to uncomment out the dev fragment in index.html for us to test in vercel.. (and re-comment prior to checking in)... otherwise np we can wait until it is in staging.. code looks good to me
> 
> ooh.. nvm, i can check out the branch locally to test

and.. looks great in gibberish

src/providers/UserProvider.tsx
sgrimm: English is the default.

karthikbtf: it is recommended to set the default https://www.w3.org/International/questions/qa-html-language-declarations
.. and i believe if unset the default is unknown
https://www.w3.org/TR/html401/struct/dirlang.html

sgrimm: Good point. I'll send out a followup.

PR #1150, opened 2023-02-07
SW-1828 Ignore diacritics in species typeahead

sgrimm:
Some species common names have accented characters, but we don't want to force
users to type the correct accent marks in the species typeahead when adding
accessions or seedling batches.


constanzauanini (approved):
Looks good!

PR #1152, opened 2023-02-07
Add Searchbar to Accessions Table

charlie-tf:
Add a searchbar to the accession table. This searches over the Accession,
Scientific Name, and Collecting Site columns. The `useDebounce` hook
was modified to accept a callback.

### DIFF
--- a/src/components/seeds/database/Filters.tsx
+++ b/src/components/seeds/database/Filters.tsx
@@ -96,7 +96,7 @@ export default function Filters(props: Props): JSX.Element {
       }
       onChange(newFilters);
     },
-    [searchColumns, filters]
+    [searchColumns, filters, onChange]
   );
   const debouncedSearchTerm = useDebounce(searchTerm, 250, searchTermCallback);
 
--- a/src/components/seeds/database/Filters.tsx
+++ b/src/components/seeds/database/Filters.tsx
@@ -245,15 +245,15 @@ export function getUpdatedFilters(
 }
 
 function getSearchTermFilter(searchCols: DatabaseColumn[], searchTerm: string): Record<string, SearchNodePayload> {
-  const orNode: SearchNodePayload = { children: [], operation: 'or' };
-  for (const col of searchCols) {
-    orNode.children.push({
+  const orNode: SearchNodePayload = {
+    children: searchCols.map((col) => ({
       operation: 'field',
       field: col.key,
       type: 'Fuzzy',
       values: [searchTerm],
-    });
-  }
+    })),
+    operation: 'or',
+  };
 
   return { searchTermFilter: orNode };
 }
--- a/src/components/seeds/database/Filters.tsx
+++ b/src/components/seeds/database/Filters.tsx
@@ -1,7 +1,7 @@
 import { ArrowDropDown } from '@mui/icons-material';
 import { Chip, Container, Divider, Link, Popover, Theme, Typography } from '@mui/material';
 import { makeStyles } from '@mui/styles';
-import React, { useState } from 'react';
+import React from 'react';
 import { FieldNodePayload, FieldValuesPayload, OrNodePayload, SearchNodePayload } from 'src/api/search';
 import strings from 'src/strings';
 import preventDefaultEvent from 'src/utils/preventDefaultEvent';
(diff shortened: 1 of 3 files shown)


src/components/seeds/database/Filters.tsx
```diff
@@ -13,6 +13,8 @@ import Search from './filters/FilterSearch';
 import SingleSelection from './filters/FilterSingleSelection';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
 import { DatabaseColumn, Option } from '@terraware/web-components/components/table/types';
+import TextField from '../../common/Textfield/Textfield';
+import useDebounce from '../../../utils/useDebounce';
```
karthikbtf, line 17: use `src/utils/..` and `src/components/common`... ?

src/components/seeds/database/Filters.tsx
```diff
@@ -56,33 +57,49 @@ const useStyles = makeStyles((theme: Theme) => ({
...
 
 export default function Filters(props: Props): JSX.Element {
+  const { columns, searchColumns, filters, availableValues, allValues, onChange } = props;
   const { isMobile, isDesktop } = useDeviceInfo();
   const classes = useStyles({ isMobile, isDesktop });
   const [popover, setPopover] = React.useState<FilterPopover>();
+  const [searchTerm, setSearchTerm] = React.useState('');
+  const debouncedSearchTerm = useDebounce(searchTerm, 250, (value) => {
+    const updatedSearchFilters = getSearchTermFilter(searchColumns, value);
+    onChange({ ...filters, ...updatedSearchFilters });
```
karthikbtf, line 90: callback is fine.. thoughts on useEffect on the debounced search term, just to keep with the react-ive pattern (vs javascript callback patterns). tbh i don't know if there are pros/cons.. more of an open comment really.
also just curious if the value of this callback changes on each render (since it's a dependency inside useDebounce's useEffect).. if so, we could make it a useCallback or something like that.. mostly curious here...

src/components/seeds/database/Filters.tsx
```diff
@@ -56,33 +57,49 @@ const useStyles = makeStyles((theme: Theme) => ({
...
 }
 
 export default function Filters(props: Props): JSX.Element {
+  const { columns, searchColumns, filters, availableValues, allValues, onChange } = props;
   const { isMobile, isDesktop } = useDeviceInfo();
   const classes = useStyles({ isMobile, isDesktop });
   const [popover, setPopover] = React.useState<FilterPopover>();
+  const [searchTerm, setSearchTerm] = React.useState('');
+  const debouncedSearchTerm = useDebounce(searchTerm, 250, (value) => {
+    const updatedSearchFilters = getSearchTermFilter(searchColumns, value);
```
karthikbtf, line 89: ignore if value is empty?

src/components/seeds/database/Filters.tsx
```diff
@@ -196,6 +233,20 @@ export function getUpdatedFilters(
   return updatedFilters;
 }
 
+function getSearchTermFilter(searchCols: DatabaseColumn[], searchTerm: string): Record<string, SearchNodePayload> {
+  const orNode: SearchNodePayload = { children: [], operation: 'or' };
```
karthikbtf, line 237: optionally children: searchCols.map(col => ({
 ...
})

karthikbtf (commented):
Looks great, a few comments here and there but nothing major except one about ignoring the updated filters if search term is empty

src/components/seeds/database/Filters.tsx
charlie-tf: I was originally trying to use a `useEffect` with basically this callback as the effect. When I did that, though, I ended up needing `filters` as a dependency of the effect. This caused an infinite recursion as calling `onChange` updated filters, which called effect and so on. This callback was the most straightforward way I could think of out of that situation.

charlie-tf: >curious if the value of this callback changes on each render

I don't think so! So having a useCallback might make sense!

karthikbtf: Got it, no worries.. I guess the callback approach makes sense then.

src/components/seeds/database/Filters.tsx
karthikbtf: i think you may have missed this one.. just a nit

karthikbtf (approved):
🎉  thanks for the changes..  looks good to merge 👍

src/components/seeds/database/Filters.tsx
charlie-tf: oop, yeah sorry

karthikbtf: 🙌

PR #1154, opened 2023-02-07
Create a 'FeatureNotifications' component to use in Notifications component

constanzauanini:
@karthikbtf is something like this the idea you have? 
I still need to figure out how to allow each "feature notification" to have their own "markAsRead" function

### DIFF
--- a/src/components/FeatureNotification/OrganizationNotification.tsx
+++ b/src/components/FeatureNotification/OrganizationNotification.tsx
@@ -1,0 +1,32 @@
+import { Box } from '@mui/material';
+import { APP_PATHS } from 'src/constants';
+import strings from 'src/strings';
+import { Notification } from 'src/types/Notifications';
+import TextWithLink from 'src/components/common/TextWithLink';
+import { useOrganization } from 'src/providers';
+import { getTodaysDateFormatted } from '@terraware/web-components/utils';
+
+export default function OrganizationNotification(orgTimeZone?: string): Notification {
+  const { selectedOrganization } = useOrganization();
+
+  return {
+    id: -2,
+    notificationCriticality: 'Info',
+    organizationId: selectedOrganization.id,
+    title: strings.REVIEW_YOUR_ORGANIZATION_SETTING,
+    body: (
+      <div>
+        <ul>
+          <li>{strings.formatString(strings.TIME_ZONE_SELECTED, orgTimeZone || '')}</li>
+        </ul>
+        <Box paddingTop={1}>
+          <TextWithLink text={strings.ORG_NOTIFICATION_ACTION} href={APP_PATHS.ORGANIZATION} />
+        </Box>
+      </div>
+    ),
+    localUrl: APP_PATHS.ORGANIZATION,
+    createdTime: getTodaysDateFormatted(),
+    isRead: false,
+    hideDate: true,
+  };
+}
--- a/src/components/FeatureNotification/OrganizationNotification.tsx
+++ b/src/components/FeatureNotification/OrganizationNotification.tsx
@@ -39,7 +39,7 @@ export default function OrganizationNotification(): Notification | undefined {
         return;
       }
 
-      const userTz: InitializedTimeZone = await UserService.initializeTimeZone(user, getDefaultTimeZone().id);
+      const userTz: InitializedTimeZone = await UserService.getInitializedTimeZone(user, getDefaultTimeZone().id);
       if (!userTz.timeZone) {
         return;
       }
--- a/src/components/FeatureNotification/OrganizationNotification.tsx
+++ b/src/components/FeatureNotification/OrganizationNotification.tsx
@@ -3,30 +3,91 @@ import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
 import { Notification } from 'src/types/Notifications';
 import TextWithLink from 'src/components/common/TextWithLink';
-import { useOrganization } from 'src/providers';
+import { useOrganization, useTimeZones, useUser } from 'src/providers';
 import { getTodaysDateFormatted } from '@terraware/web-components/utils';
+import { useEffect, useState } from 'react';
+import isEnabled from 'src/features';
+import { InitializedTimeZone, TimeZoneDescription } from 'src/types/TimeZones';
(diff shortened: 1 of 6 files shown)


src/components/FeatureNotification/OrganizationNotification.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { APP_PATHS } from 'src/constants';
+import strings from 'src/strings';
+import { Notification } from 'src/types/Notifications';
+import TextWithLink from 'src/components/common/TextWithLink';
+import { useOrganization } from 'src/providers';
+import { getTodaysDateFormatted } from '@terraware/web-components/utils';
+
+export default function OrganizationNotification(orgTimeZone?: string): Notification {
+  const { selectedOrganization } = useOrganization();
+
```
karthikbtf, line 11: Move the logic for this notification display into this component... if we don't need to show the notification we can return null in this component...
The mark read logic can be inside here as well?

src/components/FeatureNotification/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+import isEnabled from 'src/features';
+import { useOrganization, useTimeZones, useUser } from 'src/providers';
+import { InitializedTimeZone, TimeZoneDescription } from 'src/types/TimeZones';
+import { getTimeZone, getUTC } from 'src/utils/useTimeZoneUtils';
+import { OrganizationService, UserService } from 'src/services';
+import { InitializedUnits } from 'src/units';
+import OrganizationNotification from './OrganizationNotification';
+
+export default function useFeatureNotifications(): Notification[] {
+  const [unitNotification, setUnitNotification] = useState(false);
```
karthikbtf, line 13: this component should ideally only import the various feature notifications and return the list of those feature notifications.
maybe this component can also have a 'markAll' property which it can pass down to feature notificaitions.. if it's true the individual notifications can mark it as read/unread etc..if needed..

just some thoughts..

basically .. having all the feature logic in individual components will be more scalable

karthikbtf (commented):
@constanzauanini this is along the right path.. thanks for giving this a whirl.!!
Some comments...

src/components/FeatureNotification/OrganizationNotification.tsx
```diff
@@ -0,0 +1,93 @@
...
+      setOrgTimeZone(orgTz.timeZone);
+      setTimeZoneOrgNotification(!!notifyOrg);
+    };
+
+    const initializeTimeZones = async () => {
+      if (!user) {
+        return;
+      }
+
+      const userTz: InitializedTimeZone = await UserService.initializeTimeZone(user, getDefaultTimeZone().id);
```
karthikbtf, line 42: we don't want to run this multiple times... i can't find the discussion thread we had over this.

src/components/FeatureNotification/OrganizationNotification.tsx
```diff
@@ -0,0 +1,93 @@
...
+          </Box>
+        </div>
+      ),
+      localUrl: APP_PATHS.ORGANIZATION,
+      createdTime: getTodaysDateFormatted(),
+      isRead: false,
+      hideDate: true,
+      markAsRead: async () => {
+        await PreferencesService.updateUserOrgPreferences(selectedOrganization.id, {
+          timeZoneAcknowledgedOnMs: Date.now(),
```
karthikbtf, line 86: previously this applied to both org and user time zones since it was one message.. now that it is split - we probably want different names, one for org and one for user

karthikbtf (commented):
thanks for doing this..
i'd prefer we don't rush over this.. i need a bit more time to look.. some discussions seem to have been lost

karthikbtf:
Chudo added a comment that these notifications should disappear after 30 days, we need to think about how to accomplish that as well.

src/components/FeatureNotification/OrganizationNotification.tsx
karthikbtf: One way to accomplish this... add a cached promise in UserService. Example:
```
/**
 * initialize user time zone
 */
const initializeTimeZone = async (user: User, timeZone: string): Promise<InitializedTimeZone> => {
  .. the usual code
};

/**
 * return a cached promise for initialized time zone
 * this avoids executing it twice
 * /
let initializedTimeZonePromise = null;
const getInitializedTimeZone = (): Promise<InitializedTimeZone> => {
  if (!initializedTimeZonePromise) {
    initializedTimeZonePromise = initializeTimeZone();
  }
  return initializedTimeZonePromise;
};
```
now the notifications code can call getInitiailizedTimeZone() (which returns a promise to the function that was only executed once).
just a suggestion.. if you have other ideas sure..

src/components/FeatureNotification/OrganizationNotification.tsx
karthikbtf: oops ignore, this one is scoped by org id,, it works right

karthikbtf (approved):
a few suggestions if you can follow up
1. avoid calling initialize time zone multiple times (have a suggestion in there on how to do it)
2. the expire after 30 days,, we can maybe follow up separately after this PR

src/services/UserService.ts
```diff
@@ -100,6 +101,17 @@ const initializeTimeZone = async (user: User, timeZone: string): Promise<Initial
...
+ * return a cached promise for initialized time zone
+ * this avoids executing it twice
+ */
+const getInitializedTimeZone = (user: User, timeZone: string): Promise<InitializedTimeZone> => {
+  if (!initializedTimeZonePromise) {
+    initializedTimeZonePromise = initializeTimeZone(user, timeZone);
+  }
+  return initializedTimeZonePromise;
+};
+
```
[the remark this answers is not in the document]
constanzauanini: This looks good, I was trying to modify the promise with no success. Let me try with this approach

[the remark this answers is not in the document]
constanzauanini: This works! I will update the PR. Will leave the expire after 30days for a following PR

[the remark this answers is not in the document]
karthikbtf: Awesome, thanks so much 🙏

karthikbtf (approved):
🎉

PR #1155, opened 2023-02-07
SW-2906 Improve localization of accession details

sgrimm:
The accession details page had a hardwired English menu item (Delete) and was
only rendering the age at initial component render time, so it wasn't updating
when the user switched locales.

Even in English, the accession age wasn't quite ideal; it showed a value of
"1 Months" instead of "1 Month". Add support for singular month values and move
the concatenation of the number of months and the word "Months" into the strings
table to support languages with different word order.


src/components/accession2/view/Accession2View.tsx
```diff
@@ -612,11 +617,7 @@ export default function Accession2View(): JSX.Element {
           />
         </Grid>
         <Grid item xs={getOverviewGridSize(2)}>
-          <OverviewItemCard
-            isEditable={false}
-            title={strings.AGE}
-            contents={accession?.collectedDate ? `${age.value} ${age.unit}` : null}
-          />
+          <OverviewItemCard isEditable={false} title={strings.AGE} contents={age ? age : null} />
```
karthikbtf, line 620: an alternative to this ternary is `contents={age || null}` .. in case you prefer that

karthikbtf (approved):
looks good to me 👍

PR #1156, opened 2023-02-07
SW-2915 Pass locale to date picker

sgrimm:
The DatePicker component accepts an optional `locale` property, but nothing was
populating it. Pass in the current locale, or Korean if the gibberish locale is
selected (since the underlying MUI calendar component isn't trivial to extend with
custom locales).


src/components/common/DatePicker.tsx
```diff
@@ -1,2 +1,20 @@
...
+  useEffect(() => {
+    if (loadedStringsForLocale) {
+      // Adding a custom gibberish locale to MUI's date picker is nontrivial; show Korean dates
+      // in the gibberish locale to support testing that the date picker is localized.
+      const effectiveLocale = loadedStringsForLocale === 'gx' ? 'ko' : loadedStringsForLocale;
+      setPropsWithLocale({ ...props, locale: effectiveLocale });
+    }
+  }, [loadedStringsForLocale, props, setPropsWithLocale]);
+
+  return propsWithLocale && WebComponentsDatePicker(propsWithLocale);
```
karthikbtf, line 19: would propsWithLocale always be a valid object? in case we can remove the check...

karthikbtf (approved):
Looks good!!

karthikbtf (approved):
👍

PR #1158, opened 2023-02-08
SW-2813 nursery inventory service

karthikbtf:
- moved types to src/types
- wrapped inventory api around a service
- there's one bug i discovered which needs some review ,, will comment on that


src/components/common/ImportModal.tsx
```diff
@@ -220,8 +215,11 @@ export default function ImportSpeciesModal(props: ImportSpeciesModalProps): JSX.
         id: 0,
         requestSucceeded: false,
       };
+      /**
+       * Is this needed?
+       */
       if (selectedOrganization) {
-        response = await uploadApi(file, selectedOrganization.id.toString());
+        //  response = await uploadApi(file, selectedOrganization.id.toString());
```
karthikbtf, line 222: @constanzauanini can you take a look at this and let me know if it's a bug? if so, i can remove this if-block.
i identified these double uploads.. to the org id and the selected facility id..
the org id is being used as a facility id as well

constanzauanini: This bug was introduced when we starting using the useOrganization provider. Previously, the selectedOrganization was passed only by the SpeciesModal, and the facility by accessionModal and inventoryModal. So the if was necessary for that, because both props where optional. There were no double uploads. Now that we always have the org, we should change this if to check if we have a facility, use the facility and if not, use the selected org

karthikbtf: Ok got it, will update

constanzauanini (approved):
Thanks for the fix! looks good

karthikbtf:
> Thanks for the fix! looks good

Thanks @constanzauanini

PR #1159, opened 2023-02-08
SW-2921 fix inventory withdraw math on formatted numbers

karthikbtf:
- Add util to parse a number from a locale/formatted number-string. See https://observablehq.com/@mbostock/localized-number-parsing (there may be a better way to do this (npm package)? but i like the simple reverse-engineered approach in the above article)
- Add util to format a number using locale (leverage Intl.. not sure if this is totally correct for fixed precision)
- Inventory create/edit/withdraw flows had inconsistent display formatting, BE provided numbers were formatted correctly, UI generated numbers (running totals, etc.) were not formatted. Leverage above utils to render in correct format.
- UI allowed user to enter formatted numbers in the various quantity input elements - but was also sending this back to the BE as the number data, which generated format errors since BE was expecting a number (not a formatted string). Updated code to parse number from human input which could have been formatted
- UI validation functions were interpreting numbers using `Number(string)` or `+<number-string>` both of which won't work with formatted numbers. Updated code to do simper checks such as `number != '0'` and in other cases where we needed to check for `>`,, parse the number from formatted number, leveraging new utils

### DIFF
--- a/src/utils/useNumberParser.ts
+++ b/src/utils/useNumberParser.ts
@@ -1,0 +1,30 @@
+import { useMemo } from 'react';
+
+/**
+ * See: https://observablehq.com/@mbostock/localized-number-parsing
+ */
+
+const useNumberParser = (): any => {
+  const parser = (locale: string): any => {
+    const localeToUse = locale === 'gx' ? 'en' : locale;
+    const parts = new Intl.NumberFormat(localeToUse).formatToParts(12345.6);
+    const numerals = new Intl.NumberFormat(localeToUse, { useGrouping: false }).format(9876543210).split('').reverse();
+    const index = new Map(numerals.map((d, i) => [d, i]));
+    const _group = new RegExp(`[${parts?.find((d) => d?.type === 'group')?.value}]`, 'g');
+    const _decimal = new RegExp(`[${parts?.find((d) => d?.type === 'decimal')?.value}]`);
+    const _numeral = new RegExp(`[${numerals.join('')}]`, 'g');
+    const _index = (d: any) => index.get(d) as unknown as string;
+
+    const parse = (numberStr: string): number | typeof NaN => {
+      const str = numberStr.trim().replace(_group, '').replace(_decimal, '.').replace(_numeral, _index);
+
+      return str ? +str : NaN;
+    };
+
+    return { parse };
+  };
+
+  return useMemo(() => parser, []);
+};
+
+export default useNumberParser;
--- a/src/components/Inventory/InventoryCreate.tsx
+++ b/src/components/Inventory/InventoryCreate.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import { useHistory } from 'react-router-dom';
 import strings from 'src/strings';
 import { APP_PATHS } from 'src/constants';
@@ -20,6 +20,8 @@ import { Facility } from 'src/types/Facility';
 import { getNurseryById } from 'src/utils/organization';
 import { useOrganization } from 'src/providers';
 import isEnabled from 'src/features';
+import { useUser } from 'src/providers';
+import { useNumberParser, useNumberFormatter } from 'src/utils/useNumber';
 
 const MANDATORY_FIELDS = [
   'speciesId',
@@ -37,6 +39,9 @@ export default function CreateInventory(): JSX.Element {
   const theme = useTheme();
   const history = useHistory();
   const snackbar = useSnackbar();
+  const { user } = useUser();
+  const numberParser = useNumberParser();
+  const numberFormatter = useNumberFormatter();
   const [validateFields, setValidateFields] = useState<boolean>(false);
   const [totalQuantity, setTotalQuantity] = useState(0);
   const { selectedOrganization } = useOrganization();
@@ -45,6 +50,9 @@ export default function CreateInventory(): JSX.Element {
   const tz = useLocationTimeZone().get(timeZoneFeatureEnabled ? selectedNursery : undefined);
   const [timeZone, setTimeZone] = useState(tz.id);
   const [addedDateChanged, setAddedDateChanged] = useState(false);
+
+  const numericParser = useMemo(() => numberParser(user?.locale), [user?.locale, numberParser]);
+  const numericFormatter = useMemo(() => numberFormatter(user?.locale), [user?.locale, numberFormatter]);
 
   const defaultBatch = (): CreateBatchRequestPayload =>
     ({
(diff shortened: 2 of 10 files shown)


constanzauanini:
We could probably use it in accession add/edit quantity modal?

constanzauanini:
I found a bug when opening an already created seedling batch, the total quantity is not correct:
Screenshot 2023-02-08 at 10 02 58

Also, that total quantity is not formatted.

And when withdrawing from one batch the withdrawQuantity and totalReadyQuantity is not formatted:
Screenshot 2023-02-08 at 10 06 25

src/utils/useNumberParser.ts
```diff
@@ -0,0 +1,30 @@
...
+ * See: https://observablehq.com/@mbostock/localized-number-parsing
+ */
+
+const useNumberParser = (): any => {
+  const parser = (locale: string): any => {
+    const localeToUse = locale === 'gx' ? 'en' : locale;
+    const parts = new Intl.NumberFormat(localeToUse).formatToParts(12345.6);
+    const numerals = new Intl.NumberFormat(localeToUse, { useGrouping: false }).format(9876543210).split('').reverse();
+    const index = new Map(numerals.map((d, i) => [d, i]));
+    const _group = new RegExp(`[${parts?.find((d) => d?.type === 'group')?.value}]`, 'g');
```
constanzauanini, line 13: Should this "group" be the thousands separator? In Argentina it would be a "." but I'm seeing ","

karthikbtf: Thanks for looking @constanzauanini .
Just so the scope of this PR is clear.. it is to parse a formatted number.
In JS, +'1,000' returns NaN and Number('1,000') returns NaN.
Because of this, some of our Math related code is not functioning like before. (including maybe the totals that you reported as being incorrect). Our withdraw checks, our validation checks.. all use a combination of `+<number-string>` or `Number(<number-string>)`.
This utility's purpose is to return a valid number from a formatted number string.. not the other way around.
So for '1,000' it will return 1000.

Anywhere we see numbers not being formatted in a display - this PR will not fix that.. we need to keep track and fix it separately maybe.

constanzauanini: Ok, thanks for the clarification!

karthikbtf:
> We could probably use it in accession add/edit quantity modal?

in accessions, the add/edit quantity does not seem broken. Are you referring to formatting the quantity? If so, that's not the scope of this PR.

constanzauanini:
> > We could probably use it in accession add/edit quantity modal?
> 
> in accessions, the add/edit quantity does not seem broken. Are you referring to formatting the quantity? If so, that's not the scope of this PR.

Sorry I misunderstood the purpose, so this is not necessary !

karthikbtf:
> > > We could probably use it in accession add/edit quantity modal?
> > 
> > 
> > in accessions, the add/edit quantity does not seem broken. Are you referring to formatting the quantity? If so, that's not the scope of this PR.
> 
> Sorry I misunderstood the purpose, so this is not necessary !

No worries, I just saw your other comment :)
You should create a jira maybe for the formatting issues?

src/utils/useNumberParser.ts
karthikbtf: > I found a bug when opening an already created seedling batch, the total quantity is not correct: Screenshot 2023-02-08 at 10 02 58
> 
> Also, that total quantity is not formatted.
> 
> And when withdrawing from one batch the withdrawQuantity and totalReadyQuantity is not formatted: Screenshot 2023-02-08 at 10 06 25

hmm.. looks like we will need to go through our entire app and clean up how we handle numbers.. if user enters a formatted number etc.. Here the total is incorrect because you entered '2,000' and it was not added to the total count (NaN). we need to fix that for sure..
this is going to be a big sweeping change (not going to be covered in this PR) but something to bring up with @alexTerraformation maybe

constanzauanini: I entered 2000 when I created, but when I go to edit it, it is formatted like that now

constanzauanini (approved):

src/utils/useNumberParser.ts
constanzauanini: But it's not related to this change I guess. It's related to the BE change

karthikbtf: Got it... I have a change coming up that fixes the total calculation (on all the inventory withdraw views - single, batch, batch-details.. etc) and it also formats the total

constanzauanini (changes requested):
Thanks for the changes. I just try editing a batch and quantities work right! But when pressing save, I'm and getting an error: '{"error":{"message":"Field value has incorrect format: readyQuantity"},"status":"error"} '

karthikbtf:
> Thanks for the changes. I just try editing a batch and quantities work right! But when pressing save, I'm and getting an error: '{"error":{"message":"Field value has incorrect format: readyQuantity"},"status":"error"} '

that's unrelated to my changes.. if you entered a comma we are sending that back to the BE.. try it in staging
we should stop using textfields for numbers which is causing a lot of problems maybe

karthikbtf:
> > Thanks for the changes. I just try editing a batch and quantities work right! But when pressing save, I'm and getting an error: '{"error":{"message":"Field value has incorrect format: readyQuantity"},"status":"error"} '
> 
> that's unrelated to my changes.. if you entered a comma we are sending that back to the BE.. try it in staging we should stop using textfields for numbers which is causing a lot of problems maybe

we could either have BE handle it.. or we have to go through *ALL* our text fields and see which ones error out if user enters a formatted number.. this is getting messy

karthikbtf:
> > > > Thanks for the changes. I just try editing a batch and quantities work right! But when pressing save, I'm and getting an error: '{"error":{"message":"Field value has incorrect format: readyQuantity"},"status":"error"} '
> > > 
> > > 
> > > that's unrelated to my changes.. if you entered a comma we are sending that back to the BE.. try it in staging we should stop using textfields for numbers which is causing a lot of problems maybe
> > 
> > 
> > we could either have BE handle it.. or we have to go through _ALL_ our text fields and see which ones error out if user enters a formatted number.. this is getting messy
> 
> Yes, we can merge this, and fix that after ?

maybe we need a plan, i'm not going to take this on myself

src/utils/useNumberParser.ts
karthikbtf: i have an update coming for just the inventory create/edit/withdraw flow to re-parse number from user entered value.. this will fix the issue..
for rest of app we can plan that separately

karthikbtf:
> > > > Thanks for the changes. I just try editing a batch and quantities work right! But when pressing save, I'm and getting an error: '{"error":{"message":"Field value has incorrect format: readyQuantity"},"status":"error"} '
> > > 
> > > 
> > > that's unrelated to my changes.. if you entered a comma we are sending that back to the BE.. try it in staging we should stop using textfields for numbers which is causing a lot of problems maybe
> > 
> > 
> > we could either have BE handle it.. or we have to go through _ALL_ our text fields and see which ones error out if user enters a formatted number.. this is getting messy
> 
> Yes, we can merge this, and fix that after ?

should be fixed (for inventory flows alone here).. take a look when you get a chance

constanzauanini (approved):
Thank you!

PR #1160, opened 2023-02-08
SW-2716 Seed weight calculator not recognizing decimal point

constanzauanini:
Allow inserting '.' when typing on subset weight field

### DIFF
--- a/src/components/accession2/edit/CalculatorModal.tsx
+++ b/src/components/accession2/edit/CalculatorModal.tsx
@@ -76,7 +76,7 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
     onCloseHandler();
   };
 
-  const onChangeSubsetWeight = (value: number) => {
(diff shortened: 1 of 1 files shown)


src/components/accession2/edit/CalculatorModal.tsx
```diff
@@ -76,7 +76,7 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
     onCloseHandler();
   };
 
-  const onChangeSubsetWeight = (value: number) => {
+  const onChangeSubsetWeight = (value: any) => {
```
karthikbtf, line 79: why any?

karthikbtf (approved):

src/components/accession2/edit/CalculatorModal.tsx
constanzauanini: Because the value that comes from the textfield change it's "unknown". If we use number, decimal point doesn't appear when the value is "2. "

PR #1161, opened 2023-02-08
SW-2905 Use new API for storage locations

sgrimm:
There is a new API for listing a facility's storage locations; use it instead of
the old one. The new API doesn't support storage conditions; remove all references
to them from the code.

The code still has references to "Freezer" and "Refrigerator" but they are only
used in the context of sensor kit setup and monitoring.


karthikbtf (approved):
looks good to me, thanks for the changes/cleanup

PR #1162, opened 2023-02-08
SW-2916 Validation on Seed QTY causes NaN to be stuck in input field

constanzauanini:


charlie-tf (approved):
Change looks good, and tests out on Vercel 👍

PR #1164, opened 2023-02-08
SW-2925 Handle localized booleans in search results

sgrimm:
Boolean search fields, like all search fields, return string values. Currently,
these are assumed to be the English words `'true'` and `'false'`.

Update the application code to handle booleans being rendered as localized
strings.

This fixes a bug in the species list UI in English: if you had a search filter,
it could erroneously show species as rare and/or endangered because the string
`'false'` evaluates as true in JavaScript.


sgrimm:
This requires a corresponding server-side change; until that gets merged, this will break the rare/endangered species filters in Gibberish (but English will be fine).

karthikbtf (approved):
👍

PR #1165, opened 2023-02-08
SW-2917 make plants dashboard work

karthikbtf:
- use number parser
- use number formatter
- create chart only when all data is available.. to avoid constructing charts in parallel

### DIFF
--- a/src/components/Plants/DashboardChart.tsx
+++ b/src/components/Plants/DashboardChart.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { makeStyles } from '@mui/styles';
 import { Theme, useTheme } from '@mui/material';
 import { generateTerrawareRandomColors } from 'src/utils/generateRandomColor';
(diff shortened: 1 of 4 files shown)


src/components/Plants/DashboardChart.tsx
```diff
@@ -23,13 +23,36 @@ export interface DashboardChartProps {
...
+
+interface DashboardChartContentProps {
+  chartId: string;
+  chartLabels: string[];
+  chartValues: number[];
+  minHeight?: string;
+  locale: string;
+}
+
+function DashboardChartContent(props: DashboardChartContentProps): JSX.Element {
```
karthikbtf, line 50: moved the original chart into a separate component that requires a locale and the values/labels..
the parent component/wrapper now just returns null if they are not set.
this handles most cases of useEffect concurrency issues where chart was being constructed for same canvas multiple times in parallel

src/components/Plants/PlantingSiteDetails.tsx
```diff
@@ -125,12 +130,12 @@ export default function PlantingSiteDetails(props: PlantingSiteDetailsProps): JS
...
           const plantsPerSpecies: { [key: string]: number } = serverResponse.reduce((acc, population) => {
-            totalPlantsOfSite = +totalPlantsOfSite + +population.totalPlants;
+            totalPlantsOfSite = +totalPlantsOfSite + numericParser.parse(population.totalPlants);
             if (acc[population.species_scientificName]) {
               acc[population.species_scientificName] =
-                +acc[population.species_scientificName] + +population.totalPlants;
+                +acc[population.species_scientificName] + numericParser.parse(population.totalPlants);
             } else {
-              acc[population.species_scientificName] = +population.totalPlants;
+              acc[population.species_scientificName] = numericParser.parse(population.totalPlants);
```
karthikbtf, line 138: here they were all being NaN if the totalPlants was formatted (large numbers)

constanzauanini:
I tested in vercel and the first time loaded, the second chart desn't have data:
Screenshot 2023-02-08 at 15 54 12

if changing and going back:
Screenshot 2023-02-08 at 15 55 23

constanzauanini (approved):
Looks good now!

karthikbtf:
> Looks good now!

i'll get this in now... we can cleanup the error again later.. after things settle a bit

PR #1167, opened 2023-02-08
SW-2918 Format numeric chart axes in gibberish

sgrimm:
On the plants dashboard, the bar graph of number of plants by species wasn't using
localized numbers. We want localized numbers on all our charts, so extend our
custom Chart.js wrapper to add render functions for gibberish numbers on Y axis
labels and in tooltips.

The locale-aware number formatter wasn't using the formatting rules for the
gibberish locale; extract that logic out into a couple of helper functions.


sgrimm:
I'm not at all convinced this PR is a good idea. It is 100% special-case code that doesn't fix anything for any actual users. Putting it up so we can discuss at roundtable tomorrow, not proposing to merge it before then.

karthikbtf:
> I'm not at all convinced this PR is a good idea. It is 100% special-case code that doesn't fix anything for any actual users. Putting it up so we can discuss at roundtable tomorrow, not proposing to merge it before then.

I'm totally fine with this, given the debug/QA value.

karthikbtf (approved):
lgtm, thanks! we can discuss more in the roundtable if you prefer.. but certainly no harm for the near term..

sgrimm:
Abandoning this because we decided to use French number formatting; will open a separate PR for that.

PR #1168, opened 2023-02-09
SW-2681 Show withdrawal notes in accession history

sgrimm:
If an accession withdrawal has notes, show them below the system-generated
description of the history event.

### DIFF
--- a/src/components/accession2/history/Accession2History.tsx
+++ b/src/components/accession2/history/Accession2History.tsx
@@ -62,6 +62,11 @@ export default function Accession2History(props: Accession2HistoryProps): JSX.El
           </Typography>
           <Typography fontWeight={500}>
             {item.fullName || strings.NAME_UNKNOWN}&nbsp;{item.description}
+            {item.notes && (
+              <Typography fontSize={'14px'} fontWeight={300}>
+                {item.notes}
+              </Typography>
+            )}
           </Typography>
         </Box>
       ))}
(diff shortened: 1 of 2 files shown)


sgrimm:
Screenshot 2023-02-09 at 10 46 18 AM

sgrimm:
Server side of this is https://github.com/terraware/terraware-server/pull/957 but this change is harmless to merge; there just won't be any notes to show until the server starts including them in API responses.

src/components/accession2/history/Accession2History.tsx
```diff
@@ -62,6 +62,11 @@ export default function Accession2History(props: Accession2HistoryProps): JSX.El
           </Typography>
           <Typography fontWeight={500}>
             {item.fullName || strings.NAME_UNKNOWN}&nbsp;{item.description}
+            {item.notes && (
+              <Typography fontSize={'14px'} fontWeight={300}>
```
sgrimm, line 66: Not sure if setting the font size and weight to hardwired constants is the idiomatic way to do this; I was following the example of the parent `<Typography>` and `<Box>` tags.

karthikbtf: This works, we are using hardwired constants for now across the board. There is a plan to enhance all this with more size semantics. For now this is totally fine.

karthikbtf (approved):

PR #1171, opened 2023-02-09
SW-2933 Use French date/number formatting for gibberish

sgrimm:
The goal of the gibberish locale is to be visually distinct from English so we
can spot places in the UI that we haven't localized. Initially, that included
using custom formatting rules for numbers and dates, but that has turned out to
be annoying to implement in practice. Update the code to use French formatting
instead; it serves the purpose of being distinct from English but requires less
special-case code.

Also fixes SW-2918.


karthikbtf (approved):
👍

PR #1174, opened 2023-02-10
Use memoized feature notifications

karthikbtf:


karthikbtf:
@constanzauanini @charlie-tf i will get this in to avoid breaking staging.
We can follow up as needed in a separate PR

PR #1176, opened 2023-02-10
SW-2941 / SW-2942 Sort inventory/withdrawals table using BE sort order

karthikbtf:
- also applies to batch details table
- nurseries columns is a client defined column and uses it's own sort mechanism
- common name is not working so using client side for now
- the detail views within withdrawal log are static and will use client side sorting for now

### DIFF
--- a/src/components/NurseryWithdrawals/WithdrawalLogRenderer.tsx
+++ b/src/components/NurseryWithdrawals/WithdrawalLogRenderer.tsx
@@ -20,93 +20,83 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
-const WithdrawalLogRenderer = (formatter: { format: (num: number) => string }) => {
+export default function WithdrawalLogRenderer(props: RendererProps<TableRowType>): JSX.Element {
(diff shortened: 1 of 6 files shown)


constanzauanini (approved):

src/components/NurseryWithdrawals/WithdrawalLogRenderer.tsx
```diff
@@ -20,93 +20,83 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-      return <CellRenderer index={index} column={column} row={row} value='' />;
-    }
-
-    return <CellRenderer {...props} />;
-  };
-};
-
-export default WithdrawalLogRenderer;
+  return <CellRenderer {...props} />;
+}
```
karthikbtf, line 102: All these changes are basically a revert of a previous PR that introduced client-side formatting of numeric data. Removed that since we will use formatted numbers from the BE.

constanzauanini (approved):
Looks good, just tested nursery withdrawals in vercel

PR #1177, opened 2023-02-10
SW-2940 Species table should use BE sorted results when sort order/field changes

constanzauanini:
- Allow sorting on the BE
- seedStorageBehavior and ecosystemType are sort in the FE.

### DIFF
--- a/src/components/Species/index.tsx
+++ b/src/components/Species/index.tsx
@@ -22,7 +22,7 @@ import PageSnackbar from 'src/components/PageSnackbar';
 import AddSpeciesModal from './AddSpeciesModal';
 import DeleteSpeciesModal from './DeleteSpeciesModal';
 import TextField from '../common/Textfield/Textfield';
-import { FieldNodePayload, search, searchCsv, SearchNodePayload } from 'src/api/search';
+import { FieldNodePayload, search, searchCsv, SearchNodePayload, SearchSortOrder } from 'src/api/search';
 import SpeciesFilters from './SpeciesFiltersPopover';
 import useForm from 'src/utils/useForm';
 import Icon from '../common/icon/Icon';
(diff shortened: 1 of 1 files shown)


src/components/Species/index.tsx
```diff
@@ -388,26 +398,13 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
...
-            growthForm: getGrowthFormString(sp),
-            seedStorageBehavior: getSeedStorageBehaviorString(sp),
-            ecosystemTypes: getEcosystemTypesString(sp),
-          });
-        });
-        return speciesListWithCS;
-      };
-
       const params: SearchNodePayload = getParams();
-      if (params.search.children.length > 1) {
```
constanzauanini, line 410: Had to remove this condition because params length could be 1, when just sorting (and not filtering)

src/components/Species/index.tsx
```diff
@@ -388,26 +398,13 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
...
+
+      if (species) {
```
constanzauanini, line 407: Added this condition to allow refreshing the search when a species entry is edited

src/components/Species/index.tsx
```diff
@@ -285,6 +289,8 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
   const userCanEdit = !isContributor(selectedOrganization);
   const { isMobile } = useDeviceInfo();
 
+  const BE_SORTED_FIELDS = ['scientificName', 'commonName', 'familyName', 'growthForm', 'seedStorageBehavior'];
```
karthikbtf, line 292: you can use this in the fields array line 297 as well.
```
  fields: [
    ...BE_SORTED_FIELDS,
    <remaining fields>
  ],
```
updating that list will be easier in one place..

karthikbtf (approved):
looks good, one suggestion

src/components/Species/index.tsx
constanzauanini: I saw that in your PR after creating this PR haha

karthikbtf (approved):
one suggestion, feel free to merge

PR #1182, opened 2023-02-13
SW-2883 Error page copy is missing

constanzauanini:
The error page was not showing the labels, this was because locale was not set for the error page.
Adding AppBootstrap to error page to allow setting the locale

### DIFF
--- a/src/AppError.tsx
+++ b/src/AppError.tsx
@@ -3,17 +3,24 @@ import { Route, Switch } from 'react-router-dom';
 import { APP_PATHS } from 'src/constants';
 import useQuery from './utils/useQuery';
 import ErrorContent from './ErrorContent';
+import { Provider } from 'react-redux';
+import AppBootstrap from './AppBootstrap';
+import { store } from './redux/store';
 
 export default function AppError() {
   return (
-    <Switch>
-      <Route exact path={APP_PATHS.ERROR_FAILED_TO_FETCH_ORG_DATA}>
-        <ErrorContent />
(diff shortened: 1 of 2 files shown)


charlie-tf (commented):
Can these layers be placed inside `AppError` similar to the way they are layered in `App`? To avoid importing those components to `index`...

constanzauanini:
> Can these layers be placed inside `AppError` similar to the way they are layered in `App`? To avoid importing those components to `index`...

That's a good suggestion!

charlie-tf (approved):
👍

src/AppError.tsx
```diff
@@ -3,17 +3,24 @@ import { Route, Switch } from 'react-router-dom';
...
-    <Switch>
-      <Route exact path={APP_PATHS.ERROR_FAILED_TO_FETCH_ORG_DATA}>
-        <ErrorContent />
-      </Route>
-      <Route path={APP_PATHS.ERROR}>
-        <QueryParamsError />
-      </Route>
-    </Switch>
+    <AppBootstrap>
+      <Provider store={store}>
```
karthikbtf, line 13: what is the store for? app version detector?

constanzauanini: I don't think we are using the store here, just wanted to have same initialization as App, but probably we can remove it

karthikbtf: no worries, was just curious.. we can leave it if we want same initialization / consistency

PR #1185, opened 2023-02-13
SW-2618 Build Generic Filter Edit Dialog Component

charlie-tf:
Create a FilterGroup component that takes a list of filter field with
specified filter types (e.g. `multiple_selection`, `number_range`, etc).
Also take a FieldValuesPayload containing the options for each field, if
applicable. The FilterGroup may be rendered inside whichever container is
desired, such as a dialog or a popover. TODO: one item that needs to be
done still is to render the `multiple_selection` filter type as a Filter-
MultiSelect. Change the `single_selection` component to a dropdown.
These will be done in a followup.

The FilterGroup is implemented in the Accessions table page, to show it
in action. We remove the chip-based filters and also confirmed that each
of the types of filter works correctly. TODO: add pills showing active filters.

image
image
image
image
image


src/components/common/FilterGroup/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+  name: string;
+  label: string;
+  type:
+    | 'multiple_selection'
+    | 'single_selection'
+    | 'search'
+    | 'date_range'
+    | 'number_range'
+    | 'count_weight'
+    | 'hidden';
```
karthikbtf, line 42: curious.. how does hidden work?

src/components/common/FilterGroup/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+        </Typography>
+      </Box>
+      <Box flex='1 1 auto' overflow='auto'>
+        {fields.map((f, index) => (
+          <Box key={f.name}>
+            {index > 0 && <hr className={classes.divider} />}
+            <Typography fontSize='14px' fontWeight={600} margin={theme.spacing(2, 2, 0, 2)}>
+              {f.label}
+            </Typography>
+            {f.type === 'multiple_selection' && (
```
karthikbtf, line 99: this is fine .. or a switch statement if you find that to be more expressive.. all the same

src/components/common/FilterGroup/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+              <MultipleSelection
```
karthikbtf, line 100: A lot of these filter types will end up showing user defined values - which could be exhaustive. wondering if we need a scrollable sub filter area  - but that leads to a double scroll ux pattern (parent container scrolls). just calling that out since maybe the design didn't account for it

karthikbtf (approved):
looks great in vercel!!!

src/components/common/FilterGroup/index.tsx
charlie-tf: I'm going to be replacing the multiple selection checkboxes with the FilterMultiSelect, which should clean things up for that. The single select may still be exhaustive and large, so i can replace it with a Dropdown select. Those two changes should take care of this.

src/components/common/FilterGroup/index.tsx
charlie-tf: Hmm, i just copy-pasted this from the database code, the case was simply not handled there, which is the same as what I've done.

src/components/common/FilterGroup/index.tsx
karthikbtf: Awesome 🙌

PR #1187, opened 2023-02-13
SW-2911 / SW-2947 / SW-2953 Show storage locations in seed bank details view

karthikbtf:
- allow clicking on a location name to see accessions view

Terraware App 2023-02-13 13-14-34

Terraware App 2023-02-13 13-14-11

Terraware App 2023-02-13 13-13-35

### DIFF
--- a/src/types/Facility.ts
+++ b/src/types/Facility.ts
@@ -1,5 +1,4 @@
 import { components } from 'src/api/types/generated-schema';
-import strings from 'src/strings';
 
 const schemas = 'schemas';
 
@@ -16,14 +15,3 @@ export interface Facility {
   connectionState: 'Not Connected' | 'Connected' | 'Configured';
   timeZone?: string;
 }
-
-export const ActiveStatuses = () => [
-  strings.AWAITING_CHECK_IN,
-  strings.PENDING,
-  strings.AWAITING_PROCESSING,
-  strings.PROCESSING,
-  strings.PROCESSED,
-  strings.DRYING,
-  strings.DRIED,
-  strings.IN_STORAGE,
-];
--- a/src/types/Facility.ts
+++ b/src/types/Facility.ts
@@ -1,8 +1,9 @@
-import { paths, components } from 'src/api/types/generated-schema';
+import { components } from 'src/api/types/generated-schema';
+import strings from 'src/strings';
 
 const schemas = 'schemas';
 
-export type StorageLocationPayload = components[typeof schemas]['StorageLocationPayload'];
+export type StorageLocation = components[typeof schemas]['StorageLocationPayload'];
 
 export type FacilityType = components[typeof schemas]['FacilityPayload']['type'];
 
@@ -15,3 +16,14 @@ export interface Facility {
   connectionState: 'Not Connected' | 'Connected' | 'Configured';
   timeZone?: string;
 }
+
+export const ActiveStatuses = () => [
+  strings.AWAITING_CHECK_IN,
+  strings.PENDING,
+  strings.AWAITING_PROCESSING,
+  strings.PROCESSING,
+  strings.PROCESSED,
+  strings.DRYING,
+  strings.DRIED,
+  strings.IN_STORAGE,
+];
--- a/src/utils/useNumber.ts
+++ b/src/utils/useNumber.ts
@@ -6,14 +6,6 @@ import { useMemo } from 'react';
 
 const getLocaleToUse = (locale?: string) => (locale === 'gx' ? 'fr' : locale || 'en');
 
-export type NumericParser = {
-  parse: (str: string) => number | typeof NaN;
-};
-
-export type NumericFormatter = {
-  format: (num: number) => string;
-};
-
 export const useNumberParser = (): any => {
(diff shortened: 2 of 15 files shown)


src/types/Facility.ts
```diff
@@ -15,3 +16,14 @@ export interface Facility {
   connectionState: 'Not Connected' | 'Connected' | 'Configured';
   timeZone?: string;
 }
+
+export const ActiveStatuses = () => [
```
karthikbtf, line 20: This list was generated from current accession state enum definitions in the BE

karthikbtf:
Note, Jiras SW-2911 / SW-2947 / SW-2953 are handled in this one PR because the fixes are very much intertwined.

chudo76:
@karthikbtf  Is there an org on staging that I can be added to that has seed banks with sub-locations? (I'm not seeing sub-locations on vercel deployments.)

chudo76:
If someone clicks on a Sub-Location from “Locations > Seed Banks > `<seedbank>` > Sub-Locations,” we will want to show the column and filter for “Sub-Locations” in the Accessions table (if it’s not already being shown). Alternative is to show column and filter for “Sub-Locations” by default with the release of this feature. (Same note added to SW-2948.)

karthikbtf:
> If someone clicks on a Sub-Location from “Locations > Seed Banks > `<seedbank>` > Sub-Locations,” we will want to show the column and filter for "Seed Banks" and “Sub-Locations” in the Accessions table (if it’s not already being shown). Alternative is to show column and filter for "Seed Banks" and “Sub-Locations” by default with the release of this feature. (Same note added to [SW-2948](https://terraformation.atlassian.net/browse/SW-2948).)
> 
> Edit: Added "Seed Banks."

@chudo76 just curious if you are not seeing this in the vercel build or just calling out requirements..

chudo76:
> @chudo76 just curious if you are not seeing this in the vercel build or just calling out requirements..

@karthikbtf I had to turn on the columns in staging to figure out that the Accession table was being filtered by Sub-Locations. Just looking at the PRD now and although it's not called out, I think we would need to adjust setting so that column and filter are on for Seed Bank and Sub-Location to show that it is filtered by Sub-Location (if the user hadn't had those columns turned on prior).

It's noted on SW-2948 so it can be addressed as a separate issue. I'm going to ask to tweak the design so that the link to the filtered Accessions page is on the "Active Accessions" number instead of the sub-location name.

karthikbtf:
> > @chudo76 just curious if you are not seeing this in the vercel build or just calling out requirements..
> 
> @karthikbtf I had to turn on the columns in staging to figure out that the Accession table was being filtered by Sub-Locations. Just looking at the PRD now and although it's not called out, I think we would need to auto-adjust the user's setting so that column and filter are on for Seed Bank and Sub-Location to show that it is filtered by Sub-Location (if the user hadn't had those columns turned on prior).
> 
> It's noted on [SW-2948](https://terraformation.atlassian.net/browse/SW-2948) so it can be addressed as a separate issue. I'm going to ask to tweak the design so that the link to the filtered Accessions page is on the "Active Accessions" number instead of the sub-location name.

Got it.. I didn’t realize I had the columns already enabled.. thought they just showed up. Will make the enhancements.
Btw there are no designs for this.. driven by Eng and design can give feedback/suggestions. This is a starting point.
I can make the number the link, makes sense here. Also any thoughts on the Sub-Locations title? Seems redundant maybe? The table column headers seem descriptive enough?

chudo76:
> Got it.. I didn’t realize I had the columns already enabled.. thought they just showed up. Will make the enhancements. Btw there are no designs for this.. driven by Eng and design can give feedback/suggestions. This is a starting point. I can make the number the link, makes sense here. Also any thoughts on the Sub-Locations title? Seems redundant maybe? The table column headers seem descriptive enough?

I think "Sub-Locations" title is fine. We can change the column header label from "Sub-Location" to "Name" and that would remove redundancy (and probably makes sense anyway since it looks like PRD is saying we'd want people to be able to edit the name.)

src/utils/useNumber.ts
```diff
@@ -6,6 +6,14 @@ import { useMemo } from 'react';
 
 const getLocaleToUse = (locale?: string) => (locale === 'gx' ? 'fr' : locale || 'en');
 
+export type NumericParser = {
```
constanzauanini, line 9: Could we move this inside /src/types ? I thought it was a util, but it's just the type

constanzauanini (approved):
Looks good to me! Seebank and sublocation Columns on accessions page don't show yet, but guess will be implemented later

src/utils/useNumber.ts
karthikbtf: For now I'd prefer we keep the src/types specific to data modeling and API
we have a ton of types exported from utils and components that are outside of data model

karthikbtf:
> Looks good to me! Seebank and sublocation Columns on accessions page don't show yet, but guess will be implemented later

yes that will be separate

src/utils/useNumber.ts
karthikbtf: and we have a ton of types exported from services as well...

PR #1190, opened 2023-02-14
SW-2955 Remove scrollbars during progress circle animation

constanzauanini:
Avoid using display flex to center the spinner, because that was adding scrollbars apparently.


charlie-tf (approved):
Looks good. Appears centered without scrollbars in vercel 👍

PR #1191, opened 2023-02-14
SW-2962 / SW-2963 Sort timezones by display name, use localized facility type in searches

karthikbtf:
- also fix search to use localized facility enum value

### DIFF
--- a/src/components/Nurseries/TableCellRenderer.tsx
+++ b/src/components/Nurseries/TableCellRenderer.tsx
@@ -3,11 +3,9 @@ import { APP_PATHS } from 'src/constants';
 import CellRenderer, { TableRowType } from '../common/table/TableCellRenderer';
 import { RendererProps } from '../common/table/types';
 import Link from '../common/Link';
-import { useGetTimeZone } from 'src/utils/useTimeZoneUtils';
 
 export default function NurseriesCellRenderer(props: RendererProps<TableRowType>): JSX.Element {
   const { column, row, value, index } = props;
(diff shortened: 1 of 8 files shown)


constanzauanini (approved):
Thanks for fixing this!

karthikbtf:
> Thanks for fixing this!

Thanks for flagging this :) I did not notice it...

charlie-tf (commented):
Looks great! Just to make sure I understand: the idea here is that each of these facilities in question is assigned a timezone—either one that is specified by the facility or a default—and that we are sorting them correctly?

karthikbtf:
> Looks great! Just to make sure I understand: the idea here is that each of these facilities in question is assigned a timezone—either one that is specified by the facility or a default—and that we are sorting them correctly?

That is correct. We were rendering them post-sort. 
The render would use the facility's time zone (defaulting to org's time zone if facility doesn't have one). The sort was broken since the data to sort on was missing - and the data being sorted was the id and not the display name.
Moved the display logic into the data population.

charlie-tf:
> > Looks great! Just to make sure I understand: the idea here is that each of these facilities in question is assigned a timezone—either one that is specified by the facility or a default—and that we are sorting them correctly?
> 
> That is correct. We were rendering them post-sort. The render would use the facility's time zone (defaulting to org's time zone if facility doesn't have one). The sort was broken since the data to sort on was missing - and the data being sorted was the id and not the display name. Moved the display logic into the data population.

Got it. Thanks for the explanation!

PR #1192, opened 2023-02-14
SW-2621 Add Pills to Accession Filters

charlie-tf:
image
image

### DIFF
--- a/src/components/seeds/database/Filters.tsx
+++ b/src/components/seeds/database/Filters.tsx
@@ -1,6 +1,6 @@
 import { Container, IconButton, Popover, Theme, Typography } from '@mui/material';
 import { makeStyles } from '@mui/styles';
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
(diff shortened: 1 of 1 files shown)


src/components/seeds/database/Filters.tsx
```diff
@@ -171,6 +179,27 @@ export default function Filters(props: Props): JSX.Element {
...
+  for (const col of columns) {
+    const filter = filters[col.key];
+    if (filter) {
+      filterPillItems.push({
+        id: col.key,
+        label: col.name,
+        value: filter.values.join(`, `),
+      } as PillListItem<string>);
+    }
+  }
```
karthikbtf, line 201: suggestion - if it helps with perf.. move the filterPillItems into a useMemo?

karthikbtf (approved):

src/components/seeds/database/Filters.tsx
charlie-tf: That's a pretty good idea I think 👍

karthikbtf (approved):
Thank you!!

PR #1193, opened 2023-02-14
SW-2965 Add bottom padding to feature notifications

constanzauanini:
The bottom padding is given by the notification date, so add padding when no date 

Screenshot 2023-02-14 at 15 58 19


karthikbtf (approved):
🙏

PR #1194, opened 2023-02-14
SW-2901 User notification remove after 30 days

constanzauanini:
- Remove notification after 30 days
- When clicking notification, just mark it as read (not remove it)

For org notification we have to do same thing

### DIFF
--- a/src/components/FeatureNotification/UserNotification.tsx
+++ b/src/components/FeatureNotification/UserNotification.tsx
@@ -13,7 +13,7 @@ import { InitializedTimeZone, TimeZoneDescription } from 'src/types/TimeZones';
 import { getTimeZone, getUTC } from 'src/utils/useTimeZoneUtils';
 import { PreferencesService, UserService } from 'src/services';
 import { InitializedUnits } from 'src/units';
-import { DateTime } from 'luxon';
+import { featureNotificationExpired } from 'src/utils/featureNotifications';
 
 export default function UserNotification(): Notification | null {
   const [unitNotification, setUnitNotification] = useState(false);
@@ -40,11 +40,7 @@ export default function UserNotification(): Notification | null {
     };
 
     const notifyTimeZoneUpdates = (userTz: InitializedTimeZone) => {
-      const notifyUser =
-        userTz.timeZone &&
-        (!userTz.timeZoneAcknowledgedOnMs ||
-          (userTz.timeZoneAcknowledgedOnMs &&
-            DateTime.now().plus({ days: -30 }).toMillis() <= userTz.timeZoneAcknowledgedOnMs));
+      const notifyUser = userTz.timeZone && !featureNotificationExpired(userTz.timeZoneAcknowledgedOnMs);
       setUserTimeZone(getTimeZoneById(userTz.timeZone).longName);
       setTimeZoneUserNotification(!!notifyUser);
 
@@ -96,15 +92,7 @@ export default function UserNotification(): Notification | null {
         setUnitNotificationRead(true);
       }
 
-      if (
-        !userUnit.unitsAcknowledgedOnMs ||
-        (userUnit.unitsAcknowledgedOnMs &&
-          DateTime.now().plus({ days: -30 }).toMillis() <= userUnit.unitsAcknowledgedOnMs)
-      ) {
-        setUnitNotification(true);
-      } else {
-        setUnitNotification(false);
-      }
+      setUnitNotification(!featureNotificationExpired(userUnit.unitsAcknowledgedOnMs));
     };
 
     if (weightUnitsEnabled) {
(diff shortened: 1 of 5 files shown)


src/components/FeatureNotification/UserNotification.tsx
```diff
@@ -37,9 +40,18 @@ export default function UserNotification(): Notification | null {
     };
 
     const notifyTimeZoneUpdates = (userTz: InitializedTimeZone) => {
-      const notifyUser = userTz.timeZone && !userTz.timeZoneAcknowledgedOnMs;
+      const notifyUser =
+        userTz.timeZone &&
+        userTz.timeZoneNotificationCreatedMs &&
+        DateTime.now().plus({ days: -30 }).toMillis() <= userTz.timeZoneNotificationCreatedMs;
```
charlie-tf, line 46: Does this show the notification for 30 days even if the user has acknowledged it?

src/services/UserService.ts
```diff
@@ -107,6 +108,15 @@ const initializeTimeZone = async (user: User, timeZone: string): Promise<Initial
     initializedTimeZone.timeZone = user.timeZone;
   }
 
+  if (!timeZoneNotificationCreatedMs) {
+    const response = await PreferencesService.updateUserPreferences({ timeZoneNotificationCreatedMs: Date.now() });
+    if (response.requestSucceeded) {
+      initializedTimeZone.timeZoneNotificationCreatedMs = Date.now();
+    }
+  } else {
+    initializedTimeZone.timeZoneNotificationCreatedMs = timeZoneNotificationCreatedMs;
```
karthikbtf, line 117: don't need this since we have line 98

src/services/UserService.ts
```diff
@@ -149,6 +161,15 @@ const initializeUnits = async (units: string): Promise<InitializedUnits> => {
...
   }
 
+  if (!unitsNotificationCreatedMs) {
+    const response = await PreferencesService.updateUserPreferences({ unitsNotificationCreatedMs: Date.now() });
+    if (response.requestSucceeded) {
+      initializedUnits.unitsNotificationCreatedMs = Date.now();
+    }
+  } else {
+    initializedUnits.unitsNotificationCreatedMs = unitsNotificationCreatedMs;
+  }
```
karthikbtf, line 171: same.. remove?

src/components/FeatureNotification/UserNotification.tsx
karthikbtf, line 46: this is fine but won't scale for org level stuff since we can't keep it on a per user preference..
should we just stop showing 30 days "after" user acknowledges the event? if so, then we can skip the created on preference... just a thought.

karthikbtf: yes, that's the plan (similar to other events).. if acknowledged, we show the notification as 'read'

constanzauanini: We could do that, I thought the notification was removed after 30 days of creation, but maybe you are right, just use the acknowledge event

src/services/UserService.ts
```diff
@@ -143,12 +133,10 @@ const getInitializedTimeZone = (user: User, timeZone: string): Promise<Initializ
  * initialize preferred units (if not already set)
  */
 const initializeUnits = async (units: string): Promise<InitializedUnits> => {
-  const { unitsAcknowledgedOnMs, preferredWeightSystem, unitsNotificationCreatedMs } =
-    CachedUserService.getUserPreferences();
+  const { unitsAcknowledgedOnMs, preferredWeightSystem } = CachedUserService.getUserPreferences();
 
   const initializedUnits: InitializedUnits = {
     unitsAcknowledgedOnMs,
-    unitsNotificationCreatedMs,
```
karthikbtf: once we have other feature notifications, we could rethink all this.. we may need a more scalable way of doing this. maybe introduce a `firstShowedOnMs` timestamp since that will scale by user. But keeping them all in a user preference may not be a good idea.. we may need to discuss with BE, later once the feature notifications start rolling in.

karthikbtf (approved):
One suggestion in the code.
Also, I just realized we can think about a `firstShowedOnMs` timestamp (instead of created on timestamp) since that will scale across features and it's per user. But storing all this in a user preference may not scale. We can talk about this with BE once the # of features increases maybe. For now this looks good to me.

src/utils/featureNotifications.tsx
```diff
@@ -0,0 +1,4 @@
+import { DateTime } from 'luxon';
+
+export const featureNotificationExpired = (preferenceTimestamp: number | undefined) =>
```
karthikbtf, line 3: will we use this on the org notification as well?

karthikbtf (approved):
thanks for the changes

src/utils/featureNotifications.tsx
constanzauanini: Yes, didn't implement it yet

PR #1197, opened 2023-02-14
Use MultiSelect Component in FilterGroup

charlie-tf:
image
image
image

### DIFF
--- a/src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
+++ b/src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
@@ -45,8 +45,8 @@ export default function MultipleSelection(props: Props): JSX.Element {
     const valueIndex = updatesValues.findIndex((v) => v === value);
     if (valueIndex < 0) {
       updatesValues.push(value);
+      updateFilters(updatesValues);
     }
-    updateFilters(updatesValues);
   };
 
   const onRemove = (value: string | null) => {
@@ -54,8 +54,8 @@ export default function MultipleSelection(props: Props): JSX.Element {
     const valueIndex = updatesValues.findIndex((v) => v === value);
     if (valueIndex >= 0) {
       updatesValues.splice(valueIndex, 1);
+      updateFilters(updatesValues);
     }
-    updateFilters(updatesValues);
   };
 
   const updateFilters = (updatesValues: (string | null)[]) => {
--- a/src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
+++ b/src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
@@ -1,14 +1,13 @@
 import { Theme } from '@mui/material';
 import { makeStyles } from '@mui/styles';
-import React from 'react';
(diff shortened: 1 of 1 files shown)


src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
```diff
@@ -33,16 +32,33 @@ export default function MultipleSelection(props: Props): JSX.Element {
...
+  );
 
+  const onAdd = (value: string | null) => {
+    const updatesValues = [...selections];
     const valueIndex = updatesValues.findIndex((v) => v === value);
     if (valueIndex < 0) {
       updatesValues.push(value);
-    } else {
+    }
+    updateFilters(updatesValues);
```
karthikbtf, line 49: move this into the if block on line 46?

src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
```diff
@@ -33,16 +32,33 @@ export default function MultipleSelection(props: Props): JSX.Element {
     setSelections(props.values);
   }, [props.values]);
 
-  const handleChange = (value: string | null) => {
-    const updatesValues = [...selections];
+  const optionsMap = useMemo(
+    () =>
+      new Map<string, string>(
+        props.options.filter((opt) => opt.value && opt.label).map((opt) => [opt.value!, opt.label!])
```
karthikbtf, line 38: never used this construction before.. does the map get created from the array of tuples? first element key and second element value? just curious.

src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
```diff
@@ -33,16 +32,33 @@ export default function MultipleSelection(props: Props): JSX.Element {
...
+    updateFilters(updatesValues);
+  };
+
+  const onRemove = (value: string | null) => {
+    const updatesValues = [...selections];
+    const valueIndex = updatesValues.findIndex((v) => v === value);
+    if (valueIndex >= 0) {
       updatesValues.splice(valueIndex, 1);
     }
+    updateFilters(updatesValues);
```
karthikbtf, line 58: same, move it into the if block.. i understand this may not really happen (we won't remove something that doesn't exist) ..

karthikbtf (approved):
looks good, some suggestions

src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
charlie-tf: Good call; no need to perform this op if the updatesValues is unchanged 👍

src/components/common/FilterGroup/filters/FilterMultipleSelection.tsx
charlie-tf: Yep, that's how this works, `new Map<K,V>(...)` takes a list of tuples like `[[K1, V1], [K2, V2], ..., [Kn, Vn]]`

karthikbtf (approved):
Thanks for the changes!

PR #1199, opened 2023-02-15
SW-2966 Storage locations API wrappers

karthikbtf:
- introduced storage location service
- added some bulk API handling in anticipation of related UX

### DIFF
--- a/src/api/types/generated-schema.ts
+++ b/src/api/types/generated-schema.ts
@@ -203,9 +203,6 @@ export interface paths {
   "/api/v1/seedbank/values/all": {
     post: operations["listAllFieldValues"];
   };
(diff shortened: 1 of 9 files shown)


src/api/types/generated-schema.ts
```diff
@@ -203,9 +203,6 @@ export interface paths {
   "/api/v1/seedbank/values/all": {
     post: operations["listAllFieldValues"];
   };
-  "/api/v1/seedbank/values/storageLocation/{facilityId}": {
-    get: operations["getStorageLocations"];
-  };
```
karthikbtf, line 208: not sure why i updated the typespec but .. it's ok.. 🤷

PR #1200, opened 2023-02-15
SW-2946 Make modal bigger to avoid button text overflow

constanzauanini:
Screenshot 2023-02-15 at 12 17 02

We could also change the text of the button, I don't know what is the preference here? @chudo76


chudo76:
> We could also change the text of the button, I don't know what is the preference here? @chudo76

Yes, we can change to "Get Count".

(Ideally, all buttons have a minimum height instead of a fixed height so that even if we have unexpected long labels (with localization, or user's browser text settings, for example) the button grows taller. It will be a visual bug with the button looking weird, but it would be better than not being able to read the button text.)

src/components/accession2/edit/CalculatorModal.tsx
```diff
@@ -154,13 +154,15 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
...
           <Grid item xs={12} textAlign='left'>
             <Box display='flex' textAlign='left' alignItems='end'>
-              <Textfield
-                label={strings.TOTAL_WEIGHT}
-                id='remainingQuantity'
-                onChange={(value) => onChangeRemainingQuantity(Number(value))}
-                type='text'
-                value={record.remainingQuantity?.quantity}
-              />
+              <Box width='600px'>
```
karthikbtf, line 157: @constanzauanini  this doesn't work in mobile view, extends beyond the screen.
I agree with @chudo76 we should handle this at the button level with a minimum height.. not make the modal bigger.
Also yeah if Get Count solves this in the near term (even with spanish).. we can certainly revisit this later.. but min-height would be a one-time fix that works for all..

constanzauanini: I will change the text for this button, and then will update the button component to have a min-height instead of height

karthikbtf: is that a web-components change or just in this one place?

karthikbtf (approved):

src/components/accession2/edit/CalculatorModal.tsx
constanzauanini: Height is set in web-components, for each button size (small, medium, etc)

PR #1201, opened 2023-02-15
SW-2969 Fix strange scroll behavior in time zone selector

constanzauanini:
This behavior was happening with new users because when timezone feature is activated we initialize the timezone and set 'updated' value to 'true' to show the timezone notification. The problem was that after updating the timezone, the 'reloadUser' was called, and the validation to show or not the notification was being executed again, and because the 'updated' value was always true (because it comes from a chached promise) the user was being reloaded all the time (calling reloadUser), making the options for timeZones reload all the time because we need the selected timezone for the user.

The fix was to set 'updated' value to true the first time the timezone is set, and after that make the cached promise null, so the 'updated' value can be refreshed (to false, after timezone is already set)

### DIFF
--- a/src/services/UserService.ts
+++ b/src/services/UserService.ts
@@ -26,6 +26,7 @@ export type UpdateOptions = {
 type CachedTimeZone = {
   cachedOn: number;
   promise: Promise<InitializedTimeZone> | null;
+  updated: boolean;
 };
 
 // endpoint
@@ -39,6 +40,7 @@ const httpCurrentUser = HttpService.root(CURRENT_USER_ENDPOINT);
 const cachedTimeZone: CachedTimeZone = {
   cachedOn: 0,
   promise: null,
+  updated: false,
 };
 
 /**
@@ -102,9 +104,11 @@ const initializeTimeZone = async (user: User, timeZone: string): Promise<Initial
     if (response.requestSucceeded) {
       initializedTimeZone.updated = true;
       initializedTimeZone.timeZone = timeZone;
+      cachedTimeZone.updated = true;
     }
   } else {
     initializedTimeZone.timeZone = user.timeZone;
+    cachedTimeZone.updated = false;
   }
 
   return initializedTimeZone;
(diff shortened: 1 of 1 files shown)


src/services/UserService.ts
```diff
@@ -118,7 +122,10 @@ const getInitializedTimeZone = (user: User, timeZone: string): Promise<Initializ
   if (cachedTimeZone.promise) {
     // clear the promise if this was cached before user acknowledged the time zone
     const { timeZoneAcknowledgedOnMs } = CachedUserService.getUserPreferences();
-    if (!isNaN(timeZoneAcknowledgedOnMs) && timeZoneAcknowledgedOnMs > cachedTimeZone.cachedOn) {
+    if (
+      (!isNaN(timeZoneAcknowledgedOnMs) && timeZoneAcknowledgedOnMs > cachedTimeZone.cachedOn) ||
+      cachedTimeZone.updated
```
karthikbtf, line 127: this brings back a potential double setting of time zone.. no point in having this cached promise..
ugh this feature is getting messy or we need to rethink how we do this..

karthikbtf (approved):

src/services/UserService.ts
karthikbtf: might as well remove this cached promise.. maybe create a ticket to handle double setting of user time zone or something like that.. for later

constanzauanini: Yes :(

constanzauanini: But at least when timezone is already set and updated is false we are not setting time zone again

karthikbtf: the user and org notifications are initialized at the same time and both will race to set the time zone.
i'm very sad about this change.. like why even make this change when we know it's pointless.
might as well remove it

karthikbtf: another idea.. if you think it helps.. we can make the org notification take in the user time zone as a prop
and until that is valid.. the org notification renders nothing?

karthikbtf: > I mean that only with a new user the "updated" will be true, so the timezone will be initialized two times. If user has the timezone already set, it will execute it just one time, org will use the cached one (because updated will be false)

Got it.. but the two times is also not desirable :) 
This is definitely better than the scrolling issue.. so let's get this in. We can discuss offline later if we want to clean this up some more.

PR #1206, opened 2023-02-15
SW-2944 / SW-2961 / SW-2975 Add/Edit/Delete storage locations 

karthikbtf:
- in new/edit seed bank views
- the edit storage locations keeps track of full state of storage locations (instead of incremental diffs), this allows not keeping
  track of added/created/delete for new/existing seedbanks
- editing a seedbank looks at differences between existing storage locations and edited storage locations (using lodash utilities)
- uses batch API to create/update/delete storage locations
- did this in one PR since most of the code is just the plumbing/framework.. delete vs create vs edit changes are minimal
- user cannot delete locations with active accessions
- user cannot create locations that already exist

Terraware App 2023-02-15 16-01-37

Terraware App 2023-02-15 15-59-19

Terraware App 2023-02-15 15-19-21

Terraware App 2023-02-15 15-18-56

Terraware App 2023-02-15 15-18-43

Terraware App 2023-02-15 15-18-26

Terraware App 2023-02-15 15-15-23


src/types/Facility.ts
```diff
@@ -15,3 +18,12 @@ export interface Facility {
...
 }
+
+export const DEFAULT_STORAGE_LOCATIONS = (): string[] => [
+  strings.FREEZER_1,
+  strings.FREEZER_2,
+  strings.FREEZER_3,
+  strings.REFRIGERATOR_1,
+  strings.REFRIGERATOR_2,
+  strings.REFRIGERATOR_3,
+];
```
karthikbtf, line 29: This list is to render the default list of locations in the 'create new seed bank' flow.

sgrimm:
> this is needed because when we add a seedbank, there are no locations to start with

As part of the BE changes for this, I added a `storageLocationNames` field to the facility creation API payload, so you should be able to create the seed bank with its initial set of locations.

karthikbtf:
> > this is needed because when we add a seedbank, there are no locations to start with
> 
> As part of the BE changes for this, I added a `storageLocationNames` field to the facility creation API payload, so you should be able to create the seed bank with its initial set of locations.

Ok good, I missed that.. I will update this PR to not create the storage locations after the fact. Thanks!!

src/strings/strings-en.ts
```diff
@@ -878,6 +885,8 @@ export const strings = {
   STORING: 'Storing',
   STORING_START_DATE: 'Storing Start Date',
   STRATIFICATION: 'Stratification',
+  SUB_LOCATION_DETAILS: 'Sub-Location Details',
+  SUB_LOCATION_EXISTS: 'This storage location already exists',
```
sgrimm, line 889: Should this use "sub-location" instead of "storage location"?

karthikbtf: Oops, yes, good catch. Will update.

karthikbtf:
Updated the error message:

Terraware App 2023-02-15 16-01-37

@chudo76 thoughts on this empty state messaging?

Terraware App 2023-02-15 15-59-19

sgrimm:
I don't have an opinion on the React plumbing here, but the interactions with the API seem reasonable to me.

karthikbtf:
> I don't have an opinion on the React plumbing here, but the interactions with the API seem reasonable to me.

Thanks Steve for the review, and flagging those issues.

chudo76:
> @chudo76 thoughts on this empty state messaging?
> 
> Terraware App 2023-02-15 15-59-19

Looks good. Shortened message and centered text to match position similar to Seeds > Accessions > Viability Tests message.
image

Minor change request (which I realize is not part of this task): Have the same three-column layout for ` Name | Description | Time Zone ` for Edit mode to match View mode.

Everything else LGTM!

chudo76 (approved):

src/components/SeedBank/StorageLocations.tsx
```diff
@@ -36,30 +52,122 @@ export default function SubLocations({ seedBankId }: SubLocationsProps): JSX.Ele
...
+      });
     }
-  }, [seedBankId, activeLocale]);
 
-  if (!storageLocations.length) {
+    return topBarButtons;
+  };
+
+  const onStorageLocationSelected = (location: PartialStorageLocation, fromColumn?: string) => {
+    if (fromColumn === 'name') {
```
constanzauanini, line 80: If you are only calling the 'onStorageLocationSelected' from the name column, this check wouldn't be necessary, but is ok to leave it

src/components/SeedBank/StorageLocations.tsx
```diff
@@ -36,30 +52,122 @@ export default function SubLocations({ seedBankId }: SubLocationsProps): JSX.Ele
...
+
+  const getTopBarButtons = () => {
+    const topBarButtons: TopBarButton[] = [];
+    if (selectedRows.length) {
+      topBarButtons.push({
+        buttonType: 'destructive',
+        buttonText: strings.DELETE,
+        // we don't want to delete locations that have active accessions
+        disabled: !selectedRows.some((location) => !location.activeAccessions),
+        onButtonClick: () => deleteStorageLocations(selectedRows.filter((location) => !location.activeAccessions)),
```
constanzauanini, line 72: Should be show a message explaining why the active accessions are not deleted?

constanzauanini (approved):
Looks good! Some comments but nothing blocking

src/components/SeedBank/StorageLocations.tsx
karthikbtf: Current plan (separate jira) is to disable the checkbox here and in the inventory list - we don't allow withdraw on batches with 0 ready quantity. Same treatment here.

karthikbtf:
> Looks good! Some comments but nothing blocking

Thanks! Will make those changes and Chudo's change.

PR #1207, opened 2023-02-16
SW-2970 Don't redirect user to error-page after login 

karthikbtf:
- on logout, we have a race condition that sometimes attempts to fetch org data.. fails.. and sends user to error page
- this same flow redirects user to login page with redirectUri as the last visited url in the app (error page)
- after login user is landed back on the redirectUri / error page (there was no real error)
- fix is to not use the error page as a redirect uri


src/services/axios/index.ts
```diff
@@ -28,7 +28,8 @@ axios.interceptors.response.use(
   (response) => response,
   (error) => {
     if (error.response && error.response.status === 401) {
-      const redirect = encodeURIComponent(location.href);
+      const href = location.href.match('/error') ? location.origin : location.href;
```
constanzauanini, line 31: Does location origin allways exist?

constanzauanini (approved):

src/services/axios/index.ts
karthikbtf: Yes, it's a standard property on the location instance.
 https://developer.mozilla.org/en-US/docs/Web/API/Location/origin

constanzauanini: Excellent, thanks!

PR #1208, opened 2023-02-16
Fix button height on accession detail

constanzauanini:
- Buttons will take all the space of their container now (because of web-components change). 
- Add a container with height to edit button in accession detail to limit its height


karthikbtf (approved):
good catch

PR #1210, opened 2023-02-16
Don't allow users enter decimals for the seed count

constanzauanini:
This require the new web-components version to work

### DIFF
--- a/src/components/accession2/edit/CalculatorModal.tsx
+++ b/src/components/accession2/edit/CalculatorModal.tsx
@@ -150,7 +150,7 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
               onChange={(value) => onChange('subsetCount', value)}
               type='text'
               value={record.subsetCount}
-              disableDecimals={true}
+              disabledCharacters={[',', '.']}
             />
           </Grid>
           <Grid item xs={12} textAlign='left'>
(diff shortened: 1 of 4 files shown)


charlie-tf (approved):

src/components/accession2/edit/CalculatorModal.tsx
```diff
@@ -150,6 +150,7 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
               onChange={(value) => onChange('subsetCount', value)}
               type='text'
               value={record.subsetCount}
+              disableDecimals={true}
```
karthikbtf, line 153: we could use type=number and set the step size to enable/disable decimals?

constanzauanini: I tested stepSize, but allows inputing the ',' and '.' and that is what we don't want. The step just work with the arrows

src/components/accession2/edit/CalculatorModal.tsx
```diff
@@ -150,6 +150,7 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
               onChange={(value) => onChange('subsetCount', value)}
               type='text'
               value={record.subsetCount}
+              disabledCharacters={[',', '.']}
```
karthikbtf, line 153: do we need '-' here and the other place?

karthikbtf (approved):
thanks lgtm

src/components/accession2/edit/CalculatorModal.tsx
constanzauanini: Yes, we probably don't want negatives quantities haha

karthikbtf (commented):
🎉

PR #1212, opened 2023-02-17
SW-2816 / SW-2980 Notifications service refactor + bugfix

karthikbtf:
- moved/cleanedup api for notifications to a service
- updated client code
- fixed bug where we were making redundant notifications requests


constanzauanini (approved):
Looks good!

PR #1213, opened 2023-02-17
SW-2739 Disable call-to-action buttons while processing

constanzauanini:
Testing in create seedbank. 
Still have to add it to others pages

### DIFF
--- a/src/components/common/FormBottomBar.tsx
+++ b/src/components/common/FormBottomBar.tsx
@@ -42,7 +42,7 @@ export interface Props {
   cancelButtonText?: string;
   saveButtonText?: string;
   saveDisabled?: boolean;
-  handleSave: () => void;
+  onSave: () => void;
 }
 
 export default function FormBottomBar({
@@ -52,7 +52,7 @@ export default function FormBottomBar({
   cancelButtonText,
   saveButtonText,
   saveDisabled,
-  handleSave,
+  onSave,
 }: Props): JSX.Element {
   const { isMobile, isDesktop } = useDeviceInfo();
   const classes = useStyles({ isMobile, isDesktop });
@@ -77,7 +77,7 @@ export default function FormBottomBar({
         id={saveID || 'saveBottomBar'}
         size='medium'
         label={saveButtonText || strings.SAVE}
-        onClick={handleSave}
+        onClick={onSave}
         className={classes.button}
         disabled={saveDisabled}
       />
--- a/src/components/common/FormBottomBar.tsx
+++ b/src/components/common/FormBottomBar.tsx
@@ -39,20 +39,20 @@ export interface Props {
   cancelID: string;
   saveID: string;
   onCancel: () => void;
-  onSave: () => void;
   cancelButtonText?: string;
   saveButtonText?: string;
   saveDisabled?: boolean;
+  handleSave: () => void;
 }
 
 export default function FormBottomBar({
   cancelID,
   saveID,
   onCancel,
-  onSave,
   cancelButtonText,
   saveButtonText,
   saveDisabled,
+  handleSave,
 }: Props): JSX.Element {
   const { isMobile, isDesktop } = useDeviceInfo();
   const classes = useStyles({ isMobile, isDesktop });
@@ -77,7 +77,7 @@ export default function FormBottomBar({
         id={saveID || 'saveBottomBar'}
         size='medium'
         label={saveButtonText || strings.SAVE}
-        onClick={onSave}
+        onClick={handleSave}
         className={classes.button}
         disabled={saveDisabled}
       />
--- a/src/components/common/PageForm.tsx
+++ b/src/components/common/PageForm.tsx
@@ -35,7 +35,7 @@ export default function PageForm(props: PageFormProps): JSX.Element {
       <Box className={className} paddingBottom={hideEdit ? theme.spacing(4) : theme.spacing(isMobile ? 25 : 15)}>
         {children}
       </Box>
-      {!hideEdit && <FormBottomBar handleSave={handleSave} {...bottomBarProps} />}
+      {!hideEdit && <FormBottomBar onSave={handleSave} {...bottomBarProps} />}
     </>
   );
 }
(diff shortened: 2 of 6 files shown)


karthikbtf:
I have a suggestion.. do this inside the FormBottomBar...
create a new function for the save button
```
const handleSave = async () => {
  setProcessing(true); <-- while true, we can show a skrim with spinner (similar to batch withdraw on withdraw button clicked) and maybe disable the button.. the skrim will absorb all events anyways..
  await onSave(); // <-- we want all onSaves to be async 
  setProcessing(false);
}; 
```

karthikbtf:
> I have a suggestion.. do this inside the FormBottomBar... create a new function for the save button
> 
> ```
> const handleSave = async () => {
>   setProcessing(true); <-- while true, we can show a skrim with spinner (similar to batch withdraw on withdraw button clicked) and maybe disable the button.. the skrim will absorb all events anyways..
>   await onSave(); // <-- we want all onSaves to be async 
>   setProcessing(false);
> }; 
> ```

Just realized that if we want to show the skrim, we have to this inside the PageForm (since it hosts the children).. and pass the `handleSave` as `onSave` to the FormBottomBar. Either way, the general premise is the same.

src/components/common/PageForm.tsx
```diff
@@ -17,16 +18,24 @@ export interface PageFormProps {
...
+
+  const handleSave = async () => {
+    setProcessing(true);
+    await onSave(); // <-- we want all onSaves to be async
+    setProcessing(false);
+  };
 
   return (
     <>
+      {processing && <BusySpinner withSkrim={true} />}
```
karthikbtf, line 34: awesome!!! thanks for considering the suggestion

src/components/common/PageForm.tsx
```diff
@@ -17,16 +18,24 @@ export interface PageFormProps {
...
 export default function PageForm(props: PageFormProps): JSX.Element {
-  const { children, className, hideEdit, ...bottomBarProps } = props;
+  const { children, className, hideEdit, onSave, ...bottomBarProps } = props;
   const theme = useTheme();
   const { isMobile } = useDeviceInfo();
+  const [processing, setProcessing] = useState(false);
+
+  const handleSave = async () => {
+    setProcessing(true);
+    await onSave(); // <-- we want all onSaves to be async
```
karthikbtf, line 28: thanks..
in most cases onSave will redirect user to some other location and this component will unmount.. react might complain on the setProcessing(false) saying component was already unmounted.. 
we could live with that? unless there's some way to detect that..

karthikbtf (commented):
thanks for this approach @constanzauanini, appreciate it!
btw, we should remove the skrim/spinner from the batch/withdraw flow (since pageform handles it now)..
Chudo feedback?
on a related note, he wants us to catch user navigation without save/cancel and prompt user to save changes.. for another effort but just fyi.

src/components/common/PageForm.tsx
constanzauanini: Hmm that's true, I think we can live with that for now.

constanzauanini:
> thanks for this approach @constanzauanini, appreciate it! btw, we should remove the skrim/spinner from the batch/withdraw flow (since pageform handles it now).. Chudo feedback? on a related note, he wants us to catch user navigation without save/cancel and prompt user to save changes.. for another effort but just fyi.

Oh, yes, I will remove it.
@chudo76 do you want to take a look? This will apply to all forms (accesions, withdrawals, seedbanks, inventory, my account, person, planting sites, etc)

src/components/common/FormBottomBar.tsx
```diff
@@ -39,20 +39,20 @@ export interface Props {
   cancelID: string;
   saveID: string;
   onCancel: () => void;
-  onSave: () => void;
   cancelButtonText?: string;
   saveButtonText?: string;
   saveDisabled?: boolean;
+  handleSave: () => void;
```
karthikbtf, line 45: this can still be onSave.. no need to push the name change from the other place in here

src/components/common/PageForm.tsx
```diff
@@ -17,16 +18,24 @@ export interface PageFormProps {
...
+  };
 
   return (
     <>
+      {processing && <BusySpinner withSkrim={true} />}
       <Box className={className} paddingBottom={hideEdit ? theme.spacing(4) : theme.spacing(isMobile ? 25 : 15)}>
         {children}
       </Box>
-      {!hideEdit && <FormBottomBar {...bottomBarProps} />}
+      {!hideEdit && <FormBottomBar handleSave={handleSave} {...bottomBarProps} />}
```
karthikbtf, line 38: `onSave={handleSave}`

karthikbtf (approved):
nice 👍

chudo76:
Nice, I like the interaction. Is it possible to use the same "scrim"/page overlay as the one we use for delete accession warning dialog box (first example I could find in the product)?

It's a white-ish overlay instead of a green one.
image

karthikbtf:
Looks good... noticed that we need to confirm that callbacks are async. For example the inventory withdraw no longer shows the skrim when the withdraw is processing. Can we take a look/fix?

karthikbtf:
> Looks good... noticed that we need to confirm that callbacks are async. For example the inventory withdraw no longer shows the skrim when the withdraw is processing. Can we take a look/fix?

If you can. test all the page-form flows.

constanzauanini:
> > Looks good... noticed that we need to confirm that callbacks are async. For example the inventory withdraw no longer shows the skrim when the withdraw is processing. Can we take a look/fix?
> 
> If you can. test all the page-form flows.

I reviewed them all, in the select purpose and select batches steps we can't make the function async because we don't call any api

karthikbtf:
> > > Looks good... noticed that we need to confirm that callbacks are async. For example the inventory withdraw no longer shows the skrim when the withdraw is processing. Can we take a look/fix?
> > 
> > 
> > If you can. test all the page-form flows.
> 
> I reviewed them all, in the select purpose and select batches steps we can't make the function async because we don't call any api

Got it. Thanks for fixing/checking/testing.

karthikbtf (approved):

PR #1219, opened 2023-02-17
Fix e2e failures due to Node17+ upgrade in github actions

karthikbtf:
localhost was 
See https://github.com/cypress-io/github-action/issues/534#issuecomment-1278776829 (thanks @charlie-tf )

```
The default webpack configuration's host value was changed from localhost to 127.0.0.1 to support the Node 17+ changes with how DNS names are resolved. Addressed in https://github.com/cypress-io/cypress/pull/21430.
```

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,7 +1,7 @@
 import { defineConfig } from 'cypress';
 
 export default defineConfig({
-  screenshotOnRunFailure: false,
+  screenshotOnRunFailure: true,
   video: false,
   watchForFileChanges: false,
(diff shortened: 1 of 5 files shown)


charlie-tf (approved):
🎉

karthikbtf:
See https://github.com/cypress-io/github-action/issues/534#issuecomment-1278776829 (thanks @charlie-tf )

The default webpack configuration's host value was changed from localhost to 127.0.0.1 to support the Node 17+ changes with how DNS names are resolved. Addressed in https://github.com/cypress-io/cypress/pull/21430.

PR #1225, opened 2023-02-21
SW-2835 unit tests for HttpService

karthikbtf:
- moved service tests under src/services/test
- need to refactor OrganizationService.test.js will do that separately


charlie-tf (approved):
Awesome! Thanks so much for adding these!

PR #1227, opened 2023-02-21
SW-3004 Remove opt-in flag/code for time zones

karthikbtf:
- also refactor by adding a type to feature names, to catch issues
- fix bug in nurseries/planting-sites where the timezone edit view was not matching the read view (this was fixed for seedbanks earlier)


src/features.ts
```diff
@@ -1,8 +1,10 @@
 import env from 'src/utils/useEnvironment';
 import { CachedUserService } from 'src/services';
 
+export type FeatureName = 'Show Production View' | 'Weight units' | 'Locale selection' | 'Storage locations';
```
charlie-tf, line 4: Nice

PR #1229, opened 2023-02-21
zSW-3003 Refactor org service test, introduce mock utilities for http service

karthikbtf:
- various services should now not mock axios (that's handed in http service tests)
- services can directly mock http service responses
- introduced mock utilities to mock the functions (we would need to refactor http service without the 'root' function for a less hacky utility experience)
- added an empty watchman config if anyone is using watchman to run tests
- note: initialize watchman with `watchman watch <pwd>`


src/services/test/OrganizationService.test.ts
```diff
@@ -0,0 +1,54 @@
...
+          statusCode: 500,
+        }),
+    });
+
+    const { organizations, error, requestSucceeded } = await OrganizationService.getOrganizations();
+
+    expect({ organizations, error, requestSucceeded }).toEqual({
+      organizations: [],
+      // fake an error for now to test github action
+      error: 'GenericErrorr',
```
karthikbtf, line 50: this is deliberate.. to see if gh action fails on unit tests

constanzauanini (approved):

PR #1230, opened 2023-02-21
SW-2624 Use New Filters in Species Table

charlie-tf:
image
image
image


src/components/Species/index.tsx
```diff
@@ -276,6 +292,62 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
...
+              type: 'Exact',
+              values: [selectedOrganization.id],
+            },
+          ],
+        },
+        sortOrder: [{ field: 'id', direction: 'Ascending' }],
+        count: 1000,
+      };
+
+      const data = await SearchService.search(searchParams);
```
karthikbtf, line 328: It's likely I missed this before.. there's an endpoint that retrieves values for dropdowns. See how we do this for accessions table https://github.com/terraware/terraware-web/blob/main/src/services/SeedBankService.ts#L142 may need some refactor to move it out into search.

src/components/Species/index.tsx
```diff
@@ -750,12 +805,40 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
...
+              }}
+              transformOrigin={{
+                vertical: 'top',
+                horizontal: 'center',
+              }}
+              className={classes.popoverContainer}
+            >
+              <FilterGroup
+                initialFilters={filters}
+                fields={filterColumns}
```
karthikbtf, line 828: thoughts on hiding (or marking the dropdowns as disabled) if there are no values to select from? maybe not an issue in production.. in staging we have orgs with dummy species.. a lot of the filter dropdowns have no values to select from... ofc the data is mostly test data.

karthikbtf (approved):
looks good, some comments but totally fine if you want to address separately (if they need addressing)

src/components/Species/index.tsx
charlie-tf: I'll put that as a followup! Thanks

src/components/Species/index.tsx
```diff
@@ -47,6 +53,9 @@ import { DropdownItem, SortOrder } from '@terraware/web-components';
 import { useLocalization, useOrganization } from 'src/providers/hooks';
 import { PillList, PillListItem } from '@terraware/web-components';
 import { isTrue } from 'src/utils/boolean';
+import FilterGroup, { FilterField } from '../common/FilterGroup';
+import { FieldOptionsMap } from '../../services/NurseryWithdrawalService';
+import SearchService from '../../services/SearchService';
```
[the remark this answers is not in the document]
charlie-tf: oops yes

src/components/Species/index.tsx
charlie-tf: That would make sense. I'll look at doing that

karthikbtf (approved):

karthikbtf:
Looks good in vercel, thanks so much 🙏

PR #1232, opened 2023-02-21
SW-3002 Fix back to link effective area

karthikbtf:
- content alone should be link-able
- was using a flex property that gave it full width, mostly for the align items property
- inline-flex was giving it a bit more height, not suitable
- fix width to fit content


charlie-tf (approved):

charlie-tf:
looks good in vercel, too

PR #1235, opened 2023-02-22
SW-3005 Remove opt-in so that imperial units is ready for production deployment

constanzauanini:


src/components/MyAccount/index.tsx
```diff
@@ -95,7 +95,6 @@ const MyAccountContent = ({
   const snackbar = useSnackbar();
   const contentRef = useRef(null);
   const timeZonesEnabled = isEnabled('Timezones');
-  const weightUnitsEnabled = isEnabled('Weight units');
```
karthikbtf, line 98: i can merge main into my branch after you push.. and remove the isEnabled import

karthikbtf (approved):

src/components/MyAccount/index.tsx
constanzauanini: Ok! I will merge it then

PR #1238, opened 2023-02-22
Don't update collected/received value when changing other field

constanzauanini:
Now when setting an invalid value on the datepicker of accession / collected date or received date, and then changing focus to other field, the value of the datepicker is reset. 
This was happening because when the record changed one of its fields, the collected date and received date where set to the record value that is null (because the invalid value doesn't change the record value) 
So the fix to avoid "reseting" the value, is not refreshing the temporal values when the record changes.

### DIFF
--- a/src/components/accession2/properties/CollectedReceivedDate2.tsx
+++ b/src/components/accession2/properties/CollectedReceivedDate2.tsx
@@ -28,13 +28,6 @@ export default function CollectedReceivedDate2({ onChange, record, type, validat
     receivedDate: record.receivedDate,
   });
 
-  useEffect(() => {
-    setDates({
-      collectedDate: record.collectedDate,
-      receivedDate: record.receivedDate,
-    });
-  }, [record]);
-
   const datePickerStyle = {
     '.MuiFormControl-root': {
       width: '100%',


karthikbtf (approved):
good catch.. wondering why we added it.. hope it doesn't fix some other weird scenario

constanzauanini:
> good catch.. wondering why we added it.. hope it wasn't to fix some other weird scenario..

Yes, I hope it wasn't too

karthikbtf:
> > good catch.. wondering why we added it.. hope it wasn't to fix some other weird scenario..
> 
> Yes, I hope it wasn't too

Would have been good to confirm..

karthikbtf:
> > > good catch.. wondering why we added it.. hope it wasn't to fix some other weird scenario..
> > 
> > 
> > Yes, I hope it wasn't too
> 
> Would have been good to confirm..

Ah yes if you edit and cancel, the edited value sticks.. we should fix this

karthikbtf:
> > > > good catch.. wondering why we added it.. hope it wasn't to fix some other weird scenario..
> > > 
> > > 
> > > Yes, I hope it wasn't too
> > 
> > 
> > Would have been good to confirm..
> 
> Ah yes if you edit and cancel, the edited value sticks.. we should fix this

created a bug https://terraformation.atlassian.net/browse/SW-3025

constanzauanini:
> > > > > good catch.. wondering why we added it.. hope it wasn't to fix some other weird scenario..
> > > > 
> > > > 
> > > > Yes, I hope it wasn't too
> > > 
> > > 
> > > Would have been good to confirm..
> > 
> > 
> > Ah yes if you edit and cancel, the edited value sticks.. we should fix this
> 
> created a bug https://terraformation.atlassian.net/browse/SW-3025

Thanks for fixing it!

karthikbtf:
> > > > > > good catch.. wondering why we added it.. hope it wasn't to fix some other weird scenario..
> > > > > 
> > > > > 
> > > > > Yes, I hope it wasn't too
> > > > 
> > > > 
> > > > Would have been good to confirm..
> > > 
> > > 
> > > Ah yes if you edit and cancel, the edited value sticks.. we should fix this
> > 
> > 
> > created a bug https://terraformation.atlassian.net/browse/SW-3025
> 
> Thanks for fixing it!

Np.. actually the problem was in all the fields, not just dates. The cleanup in this PR was correct, it didn't introduce any new bug.

PR #1240, opened 2023-02-22
SW-3017 Allow decimals in total weight / calculator 

karthikbtf:
- use number type inputs so we can avoid text<-->number processing issues
- also show validation so we don't send bad data to BE and error out

Terraware App 2023-02-22 15-13-09

Terraware App 2023-02-22 14-37-55

### DIFF
--- a/src/components/accession2/edit/CalculatorModal.tsx
+++ b/src/components/accession2/edit/CalculatorModal.tsx
@@ -60,8 +60,8 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
     }
 
     setTotalWeightError('');
-    const totalWeightUnits = record.remainingQuantity?.units ?? 'Seeds';
-    if (!record.remainingQuantity?.quantity || totalWeightUnits === 'Seeds') {
+    const totalWeightUnits = record.remainingQuantity?.units;
+    if (!record.remainingQuantity?.quantity || !totalWeightUnits || totalWeightUnits === 'Seeds') {
       setTotalWeightError(strings.REQUIRED_VALUE_AND_UNITS_FIELD);
       hasErrors = true;
     }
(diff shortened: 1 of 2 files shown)


src/components/accession2/edit/CalculatorModal.tsx
```diff
@@ -147,29 +187,36 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
             <Textfield
               label={strings.SUBSET_COUNT}
               id='subsetCount'
-              onChange={(value) => onChange('subsetCount', value)}
-              type='text'
+              onChange={(value) => onChangeSubsetCount(value as number)}
+              type='number'
+              min={0}
+              disabledCharacters={['.', '-']}
```
sgrimm, line 193: Will this do the right thing if you're using a locale that uses comma as the decimal separator?

karthikbtf: oops good catch, this list had a comma not sure why I removed it.. will bring it back

src/components/accession2/edit/CalculatorModal.tsx
```diff
@@ -33,17 +33,40 @@ export default function CalculatorModal(props: CalculatorModalProps): JSX.Elemen
...
-    return true;
+
+    setSubsetCountError('');
+    if (!record.subsetCount) {
+      setSubsetCountError(strings.REQUIRED_FIELD);
+      hasErrors = true;
+    }
+
+    setTotalWeightError('');
+    const totalWeightUnits = record.remainingQuantity?.units ?? 'Seeds';
```
constanzauanini, line 63: Why are we settings 'seeds' here and not just undefined if no unit is selected? Then we could just check if totalWeightUnits.  I don't think it can we undefined anyways

constanzauanini (approved):

src/components/accession2/edit/CalculatorModal.tsx
karthikbtf: yes, it defaults to Seeds.. i possibly don't understand this code very well

karthikbtf: changing this check to `(!totalWeightUnits || totalWeightUnits === 'Seeds')` even though it can't be undefined theoretically but according to the typing it can be

PR #1242, opened 2023-02-23
SW-3025 fix edit accession view stickiness upon cancel

karthikbtf:
- pretty much all editable fields were sticking
- the open property on the modal is not working to our advantage here
- also using open for conditional rendering which is really the best solution here
- i propose we do similar conditional rendering in other places where we have stickiness instead of so much cleanup code with little ROI


constanzauanini (approved):
Thanks! I think we did this approach in other parts, definitely is better.

PR #1243, opened 2023-02-23
Reports service to abstract APIs

constanzauanini:
Add a report service with all the apis
Files and photos management may be reviewed when using it

### DIFF
--- a/src/services/ReportService.ts
+++ b/src/services/ReportService.ts
@@ -1,0 +1,361 @@
+import { paths } from 'src/api/types/generated-schema';
+import { Id } from 'src/types/Id';
+import { ListReport, Report, ReportFile, ReportPhoto } from 'src/types/Report';
+import HttpService, { Response } from './HttpService';
+import { getPromisesResponse } from './utils';
+
+/**
+ * Report related services
+ */
+
+const REPORTS_ENDPOINT = '/api/v1/reports';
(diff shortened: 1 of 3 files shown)


src/services/ReportService.ts
```diff
@@ -0,0 +1,361 @@
...
+      '{reportId}': reportId.toString(),
+      '{fileId}': fileId.toString(),
+    },
+  });
+};
+
+/**
+ * Exported functions
+ */
+const ReportService = {
```
karthikbtf, line 342: lgtm.. would it be worth splitting this into ReportService, ReportLockingService and ReportPhotosService? or is that overkill.. i guess my thoughts are mostly coming from size of service (maybe that's not a valid issue).. cc @charlie-tf @constanzauanini

karthikbtf: not requesting any changes.. was mostly a question to hear y'alls thoughts

constanzauanini: Yes, I don't think it's completely necessary, but if we prefer small files I can split it

charlie-tf: I'm kinda on the side of keeping it all in one. Then there's just one service to interact with. But either way would completely work 🤷

karthikbtf: Thanks Constanza/Charlie.. one files sounds good 👍

charlie-tf (approved):
lgtm!

PR #1244, opened 2023-02-23
SW-3022 / SW-3023 Add/edit for inventory and reassign should send use…

karthikbtf:
…r back to context

- add inventory takes user to species view after new batch is addded
- cancel on add inventory goes back to inventory list
- reassign cancel/save will take user back to originating view, which could be the withdrawal log or the withdrawal detail
- leveraging a url query param for the reassign redirect


src/components/NurseryWithdrawals/NurseryReassignment.tsx
```diff
@@ -258,7 +256,6 @@ export default function NurseryReassignment(): JSX.Element {
             minWidth='fit-content'
             ref={contentRef}
           >
-            {reassigning && <BusySpinner withSkrim={true} />}
```
constanzauanini, line 261: oops, i missed this one

PR #1245, opened 2023-02-23
SW-3010 Use AND Logic Across Different Filters

charlie-tf:
It seems like we're currently using the "or" logic across different filters but we should be using the "and" logic. In other words, if a user filters by the species Koa and the destination Site A on the withdrawal log page, then the results should show all withdrawals with Koa as the species AND Site A as the destination.

The only time we use the "or" logic should be within a single filter. For example, if a user filters by the status Awaiting Check-In, Drying, and In Storage then we should show all results whose status is any of those options.

* This has been fixed on the Withdrawal Log filters. It was already being done correctly in the Species and Accessions filters


karthikbtf (approved):
good find

PR #1248, opened 2023-02-24
SW-2729 Species service

constanzauanini:
Create a Species service from api/species code
Still have to replace all code with the new service


src/api/species/species.ts
```diff
@@ -9,45 +9,6 @@ import { GetUploadStatusResponsePayload, ResolveResponse, UploadFileResponse } f
  * surfaced to the caller via the requestSucceeded or error field.
  */
 
-const GET_SPECIES_ENDPOINT = '/api/v1/species';
```
karthikbtf, line 12: remove this file and directory?

karthikbtf (approved):
looks good to me.. what are your thoughts so far? was it tedious or useful? any unit test worthy functions that might need a follow up ticket?

src/api/species/species.ts
constanzauanini: Yes, I didn't replace all the functions yet, that's why I didn't delete the file, but I will do it and remove it after that

constanzauanini:
> looks good to me.. what are your thoughts so far? was it tedious or useful? any unit test worthy functions that might need a follow up ticket?

It was not tedious, I like that the requestSucceded/error managment is much better now, we are not repeating code for that. Also adding parameters or replace parts of the url is easier

src/types/Species.ts
```diff
@@ -149,3 +151,5 @@ export enum SpeciesRequestError {
   // setting up the request. In other words, there was a developer error or server outage.
   RequestFailed = 'AN_UNRECOVERABLE_ERROR_OCCURRED',
 }
+
+export type SpeciesDetails = components[typeof schemas]['SpeciesLookupDetailsResponsePayload'];
```
karthikbtf, line 155: i believe `'schemas'` might work as well.. we are using `typeof schemas` in many places.. not sure why.. a lot of it was mostly copy paste lol

src/components/Species/ImportSpeciesModal.tsx
```diff
@@ -33,12 +28,12 @@ export default function ImportSpeciesModal(props: ImportSpeciesModalProps): JSX.
...
-      resolveApi={resolveSpeciesUpload}
+      resolveApi={SpeciesService.resolveSpeciesUpload}
       uploaderTitle={strings.IMPORT_SPECIES_LIST}
       uploaderDescription={strings.IMPORT_SPECIES_LIST_DESC}
-      uploadApi={uploadSpeciesFile}
-      templateApi={downloadSpeciesTemplate}
-      statusApi={getUploadStatus}
+      uploadApi={SpeciesService.uploadSpecies}
+      templateApi={SpeciesService.downloadSpeciesTemplate}
+      statusApi={SpeciesService.getSpeciesUploadStatus}
```
karthikbtf, line 36: in the import modal component, i added if/else handling for both types of responses.. now that species is also handled.. we can cleanup the if/else handling

karthikbtf (approved):

src/types/Species.ts
constanzauanini: Yes, i copied/pasted that

karthikbtf (approved):
🙏

PR #1253, opened 2023-02-27
Report View and Edit

charlie-tf:
- create view and edit wrapper components for the report form
- create a report form and annual report form
- put a couple dummy fields in the forms
- allow transitioning from view to edit, locking the report
- warn if someone else is already editing (locked) the report
- allow saving and quitting editing
- allow report submission, with a confirmation dialog
- show pages of the report if annual

## View Report
image

### Annual Report
image image

## Edit report
image

### Takeover Editing
image

### Annual Report
image
image

### Submit Report
image

## TODO:
- Add actual report fields to form
- Validate form fields (make sure required fields are filled) on submit
- Poll the BE during editing for who has the lock; nav back to view if lock is lost
- Extract Dialogs into separate components


src/components/Reports/ReportEdit.tsx
```diff
@@ -1,6 +1,145 @@
...
+
+  const [confirmSubmitDialogOpen, setConfirmSubmitDialogOpen] = useState(false);
+
+  const gotoReportView = async (saveChanges: boolean) => {
+    if (saveChanges && report) {
+      await ReportService.updateReport(report);
+    }
+
+    // unlock the report
+    await ReportService.unlockReport(reportIdInt);
```
constanzauanini, line 45: Should we check if ReportService.updateReport succedded before unlocking?

src/components/Reports/ReportEdit.tsx
```diff
@@ -1,6 +1,145 @@
...
+  const [report, setReport] = useState<Report>();
+  useEffect(() => {
+    const getReport = async () => {
+      const result = await ReportService.getReport(reportIdInt);
+      if (result.requestSucceeded) {
+        setReport(result.report);
+      }
+    };
+
+    getReport();
```
karthikbtf, line 32: probably check for a non-empty id before calling get report?

src/components/Reports/ReportEdit.tsx
```diff
@@ -1,6 +1,145 @@
...
+      await ReportService.updateReport(report);
+      setShowAnnual(false);
+    }
+  };
+
+  const submitReport = async () => {
+    if (report) {
+      await ReportService.updateReport(report);
+      await ReportService.submitReport(reportIdInt);
+      await ReportService.unlockReport(reportIdInt);
```
karthikbtf, line 73: Will be good to have one BE API that can handle all this... often times user may click a button and close the browser.. or may lose internet connection in between API requests.. etc.
Generally, workflows consisting of multiple API should be batched in the BE and not FE. 
Not a show stopper.. we have other use-cases in our FE today (storage locations, add/remove orgs, etc.). Probably should revisit those at some point.

src/components/Reports/ReportForm.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <Typography key='thing1'>Thing 1 for report {`${report.year}-Q${report.quarter}`}</Typography>
+      <Typography key='thing2'>Thing 2 for report {`${report.year}-Q${report.quarter}`}</Typography>
+      <Textfield
+        key='testInput'
+        placeholder={'Test Input...'}
+        label='Details'
+        id='testInput'
+        type='text'
+        disabled={!editable}
+      />
```
karthikbtf, line 34: these are all test stuff yes?

src/components/Reports/ReportEdit.tsx
```diff
@@ -1,6 +1,145 @@
...
+  };
+
+  const handleBack = async () => {
+    if (report) {
+      await ReportService.updateReport(report);
+      setShowAnnual(false);
+    }
+  };
+
+  const submitReport = async () => {
```
karthikbtf, line 69: this can fail, if some other user grabbed the lock (and we haven't polled yet)

karthikbtf (commented):
nice progress, some comments/thoughts

src/components/Reports/ReportEdit.tsx
charlie-tf: Yes, I will have to add checking the response to service calls in many places.

src/components/Reports/ReportEdit.tsx
karthikbtf: sounds good, i understand this is WIP

karthikbtf (approved):
This is good for a WIP stop.

karthikbtf (approved):
Thanks for the changes!

PR #1254, opened 2023-02-27
SW-3069 localize app title

karthikbtf:
- update document's title when locale changes
- also works for initializing the title using saved locale

QXBw VGVycmF3YXJl 2023-02-27 13-42-30

### DIFF
--- a/src/AppBootstrap.tsx
+++ b/src/AppBootstrap.tsx
@@ -9,6 +9,7 @@ import {
   OrganizationProvider,
   UserProvider,
 } from 'src/providers';
+import strings from 'src/strings';
 
 const useStyles = makeStyles((theme: Theme) => ({
   spinner: {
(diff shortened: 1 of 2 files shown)


src/AppBootstrap.tsx
```diff
@@ -62,6 +63,10 @@ export default function AppBootstrap({ children }: AppBootstrapProps): JSX.Eleme
   const [selectedLocale, setSelectedLocale] = useState('en');
   const [activeLocale, setActiveLocale] = useState<string | null>(null);
 
+  useEffect(() => {
+    window.document.title = strings.APP_TITLE;
```
karthikbtf, line 67: i don't know if this is the best place, but works

karthikbtf:
> I don't think the app name will change in Spanish for example, but we still need it for Japanese or Chinese

good point, probably don't need this extra string/change for now. we can revisit later for sure.
i will close this for now.

PR #1255, opened 2023-02-28
SW-3071 Fix Can't upload species list

constanzauanini:
This bug was introduced with the new species service. Species upload file api uses organizationId instead of faciityId. This was causing the error

### DIFF
--- a/src/components/Species/index.tsx
+++ b/src/components/Species/index.tsx
@@ -14,7 +14,6 @@ import {
   getGrowthFormString,
   getSeedStorageBehaviorString,
   Species,
-  SpeciesProblemElement,
 } from 'src/types/Species';
 import TfMain from 'src/components/common/TfMain';
 import PageSnackbar from 'src/components/PageSnackbar';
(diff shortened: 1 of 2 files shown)


src/components/Species/index.tsx
```diff
@@ -455,7 +454,7 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
           searchResults?.forEach((result) => {
             speciesResults.push({
               id: result.id as number,
-              problems: result.problems as SpeciesProblemElement[],
+              problems: species.find((sp) => sp.id.toString() === (result.id as string))?.problems,
```
charlie-tf, line 457: What was happening here? Were there multiple unrelated problems getting attached to a species?

charlie-tf (approved):
Changes look good 👍

src/components/Species/index.tsx
constanzauanini: No, this never worked before because the search api doesn't include the problems in the response, so problems was always undefined. It used to work with not filtering, because we used the species array that comes from App (in the props), but when filtering an calling search it was not working. Now, we removed that difference from filtering or not, and use search api always, so we were not seeing the problems at any moment

charlie-tf: Ah makes sense. Thank you

PR #1256, opened 2023-02-28
SW-2884 Report list page

constanzauanini:
Screenshot 2023-02-28 at 12 36 54

I didn't add any report to test this, so probably this will need testing after create flow is finished.
Also, the listReports api doesn't support a query, so I didn't add the search bar. Should we use the search api for that??
Should we support filters too?


charlie-tf (approved):
Looks nice!

src/components/Reports/ReportList.tsx
```diff
@@ -1,6 +1,63 @@
...
+    { key: 'name', name: strings.REPORT, type: 'string' },
+    { key: 'status', name: strings.STATUS, type: 'string' },
+    { key: 'modifiedByName', name: strings.LAST_EDITED_BY, type: 'string' },
+    { key: 'submittedByName', name: strings.SUBMITTED_BY, type: 'string' },
+    { key: 'submittedTime', name: strings.DATE_SUBMITTED, type: 'string' },
+  ];
+
+  useEffect(() => {
+    const refreshSearch = async () => {
+      const requestId = Math.random().toString();
```
karthikbtf, line 30: probably don't need this since there's no search where we send multiple requests..

src/components/Reports/ReportList.tsx
```diff
@@ -1,6 +1,63 @@
...
+            <Typography sx={{ fontSize: '24px', fontWeight: 600 }}>{strings.REPORTS}</Typography>
+          </Grid>
+        </PageHeaderWrapper>
+        <Container
+          ref={contentRef}
+          maxWidth={false}
+          sx={{ padding: 3, borderRadius: 4, backgroundColor: theme.palette.TwClrBaseWhite }}
+        >
+          <Grid item xs={12}>
+            <Table id='reports-table' columns={columns} rows={results} orderBy='name' Renderer={ReportsCellRenderer} />
```
karthikbtf, line 57: use pre-sorted option?

src/components/Reports/TableCellRenderer.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { APP_PATHS } from 'src/constants';
+import CellRenderer, { TableRowType } from '../common/table/TableCellRenderer';
+import { RendererProps } from '../common/table/types';
+import Link from '../common/Link';
+
+export default function ReportsCellRenderer(props: RendererProps<TableRowType>): JSX.Element {
+  const { column, row, index } = props;
+
+  const createLinkToSeedBank = () => {
```
karthikbtf, line 10: rename :)

karthikbtf (approved):
some comments.. otherwise cool 👍

src/components/Reports/ReportList.tsx
constanzauanini: This is not using the search api, it is using a reports specific api

src/components/Reports/TableCellRenderer.tsx
constanzauanini: oops, thanks!

charlie-tf:
shiny ✨ 
image

src/components/Reports/ReportList.tsx
karthikbtf: Oops.. ok good point. 
Realized we don't use [locale compare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) for strings in the table comparator. For spanish we should be ok? I can create a ticket to pass locale to table..

charlie-tf:
>this will need testing after create flow is finished.

So reports are kinda odd in this respect. There isn't a create flow. Rather, reports are created automatically by the backend on a schedule. They _can_ be manually created through the admin page, however, if you are a super-admin

src/components/Reports/ReportList.tsx
constanzauanini: In the example it is comparing two strings in different locales. Table has same locale for all its rows, so why is this necessary?

karthikbtf: The example may be doing that.. locale compare uses language specific rules for comparison. Rules vary by language.
Uses Intl.Collator behind the scenes.. see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator

karthikbtf: Didn't mean to add any distraction.. not requesting any changes here right now.. just brought that up for discussion. We can discuss this offline with the team maybe?

constanzauanini: Ok! I see now, yes, we may need it

PR #1260, opened 2023-03-01
Allow Editing Actual Report Fields

charlie-tf:
- bring in the immer library, which handles deep copies of deeply nested objects
- start layout of actual report form fields
- debounce inputs to report fields, so that we aren't calling the update function too frequently
- fix report saving
image
image


src/components/Reports/ReportEdit.tsx
```diff
@@ -122,6 +123,22 @@ export default function ReportEdit(): JSX.Element {
     buttonType: 'passive',
   });
 
+  /**
+   * Report update functions
+   */
+  const updateReport = (field: string, value: any) => {
+    if (report) {
+      setReport(
+        produce((draft) => {
```
charlie-tf, line 132: Brought in this function from the `immer` library. It automatically handles creating a deep copy, keeping track of updates, and returning it to update state.

src/components/Reports/ReportEdit.tsx
```diff
@@ -122,6 +123,22 @@ export default function ReportEdit(): JSX.Element {
...
+          // @ts-ignore
+          draft[field] = value;
```
charlie-tf, line 134: Typescript didn't like that I was using a string to index this object

src/components/Reports/ReportEdit.tsx
karthikbtf: this is cool..
in case it helps.. we have a similar pattern in accession edit.. there's a `useForm` util that is used across the board.

src/strings/strings-en.ts
```diff
@@ -915,6 +920,8 @@ export const strings = {
   SUBTITLE_GET_STARTED:
     "To get started, please create your organization now. You'll also be able to set up and manage information on people and species.",
   SUGGESTION: 'Suggestion',
+  SUMMARY_OF_PROGRESS: 'Summary of Progress',
+  SUMMARY_OF_PROGRESS_DESCRIPTION: 'Give a brief summary this project.',
```
karthikbtf, line 924: summary of?

src/components/Reports/ReportForm.tsx
```diff
@@ -1,37 +1,96 @@
...
+      <Grid item xs={12} marginLeft={theme.spacing(-3)} marginTop={theme.spacing(-3)}>
+        <OverviewItemCard isEditable={false} title={strings.ORGANIZATION} contents={report.organizationName ?? ''} />
+      </Grid>
+      <Grid item xs={isMobile ? 12 : 4} marginLeft={theme.spacing(-3)} marginTop={theme.spacing(-3)}>
+        <OverviewItemCard isEditable={false} title={strings.SEED_BANKS} contents={`${report.totalSeedBanks}` ?? '0'} />
+      </Grid>
+      <Grid item xs={isMobile ? 12 : 4} marginLeft={theme.spacing(-3)} marginTop={theme.spacing(-3)}>
+        <OverviewItemCard isEditable={false} title={strings.NURSERIES} contents={`${report.totalNurseries}` ?? '0'} />
+      </Grid>
+      <Grid item xs={isMobile ? 12 : 4} marginLeft={theme.spacing(-3)} marginTop={theme.spacing(-3)}>
```
karthikbtf, line 63: if you want to avoid the `isMobile ? 12 : 4` repetition.. extract out int a func?

karthikbtf (approved):
lgtm! some thoughts/comments

src/components/Reports/ReportEdit.tsx
charlie-tf: Oh, i'll go take a look at that. Thanks

charlie-tf: Ah, okay. `useForm` looks pretty nice, but I'm not sure how well it would play with an object like a report that is nested 3 or 4 levels deep. For instance, if I have an object,
```javascript
obj0 = {
    obj1: {
        key: value
    }
}
```
Then `useForm`, which employs the spread operator, would only make a copy of the `obj1` reference, but `key: value` would not be copied. I think this could be a problem because updating the "copy" would modify the original react state at that level, which is a no-no. If I'm understanding correctly. I used `produce` because it deep-copies the object through its entire hierarchy.

charlie-tf: >We could theoretically use produce inside useForm (for the deep copy).. or leverage lodash's deepClone..

Oh cool, yeah we could enhance `useForm` even more

karthikbtf: All good, we can keep things as-is and refactor later if we want to.
Btw, in this example.. is draft the root level object and field a hierarchical property or a leaf property?
Ex.
draft = { 
  abc: {
      def: 123
  }
}

if we want to change def, is the field = abc.def? or just def?

karthikbtf: nvm, i see it is leaf level from usage 👍

charlie-tf: Yes. I will need to make some deeper modifications, so `produce` will show its worth more in future updates

PR #1261, opened 2023-03-01
SW-3074 Poll BE for editing lock to make sure the user is still the current editor

constanzauanini:
Screenshot 2023-03-01 at 16 15 08


src/strings/strings-en.ts
```diff
@@ -399,6 +399,7 @@ export const strings = {
   INITIAL_SEEDS: 'Initial Seeds',
   INTERMEDIATE: 'Intermediate',
   INVALID_DATE: 'Invalid date',
+  INVALID_EDITOR: 'This report is beign edited by another user',
```
karthikbtf, line 402: `being`

src/components/Reports/ReportEdit.tsx
```diff
@@ -24,6 +26,8 @@ export default function ReportEdit(): JSX.Element {
   const snackbar = useSnackbar();
 
   const [report, setReport] = useState<Report>();
+
+  const [checkUserInterval, setCheckUserInterval] = useState<NodeJS.Timer>();
```
karthikbtf, line 30: we could use `ReturnType<typeof setInterval>` .. Charlie flagged a similar issue in my PR.. (NodeJS is more BE specific)..

src/components/Reports/ReportEdit.tsx
```diff
@@ -39,8 +43,20 @@ export default function ReportEdit(): JSX.Element {
     } else {
       snackbar.toastError(strings.GENERIC_ERROR, strings.REPORT_COULD_NOT_OPEN);
     }
+
+    setCheckUserInterval(setInterval(getReport, 60000));
```
karthikbtf, line 47: do we need to return a () => clearInterval ? when component dismounts?
something like that.. so we aren't polling forever

src/components/Reports/ReportEdit.tsx
```diff
@@ -39,8 +43,20 @@ export default function ReportEdit(): JSX.Element {
...
   }, [reportIdInt, snackbar]);
 
+  useEffect(() => {
+    if (report && user && report?.lockedByUserId !== user?.id) {
+      if (checkUserInterval) {
+        clearInterval(checkUserInterval);
+        setCheckUserInterval(undefined);
+      }
+      history.push(APP_PATHS.REPORTS_VIEW.replace(':reportId', reportId).concat('?invalidEditor=true'));
```
karthikbtf, line 56: we should probably let user know before redirecting? a popup.. user hits ok.. then redirect?
i understand we are showing a popup info on the next page instead...

src/components/Reports/ReportView.tsx
```diff
@@ -12,9 +12,13 @@ import BackToLink from 'src/components/common/BackToLink';
 import ReportFormAnnual from 'src/components/Reports/ReportFormAnnual';
 import useSnackbar from 'src/utils/useSnackbar';
 import ConcurrentEditorWarningDialog from 'src/components/Reports/ConcurrentEditorWarningDialog';
+import useQuery from 'src/utils/useQuery';
 
 export default function ReportView(): JSX.Element {
   const { reportId } = useParams<{ reportId: string }>();
+
+  const query = useQuery();
+  const invalidEditor = !!query.get('invalidEditor');
```
karthikbtf, line 21: we should remove the param.. so reloading the page won't keep showing that message

karthikbtf (approved):
looks good overall.. a few comments/suggestions.. 
for UX stuff.. that's not required in this review..

src/components/Reports/ReportEdit.tsx
charlie-tf: I was thinking of just the redirect for an initial pass when I described to @constanzauanini. If we want to have a bit more interaction, like maybe popup confirmation as you suggest, or allow to steal the lock back, we can. I agree that this could be a very jarring UX.

karthikbtf: Got it. A simpler initial pass does sound good. Stealing locks is interesting.. users battling it out 🍿 
At some later point yes we should redirect after informing the user.

src/components/Reports/ReportEdit.tsx
constanzauanini: I changed the implementation because using useState was causing more than a interval at the same time. Modified to use useRef and added a clean up

src/components/Reports/ReportEdit.tsx
```diff
@@ -40,8 +45,30 @@ export default function ReportEdit(): JSX.Element {
...
+    if (intervalref.current !== null) {
+      return;
+    }
+    intervalref.current = window.setInterval(getReport, 600);
+
+    return () => {
+      if (intervalref.current) {
+        window.clearInterval(intervalref.current);
+        intervalref.current = null;
+      }
```
karthikbtf, line 58: we could extract this out since it is used in multiple places..
btw, saw this earlier https://usehooks-ts.com/react-hook/use-interval in case it helps.. also the NotificationsDropdown clears intervals as well if you want to see how we do it there.
in case it helps with consistency..

src/components/Reports/ReportEdit.tsx
karthikbtf: btw, if you think we won't revisit this to cleanup later... my suggestion is to clean this up now.. while we are working on this anyway.

karthikbtf (approved):
some thoughts

src/components/Reports/ReportEdit.tsx
constanzauanini: So we don't want to show the toast message and show a message here?

karthikbtf: Or just create a ticket to revisit later.

karthikbtf:
Thanks for the changes, lgtm 👍

PR #1262, opened 2023-03-01
SW-3070 Apply ExactOrFuzzy Search to Accession Number Only

charlie-tf:
- Now `Fuzzy` searches are always fuzzy, even if an exact match is found
- A new search type, `ExactOrFuzzy`, will show only an exact match if it
  exists, otherwise it will show fuzzy results
- In the Accessions list page apply `ExactOrFuzzy` to the Accession
  Number column only.

### DIFF
--- a/src/components/seeds/database/Filters.tsx
+++ b/src/components/seeds/database/Filters.tsx
@@ -305,7 +305,7 @@ function getSearchTermFilter(searchCols: DatabaseColumn[], searchTerm: string): 
     children: searchCols.map((col) => ({
       operation: 'field',
       field: col.key,
-      type: 'Fuzzy',
+      type: col.key === 'accessionNumber' ? 'ExactOrFuzzy' : 'Fuzzy',
       values: [searchTerm],
     })),
     operation: 'or',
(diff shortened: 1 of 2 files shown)


src/components/seeds/database/Filters.tsx
```diff
@@ -305,7 +305,7 @@ function getSearchTermFilter(searchCols: DatabaseColumn[], searchTerm: string):
     children: searchCols.map((col) => ({
       operation: 'field',
       field: col.key,
-      type: 'Fuzzy',
+      type: col.key === 'accessionNumber' ? 'ExactOrFuzzy' : 'Fuzzy',
```
karthikbtf, line 308: for future consideration.. the column definitions can set the search type (defaulting to Fuzzy if undefined)

karthikbtf (approved):
🎉 nice! thanks for fixing.

src/components/seeds/database/Filters.tsx
charlie-tf: Hmm, i didn't see how to define this. There wasn't a field on the DatabaseColumn type

karthikbtf: Correct, hence 'future consideration'.. it was just a thought :)
Maybe if/when we end up special casing any more keys, we could lift the search type semantics into the column def.

charlie-tf: Ah, okay I see what you meant.

PR #1264, opened 2023-03-03
SW-3102 query param to bust cache when requesting app version data

karthikbtf:
- this was accidentally removed as part of the service refactor 😿


constanzauanini (approved):

karthikbtf:
I will hotfix the UI today .. so we can probably avoid this issue next week..

PR #1266, opened 2023-03-03
SW-3087 Add photos upload section to report form

constanzauanini:
Allow upload and view report photos

Upload 
Screenshot 2023-03-03 at 16 40 40
Screenshot 2023-03-03 at 16 41 39

View
Screenshot 2023-03-03 at 16 42 27


src/components/common/SelectPhotos.tsx
```diff
@@ -51,8 +51,8 @@ const useStyles = makeStyles((theme: Theme) => ({
 }));
 
 type SelectPhotosProps = {
-  title: string;
-  description: string | string[];
+  title?: string;
+  description?: string | string[];
```
karthikbtf, line 55: do conditional rendering below? otherwise we will be using up some space/margin, etc...
```
{title ? (...) : null}
{description ? (....) : null}
```

src/services/ReportService.ts
```diff
@@ -287,8 +292,21 @@ const submitReport = async (id: number): Promise<Response> => {
...
+    'content-type': 'multipart/form-data',
+  };
+
+  const response: UploadReportPhotoResponse = await httpUploadReportPhoto.post({
+    entity,
+    headers,
+    urlReplacements: {
+      '{reportId}': reportId.toString(),
+    },
+  });
```
karthikbtf, line 309: we could create a PhotoService that does an upload + bulk uploads (maybe it takes in the url as an argument)..  this is a lot of redundant code from withdrawal photo uploads

src/components/Reports/UploadPhotos.tsx
```diff
@@ -0,0 +1,16 @@
...
+type UploadPhotosProps = {
+  onPhotosChanged: (photos: File[]) => void;
+};
+export default function UploadPhotos(props: UploadPhotosProps): JSX.Element {
+  const { onPhotosChanged } = props;
+
+  return (
+    <Container maxWidth={false}>
+      <SelectPhotos onPhotosChanged={onPhotosChanged} multipleSelection={true} />
+    </Container>
```
karthikbtf, line 14: this is fine as a component but maybe can just not be a separate component? only addition here is the container

karthikbtf (commented):
looks great, some comments

src/components/Reports/ViewPhotos.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    getPhotos();
+  }, [reportId, snackbar]);
+
+  return (
+    <>
+      <ViewPhotosModal
+        photosUrls={photoUrls}
+        open={photosModalOpened}
+        onClose={() => setPhotosModalOpened(false)}
```
karthikbtf, line 44: reset the selectedSlide?

src/components/Reports/UploadPhotos.tsx
constanzauanini: Yes, I think I created this because I thought it was going to do a lot of stuff and then it resulted really simple

karthikbtf (approved):
🎉

src/services/ReportService.ts
```diff
@@ -356,6 +356,7 @@ const seedbankFromFacility = (facility: Facility, originalReport: Report): Repor
     operationStartedDateEditable: true,
     totalSeedsStored: 0,
     workers: {},
+    selected: true,
```
constanzauanini, line 359: I added this here to compile, but this function is removed in charlie's PR (https://github.com/terraware/terraware-web/pull/1267)

PR #1267, opened 2023-03-03
Updates to Report Seedbanks

charlie-tf:
- Seedbanks have the "selected" field, which we toggle to show whether a
  seedbank is included in the report or not
- Let the seedbank InfoField component use the useStyles hook directly
  instead of passing the className through props
- No longer need two copies, report and draftReport, in ReportEdit
- Do not let the polling for lock status change the report data out from
  under the ReportEdit component

### DIFF
--- a/src/components/Reports/ReportEdit.tsx
+++ b/src/components/Reports/ReportEdit.tsx
@@ -14,7 +14,6 @@ import useSnackbar from 'src/utils/useSnackbar';
 import SubmitConfirmationDialog from 'src/components/Reports/SubmitConfirmationDialog';
 import { useUser } from 'src/providers';
 import produce from 'immer';
-import { getAllSeedBanks } from 'src/utils/organization';
 import { Organization } from 'src/types/Organization';
 import CannotEditReportDialog from './InvalidUserModal';
 
@@ -36,13 +35,11 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
   const [showInvalidUserModal, setShowInvalidUserModal] = useState(false);
 
   const [report, setReport] = useState<Report>();
-  const [draftReport, setDraftReport] = useState<Report>();
   useEffect(() => {
     const getReport = async () => {
       const result = await ReportService.getReport(reportIdInt);
       if (result.requestSucceeded && result.report) {
         setReport(result.report);
-        setDraftReport(structuredClone(result.report));
       } else {
         snackbar.toastError(strings.GENERIC_ERROR, strings.REPORT_COULD_NOT_OPEN);
       }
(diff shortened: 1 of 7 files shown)


src/components/Reports/ReportEdit.tsx
```diff
@@ -53,6 +50,16 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
     } else {
       snackbar.toastError(strings.GENERIC_ERROR, strings.REPORT_COULD_NOT_OPEN);
     }
+  }, [reportIdInt, snackbar]);
+
+  const [currentUserEditing, setCurrentUserEditing] = useState(true);
+  useEffect(() => {
```
charlie-tf, line 56: Moved the `setInterval` into its own `useEffect` and also don't allow its handler to write to the report state. This was changing the state out from under the component and invisibly reverting edits 😱

src/components/Reports/SeedbankSection.tsx
```diff
@@ -148,18 +134,24 @@ export default function SeedbankSection(props: SeedbankSectionProps): JSX.Elemen
...
   value: string | number;
   onChange: (value: any) => void;
-  className: string;
   type: 'text' | 'date';
 };
 
 function InfoField(props: InfoFieldProps): JSX.Element {
-  const { id, label, editable, value, onChange, className, type } = props;
+  const { id, label, editable, value, onChange, type } = props;
+  const classes = useStyles();
```
charlie-tf, line 154: This class didn't have to be passed in through props. Use `useStyles` directly in this component.

src/components/Reports/ReportEdit.tsx
```diff
@@ -226,15 +236,14 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
...
             (showAnnual ? (
               <ReportFormAnnual editable={true} report={report} />
             ) : (
               <ReportForm
                 editable={true}
-                draftReport={draftReport}
+                draftReport={report}
                 onUpdateReport={updateReport}
-                allSeedbanks={getAllSeedBanks(organization).map((f) => ReportService.seedbankFromFacility(f, report))}
+                allSeedbanks={report.seedBanks ?? []}
```
charlie-tf, line 246: `report.seedBanks` will always contain the full list of seedbanks.

src/components/Reports/SeedbankSection.tsx
```diff
@@ -169,6 +161,11 @@ function InfoField(props: InfoFieldProps): JSX.Element {
       <></>
     )
   ) : (
-    <OverviewItemCard isEditable={false} title={label} contents={value.toString() ?? '0'} className={className} />
+    <OverviewItemCard
+      isEditable={false}
+      title={label}
+      contents={value.toString() ?? '0'}
+      className={classes.infoCardStyle}
```
karthikbtf, line 168: cool

src/components/Reports/ReportEdit.tsx
karthikbtf: good catch!

karthikbtf (approved):
👍

PR #1275, opened 2023-03-08
Save photos when clicking save and close

constanzauanini:


src/components/Reports/ReportEdit.tsx
```diff
@@ -89,6 +89,8 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
       saveResult = await ReportService.updateReport(report);
       if (!saveResult.requestSucceeded) {
         snackbar.toastError(strings.GENERIC_ERROR, strings.REPORT_COULD_NOT_SAVE);
+      } else {
+        ReportService.uploadReportPhotos(report.id, photos);
```
karthikbtf, line 93: `await`
guessing the pageform shows the spinner while all this is happening

karthikbtf (approved):
one request

charlie-tf (commented):
Can you add to the other places we save, i.e. the ["save and next"](https://github.com/terraware/terraware-web/blob/2e690536af0300e99a6b5bbaa52d87c3c0e0d7c1/src/components/Reports/ReportEdit.tsx#L113) button press when there is an annual portion to the report?

constanzauanini:
> Can you add to the other place we save, i.e. the ["save and next"](https://github.com/terraware/terraware-web/blob/2e690536af0300e99a6b5bbaa52d87c3c0e0d7c1/src/components/Reports/ReportEdit.tsx#L113) button press when there is an annual portion to the report? (realizing that the save logic is a bit scattered and could perhaps be consolidated. Not a request for this PR, but something we can do in a followup perhaps)

Sure, I wasn't really sure if I needed to add it there!

charlie-tf (approved):

PR #1276, opened 2023-03-08
SW-3091 Add Annual Details

charlie-tf:
image
image
image
image


src/components/Reports/useSDGProgress.ts
```diff
@@ -0,0 +1,236 @@
+import React, { useState } from 'react';
+import useDebounce from 'src/utils/useDebounce';
+import { Report, SDG } from 'src/types/Report';
+
+const DEBOUNCE_TIME_MS = 500;
+
+export default function useSDGProgress(
```
charlie-tf, line 7: This is a big ugly hook which adds state and debounce for all of the Sustainable Development Goal Progress fields

src/components/Reports/ReportFormAnnual.tsx
```diff
@@ -1,36 +1,389 @@
...
+
+  const handleIsCatalyticChange = (catalytic: boolean) => {
+    if (updateDetails) {
+      setIsCatalytic(catalytic);
+      updateDetails('isCatalytic', catalytic);
+    }
+  };
+
+  const handleFileUploadButtonClick = () => {
+    // TODO
```
charlie-tf, line 181: Will do this in a followup—this PR was getting big and I hadn't quite gotten this figured out

src/components/Reports/ReportEdit.tsx
```diff
@@ -222,6 +222,33 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
...
+      );
+    }
+  };
+
+  const updateSDGProgress = (index: number, value: string) => {
+    if (
+      report &&
+      report.annualDetails &&
+      report.annualDetails.sustainableDevelopmentGoals &&
+      report.annualDetails.sustainableDevelopmentGoals[index]
```
karthikbtf, line 241: if you prefer.. `report?.annualDetails?.sustainableDevelopmentGoals?[index]`

src/strings/strings-en.ts
```diff
@@ -790,6 +830,23 @@ export const strings = {
   SCIENTIFIC_NAME_NOT_FOUND: '“{0}” was not found in our database, but you can still add it.',
   SCIENTIFIC_NAME_REQUIRED: 'Scientific Name *',
   SCIENTIFIC_NAME: 'Scientific Name',
+  SDG_NO_POVERTY: '1. No Poverty',
```
karthikbtf, line 833: similar to MONTH_{num} these could be SDG_{num}

src/components/Reports/ReportFormAnnual.tsx
```diff
@@ -1,36 +1,389 @@
...
+    ResponsibleConsumption: strings.SDG_RESPONSIBLE_CONSUMPTION_AND_PRODUCTION,
+    ClimateAction: strings.SDG_CLIMATE_ACTION,
+    LifeBelowWater: strings.SDG_LIFE_BELOW_WATER,
+    LifeOnLand: strings.SDG_LIFE_ON_LAND,
+    Peace: strings.SDG_PEACE_JUSTICE_AND_STRONG_INSTITUTIONS,
+    Partnerships: strings.SDG_PARTNERSHIPS_FOR_THE_GOALS,
+  };
+
+  const [projectSummary, setProjectSummary] = useState(report.annualDetails?.projectSummary ?? '');
+  useDebounce(projectSummary, DEBOUNCE_TIME_MS, (value) => {
```
karthikbtf, line 61: just curious.. why debounce? is it too slow to edit otherwise?

src/components/Reports/useSDGProgress.ts
```diff
@@ -0,0 +1,236 @@
...
+    report.annualDetails?.sustainableDevelopmentGoals?.find((s) => s.goal === SDG[0])?.progress ?? ''
+  );
+  sdgProgressStates.push(sdgProgress0);
+  setSdgProgressStates.push(setSdgProgress0);
+  useDebounce(sdgProgress0, DEBOUNCE_TIME_MS, (value) => {
+    const index = report.annualDetails?.sustainableDevelopmentGoals?.findIndex((s) => s.goal === SDG[0]);
+    if (updateSDGProgress && index !== undefined && index >= 0) {
+      updateSDGProgress(index, value);
+    }
+  });
```
karthikbtf, line 25: if it helps.. you can extract this out into another hook... and call that hook sequentially, passing in the sdgProgressStates array and sdg index

karthikbtf (approved):
wow, this is a big effort..  some comments but looks good to get in and follow up

src/components/Reports/ReportEdit.tsx
charlie-tf: oh.. yeah 😄

src/components/Reports/ReportFormAnnual.tsx
charlie-tf: I did it for performance reasons. I didn't want to call updates to the report every keystroke. I thought it would be better to update state locally and debounce those updates to full report state. But honestly I didn't test without debounce, so it could be a premature optimization.

karthikbtf: i figured..  my thought is it's worth testing.. since user will be typing in one input at a time. should be ok to call onChange for every key stroke. 
However,, since we are updating the parent report object.. not sure if that change is triggering a lot of re-renders across the report.

src/components/Reports/useSDGProgress.ts
charlie-tf: Made it much less big and less ugly. Thanks @karthikbtf!

src/components/Reports/ReportFormAnnual.tsx
charlie-tf: My sense is that it would cause a lot of rerenders. If `updateDetails` is called each keystroke, it updates the entire report object, which is a dependency of every item in the form. I'll test and see what it's like without the debounce though!

src/components/Reports/useSDGProgress.ts
karthikbtf: awesome, thanks @charlie-tf !
can we iterate over `0..16` and call this with the index? or does react want a more deterministic order of hooks?

src/strings/strings-en.ts
```diff
@@ -790,6 +830,23 @@ export const strings = {
   SCIENTIFIC_NAME_NOT_FOUND: '“{0}” was not found in our database, but you can still add it.',
   SCIENTIFIC_NAME_REQUIRED: 'Scientific Name *',
   SCIENTIFIC_NAME: 'Scientific Name',
+  SDG_01: '1. No Poverty',
```
karthikbtf, line 833: thanks so much!

karthikbtf (approved):
🎉

src/components/Reports/useSDGProgress.ts
charlie-tf: I initially tried putting this into a map expression, then a for-loop. React didn't like either. I think deterministically calling each is what it wants here.

karthikbtf: Got it.. fine as-is. Thanks 👍

src/components/Reports/ReportFormAnnual.tsx
charlie-tf: Actually, it is smart enough not to repaint everything besides the input that is being interacted with. Thus, this seems to work fine without the debounce. So my intuition was wrong here and I can change this to get rid of the debounces 😄

karthikbtf: Thanks for confirming!!! Sounds good to me :)  I think this can be done as a follow up if you want?

charlie-tf: I can get it done pretty quickly for the annual details and follow up for the other part of the report.

karthikbtf: Looks good 🙌

PR #1277, opened 2023-03-08
SW-3112 Allow deleting photos when editing report

constanzauanini:
Screenshot 2023-03-08 at 16 31 48


src/services/ReportService.ts
```diff
@@ -326,14 +326,37 @@ const updateReportPhoto = async (reportId: number, photoId: number, caption: str
   });
 };
 
+/**
+ * Delete multiple photos for a report
+ */
+const deleteReportPhotos = async (reportId: number, photosId: number[]): Promise<(Response | string)[]> => {
+  const deletePhotoPromises = photosId.map((photoId) => deleteReportPhoto(reportId, photoId));
+  try {
+    const promiseResponses = await Promise.allSettled(deletePhotoPromises);
```
karthikbtf, line 335: there's a util in `src/services/utils` called `getPromisesResponse` which you can reuse here

src/components/Reports/ReportEdit.tsx
```diff
@@ -222,6 +231,12 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
     }
   };
 
+  const onAddPhotoIdToRemove = (id: number) => {
```
karthikbtf, line 234: `onRemovePhoto` ..adding the photo id for later deletion is an internal implementation detail,, ok to skip that in the name i think

src/components/Reports/ReportEdit.tsx
```diff
@@ -83,14 +85,21 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
...
+
   const gotoReportView = async (saveChanges: boolean) => {
     let saveResult;
     if (saveChanges && report) {
       saveResult = await ReportService.updateReport(report);
       if (!saveResult.requestSucceeded) {
         snackbar.toastError(strings.GENERIC_ERROR, strings.REPORT_COULD_NOT_SAVE);
       } else {
-        await ReportService.uploadReportPhotos(report.id, photos);
+        updatePhotos(report.id);
```
karthikbtf, line 102: `await` here and in other places

src/components/Reports/ReportEdit.tsx
```diff
@@ -36,6 +36,8 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
 
   const [photos, setPhotos] = useState<File[]>([]);
 
+  const [photosIdToRemove, setPhotosIdToRemove] = useState<number[]>([]);
```
karthikbtf, line 39: photoIdsToRemove?

src/components/Reports/ViewPhotos.tsx
```diff
@@ -1,35 +1,64 @@
...
   reportId: number;
+  onPhotoRemove?: (id: number) => void;
+  editable: boolean;
 };
 
-export default function ViewPhotos({ reportId }: PhotosSectionProps): JSX.Element {
+type ReportPhotoWithUrl = ReportPhoto & { url: string };
+
+const useStyles = makeStyles((theme: Theme) => ({
+  removePhoto: {
```
karthikbtf, line 19: don't want to create more work for you but seems like we can extract out a component to reuse between report and withdraw batch photos (same display to remove a photo)?
if it's a big risk .. we can follow up with a cleanup ticket

karthikbtf (commented):
looks good, a few comments, some that can be fixed in this PR.. some for later

src/components/Reports/ViewPhotos.tsx
constanzauanini: From the withdrawal log we can't remove photos of the withdrawals. We can remove photos when creating a new withdraw and uploading photos, and we are reusing that component to upload here too. Is that what you mean?

karthikbtf: the display only.. attaching screenshot 
Terraware App 2023-03-08 12-13-36

karthikbtf: this deserves to be a separate ticket.. lets just create one and not do it here..

PR #1279, opened 2023-03-09
SW-2890 Adding required validation to quarterly fields when submitting

constanzauanini:
- Adding required validation only to first report step

### DIFF
--- a/src/components/Reports/ReportEdit.tsx
+++ b/src/components/Reports/ReportEdit.tsx
@@ -169,7 +169,9 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
       if (hasEmptyRequiredFields(report)) {
         setConfirmSubmitDialogOpen(false);
         setValidateFields(true);
-        handleBack();
+        if (showAnnual) {
+          handleBack();
+        }
         return;
       }
       const saveResult = await ReportService.updateReport(report);
--- a/src/components/Reports/ReportEdit.tsx
+++ b/src/components/Reports/ReportEdit.tsx
@@ -144,16 +144,16 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
     }
   };
 
-  const hasEmptyRequieredFields = (report: Report) => {
-    const emptySeedbankFields = report.seedBanks?.some((sb) => {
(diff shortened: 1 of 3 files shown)


src/components/Reports/ReportEdit.tsx
```diff
@@ -141,8 +144,33 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
     }
   };
 
+  const hasEmptyRequieredFields = (iReport: Report) => {
```
charlie-tf, line 147: #spelling `hasEmptyRequiredFields`

src/components/Reports/ReportEdit.tsx
```diff
@@ -141,8 +144,33 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
...
+        !plantingSite.mortalityRate
+      );
+    });
+    return emptySeedbankFields || emptyNurseryFields || emptyPlantingSitesFields;
+  };
   const submitReport = async () => {
     if (report) {
+      if (hasEmptyRequieredFields(report)) {
+        setValidateFields(true);
+        setShowAnnual(false);
```
charlie-tf, line 171: Maybe use the `handleBack` function here if `showAnnual` is true, so that any annual details that were entered are saved

charlie-tf (approved):
Looks good to me :)

src/components/Reports/ReportEdit.tsx
```diff
@@ -141,8 +144,34 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
...
+      );
+    });
+    return emptySeedbankFields || emptyNurseryFields || emptyPlantingSitesFields;
+  };
   const submitReport = async () => {
     if (report) {
+      if (hasEmptyRequiredFields(report)) {
+        setConfirmSubmitDialogOpen(false);
+        setValidateFields(true);
+        handleBack();
```
charlie-tf, line 172: Actually, would you mind doing
```typescript
if (showAnnual) {
  handleBack();
}
```
here? My thought being that if they are already on the quarterly page a Save might not be intended.

constanzauanini: Oh, you are right!

PR #1280, opened 2023-03-09
SW-3124 Add Upload Budget File(s) to Report

charlie-tf:
## Edit:
image
image

## View:
image

### DIFF
--- a/src/components/Reports/ReportEdit.tsx
+++ b/src/components/Reports/ReportEdit.tsx
@@ -16,6 +16,7 @@ import { useUser } from 'src/providers';
 import produce from 'immer';
 import { Organization } from 'src/types/Organization';
 import CannotEditReportDialog from './InvalidUserModal';
+import useReportFiles from 'src/components/Reports/useReportFiles';
 
 export type ReportEditProps = {
   organization: Organization;
@@ -56,11 +57,11 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
     }
   }, [reportIdInt, snackbar]);
 
-  const [newReportFiles, setNewReportFiles] = useState<File[]>();
+  const [newReportFiles, setNewReportFiles] = useState<File[]>([]);
 
-  const [initialReportFiles, setInitialReportFiles] = useState<ReportFile[]>();
+  const [updatedReportFiles, setUpdatedReportFiles] = useState<ReportFile[]>([]);
 
-  const [updatedReportFiles, setUpdatedReportFiles] = useState<ReportFile[]>();
+  const initialReportFiles = useReportFiles(report, setUpdatedReportFiles);
 
   const updateFiles = async () => {
     await Promise.all(
@@ -73,28 +74,6 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
     );
     await Promise.all(newReportFiles?.map((f) => ReportService.uploadReportFile(reportIdInt, f)) ?? []);
   };
-
-  useEffect(() => {
-    const getFiles = async () => {
-      if (report) {
-        const fileListResponse = await ReportService.getReportFiles(report.id);
-        if (!fileListResponse.requestSucceeded || fileListResponse.error) {
-          setInitialReportFiles([]);
-          snackbar.toastError();
-        } else {
-          const fileArray: ReportFile[] = [];
-          fileListResponse.files?.forEach((f) => {
-            fileArray.push(f);
-          });
-
-          setInitialReportFiles(fileArray);
-          setUpdatedReportFiles(fileArray);
-        }
-      }
-    };
-
-    getFiles();
-  }, [report, snackbar]);
 
   const [currentUserEditing, setCurrentUserEditing] = useState(true);
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/components/Reports/ReportEdit.tsx
```diff
@@ -56,6 +56,46 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
...
+      }) ?? []
+    );
+    await Promise.all(newReportFiles?.map((f) => ReportService.uploadReportFile(reportIdInt, f)) ?? []);
+  };
+
+  useEffect(() => {
+    const getFiles = async () => {
+      if (report) {
+        const fileListResponse = await ReportService.getReportFiles(report.id);
+        if (!fileListResponse.requestSucceeded || fileListResponse.error) {
```
karthikbtf, line 81: i think we can skip the || check since we won't have a success case with an error message

src/components/Reports/ReportView.tsx
```diff
@@ -65,6 +65,28 @@ export default function ReportView(): JSX.Element {
     }
   };
 
+  const [initialReportFiles, setInitialReportFiles] = useState<ReportFile[]>();
+  useEffect(() => {
+    const getFiles = async () => {
+      if (report) {
+        const fileListResponse = await ReportService.getReportFiles(report.id);
+        if (!fileListResponse.requestSucceeded || fileListResponse.error) {
```
karthikbtf, line 73: same.. also looks like an opportunity for a hook (to use across view and edit.. with that one difference of updating the updatedFiles state).. just a thought.. not necessary

src/components/Reports/ReportFormAnnual.tsx
```diff
@@ -199,11 +254,42 @@ export default function ReportFormAnnual(props: ReportFormAnnualProps): JSX.Elem
...
+            />
+            <Typography>{f.name}</Typography>
+          </Box>
+        ))}
+        {existingFiles.map((f) => (
+          <Box key={f.id} display='flex' alignItems='center'>
+            {editable ? (
+              <Button priority='ghost' type='passive' icon='iconTrashCan' onClick={() => onRemoveExistingFile(f.id)} />
+            ) : (
+              <a href={getReportFileUrl(report.id, f.id)}>
```
karthikbtf, line 286: Not sure how permissions for this report file url works,, guessing it isn't public. Check with @sgrimm ?
For authenticated routes, may want to issue a GET within this session and make it downloadable as a data blob.

karthikbtf (commented):
some thoughts.. there's one permission question

src/components/Reports/ReportFormAnnual.tsx
sgrimm: Not 100% sure what the question is here, but all the `/api/v1/reports` endpoints require authentication and enforce the usual permissions. The GET endpoint for retrieving files sets the `Content-Disposition` header to `attachment` with a filename so the browser won't display it inline.

sgrimm: Server side of this, for reference: https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/report/api/ReportsController.kt#L255

karthikbtf: I thought the `href=<authenticated-route>` link/button would not work but it actually does since we use cookies and same url host for API. My question about permissions is no longer an issue. 
Glad I asked this question though, the `Content-Disposition` needs to be considered.

Charlie, if it helps -  [blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob) 
Setting an href with blob data.. https://javascript.info/blob#blob-as-url

sgrimm: I'd assumed we'd want to do more or less what this PR does, where the browser treats the file as a download and the React code never touches the file contents. If instead there's an advantage to loading the file into memory in JavaScript and then generating a copy of it for download, we can easily get rid of the `Content-Disposition` header.

charlie-tf: I think we're good to go with regards to the `Content-Disposition` it causes the file to be downloaded as an attachment as desired
image
I'll take a look at the Blob things

src/components/Reports/ReportView.tsx
charlie-tf: That's a good idea. Saves some code repetition

src/components/Reports/ReportFormAnnual.tsx
karthikbtf: Thanks for the clarification and answering my questions, makes sense to me now. Looks good then 👍

karthikbtf (approved):
Looks good to me, a few suggestions in there.

src/components/Reports/ReportFormAnnual.tsx
charlie-tf: Yup I think leaving as-is is the way to go here.

src/components/Reports/useReportFiles.ts
```diff
@@ -0,0 +1,37 @@
+import React, { useEffect, useState } from 'react';
+import { Report, ReportFile } from 'src/types/Report';
+import ReportService from 'src/services/ReportService';
+import useSnackbar from 'src/utils/useSnackbar';
+
+export default function useReportFiles(
```
karthikbtf, line 6: Thanks for the changes 👍

PR #1281, opened 2023-03-10
SW-3129 Withdrawing seeds to nursery with weight based accessions

constanzauanini:
When withdrawing for Nursery purpose, we were not setting the correct unit to validate, and it was using 'Grams' instead of 'Seeds'. Fix that, and also added a useEffect to set unit value correctly when changing purpose

### DIFF
--- a/src/components/accession2/withdraw/WithdrawModal.tsx
+++ b/src/components/accession2/withdraw/WithdrawModal.tsx
@@ -110,6 +110,27 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element 
     getOrgUsers();
   }, [selectedOrganization]);
 
+  useEffect(() => {
+    setRecord((previousRecord: Withdrawal): Withdrawal => {
+      if (previousRecord.withdrawnQuantity) {
+        return {
+          ...previousRecord,
+          withdrawnQuantity: {
+            quantity: previousRecord.withdrawnQuantity?.quantity,
+            units:
+              isNurseryTransfer || record.purpose === 'Viability Testing'
+                ? 'Seeds'
+                : previousRecord.withdrawnQuantity?.units !== 'Seeds'
+                ? previousRecord.withdrawnQuantity?.units
+                : 'Grams',
+          },
+        };
+      } else {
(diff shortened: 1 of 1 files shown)


src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -159,7 +180,11 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
           ? undefined
           : {
               quantity: value || 0,
-              units: prev.withdrawnQuantity?.units || accession.remainingQuantity?.units || 'Grams',
+              units:
+                prev.withdrawnQuantity?.units ||
+                (isNurseryTransfer || record.purpose === 'Viability Testing'
+                  ? 'Seeds'
+                  : accession.remainingQuantity?.units || 'Grams'),
```
karthikbtf, line 187: do we need to check if units !== 'Seeds' here just like on line 123?..
wondering if this can be extracted out into a reusable function

karthikbtf (approved):
Correct me if I'm wrong.. feels like this code is getting complicated.. a lot of bandaids have been applied to this modal. If you agree, we can revisit later for a refactor/cleanup.
Otherwise good 👍

constanzauanini:
> Correct me if I'm wrong.. feels like this code is getting complicated.. a lot of bandaids have been applied to this modal. If you agree, we can revisit later for a refactor/cleanup. Otherwise good 👍

Yes, totally, this is getting more complex everytime. Let's clean up later

src/components/accession2/withdraw/WithdrawModal.tsx
constanzauanini: I don't think so, because the change purpose use effect would have already changed this to grams if units was seeds before. But we can add it just in case

constanzauanini: I will merge it like this, we are revisiting this later anyway

karthikbtf: 🎉

PR #1282, opened 2023-03-10
Report Annual Validation

charlie-tf:
- quarterly section is validated first and navigated to if there are problems
- then the annual section is validated
- added quarterly project summary to required fields
- turn off validation when switching pages until submit retry


constanzauanini (approved):
Lgtm!

PR #1283, opened 2023-03-10
E2e Cypress Test Improvements

charlie-tf:
Some changes to make the e2e cypress tests more robust and debuggable:
- Capture screenshots and store them in the event of a test failure
  - These can be found on the workflow summary page after the run completes
- Increase the `pageLoadTimeout` and `responseTimeout` ([ref](https://docs.cypress.io/guides/references/configuration#Timeouts))
  - Passing tests were sometimes running close to these limits, so increasing ought to help with the flakiness
- Add a dummy test to visit `/accessions` at the beginning of the `10-database` tests
  - Noticed that the first attempt to `cy.visit('/accessions')` often fails, with subsequent attempts succeeding
  - The dummy tests ignores exceptions thrown by the visit, so the idea is to clear any network issues before running the actual tests.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -87,6 +87,13 @@ jobs:
           REACT_APP_SNACKBAR_TIMEOUT: 1000
           REACT_APP_TERRAWARE_FE_BUILD_VERSION: ${{ env.APP_VERSION }}
 
+      - name: save screenshots of e2e test failures
+        uses: actions/upload-artifact@latest
+        if: failure()
+        with:
+          name: cypress-snapshots
+          path: cypress/snapshots
+
       - name: build
         run: yarn build
         env:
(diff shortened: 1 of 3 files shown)


charlie-tf:
As additional background, a few successful runs I've found:
image
image
image
image


One could easily imagine these bumping up over 30/60s timeout limits

charlie-tf:
hmmm, got this failure even with the longer timeouts
image

constanzauanini:
> hmmm, got this failure even with the longer timeouts image

I was trying to fix this yesterday, but I couldn't. I don't know what happened that we are getting this errors sometimes

constanzauanini:
Maybe run this again to check if the accession increased time works?

karthikbtf (approved):
🎉 🥳

PR #1284, opened 2023-03-10
SW-3151 Fix add inventory manually entered dates

constanzauanini:
Timezone was not set correctly, so when manually entering dates, they were off by one


karthikbtf (approved):
👍

PR #1286, opened 2023-03-10
SW-3152 Dashboard status filters should stick with pre-exposed-filter setup

karthikbtf:
- when going from a dashboard quick filter to accessions table, the status does not stick with the pre-exposed filter
- update code such that it sticks, previously the filters were keyed off different names `state` and `preExpFilter`.. converge them all to use `state`

screencast 2023-03-10 13-49-53

### DIFF
--- a/src/components/seeds/database/Filters.tsx
+++ b/src/components/seeds/database/Filters.tsx
@@ -99,6 +99,7 @@ export default function Filters(props: Props): JSX.Element {
   const { columns, searchColumns, preExpFilterColumn, filters, availableValues, allValues, onChange } = props;
   const { isMobile, isDesktop } = useDeviceInfo();
   const classes = useStyles({ isMobile, isDesktop });
+  const preExpFilterKey = preExpFilterColumn.key;
   const [searchTerm, setSearchTerm] = React.useState(getSearchTermFromFilters(filters));
   const searchTermCallback = useCallback(
     (value: string) => {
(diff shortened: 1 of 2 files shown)


karthikbtf:
@charlie-tf I started out by renaming `state` to `preExpFilter` (see first commit in this PR).. I eventually thought it might be cleaner to use the `preExpFilterColumn` information to set the pre-exposed-filter information (remove hard-coded names) - this is the second commit. Happy to use the first commit instead of the second but thought I'd call it out.

charlie-tf:
> @charlie-tf I started out by renaming `state` to `preExpFilter` (see first commit in this PR).. I eventually thought it might be cleaner to use the `preExpFilterColumn` information to set the pre-exposed-filter information (remove hard-coded names) - this is the second commit. Happy to use the first commit instead of the second but thought I'd call it out.

This looks good to me, Karthik. Thanks so much for fixing this!

charlie-tf (approved):

karthikbtf:
> > @charlie-tf I started out by renaming `state` to `preExpFilter` (see first commit in this PR).. I eventually thought it might be cleaner to use the `preExpFilterColumn` information to set the pre-exposed-filter information (remove hard-coded names) - this is the second commit. Happy to use the first commit instead of the second but thought I'd call it out.
> 
> This looks good to me, Karthik. Thanks so much for fixing this!

Thanks Charlie!

PR #1287, opened 2023-03-11
SW-2485 / SW-2501 / SW-2529

alexTerraformation:
A few relatively minor tweaks leftover from visual updates.
Screen Shot 2023-03-10 at 4 05 27 PM
Screen Shot 2023-03-10 at 4 05 52 PM
Screen Shot 2023-03-10 at 4 07 34 PM

### DIFF
--- a/src/components/Inventory/view/QuantitiesMenu.tsx
+++ b/src/components/Inventory/view/QuantitiesMenu.tsx
@@ -71,7 +71,7 @@ export default function QuantitiesMenu(props: QuantitiesMenuProps): JSX.Element 
           icon='menuVertical'
           type='passive'
           priority='ghost'
-          size='small'
+          size='medium'
         />
       </Tooltip>
     </>
--- a/src/components/common/OverviewItemCard.tsx
+++ b/src/components/common/OverviewItemCard.tsx
@@ -1,19 +1,13 @@
-import { Box, Theme, Typography, useTheme } from '@mui/material';
-import { makeStyles } from '@mui/styles';
-import { Icon, IconTooltip } from '@terraware/web-components';
+import { Box, Typography, useTheme } from '@mui/material';
+import { Button, Tooltip, IconTooltip } from '@terraware/web-components';
 import useDeviceInfo from '../../utils/useDeviceInfo';
-
-const useStyles = makeStyles((theme: Theme) => ({
-  editIcon: {
-    display: 'none',
-    fill: theme.palette.TwClrIcn,
-  },
-}));
+import strings from 'src/strings';
 
 type OverviewItemCardProps = {
   isEditable: boolean;
   hideEditIcon?: boolean;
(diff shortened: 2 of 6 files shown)


src/components/Inventory/view/QuantitiesMenu.tsx
```diff
@@ -71,7 +71,7 @@ export default function QuantitiesMenu(props: QuantitiesMenuProps): JSX.Element
           icon='menuVertical'
           type='passive'
           priority='ghost'
-          size='small'
+          size='medium'
```
alexTerraformation, line 74: This makes the row size grow quite a bit, which prompted the change in https://github.com/terraware/web-components/pull/306 to match the table spec.

src/components/common/OverviewItemCard.tsx
```diff
@@ -93,8 +74,16 @@ export default function OverviewItemCard({
         </Box>
       </Box>
       {isEditable && !hideEditIcon ? (
-        <Box sx={{ marginLeft: 1, height: '16px', width: '16px' }}>
-          <Icon name='iconEdit' className={`${classes.editIcon} edit-icon`} />
+        <Box sx={{ marginLeft: 1, position: 'relative', top: '-11px' }}>
```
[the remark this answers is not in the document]
karthikbtf: This is totally fine here, the button has a padding but no border.. we have to align with the icon inside of it.

src/components/common/PageCard.tsx
```diff
@@ -87,15 +80,8 @@ export default function PageCard(props: PageCardProps): JSX.Element {
...
-  const goToPage = (event?: React.MouseEvent) => {
-    if (linkStyle === 'button') {
-      return;
-    }
-    history.push({ pathname: link });
-  };
-
   return (
-    <Box className={`${classes.container} ${isMobile ? '' : 'min-height'}`} onClick={goToPage}>
+    <Box className={`${classes.container} ${isMobile ? '' : 'min-height'}`}>
```
karthikbtf, line 84: interesting.. i should let Ming know.. current funder portal has the entire card as clickable but similar design as here (bottom link but in a card)...

karthikbtf (approved):
Thanks for the changes, looks great 👍

src/components/common/OverviewItemCard.tsx
```diff
@@ -1,19 +1,13 @@
...
-    display: 'none',
-    fill: theme.palette.TwClrIcn,
-  },
-}));
+import strings from 'src/strings';
 
 type OverviewItemCardProps = {
   isEditable: boolean;
   hideEditIcon?: boolean;
   onClick?: () => void;
```
constanzauanini, line 9: Are we using the onClick somewhere in this component? If not, we could remove this prop

karthikbtf (approved):

PR #1294, opened 2023-03-15
SW-3185 Fix delete photo upload trash icon is stretched

constanzauanini:
Screenshot 2023-03-15 at 16 27 19

Remove styles that are not needed anymore since web-components provide this styles for only icon buttons


karthikbtf (approved):
Thanks for fixing this!

PR #1297, opened 2023-03-18
SW-3236 Add Subzones and Monitoring Plots

charlie-tf:
Make necessary changes to FE code to accomodate the renaming of plots to
planting subzones and the addition of monitoring plots inside of those.
We update the types from the OpenAPI spec, but the bulk of the changes
here are renames. I've gone through the withdrawal and reassignment flows,
as well as verifying that the maps are working correctly.

Note, this won't work correctly without the BE changes in https://github.com/terraware/terraware-server/pull/1028


karthikbtf (approved):
lgtm 👍 and thanks for this sweeping effort 🙏 
like you mentioned, was unable to test but looks good..

karthikbtf:
The subzone selector label in plants dashboard chart may need a style lift.. i think the label has a min-width that needs to be increased.

Terraware App 2023-03-19 07-45-27

Terraware App 2023-03-19 07-45-06

Terraware App 2023-03-19 07-44-45

charlie-tf:
>The subzone selector label in plants dashboard chart may need a style lift.. i think the label has a min-width that needs to be increased.

I'll take a look at this. Thanks!

karthikbtf:
> image
> 
> image

👍🙏🏻

PR #1303, opened 2023-03-21
SW-3194 Fix Input Behavior on Planting Site Species Table

charlie-tf:
- onRowClick was being called on every change, which caused the input to
  lose focus.
- break the input out into a separate component
- the input component keeps track of its own input value state
- onRowClick is only called on "confirming" the input, which we define as
  when the input is blurred
- the UX is good in testing, blur happens when anything else in the form is
  clicked, including the save and quit button

image


constanzauanini (approved):
Looks good! Thanks for fixing this! Tested in Vercel

PR #1305, opened 2023-03-21
SW-3201, SW-3196 Some Updates to Report Validation

charlie-tf:
- Switch validation to before showing the submit confirmation
- Show a toast message if there are missing required fields

image


karthikbtf (approved):
lgtm 👍

PR #1307, opened 2023-03-21
SW-3197, SW-3198 Validate That Facility Dates are in Order

charlie-tf:
- Build start should be before build complete
- Operation start should be after build complete

Datepicker limits:
image

Form validation:
image


src/components/Reports/LocationSection.tsx
```diff
@@ -437,3 +453,27 @@ function InfoField(props: InfoFieldProps): JSX.Element {
     />
   );
 }
+
+export const buildStartedDateValid = (loc: ReportSeedBank | ReportNursery) => {
+  if (loc.buildStartedDate && loc.buildCompletedDate) {
+    return Date.parse(loc.buildStartedDate) <= Date.parse(loc.buildCompletedDate);
+  }
+  return true;
```
karthikbtf, line 461: should this be false? assuming we want both dates to be present and the condition to be met

src/components/Reports/LocationSection.tsx
```diff
@@ -437,3 +453,27 @@ function InfoField(props: InfoFieldProps): JSX.Element {
...
+
+export const buildCompletedDateValid = (loc: ReportSeedBank | ReportNursery) => {
+  if (loc.buildStartedDate && loc.buildCompletedDate) {
+    return Date.parse(loc.buildStartedDate) <= Date.parse(loc.buildCompletedDate);
+  }
+  if (loc.buildCompletedDate && loc.operationStartedDate) {
+    return Date.parse(loc.buildCompletedDate) <= Date.parse(loc.operationStartedDate);
+  }
+  return true;
+};
```
karthikbtf, line 472: do we want both checks to be executed and not short-circuit on the first one?

src/strings/strings-en.ts
```diff
@@ -781,9 +781,12 @@ export const strings = {
   REPORT_EDIT: 'Edit Report',
   REPORT_NAME: 'Report Name',
   REPORT_NURSERY_BUILD_COMPLETION_DATE: 'Nursery Build Completion Date *',
+  REPORT_NURSERY_BUILD_COMPLETION_DATE_INVALID: 'Must be between build start and operation start dates.',
   REPORT_NURSERY_BUILD_START_DATE: 'Nursery Build Start Date *',
+  REPORT_NURSERY_BUILD_START_DATE_INVALID: 'Must be before build completed date.',
```
karthikbtf, line 786: `on or before`?

src/strings/strings-en.ts
```diff
@@ -781,9 +781,12 @@ export const strings = {
...
   REPORT_NURSERY_CAPACITY: 'Nursery Capacity *',
   REPORT_NURSERY_OPERATION_START_DATE: 'Nursery Operation Start Date *',
+  REPORT_NURSERY_OPERATION_START_DATE_INVALID: 'Must be after build completed date.',
```
karthikbtf, line 789: `on or after`?

karthikbtf (approved):
Looks good to me, easy to follow.. hence a few suggestions/thoughts... 👍

src/components/Reports/LocationSection.tsx
charlie-tf: I figured `true` here since if either of the dates is not present then it'll throw a missing required field error anyways.

src/components/Reports/LocationSection.tsx
charlie-tf: hm, good point, if the first one resolves to `true` and the second one would have resolved to `false` then it doesn't get to return that. I'll think about how to restructure this logic for that.

src/components/Reports/LocationSection.tsx
karthikbtf: That's what I figured. That's fine too. From an isolated perspective might be good to return false (assuming this code doesn't know what's happening outside of it).

charlie-tf: Good point. I will make that change

src/components/Reports/LocationSection.tsx
karthikbtf: The changes look good 👍

PR #1310, opened 2023-03-23
SW-3272 Generate string tables from CSV files

sgrimm:
The translation service we're going to use doesn't support TypeScript source
files. Move the English strings to a CSV file that has an additional column
(currently empty for all our strings) where we can add comments to give additional
context to translators.

We will send the English CSV file out for translation, and get back CSV files in
other languages; these will then be checked into the repo.

There's now a build step that converts the CSV files to JavaScript so they can be
imported from the rest of the code as usual. This is done as an explicit one-off
step in the CI workflow; in dev environments, the CSV files are monitored for
changes and processed on the fly so that edits to strings will be hot-reloaded
the same way they were reloaded when the strings table was a TypeScript source
file.

This change doesn't introduce any actual translation workflow steps, and for now,
the only locale with a CSV file is US English. But this will enable us to start
adding comments to our strings right away.


sgrimm:
Hopefully this is a reasonable approach; I'm not super familiar with the ins and outs of the frontend build pipeline but I didn't immediately see a better way to do it.

sgrimm:
Looks like I need to figure out how Vercel builds the code. Will look at that tomorrow.

karthikbtf:
> Looks like I need to figure out how Vercel builds the code. Will look at that tomorrow.

It's under settings...

tree-location-web – Settings - Vercel 2023-03-22 17-17-08

karthikbtf:
> > Looks like I need to figure out how Vercel builds the code. Will look at that tomorrow.
> 
> It's under settings...
> 
> tree-location-web – Settings - Vercel 2023-03-22 17-17-08

might break other vercel builds but ok for a one-off to test I'm sure

karthikbtf:
> The translation service we're going to use doesn't support TypeScript source files. Move the English strings to a CSV file that has an additional column (currently empty for all our strings) where we can add comments to give additional context to translators.
> 
> We will send the English CSV file out for translation, and get back CSV files in other languages; these will then be checked into the repo.
> 
> There's now a build step that converts the CSV files to JavaScript so they can be imported from the rest of the code as usual. This is done as an explicit one-off step in the CI workflow; in dev environments, the CSV files are monitored for changes and processed on the fly so that edits to strings will be hot-reloaded the same way they were reloaded when the strings table was a TypeScript source file.
> 
> This change doesn't introduce any actual translation workflow steps, and for now, the only locale with a CSV file is US English. But this will enable us to start adding comments to our strings right away.

This approach is great 👍 

Will the initial strings file get auto created in the **dev** environment as part of the chokidar watcher? Or do we need to run `yarn generate-strings` before `yarn start`?

sgrimm:
> Will the initial strings file get auto created in the **dev** environment as part of the chokidar watcher? Or do we need to run `yarn generate-strings` before `yarn start`?

Yes, Chokidar generates an `add` event as it finds each CSV file initially, so the JS gets generated while the dev server is initializing itself before the compilation phase.

sgrimm:
Rebased on current main and updated the CSV file to include `strings-en.ts` changes on main since my initial PR.

sgrimm:
Vercel build worked once, then I pushed the branch again and it started failing with no output. I'm not sure how to diagnose the failure with no error messages or anything. (I already updated the settings to use the new `build-collateral` task.)

karthikbtf:
> Vercel build worked once, then I pushed the branch again and it started failing with no output. I'm not sure how to diagnose the failure with no error messages or anything. (I already updated the settings to use the new `build-collateral` task.)

looks like `btoa` is not found (might only be available in browsers but not in nodejs environments).. I think we need the npm package btoa installed? https://www.npmjs.com/package/btoa

sgrimm:
Ah, thanks. How did you find that? I didn't see an error about it (or anything else).

karthikbtf:
> Ah, thanks. How did you find that? I didn't see an error about it (or anything else).

There was a delay in the logs showing up in Vercel.. have noticed that at times.

sgrimm:
Not sure what's going on with Node memory here. My initial tweak to the memory limit accidentally lowered it instead of raising it, but the latest commit should have bumped it from the default 1.4GB to 2GB and it's still running out. Feels like there's a leak somewhere but it's hard to see how my changes could have introduced one.

karthikbtf:
> Not sure what's going on with Node memory here. My initial tweak to the memory limit accidentally lowered it instead of raising it, but the latest commit should have bumped it from the default 1.4GB to 2GB and it's still running out. Feels like there's a leak somewhere but it's hard to see how my changes could have introduced one.

Anything to do with chokidar? We could try react-scripts instead of craco for running cypress `start:dev`?

sgrimm:
It does indeed seem to be related to Chokidar.

karthikbtf:
> It does indeed seem to be related to Chokidar.

Looks like everything is working now 🎉

sgrimm:
I've been unable to figure out why Chokidar is eating that much memory, but I don't think it's worth spending a bunch of time tracking down right now. There are various bug reports about its memory usage but they're all along the lines of, "It eats lots of memory when I monitor a directory with half a million files."

PR #1311, opened 2023-03-23
yarn script to generate build collateral (for vercel)

karthikbtf:
- we can use this as an entry point to add more commands without having to update vercel build settings


sgrimm (approved):
This makes sense!

PR #1314, opened 2023-03-23
SW-3208 Scroll to Top When Switching Report Pages

charlie-tf:


src/components/Reports/ReportEdit.tsx
```diff
@@ -144,10 +144,22 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
...
+  const goToTop = () => {
+    window.scrollTo({
+      top: 0,
+      behavior: 'smooth',
+    });
+  };
+
+  const switchPages = (doShowAnnual: boolean) => {
+    setShowAnnual(doShowAnnual);
+    goToTop();
```
karthikbtf, line 156: do we need a setTimeout of 0 on the scroll .. wait until setShowAnnual does it re-rendering?

karthikbtf (approved):
looks clear to me 👍

src/components/Reports/ReportEdit.tsx
charlie-tf: It seems to render before scrolling already. I think this is okay

PR #1320, opened 2023-03-27
SW-3202, SW-3204 Handle Missing Planting Site Species Data

charlie-tf:
image


karthikbtf (approved):
👍 
Do you want to increase the width of those inputs a bit? The required field error might fit on the same line.

charlie-tf:
> 👍 Do you want to increase the width of those inputs a bit? The required field error might fit on the same line.

That's a good call. It does look squished as is.

PR #1321, opened 2023-03-28
SW-2874 Use componentized popover in terraware-web

karthikbtf:
- no change in functionality just some cleanup / reuse


src/components/OrganizationsDropdown.tsx
```diff
@@ -69,18 +43,13 @@ export default function OrganizationsDropdown(): JSX.Element {
   return (
     <div>
       <AddNewOrganizationModal open={newOrganizationModalOpened} onCancel={onCloseCreateOrganizationModal} />
-      <IconButton onClick={handleClick} size='small' className={classes.iconContainer}>
-        <p>{selectedOrganization.name}</p>
-        <Icon name='chevronDown' size='medium' className={classes.icon} />
-      </IconButton>
       <PopoverMenu
-        sections={[
+        anchor={<p>{selectedOrganization.name}</p>}
```
constanzauanini, line 47: I'm seeing the name a little bigger now, I think the iconContainer class was giving the fontSize to this <p>

src/components/Species/index.tsx
```diff
@@ -40,8 +40,7 @@ import TooltipLearnMoreModal, {
   TooltipLearnMoreModalData,
 } from 'src/components/TooltipLearnMoreModal';
 import PageHeaderWrapper from 'src/components/common/PageHeaderWrapper';
-import PopoverMenu from '../common/PopoverMenu';
-import { DropdownItem, SortOrder } from '@terraware/web-components';
+import { DropdownItem, Popover as PopoverMenu, SortOrder } from '@terraware/web-components';
```
constanzauanini, line 43: Isn't Popover as PopoverMenu confusing here? Because we have the other component that is PopoverMenu

src/components/OrganizationsDropdown.tsx
karthikbtf: ok i will add a class name for this

src/components/Species/index.tsx
karthikbtf: I don't think so but I can rename it

constanzauanini: But why don't we use the Popover name as in other places? Are we using other Popover here?

karthikbtf: Yes, Popover is also imported :)

karthikbtf: from mui/material

src/components/OrganizationsDropdown.tsx
karthikbtf: the other changes seem ok, they all have explicit font sizes.. good catch

PR #1326, opened 2023-03-29
SW-3177 Fix Positioning of Skrim and Spinner

charlie-tf:
The skrim and spinner that are there when, e.g., uploading or deleting photos were stuck at the top of the page, even when scrolled. To fix this, and make them always visible and blocking interactions, change to `'fixed'` positioning.

image


karthikbtf (approved):
good catch.. looks good on vercel.. tested inventory/withdraw with image uploads.

PR #1330, opened 2023-03-29
SW-3268 Visual Bugfixes

charlie-tf:
- fix margin around report content frame
- fix mobile button layout (stacked instead of horizontal)

image
image

### DIFF
--- a/src/components/Reports/ReportForm.tsx
+++ b/src/components/Reports/ReportForm.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { Container, Grid, Theme, Typography, useTheme } from '@mui/material';
+import { Box, Container, Grid, Theme, Typography, useTheme } from '@mui/material';
 import { Checkbox, Textfield } from '@terraware/web-components';
 import { Report, ReportNursery, ReportPlantingSite } from 'src/types/Report';
 import strings from 'src/strings';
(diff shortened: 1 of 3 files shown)


charlie-tf:
@chudo76 I commented on this in the jira, but please let me know if you want 32px around the content frame. Thanks!

src/components/Reports/ReportForm.tsx
```diff
@@ -311,5 +313,6 @@ export default function ReportForm(props: ReportFormProps): JSX.Element {
         )}
       </Grid>
     </Grid>
+    // </Box>
```
karthikbtf, line 316: remove the comments here and above?

src/components/common/FormBottomBar.tsx
```diff
@@ -90,7 +94,7 @@ export default function FormBottomBar({
         type='passive'
         className={classes.button}
       />
-      <div>
+      <div className={classes.rightButtonGroup}>
```
karthikbtf, line 97: thanks for fixing.. i see it broken in staging for other pages with a form bottom bar.
curious.. when/how did this break?

karthikbtf (approved):
👍

src/components/Reports/ReportForm.tsx
charlie-tf: Oops, thanks

src/components/common/FormBottomBar.tsx
charlie-tf: It's likely i broke it when i was adding the ability to display more buttons in the form

karthikbtf: all cool, glad you caught it, thanks much 🙏

PR #1331, opened 2023-03-29
SW-3344 Prevent Browser Back When Editing Report

charlie-tf:
When editing a report, allowing the back button to fire a navigation can result in work being lost. Here, we disable that action when in edit mode, and display a warning to the user to use the form controls instead. We also refactor the type of the switch pages function in ReportEdit.

image


karthikbtf (approved):
Nice!
What happened when user hit the back button (prior to this change)?

charlie-tf:
> Nice! What happened when user hit the back button (prior to this change)?

Prior to this, they would navigate back to the reports list, whether they were on the quarterly or annual page of the report. But the report wouldn't get unlocked and none of the changes would be saved. This seemed like the simplest approach to make sure they use the form controls and don't lose that work.

PR #1332, opened 2023-03-30
SW-3355 Limit Photos Attached to a Report to 30

charlie-tf:
number of photos to upload is limited to a max value, (30 - existing # of photos)
image

show # of additional photos allowed and also disable button if limit reached
image


src/components/common/SelectPhotos.tsx
```diff
@@ -72,7 +73,7 @@ export default function SelectPhotos(props: SelectPhotosProps): JSX.Element {
     }
 
     if (newFiles.length) {
-      updateSelection([...files, ...newFiles]);
+      updateSelection([...files, ...newFiles].slice(0, maxPhotos));
```
karthikbtf, line 76: Some thoughts here:
1. show the max count allowed with some messaging in this widget (if a max is set)?
2. disable the photo selector button if max was reached?

karthikbtf (approved):
Looks great @charlie-tf !

PR #1333, opened 2023-03-30
SW-3350 SW-3354 Lots of Report Field Validation Fixes

charlie-tf:
- fix numeric fields to not allow a negative number to be input
- allow numeric fields to be cleared out all the way
- validate species table fields and worker fields correctly
- check for null instead of falsey (problems with zeros)

### DIFF
--- a/src/components/Reports/LocationSection.tsx
+++ b/src/components/Reports/LocationSection.tsx
@@ -41,9 +41,11 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
   const isNursery = locationType === 'nursery';
   const isPlantingSite = locationType === 'plantingSite';
 
-  const [paidWorkers, setPaidWorkers] = useState<number | undefined>(location.workers?.paidWorkers);
-  const [femalePaidWorkers, setFemalePaidWorkers] = useState<number | undefined>(location.workers?.femalePaidWorkers);
-  const [volunteers, setVolunteers] = useState<number | undefined>(location.workers?.volunteers);
+  const [paidWorkers, setPaidWorkers] = useState<number | null>(location.workers?.paidWorkers ?? null);
+  const [femalePaidWorkers, setFemalePaidWorkers] = useState<number | null>(
+    location.workers?.femalePaidWorkers ?? null
+  );
+  const [volunteers, setVolunteers] = useState<number | null>(location.workers?.volunteers ?? null);
   const [locationNotes, setLocationNotes] = useState(location.notes ?? '');
 
   const smallItemGridWidth = () => (isMobile ? 12 : 4);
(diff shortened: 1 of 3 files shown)


src/components/Reports/LocationSection.tsx
```diff
@@ -114,7 +116,7 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
   ];
 
   const onEditPlantingSiteReport = (species: PlantingSiteSpecies, fromColumn?: string, value?: any) => {
-    if (fromColumn && value) {
+    if (fromColumn) {
```
charlie-tf, line 119: Allow the species table fields to be cleared if desired.

src/components/Reports/LocationSection.tsx
```diff
@@ -224,9 +226,11 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
               value={(location as ReportNursery).capacity ?? ''}
               minNum={0}
               editable={editable}
-              onChange={(value) => onUpdateLocation('capacity', Math.floor(value as number))}
+              onChange={(value) =>
+                onUpdateLocation('capacity', value === '' ? null : Math.max(0, Math.floor(value as number)))
```
charlie-tf, line 230: Setting to null allows the field to be cleared, otherwise it gets stuck with a zero in the field

src/components/Reports/ReportEdit.tsx
```diff
@@ -193,8 +193,14 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
       snackbar.toastWarning(strings.REPORT_BACK_WARNING);
     });
   }, [snackbar]);
-
   const hasEmptyRequiredFields = (iReport: Report) => {
+    const emptyWorkerField = (location: any) => {
```
charlie-tf, line 197: Use this helper function to check for worker fields being filled out. This was missed before somehow.

src/components/Reports/ReportEdit.tsx
```diff
@@ -212,23 +219,25 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
...
+        plantingSite.species?.some((sp) => sp.totalPlanted === null || sp.mortalityRateInField === null) ?? false;
       return (
         plantingSite.selected &&
-        (!plantingSite.totalPlantingSiteArea ||
-          !plantingSite.totalPlantedArea ||
-          !plantingSite.totalTreesPlanted ||
-          !plantingSite.totalPlantsPlanted ||
-          !plantingSite.mortalityRate ||
-          speciesDataMissing)
+        (plantingSite.totalPlantingSiteArea === null ||
```
charlie-tf, line 234: Check for null instead of falsey. Zero was getting interpreted as an invalid value

src/components/Reports/LocationSection.tsx
karthikbtf: @constanzauanini had a related fix i think, in web-comp

karthikbtf: looks like this could be extracted out into a function and reused in other places?
func(value) => value === '' ? null : Math.max(...)

karthikbtf (approved):
Looks good 👍

src/components/Reports/LocationSection.tsx
charlie-tf: yeah, looks like I have quite a lot of duplication. I can definitely extract the function

PR #1336, opened 2023-03-31
SW-3360 Fix Submission Status

charlie-tf:
- unlock was being called unnecessarily after submitting
  - this kicked the report into InProgress status
  - the BE should not have allowed this, and I will put up a fix for that
- also prettified the submit date display in the table

### DIFF
--- a/src/components/Reports/ReportEdit.tsx
+++ b/src/components/Reports/ReportEdit.tsx
@@ -293,7 +293,6 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
         await updatePhotos(report.id);
         const submitResult = await ReportService.submitReport(reportIdInt);
         if (submitResult.requestSucceeded) {
-          await ReportService.unlockReport(reportIdInt);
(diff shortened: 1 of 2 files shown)


src/components/Reports/ReportEdit.tsx
```diff
@@ -293,7 +293,6 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
         await updatePhotos(report.id);
         const submitResult = await ReportService.submitReport(reportIdInt);
         if (submitResult.requestSucceeded) {
-          await ReportService.unlockReport(reportIdInt);
```
karthikbtf, line 296: at this point is it already locked? if so, maybe BE should unlock but without updating the status?

karthikbtf (approved):
one question

src/components/Reports/ReportEdit.tsx
charlie-tf: It gets unlocked in the submit call. This should have been a no-op, but it was changing the status

karthikbtf: got it

PR #1338, opened 2023-04-01
SW-2704 / SW-2705 Update styles of nav and user menus on mobile

alexTerraformation:
Make account menu look more like nav on mobile.
Fix slide-in on navbar, which wasn't happening


charlie-tf (approved):
Needs a format. But looks much nicer—thanks for doing this!

PR #1346, opened 2023-04-04
SW-3367 Error Toast and Scroll to Error

charlie-tf:
- omit showing the changes saved toast when switching pages for an error
- add ids to all fields and scroll to the first id with an error

### DIFF
--- a/src/components/Reports/ReportEdit.tsx
+++ b/src/components/Reports/ReportEdit.tsx
@@ -58,6 +58,15 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
 
   const [validateFields, setValidateFields] = useState(false);
 
+  const [idInView, setIdInView] = useState('');
+  useEffect(() => {
+    const el = document.getElementById(idInView);
+    if (el) {
+      el.scrollIntoView({ behavior: 'smooth' });
+      setIdInView('');
+    }
+  }, [idInView]);
+
   useEffect(() => {
     const getReport = async () => {
       const result = await ReportService.getReport(reportIdInt);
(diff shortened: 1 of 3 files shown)


src/components/Reports/ReportEdit.tsx
```diff
@@ -58,6 +58,15 @@ export default function ReportEdit({ organization }: ReportEditProps): JSX.Eleme
 
   const [validateFields, setValidateFields] = useState(false);
 
+  const [idInView, setIdInView] = useState('');
+  useEffect(() => {
+    const el = document.getElementById(idInView);
+    if (el) {
+      el.scrollIntoView({ behavior: 'smooth' });
+      setIdInView('');
```
charlie-tf, line 66: reset the `idInView` so that if the same field errors it will get scrolled to again. Otherwise this effect isn't triggered.

src/components/Reports/ReportFormAnnual.tsx
```diff
@@ -561,17 +571,17 @@ type ReportFieldProps = {
...
   const theme = useTheme();
   const classes = useStyles();
   return (
     <>
-      <Typography fontSize='20px' fontWeight={600} marginBottom={theme.spacing(1)}>
+      <Typography id={id} fontSize='20px' fontWeight={600} marginBottom={theme.spacing(1)}>
         {title}
       </Typography>
       <Textfield
+        id={`${id}-input`}
```
charlie-tf, line 583: Setting the `id` prop doesn't actually set the `id` attribute for some reason.

karthikbtf (approved):

PR #1347, opened 2023-04-04
SW-3371 Fix Untranslated Strings

charlie-tf:
- use localized strings for report statuses
- use localized strings for species growth forms

image
image


src/components/Reports/PlantingSitesSpeciesCellRenderer.tsx
```diff
@@ -47,6 +47,28 @@ export default function PlantingSiteSpeciesCellRenderer({ editMode, validate }:
       }
     };
 
+    if (column.key === 'growthForm') {
```
karthikbtf, line 50: move the mapping into the types as a util?
example: https://github.com/terraware/terraware-web/blob/main/src/types/Accession.ts#L33
in case it helps with reuse later

karthikbtf (approved):
lgtm, one thought but not required

src/components/Reports/PlantingSitesSpeciesCellRenderer.tsx
karthikbtf: Changes look good, thanks!!

PR #1350, opened 2023-04-04
SW-3272 Use Phrase export format for English strings

sgrimm:
Phrase's GitHub export includes the English strings file as well as the files
for other languages. To avoid cluttering Phrase's PRs with no-op edits to the
English strings file, update the file to match what would be exported by Phrase.

This differs from the previous version of the file in two main ways:

1. The comment column is omitted for strings without comments. That is, if a
   string has a comment, its CSV row will have 3 columns. If a string does not
   have a comment, its CSV row will have 2 columns.
2. Underscores are treated as spaces when Phrase sorts the strings, so they
   come earlier than letters. This differs from the "sort by ASCII" sort order,
   which sorts underscores between the upper and lower case letters.

### DIFF
--- a/src/strings/csv/en.csv
+++ b/src/strings/csv/en.csv
@@ -1,1153 +1,1166 @@
-Key,English,Comments
-ACCEPT,Accept,
-ACCESSION,Accession,
-ACCESSIONS,Accessions,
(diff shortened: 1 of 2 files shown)


sgrimm:
No actual strings should be changed here, but removing the trailing comma from most of them and shuffling some of them around to match Phrase's sort order basically means every line in `en.csv` was edited.

I produced the new `en.csv` by exporting to a test branch from Phrase and copying the file from its branch.

PR #1351, opened 2023-04-04
SW-3272 Include all English keys in JS strings

sgrimm:
If we have CSV files for multiple locales, we generate a JavaScript source file
with the strings for each locale. Previously, this was just an alternate rendering
locale's CSV file.

Unfortunately, that simple approach means that if you add or remove an English
string, the JS strings objects for the other locales won't have the same keys as
the English one. At best, this will cause a TypeScript error because the type
signatures don't match; at worst, it'll be a runtime error.

Instead, generate each locale's JS file using both the English and localized CSV
files. The list of keys will always be based on the English file. The values will
be the localized strings if available, or the original English strings if not,
such that the app effectively falls back to English for strings that aren't
translated yet.

Print a warning for each untranslated string; that'll let us easily check the CI
build logs before a release to see if we're missing any translations.

### DIFF
--- a/src/strings/export.js
+++ b/src/strings/export.js
@@ -83,8 +83,8 @@ async function convertCsvFile(csvPath, targetDir) {
   if (locale === 'en') {
     englishStringsMap = stringsMap;
   } else {
-    const englishPath = path.resolve(path.dirname(csvPath), "en.csv");
-    const englishCsvData = await fs.readFile(englishPath, { encoding: 'utf-8'});
+    const englishPath = path.resolve(path.dirname(csvPath), 'en.csv');
+    const englishCsvData = await fs.readFile(englishPath, { encoding: 'utf-8' });
     englishStringsMap = csvToStrings(englishCsvData);
   }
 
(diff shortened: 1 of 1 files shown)


src/strings/export.js
```diff
@@ -41,7 +41,18 @@ function stringsToJS(stringsMap) {
...
+async function exportStrings(englishStrings, localizedStrings, locale, targetDir) {
+  const stringsMap = {};
+  for (let key in englishStrings) {
+    if (key in localizedStrings) {
+      stringsMap[key] = localizedStrings[key];
+    } else {
+      // tslint:disable-next-line:no-console
+      console.warn(`Locale ${locale} has no translation for ${key}`);
+      stringsMap[key] = englishStrings[key];
+    }
```
karthikbtf, line 53: should we warn if there are extra keys in the localizedStrings?

karthikbtf (approved):
lgtm, thanks for doing this!

src/strings/export.js
sgrimm: I think it's harmless if we still have translations for strings we've removed from the English strings file. With this change, they won't be included in the generated JS.

There are questions about deleting strings that I need to address more broadly, but I don't think they'll impact this code. (Basically: how and when strings get completely removed from Phrase. It needs to involve some kind of delay because removing a string from main shouldn't cause us to be unable to do localized patch releases if the prod code still needs the string.)

PR #1352, opened 2023-04-05
SW-3272 Remove translatable punctuation strings

sgrimm:
Phrase doesn't handle punctuation-only strings very well; orders that include them
never get marked as completed, and it's unclear whether or not they ever get sent
to translators. Also, Phrase's GitHub export sometimes (though not always) trims
trailing whitespace from strings.

Remove the three delimiter strings from the strings CSV file and hardwire them into
the code that generates the JavaScript string objects. The delimiters are the same
in English and Spanish, so include delimiters for gibberish (to test that we're
using localized delimiters) and Chinese (to test that we're not assuming delimiters
include trailing whitespace).


karthikbtf (approved):
Thanks!!

PR #1358, opened 2023-04-07
SW-3408 Keep "Add Batch" button on screen

alexTerraformation:


src/components/Inventory/view/InventorySeedlingsTable.tsx
```diff
@@ -243,6 +243,8 @@ export default function InventorySeedslingsTable(props: InventorySeedslingsTable
             justifyContent: 'space-between',
             alignItems: 'center',
             marginBottom: theme.spacing(4),
+            width: (isDesktop ? 'calc(100vw - 300px)' : 'calc(100vw - 76px)'),
```
alexTerraformation, line 246: Open to alternate suggestions on this approach.  We use this pattern in a couple other places, but I don't love how fragile it is.

karthikbtf: We could move the table inside another horizontally scrollable div perhaps - or maybe that's how it was at the beginning?
cc @constanzauanini / @charlie-tf I know we tried some experiments.. I recall calling out that these call-to-action buttons will scroll out of view, if we made the entire page scroll with the table. I think maybe PMs were fine with that.

karthikbtf (approved):
I guess this is fine for now. Might be good to revisit why we want the entire page to scroll with the table.

src/components/Inventory/view/InventorySeedlingsTable.tsx
alexTerraformation: IIRC, @nicolewilke made the call on the horizontal scroll.  Perhaps it is something we can revisit since that decision was made though.  I'll merge this for now and bring it up with PMs.

karthikbtf: Sounds good 👍

PR #1359, opened 2023-04-07

terraware-phrase:


sgrimm:
@karthikbtf This is me exporting strings after copying the backend accession state translations over. We'll have to remember to keep them in sync if/when we decide we don't like the translations for any of them and ask for new ones.

karthikbtf (approved):
Thanks @sgrimm , this does cover the list of hardcoded states for url purposes.

PR #1366, opened 2023-04-11
SW-3430 Fix species template download and species import resolutions

karthikbtf:
- the response needed to be parsed for csv template
- the payload for import resolutions needed to be an entity instead of search param


charlie-tf (approved):

karthikbtf:
Thanks @charlie-tf

PR #1368, opened 2023-04-13
SW-3434 Show Linebreaks in Display Text

charlie-tf:
## Example:
image

### DIFF
--- a/src/components/Reports/ReportForm.tsx
+++ b/src/components/Reports/ReportForm.tsx
@@ -23,6 +23,11 @@ const useStyles = makeStyles((theme: Theme) => ({
     marginLeft: theme.spacing(3),
     borderBottom: `solid 1px ${theme.palette.TwClrBrdrTertiary}`,
   },
+  textfieldDisplay: {
+    '& p': {
+      whiteSpace: 'pre-line',
+    },
+  },
 }));
 
 export type ReportFormProps = {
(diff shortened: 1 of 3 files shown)


src/components/Reports/ReportForm.tsx
```diff
@@ -23,6 +23,11 @@ const useStyles = makeStyles((theme: Theme) => ({
     marginLeft: theme.spacing(3),
     borderBottom: `solid 1px ${theme.palette.TwClrBrdrTertiary}`,
   },
+  textfieldDisplay: {
+    '& p': {
+      whiteSpace: 'pre-line',
+    },
+  },
```
karthikbtf, line 30: Interesting.. wondering if this should be part of the default css in web-components - the use-case seems legit for all textarea values in display mode.
Thoughts @charlie-tf @constanzauanini ?

karthikbtf (approved):

src/components/Reports/ReportForm.tsx
charlie-tf: Oh you're right. That might be better than sprinkling this custom css here and there. Are there any use cases where we'd want to collapse all the newlines? I could add it as a prop to `Textfield` 🤔

constanzauanini: Yes, we could add it to the textfield (textarea type) component

PR #1369, opened 2023-04-13
SW-3434 Preserve Newlines in Report Textfields

charlie-tf:
## Example
image


karthikbtf (approved):
Nice 🎉 
Thanks @charlie-tf for considering the feedback.

PR #1370, opened 2023-04-13
SW-3433 Save SeedBank and Nursery Info on Submit

charlie-tf:
Must reload org after submitting for changes in https://github.com/terraware/terraware-server/pull/1080 to appear.


src/components/Reports/ReportEdit.tsx
```diff
@@ -31,11 +30,8 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export type ReportEditProps = {
-  organization: Organization;
```
karthikbtf, line 35: Thanks 🙏 the changes are much cleaner.

karthikbtf (approved):
Nice changes, the flow seems straightforward.

PR #1380, opened 2023-04-24
Update typescript to 5.0.0

charlie-tf:
Like it says in the title


karthikbtf (approved):
Thanks for this upgrade 🙏 
We should think about which new features we can use..  [some slack context](https://terraformationteam.slack.com/archives/C01C69KA23Y/p1678998752933709) around

PR #1384, opened 2023-04-28
SW-3478 Allow user to delete account in terraware-web

karthikbtf:
- add delete account option in MyAccount
- delete user and send to logout


src/components/MyAccount/DeleteAccountModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+            icon='iconTrashCan'
+            type='destructive'
+            onClick={deleteUser}
+            size='medium'
+            key='button-2'
+          />,
+        ]}
+        skrim={true}
+      >
+        <p className={classes.mainContent}>{strings.DELETE_ACCOUNT_CONFIRMATION}</p>
```
constanzauanini, line 67: DialogBox has a message property, not sure if that would work for this

constanzauanini (approved):

src/components/MyAccount/DeleteAccountModal.tsx
karthikbtf: Let me try.. I copied this from another modal.

karthikbtf: That works, thanks! I should probably update the other modals

karthikbtf: The only difference is it uses less padding, maybe that's ok?

using-custom-content

using-message-property

constanzauanini: Yes, maybe it's ok. Dialog box uses styles specified in zero height

karthikbtf: Awesome, cool!

constanzauanini (approved):
Thanks for fixing the others too

karthikbtf:
Updated to use the more-options menu style as per @chudo76 's request:

Terraware App 2023-05-01 09-22-39

Terraware App 2023-05-01 09-22-24

@chudo76 should this be in red color?

src/components/common/OptionsMenu.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Button, DropdownItem, Popover, Tooltip } from '@terraware/web-components';
+import strings from 'src/strings';
+
+export type OptionsMenuProps = {
+  optionItems: DropdownItem[];
+  onOptionItemClick?: (optionItem: DropdownItem) => void;
+  size?: 'medium' | 'small';
+};
+
+export default function OptionsMenu({ optionItems, onOptionItemClick, size }: OptionsMenuProps): JSX.Element {
```
karthikbtf, line 12: Extracted the OptionsMenu into a separate component since it was being implemented in every use-case.

chudo76:
> @chudo76 should this be red in color?

Thanks, that is a good call.

There is a separate issue of the menu positioning. I think we want either a right-aligned or left-aligned depending on the position of the kebab icon button, for _all_ popover menus. Would that be a separate issue?

image
vs.
image

karthikbtf:
> > @chudo76 should this be red in color?
> 
> Thanks, that is a good call.
> 
> There is a separate issue of the menu positioning. I think we want either a right-aligned or left-aligned depending on the position of the kebab icon button, for _all_ popover menus. Would that be a separate issue?
> 
> image
> 
> vs.
> image

@chudo76 yes, this is an existing implementation in the codebase. I sent out a PR with fix https://github.com/terraware/web-components/pull/360 in web-components to address this

also updated the colors of the delete buttons in these options menus
Terraware App 2023-05-01 14-25-16

Terraware App 2023-05-01 14-24-58

karthikbtf:
> > > @chudo76 should this be red in color?
> > 
> > 
> > Thanks, that is a good call.
> > There is a separate issue of the menu positioning. I think we want either a right-aligned or left-aligned depending on the position of the kebab icon button, for _all_ popover menus. Would that be a separate issue?
> > image
> > vs.
> > image
> 
> @chudo76 yes, this is an existing implementation in the codebase. I sent out a PR with fix [terraware/web-components#360](https://github.com/terraware/web-components/pull/360) in web-components to address this
> 
> also updated the colors of the delete buttons in these options menus Terraware App 2023-05-01 14-25-16
> 
> Terraware App 2023-05-01 14-24-58

Ok, the delete buttons now show in red color. The alignment is not centered anymore but not perfectly left or right, may need to use another component to achieve that. We can do that out-of-scope from this ticket.

PR #1390, opened 2023-05-01
SW-3126 Allow searching species by common name

karthikbtf:
- updated the search payload to also look up common name
- tested with a new species that also had diacritics in the name


charlie-tf (approved):
👍

PR #1396, opened 2023-05-04
Updates to Use Photo Viewer from web-components

charlie-tf:
image
image


src/components/common/ViewPhotosModal.tsx
```diff
@@ -1,108 +0,0 @@
...
-  const myCarousel = useRef<Carousel>(null);
-  const responsive = {
-    mobile: {
-      breakpoint: { max: 4000, min: 0 },
-      items: 1,
-    },
-  };
-
-  const disableButtons = () => {
-    if (myCarousel.current) {
```
constanzauanini, line 30: I think it was a requirement to disable the next and previous button, is this still working?

charlie-tf: Oh interesting. I transferred this code directly over to the web-components version, so it should be working the same. I'll look into it. Thanks!

charlie-tf: It only works _after_ clicking one of those buttons. I may have messed this up—I did some things with the dialog state that might have changed this. Thanks for catching it

karthikbtf (approved):
Thanks for the cleanup 👍 
Besides Constanza's comments.. looks good. Nice dandelion!!

src/components/common/ViewPhotosModal.tsx
karthikbtf: I don't know if it was a requirement, I had suggested this behavior when Constanza wrote this code.

PR #1398, opened 2023-05-04

terraware-phrase:


karthikbtf (approved):
For the ViewPhotos dialog

PR #1399, opened 2023-05-04
SW-3529 Update accessions columns configuration 

karthikbtf:
- We have a [new list and order](https://docs.google.com/spreadsheets/d/163ZfE71_o-iRHjKFF4Uoboc8oEdNDpml0PXLBpus03Y/edit#gid=0) from Ming 
- Removed and updated the order as per Ming's request
- also persist columns when they are reordered
- add viability percent as new column
- there's an outstanding task to keep the column order when columns are added/removed (don't append to the end), will be a follow-up
- best to preview this in vercel to see changes

### DIFF
--- a/cypress/e2e/10-database.cy.js
+++ b/cypress/e2e/10-database.cy.js
@@ -123,20 +123,21 @@ describe('Database', () => {
         cy.get('#more-options').click();
         cy.get('.MuiList-root > :nth-child(3)').click();
 
-        cy.get('#General\\ Inventory').click();
         cy.get('#Default').click();
         cy.get('#saveColumnsButton').click();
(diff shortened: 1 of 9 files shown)


karthikbtf:
The e2e tests will fail for now, I will clean them up shortly.

constanzauanini (approved):
Looks good! Tested in vercel

constanzauanini:
> > Looks good! Tested in vercel
> 
> Thanks @constanzauanini ! I just showed it to @minqingzhou - need to remove 'active/inactive' from export and move 'count' into the weights category and call that 'Quantity'. Small change coming.

Just checked changes and looks ok!

PR #1401, opened 2023-05-05
SW-3541 Preserve system order of accession columns where possible

karthikbtf:
- in a previous commit, we added support to persist shuffled columns
- same commit introduced the new/updated preferred system order of column names
- in this commit, we try to preserve the shuffled columns and system order when columns are added/removed.. as much as possible using simple math
- approach is documented in the code
- a) when editing columns, preserve the current order (minus removed columns)
- b) determine which columns were shuffled and keep track of shuffled index
- b) sort new columns by order of appearance in the preferred system order, remove shuffled columns from this list
- c) re-insert shuffled columns at shuffled index
- this works well for most common cases without requiring us to persist any more information, after many attempts i like the latest version of the algorithm

### DIFF
--- a/cypress/e2e/10-database.cy.js
+++ b/cypress/e2e/10-database.cy.js
@@ -58,11 +58,11 @@ describe('Database', () => {
       cy.get('#table-header').children().should('have.length', 7);
       cy.get('#table-header > :nth-child(1)').contains('ACCESSION');
       cy.get('#table-header > :nth-child(2)').contains('STATUS');
-      cy.get('#table-header > :nth-child(7)').contains('SEED BANKS');
-      cy.get('#table-header > :nth-child(3)').contains('COLLECTING SITE');
-      cy.get('#table-header > :nth-child(4)').contains('AGE (MONTH');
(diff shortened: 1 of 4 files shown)


karthikbtf:
@chudo76 give this preview a try

charlie-tf (approved):
After staring at this for some time, I think I know what's going on. Thanks for the helpful comments.

karthikbtf:
> After staring at this for some time, I think I know what's going on. Thanks for the helpful comments.

Thanks @charlie-tf , let me see if i can clean up the comments a bit more with examples.

karthikbtf:
> > After staring at this for some time, I think I know what's going on. Thanks for the helpful comments.
> 
> Thanks @charlie-tf , let me see if i can clean up the comments a bit more with examples.

Added more comments and an example. Will get this in for now, if we find an easier way I can certainly update it then.

chudo76 (approved):
Tricky logic but LGTM!

karthikbtf:
@charlie-tf @chudo76 I reverted this on main. Feel like this can be improved and should. Will revisit with help.

PR #1402, opened 2023-05-08
SW-3540 Auto save table columns order in all tables

karthikbtf:
- tables in terraware-web can be reshuffled
- added a wrapper component around the base table to save the order of columns upon reshuffling, in user preferences for the specific org
- added another wrapper with some default implementation,this wrapper component is now the default table as it is the most common use-case
- the species table uses a the original wrapper with some overridden implementation
- the accessions table does not use these wrapper, it has its own extended/customized behavior

### DIFF
--- a/src/components/Species/index.tsx
+++ b/src/components/Species/index.tsx
@@ -4,7 +4,7 @@ import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { useRecoilState } from 'recoil';
 import Button from 'src/components/common/button/Button';
 import EmptyMessage from 'src/components/common/EmptyMessage';
-import Table from 'src/components/common/table';
+import { OrderPreserveableTable as Table } from 'src/components/common/table';
 import { TableColumnType } from 'src/components/common/table/types';
 import speciesAtom from 'src/state/species';
 import strings from 'src/strings';
@@ -48,6 +48,7 @@ import FilterGroup, { FilterField } from 'src/components/common/FilterGroup';
 import { FieldOptionsMap } from 'src/services/NurseryWithdrawalService';
 import { SpeciesService } from 'src/services';
 import OptionsMenu from 'src/components/common/OptionsMenu';
+import _ from 'lodash';
 
 type SpeciesListProps = {
   reloadData: () => void;
@@ -626,7 +627,7 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps): 
       if (hasErrors?.length) {
         newColumns = [problemsColumn, ...columns];
       }
-      if (selectedColumns[0].key !== newColumns[0].key) {
+      if (!_.isEqual(selectedColumns, newColumns)) {
         setSelectedColumns(newColumns);
       }
       setHandleProblemsColumn(false);
@@ -819,6 +820,13 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps): 
             <Grid item xs={12}>
               {results && (
                 <Table
+                  setColumns={(columnsToSet: TableColumnType[]) => {
+                    // show the check-data error column only if it was already exposed
+                    const showProblemsColumn = selectedColumns.find((column) => column.key === problemsColumn.key);
+                    setSelectedColumns(
+                      columnsToSet.filter((column) => showProblemsColumn || column.key !== problemsColumn.key)
+                    );
+                  }}
                   id='species-table'
                   columns={selectedColumns}
                   rows={results}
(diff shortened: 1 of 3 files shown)


src/components/Species/index.tsx
```diff
@@ -819,6 +820,13 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
             <Grid item xs={12}>
               {results && (
                 <Table
+                  setColumns={(columnsToSet: TableColumnType[]) => {
```
karthikbtf, line 823: species table overrides the implementation to set columns upon reshuffling (it filters out error column table if it is not being shown)

src/components/common/table/index.tsx
```diff
@@ -28,3 +32,90 @@ export default function Table<T extends TableRowType>(props: TableProps<T>): JSX
...
+ * Ordered columns component that preserves reordered columns using callback functions.
+ * This table fetches column names from preferences and sets them if available.
+ * Also saves new columns order upon reordering, into user preferences (for the org scope).
+ */
+
+export type OrderPreserveableTableProps = {
+  setColumns: (columns: TableColumnType[]) => void;
+};
+
+export function OrderPreserveableTable<T extends TableRowType>(
```
karthikbtf, line 46: this table listens for reorder changes (columns being shuffled) and saves them in the user preference..
on initialization it uses these saved columns (if available) to set the columns order (using a callback - which can have different implementations, species table has its own)

src/components/common/table/index.tsx
```diff
@@ -18,7 +22,7 @@ interface TableProps<T> extends Omit<Props<T>, keyof LocalizationProps> {
   showPagination?: boolean;
 }
 
-export default function Table<T extends TableRowType>(props: TableProps<T>): JSX.Element {
+export function BaseTable<T extends TableRowType>(props: TableProps<T>): JSX.Element {
```
karthikbtf, line 25: accessions table still uses this

src/components/common/table/index.tsx
```diff
@@ -28,3 +32,90 @@ export default function Table<T extends TableRowType>(props: TableProps<T>): JSX
...
+    onReorderEnd: reorderHandler,
+  });
+}
+
+/**
+ * Ordered columns table with implementation to set columns - this handles the most common cases
+ * where client does not need extra semantics to filter columns.
+ * Species table is an example where this won't work, uses its own setColumns implementation.
+ */
+export default function OrderPreservedTable<T extends TableRowType>(props: TableProps<T>): JSX.Element {
```
karthikbtf, line 112: This is the new default table in terraware web. It uses the wrapped table that saves/initializes-from preferences.. with a default implementation to set the columns in state. Nothing fancy but doing it this way allows us to avoid changing any of the existing tables that conform to this generic use-case (no special semantics in setting columns order).

karthikbtf:
@chudo76 give this preview a try as well (this is for non-accessions tables)

chudo76 (approved):
Columns orders for tables are being preserved. Thanks!

karthikbtf:
> Columns orders for tables are being preserved. Thanks!

Awesome, thanks @chudo76

karthikbtf:
@chudo76 this PR now supports shuffling columns for all tables (which you already tested) - and, sorts accessions columns by system order (when reconfigured). Columns won't be added to the end but user may need to reshuffle columns again.
I will create a separate ticket out-of-band of current parent JIRA to discuss supporting both, if we need it later.

src/components/common/table/index.tsx
```diff
@@ -28,3 +32,94 @@ export default function Table<T extends TableRowType>(props: TableProps<T>): JSX
...
+      onReorderEnd(reorderedColumns);
+    }
+  };
+
+  useEffect(() => {
+    const fetchSavedColumns = () => {
+      if (!orgPreferences) {
+        return;
+      }
+      setInitialized(true);
```
charlie-tf, line 82: Minor nitpick, but I wondered about setting initialized to true before finishing the initialization? It's unlikely that any of the subsequent steps would fail, though, since you programmed it defensively.

charlie-tf (approved):
Looks great!

src/components/common/table/index.tsx
karthikbtf: Good call, for readability I can move this statement to the end after all the checks have completed. That is the correct logical order anyways.

karthikbtf (commented):
Thanks for the review @charlie-tf

charlie-tf (approved):

PR #1404, opened 2023-05-09
SW-3409 Save nursery capacity on new nursery creation

karthikbtf:
- pass the capacity in the api payload, it was missed somehow


karthikbtf:
@charlie-tf this should be a quick one.. so I can send it for QA

charlie-tf (approved):
Thanks

PR #1405, opened 2023-05-09
SW-3440 Fix change status/quantities modal in seedlings batches when the numbers are large

karthikbtf:
- code was using the localized display value (with commas) for math instead of the raw values..
- switched to raw
- also fixed some UX issues like error validation and spinner

### DIFF
--- a/src/components/Inventory/view/BatchesCellRenderer.tsx
+++ b/src/components/Inventory/view/BatchesCellRenderer.tsx
@@ -75,13 +75,14 @@ export default function BatchesCellRenderer(props: RendererProps<TableRowType>):
         row={row}
         value={
           <>
-            <ChangeQuantityModal
-              open={modalValues.openChangeQuantityModal}
-              onClose={() => setModalValues({ openChangeQuantityModal: false, type: 'germinating' })}
-              modalValues={modalValues}
-              row={row as Batch}
-              reload={onRowClick}
-            />
+            {modalValues.openChangeQuantityModal && (
+              <ChangeQuantityModal
+                onClose={() => setModalValues({ openChangeQuantityModal: false, type: 'germinating' })}
+                modalValues={modalValues}
+                row={row as Batch}
+                reload={onRowClick}
+              />
+            )}
             <QuantitiesMenu setModalValues={setModalValues} batch={row} />
           </>
         }
--- a/src/components/Inventory/view/ChangeQuantityModal.tsx
+++ b/src/components/Inventory/view/ChangeQuantityModal.tsx
@@ -1,32 +1,48 @@
-import React, { useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import strings from 'src/strings';
 import Button from 'src/components/common/button/Button';
 import DialogBox from 'src/components/common/DialogBox/DialogBox';
 import { Box, Grid } from '@mui/material';
-import { Icon, Textfield } from '@terraware/web-components';
+import { BusySpinner, Icon, Textfield } from '@terraware/web-components';
 import { ModalValuesType } from './BatchesCellRenderer';
-import { Batch } from 'src/types/Batch';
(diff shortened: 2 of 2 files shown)


karthikbtf:
@chudo76 give this a try, also showing a spinner on save now.. and if you save without an entry we flag that now.

charlie-tf (approved):
Mostly makes sense. Just curious why the modal was made to unmount/remount instead of controlling the open prop

karthikbtf:
> Mostly makes sense. Just curious why the modal was made to unmount/remount instead of controlling the open prop

Thanks @charlie-tf  - in case it helps, the "xyz(raw)" properties are the actual numbers (as provided by terraware-server search). The "xyz" properties are localized for display only. We had switched to using the '(raw)' properties sometime back but missed this modal. We cannot do "30,000 - 5" in javascript, it yields 'NaN'. The 'raw' value is '30000' without the comma.

Re: the open boolean property, there are two reasons for why I did it this way:

1) Here specifically, we have one of them per row in the table. The cost of having one mounted but invisible per row is more expensive that mounting/unmounting one at a time upon menu click.

2) I have been switching to this approach every where we have modals - the cost of unmount/remount is less than ensuring a cleaned up modal - which often they aren't.. just buggy. It it was one modal that's fine.. we have a lot of modals with a lot of content.. and often with too much context (when the modals should be separated by context). A lot of bugs have been uncovered, I wonder if the boolean open pattern is best for static content.

charlie-tf:
> > Mostly makes sense. Just curious why the modal was made to unmount/remount instead of controlling the open prop
> 
> Thanks @charlie-tf - in case it helps, the "xyz(raw)" properties are the actual numbers (as provided by terraware-server search). The "xyz" properties are localized for display only. We had switched to using the '(raw)' properties sometime back but missed this modal. We cannot do "30,000 - 5" in javascript, it yields 'NaN'. The 'raw' value is '30000' without the comma.
> 
> Re: the open boolean property, there are two reasons for why I did it this way:
> 
> 1. Here specifically, we have one of them per row in the table. The cost of having one mounted but invisible per row is more expensive that mounting/unmounting one at a time upon menu click.
> 2. I have been switching to this approach every where we have modals - the cost of unmount/remount is less than ensuring a cleaned up modal - which often they aren't.. just buggy. It it was one modal that's fine.. we have a lot of modals with a lot of content.. and often with too much context (when the modals should be separated by context). A lot of bugs have been uncovered, I wonder if the boolean open pattern is best for static content.

Great reasons! Thanks for taking the time to explain it to me :)

karthikbtf:
> > > Mostly makes sense. Just curious why the modal was made to unmount/remount instead of controlling the open prop
> > 
> > 
> > Thanks @charlie-tf - in case it helps, the "xyz(raw)" properties are the actual numbers (as provided by terraware-server search). The "xyz" properties are localized for display only. We had switched to using the '(raw)' properties sometime back but missed this modal. We cannot do "30,000 - 5" in javascript, it yields 'NaN'. The 'raw' value is '30000' without the comma.
> > Re: the open boolean property, there are two reasons for why I did it this way:
> > 
> > 1. Here specifically, we have one of them per row in the table. The cost of having one mounted but invisible per row is more expensive that mounting/unmounting one at a time upon menu click.
> > 2. I have been switching to this approach every where we have modals - the cost of unmount/remount is less than ensuring a cleaned up modal - which often they aren't.. just buggy. It it was one modal that's fine.. we have a lot of modals with a lot of content.. and often with too much context (when the modals should be separated by context). A lot of bugs have been uncovered, I wonder if the boolean open pattern is best for static content.
> 
> Great reasons! Thanks for taking the time to explain it to me :)

Thanks :) it was a good call-out we may still need to address. Once Constanza is back we can bring this up in a discussion. For consistency we need to pick one over the other, evaluate costs..etc. maybe start a styleguide/best-practices-doc at some point.

PR #1406, opened 2023-05-09
SW-3553 Replace Arrows on Photo Dialog Buttons

charlie-tf:
image

(adding a loading spinner will come next, but in web-components)


charlie-tf:
> A suggestion on that slack thread from yesterday is to use icons instead of text for < and > This might be an opportunity? add it in web-components perhaps

Oh, I overlooked that bit of the discussion. I'll take a look at adding icons 👍

karthikbtf (approved):
Thanks 👍 @charlie-tf can we do this for funder web as well?

PR #1407, opened 2023-05-09
SW-3431 Re-orient organization-created message content

karthikbtf:
- refer to correct position of menus

Screen Shot 2023-05-09 at 9 41 07 AM

Screen Shot 2023-05-09 at 9 41 55 AM


charlie-tf (approved):
Nice improvement!

PR #1408, opened 2023-05-09

terraware-phrase:

### DIFF
--- a/src/strings/csv/es.csv
+++ b/src/strings/csv/es.csv
@@ -618,6 +618,7 @@ ORG_NOTIFICATION_ACTION,"Para cambiar la preferencia, vaya a [""Ajustes > Organi
 ORGANIZATION,Organización
 ORGANIZATION_CREATE_FAILED,No se puede crear la organización.
 ORGANIZATION_CREATED_MSG,"Puede acceder a las organizaciones a las que pertenece haciendo clic en la flecha de la esquina superior derecha, junto a su perfil."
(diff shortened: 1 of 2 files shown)


src/strings/csv/es.csv
```diff
@@ -618,6 +618,7 @@ ORG_NOTIFICATION_ACTION,"Para cambiar la preferencia, vaya a [""Ajustes > Organi
 ORGANIZATION,Organización
 ORGANIZATION_CREATE_FAILED,No se puede crear la organización.
 ORGANIZATION_CREATED_MSG,"Puede acceder a las organizaciones a las que pertenece haciendo clic en la flecha de la esquina superior derecha, junto a su perfil."
```
karthikbtf, line 620: @sgrimm  - the `ORGANIZATION_CREATED_MSG` message was updated as well but not reflected here in the spanish changes. 
Not a new string, just an update.
Is there something we can do in the Phrase UI to trigger an update?

karthikbtf (approved):

src/strings/csv/es.csv
sgrimm: Ah, this was my mistake -- I hadn't set English as the "main language" for terraware-web in Phrase, so updates to English strings weren't causing the Spanish translations to be included in orders. Fixed now, and I'll manually place an order to cover the recently-modified strings.

karthikbtf: Great, thanks Steve 🙏

sgrimm: Done. The automator should place an order for these strings in the next hour or so.

PR #1410, opened 2023-05-11
SW-3070 Update Exact Matching in Accessions Search

charlie-tf:
Set "exact or fuzzy" search for accessions table "Accession Number" and "Collection Site Name" columns.
Exact search now matches exact substrings.

image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@mui/styled-engine-sc": "^5.12.0",
     "@mui/styles": "^5.11.9",
     "@reduxjs/toolkit": "^1.9.3",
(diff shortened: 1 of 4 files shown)


karthikbtf (approved):
lgtm 👍 
Guessing this only affects the accessions table, other tables don't have similar search/type use-cases.

charlie-tf:
> lgtm 👍 Guessing this only affects the accessions table, other tables don't have similar search/type use-cases.

Correct. This is the place where the updates were requested. Can add to other places as needed and as feedback is received.

PR #1412, opened 2023-05-19
SW-3626 Storybook Integration

charlie-tf:
- Added storybook using `npx storybook init`
- Correctly initialize theme in preview.tsx
- Add a story for the TfMain component to prove things are working

image

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,0 +1,28 @@
+import type { Preview } from '@storybook/react';
+import { StyledEngineProvider, ThemeProvider } from '@mui/material';
+
+const preview: Preview = {
+  parameters: {
+    actions: { argTypesRegex: '^on[A-Z].*' },
+    controls: {
+      matchers: {
+        color: /(background|color)$/i,
+        date: /Date$/,
+      },
+    },
+  },
+};
+
+export default preview;
+
+import theme from '../src/theme';
+
+export const decorators = [
+  (Story) => (
+    <ThemeProvider theme={theme}>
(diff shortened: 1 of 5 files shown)


charlie-tf:
Ack, the build is broken for the app now 😭 digging into this

.storybook/preview.tsx
```diff
@@ -0,0 +1,28 @@
...
+        color: /(background|color)$/i,
+        date: /Date$/,
+      },
+    },
+  },
+};
+
+export default preview;
+
+import theme from '../src/theme';
```
karthikbtf, line 18: nice 👍

karthikbtf (commented):
Looks great, nice upgrade for us and enables new workflows, thanks!
The build issue sure is strange.. not sure why the node version confusion.

karthikbtf (approved):
🎉

charlie-tf:
So the webpack dependency addition was causing a different error locally. Taking that out solved local build. However, the Vercel deploy still failed. I updated the node version Vercel is using to 16 and attempting to redeploy...

karthikbtf:
> So the webpack dependency addition was causing a different error locally. Taking that out solved local build. However, the Vercel deploy still failed. I updated the node version Vercel is using to 16 and attempting to redeploy...

Ok 🤞

charlie-tf:
> So the webpack dependency addition was causing a different error locally. Taking that out solved local build. However, the Vercel deploy still failed. I updated the node version Vercel is using to 16 and attempting to redeploy...

And it worked! 🎉

PR #1415, opened 2023-05-24
SW-3345 bump web-components to pick up date format change

karthikbtf:


karthikbtf:
Or we can wait for the PopoverMultiSelect version which I'm guessing will be very soon...

charlie-tf (approved):
Go ahead and bump now so this can bake. I'll update for the popover stuff when I build the MapLayerSelect component

PR #1421, opened 2023-05-25
SW-3666 Add planting site selector to observations page

karthikbtf:
- Plants dashboard and plants observations have the same page structure
- extracted out plants dashboard page into a wrapper page component
- reused that for observations
- added a concept of 'All' as a selection
- added a story for the page view

PlantsPrimaryPageView - Default ⋅ Storybook 2023-05-26 10-09-10

PlantsPrimaryPageView - Default ⋅ Storybook 2023-05-26 10-08-54

Terraware App 2023-05-26 09-50-54

### DIFF
--- a/src/components/Observations/index.tsx
+++ b/src/components/Observations/index.tsx
@@ -5,24 +5,31 @@ import { APP_PATHS } from 'src/constants';
 import PlantsPrimaryPage from 'src/components/PlantsPrimaryPage';
 
 export default function PlantsDashboard(): JSX.Element {
+  const [plantingSites, setPlantingSites] = useState<PlantingSite[]>();
   const [selectedPlantingSite, setSelectedPlantingSite] = useState<PlantingSite>();
   const [plantsSitePreferences, setPlantsSitePreferences] = useState<Record<string, unknown>>();
+  const hasObservations = false;
 
   const onSelect = useCallback((site: PlantingSite) => setSelectedPlantingSite(site), [setSelectedPlantingSite]);
+
   const onPreferences = useCallback(
     (preferences: Record<string, unknown>) => setPlantsSitePreferences(preferences),
     [setPlantsSitePreferences]
   );
+
+  const onPlantingSites = useCallback((sites: PlantingSite[]) => setPlantingSites(sites), [setPlantingSites]);
 
   return (
     <PlantsPrimaryPage
       title={strings.OBSERVATIONS}
       onSelect={onSelect}
       pagePath={APP_PATHS.PLANTING_SITE_OBSERVATIONS}
-      lastVisitedPreferenceName='lastObservationsPlantingSite'
+      lastVisitedPreferenceName='plants.observations.lastVisitedPlantingSite'
       plantsSitePreferences={plantsSitePreferences}
       setPlantsSitePreferences={onPreferences}
       allowAllAsSiteSelection={true}
+      onPlantingSites={onPlantingSites}
+      isEmptyState={!plantingSites?.length || !hasObservations}
     >
       <div>placeholder for selected planting site {selectedPlantingSite?.id}</div>
     </PlantsPrimaryPage>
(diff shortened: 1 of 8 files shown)


src/components/Observations/index.tsx
```diff
@@ -1,3 +1,30 @@
...
+    (preferences: Record<string, unknown>) => setPlantsSitePreferences(preferences),
+    [setPlantsSitePreferences]
+  );
+
+  return (
+    <PlantsPrimaryPage
+      title={strings.OBSERVATIONS}
+      onSelect={onSelect}
+      pagePath={APP_PATHS.PLANTING_SITE_OBSERVATIONS}
+      lastVisitedPreferenceName='lastObservationsPlantingSite'
```
charlie-tf, line 22: Thoughts on namespacing these in a structured way to keep things organized? Such as `'<page>.<preferenceName>'`, e.g. `'observations.lastPlantingSite'`. Realize that this might break with how preferences have been organized in previous pages/components, if they are organized at all...

src/components/Plants/index.tsx
```diff
@@ -1,153 +1,36 @@
...
-        <PlantingSiteDetails
-          plantingSite={selectedPlantingSite}
-          plantsDashboardPreferences={plantsDashboardPreferences}
-          setPlantsDashboardPreferences={(newPreferences) => {
-            setPlantsDashboardPreferences((oldPreferences) => ({ ...oldPreferences, ...newPreferences }));
-          }}
-        />
-      </Box>
-    </TfMain>
+    <PlantsPrimaryPage
```
charlie-tf, line 19: Nice consolidation

src/components/PlantsPrimaryPage/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+        snackbar.toastError();
+      }
+    };
+    populatePlantingSites();
+  }, [selectedOrganization.id, snackbar, allowAllAsSiteSelection]);
+
+  const setActivePlantingSite = useCallback(
+    (site: PlantingSite | undefined) => {
+      if (site) {
+        history.push(pagePath.replace(':plantingSiteId', site.id.toString()));
```
charlie-tf, line 78: Should this be a `replace` type of history update? Do we want the user to go to the sequence of last selected planting sites when they hit the back button?

charlie-tf (commented):
Nice work, just a couple suggestions

src/components/PlantsPrimaryPage/index.tsx
karthikbtf: Yes we do want them to go back. The selections are url driven today.

src/components/Observations/index.tsx
karthikbtf: That's a good suggestion, I understand the structure. I prefer to keep preference names flatter for now, let each component decide how they want to name the preferences. Also yes, this will break existing preferences that are already in the database. The preference names are not parsed for semantics anywhere.

karthikbtf:
I am going to clean this up some more - see if i can create a story without any BE dependencies

src/components/Observations/index.tsx
karthikbtf: Here's a suggestion, since we aren't using plants dashboard in production yet.
I can make the preferences for the planting site pages hierarchical for sure. We can apply this going forward to other preferences.
For existing preferences (like accessions table, etc.) we will keep them as they are today.

karthikbtf:
> I am going to clean this up some more - see if i can create a story without any BE dependencies

Was able to refactor this, extracted out the rendering and created a story for that. Also updated the preference names.

charlie-tf (approved):
🎉

src/components/Observations/index.tsx
charlie-tf: Sounds lovely. Thank you!

PR #1422, opened 2023-05-26
SW-3632 Observation Date Selection Slider

charlie-tf:
## Single Option Variant
image

## Multiple Option Variant
image

### DIFF
--- a/src/components/common/MapDateSelect/DateSlider.tsx
+++ b/src/components/common/MapDateSelect/DateSlider.tsx
@@ -1,0 +1,44 @@
+import { Slider } from '@terraware/web-components';
+import { useMemo } from 'react';
+
+type DateSliderProps = {
+  dates: Date[];
+  onChange: (newDate: Date) => void;
+  initialSelectionIndex: number;
+};
+
+export default function DateSlider(props: DateSliderProps): JSX.Element {
+  const { dates, onChange, initialSelectionIndex } = props;
+
+  const minTime = useMemo(() => {
+    return Math.min(...dates.map((d) => d.getTime()));
+  }, [dates]);
+
+  const maxTime = useMemo(() => {
+    return Math.max(...dates.map((d) => d.getTime()));
+  }, [dates]);
+
+  const marks = useMemo(() => {
+    return dates.map((date) => ({ value: date.getTime() }));
+  }, [dates]);
+
+  const timestampForDate = useMemo(() => {
+    const tsMap = new Map<number, Date>();
+    for (const date of dates) {
+      tsMap.set(date.getTime(), date);
+    }
+
+    return tsMap;
+  }, [dates]);
+
+  return (
+    <Slider
+      defaultValue={marks[initialSelectionIndex].value}
+      min={minTime}
+      max={maxTime}
+      marks={marks}
+      valueLabelDisplay='off'
+      onChange={(ts) => onChange(timestampForDate.get(ts) ?? dates[initialSelectionIndex])}
+    />
+  );
+}
--- a/src/components/common/MapDateSelect/DateSlider.tsx
+++ b/src/components/common/MapDateSelect/DateSlider.tsx
@@ -2,8 +2,8 @@ import { Slider } from '@terraware/web-components';
 import { useMemo } from 'react';
 
 type DateSliderProps = {
-  dates: Date[];
-  onChange: (newDate: Date) => void;
+  dates: string[]; // date strings in the format 'YYYY-MM-DD'
+  onChange: (newDate: string) => void;
   initialSelectionIndex: number;
 };
 
@@ -11,21 +11,21 @@ export default function DateSlider(props: DateSliderProps): JSX.Element {
   const { dates, onChange, initialSelectionIndex } = props;
 
   const minTime = useMemo(() => {
-    return Math.min(...dates.map((d) => d.getTime()));
+    return Math.min(...dates.map((d) => Date.parse(d)));
   }, [dates]);
 
   const maxTime = useMemo(() => {
-    return Math.max(...dates.map((d) => d.getTime()));
+    return Math.max(...dates.map((d) => Date.parse(d)));
   }, [dates]);
 
   const marks = useMemo(() => {
-    return dates.map((date) => ({ value: date.getTime() }));
+    return dates.map((date) => ({ value: Date.parse(date) }));
   }, [dates]);
 
   const timestampForDate = useMemo(() => {
-    const tsMap = new Map<number, Date>();
+    const tsMap = new Map<number, string>();
     for (const date of dates) {
-      tsMap.set(date.getTime(), date);
+      tsMap.set(Date.parse(date), date);
     }
 
     return tsMap;
--- a/src/components/common/MapDateSelect/index.tsx
+++ b/src/components/common/MapDateSelect/index.tsx
@@ -2,6 +2,7 @@ import DateSlider from './DateSlider';
 import strings from 'src/strings';
 import { Box, Typography, useTheme } from '@mui/material';
 import { useMemo, useState } from 'react';
+import Card from 'src/components/common/Card';
 
 type MapDateSelectProps = {
   dates: Date[];
@@ -58,16 +59,19 @@ export default function MapDateSelect({ dates, onChange }: MapDateSelectProps): 
   const getDateString = (date: Date) => {
     return `${monthStrings[date.getUTCMonth()]} ${date.getUTCFullYear()}`;
   };
+  const getDateLabel = (date: Date) => <Typography fontSize='12px'>{getDateString(date)}</Typography>;
 
   return (
-    <Box
-      border={`1px solid ${theme.palette.TwClrBrdrTertiary}`}
-      borderRadius='8px'
-      display='flex'
-      flexDirection='column'
-      padding={theme.spacing(2)}
-      maxWidth='270px'
-      width={dates.length > 1 ? '270px' : undefined}
+    <Card
+      style={{
+        border: `1px solid ${theme.palette.TwClrBrdrTertiary}`,
+        borderRadius: '8px',
+        display: 'flex',
+        flexDirection: 'column',
+        padding: theme.spacing(2),
+        maxWidth: '270px',
+        width: dates.length > 1 ? '270px' : undefined,
+      }}
     >
       {selectedDate && (
         <Typography fontSize='14px' fontWeight={600} textAlign='right'>
@@ -78,11 +82,11 @@ export default function MapDateSelect({ dates, onChange }: MapDateSelectProps): 
         <>
           <DateSlider dates={dates} onChange={handleChange} initialSelectionIndex={latestDateIndex} />
           <Box display='flex' justifyContent='space-between'>
-            <Typography fontSize='12px'>{getDateString(earliestDate)}</Typography>
-            <Typography fontSize='12px'>{getDateString(latestDate)}</Typography>
+            {getDateLabel(earliestDate)}
+            {getDateLabel(latestDate)}
           </Box>
         </>
       )}
-    </Box>
+    </Card>
   );
 }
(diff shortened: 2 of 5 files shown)


charlie-tf:
@chudo76, can you take a look at this when you get a chance and tell me whether it is behaving as you'd expect? You can take a look at the storybook preview at https://terraware-web-storybook-git-charlie-sw-3632-terraformation.vercel.app/

src/components/common/MapDateSelect/DateSlider.tsx
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    return tsMap;
+  }, [dates]);
+
+  return (
+    <Slider
+      defaultValue={marks[initialSelectionIndex].value}
+      min={minTime}
+      max={maxTime}
```
karthikbtf, line 38: probably don't need this since we allow the entire range of marks as valid values?

src/components/common/MapDateSelect/DateSlider.tsx
```diff
@@ -0,0 +1,44 @@
...
+      marks={marks}
+      valueLabelDisplay='off'
+      onChange={(ts) => onChange(timestampForDate.get(ts) ?? dates[initialSelectionIndex])}
```
karthikbtf, line 41: the second part won't be needed since we are snapping to legit values always?

src/components/common/MapDateSelect/DateSlider.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const minTime = useMemo(() => {
+    return Math.min(...dates.map((d) => d.getTime()));
+  }, [dates]);
+
+  const maxTime = useMemo(() => {
+    return Math.max(...dates.map((d) => d.getTime()));
+  }, [dates]);
+
+  const marks = useMemo(() => {
+    return dates.map((date) => ({ value: date.getTime() }));
```
karthikbtf, line 22: Just realized that the user has no idea what the intermediate observation-dates are.. this might be a ux issue.. something to discuss with @chudo76 perhaps. User has to complete the selection in order to see the date, instead of making an informed selection.

src/components/common/MapDateSelect/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+    return dates.reduce((prev, curr) => {
+      if (prev.getTime() < curr.getTime()) {
+        return prev;
+      }
+      return curr;
+    });
+  }, [dates]);
+
+  const latestDate = useMemo(() => {
+    return dates.reduce((prev, curr) => {
```
karthikbtf, line 39: `Math` has a min and max function which supports array spread operator, you can optionally do:
```
  return Math.max([... dates.map(date => date.getTime())])
```

src/components/common/MapDateSelect/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+    setSelectedDate(newDate);
+    onChange(newDate);
+  };
+
+  const getDateString = (date: Date) => {
+    return `${monthStrings[date.getUTCMonth()]} ${date.getUTCFullYear()}`;
+  };
+
+  return (
+    <Box
```
karthikbtf, line 63: may need a bgcolor (if it will be overlaid on the map), there's a `Card` component if you want to leverage that

src/components/common/MapDateSelect/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+      {selectedDate && (
+        <Typography fontSize='14px' fontWeight={600} textAlign='right'>
+          {getDateString(selectedDate)}
+        </Typography>
+      )}
+      {dates.length > 1 && (
+        <>
+          <DateSlider dates={dates} onChange={handleChange} initialSelectionIndex={latestDateIndex} />
+          <Box display='flex' justifyContent='space-between'>
+            <Typography fontSize='12px'>{getDateString(earliestDate)}</Typography>
```
karthikbtf, line 81: create a function that returns this component, for reusability?
```
getDateLabel = (date: Date) => (<Typography fontSize='12px'>{getDateString(date)}</Typography>);
```

src/components/common/MapDateSelect/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const [selectedDate, setSelectedDate] = useState(latestDate);
+
+  const handleChange = (newDate: Date) => {
+    setSelectedDate(newDate);
+    onChange(newDate);
+  };
+
+  const getDateString = (date: Date) => {
+    return `${monthStrings[date.getUTCMonth()]} ${date.getUTCFullYear()}`;
```
karthikbtf, line 59: Nice, we will get localized months 👍

karthikbtf (commented):
This looks great and functions great 🎉 
Some comments if they make sense.

charlie-tf:
Thanks for the review, @karthikbtf

src/components/common/MapDateSelect/DateSlider.tsx
charlie-tf: I started without these, but it wasn't displaying correctly 🤔

src/components/common/MapDateSelect/DateSlider.tsx
charlie-tf: I added that because `.get()` can return undefined

src/components/common/MapDateSelect/DateSlider.tsx
karthikbtf: ok, we can leave it in and figure it out later maybe :)

src/components/common/MapDateSelect/index.tsx
charlie-tf: Nice, thanks!

charlie-tf: Nice, thanks!

charlie-tf: Doesn't work in this case, since I want to return the `Date`, and `Math.max/min` will return the timestamp.

karthikbtf: That makes sense. Missed the date as result.

src/components/common/MapDateSelect/DateSlider.tsx
charlie-tf: I thought it might be because the marks were sorted backwards, but switching that didn't help either

karthikbtf (approved):
Nice 🙌 
Making these look easier than what I expected it to be.

chudo76:
> @chudo76, can you please take a look at this when you get a chance and tell me whether it is behaving as you'd expect? You can find at the storybook preview at https://terraware-web-storybook-git-charlie-sw-3632-terraformation.vercel.app/

Looking good! (Tagging @rachel-tf-design since she designed the component.)

src/components/common/MapDateSelect/index.tsx
charlie-tf: Oh, okay. I see what you mean. So would you suggest changing the type of `props.dates` to a `number[]`? And passing in a list of timestamps? That would simplify several places where I have to convert between dates and timestamps (sorting). Passing in the planting site timezone as a prop would probably be the way to go here as well, I'd think, and a good idea since a timezone mismatch could land the observation in the wrong month.

karthikbtf: Correct, timestamps passed in as props - or the date strings "yyyy-mm-dd"  as-is and you can have the component convert to timestamps. Or keep it as-is with date objects. All theoretically correct. Throughout the product we use "yyyy-mm-dd" to represent date values, maybe timestamp semantics should be local to this component? Will let you decide, just some thoughts.

Primary suggestion - support time zone, locale changes (component should re-render on locale change), and use Intl.DateTimeFormat instead of constructing the localized string ourselves.

If you want to keep the current implementation as-is and make minimal changes, Constanza added this [util](https://github.com/terraware/terraware-web/blob/main/src/utils/useTimeZoneUtils.ts#L100) to work off date objects (primarily for monitoring charts). It's more of a hack but will still work for display purposes here. With this approach you would only need the time zone prop change.

charlie-tf: I like the idea of using date strings (yyyy-mm-dd), if that is what is used throughout the app. The logic of converting to timestamps can occur locally within this component, hiding those implementation details away.

charlie-tf: @karthikbtf I'm playing around with this a little, and I'm thinking: if I use the date string as the props, then technically I can perform the calculations internally using whatever timezone I want (e.g. UTC) and they'll still come out correctly.

For example, `new Date('2023-06-01')` creates a date in UTC by default, so it's possible to do
```typescript
new Intl.DateTimeFormat('en-US', { month: 'short', year: 'numeric', timeZone: 'UTC' }).format(new Date('2023-06-01'))
```
which returns `Jun 2023` regardless. And comparing timestamps is easy by comparing timestamps produced by `Date.parse('yyyy-mm-dd')`, which is UTC by default. 

Finally, the `onChange` callback can return yyyy-mm-dd datestrings. This way, the component transparently handles dates and doesn't need to know anything about or handle timezones at all. Thoughts on this approach?

karthikbtf: Yes, we can absolutely use date strings for the calculations and across the board.
My bad, the time zone suggestion wasn't for date related calculations, it was strictly for displaying a UTC date in the planting site's time zone (the Intl.DateTimeFormat's timeZone option handles this display/conversion). Under the hood all dates are timestamps, the time zone is just a presentation layer on top of that.
My assumption is the observation dates are UTC and we need to display that in the planting site's time zone. If my assumption is incorrect then we don't need time zone handling.

karthikbtf: Your idea sounds great for now, we can iterate once we have a better idea of the nature of observation dates. Skip time zone handling.

src/components/common/MapDateSelect/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const [selectedDate, setSelectedDate] = useState(latestDate);
+
+  const handleChange = (newDate: string) => {
+    setSelectedDate(newDate);
+    onChange(newDate);
+  };
+
+  const getDateString = (date: string) => {
+    return new Intl.DateTimeFormat(locale, { month: 'short', year: 'numeric', timeZone: 'UTC' }).format(new Date(date));
```
karthikbtf, line 46: Add a note/TODO to re-visit planting site's timezone here? Example: if date string is UTC and planting site time zone is hawaii, we have to use hawaii time zone in the options.
Remove 'UTC' timezone for now since that's the default anyways?

charlie-tf: If `timeZone: UTC` is left out, then the formatter localizes to the browser timezone, so it needs to stay in.

I'm not sure that this needs to be revisited. If the calculations internal to this component are all in UTC, and it accepts and returns datestrings only, then the timezone is essentially factored out.

karthikbtf: > If timeZone: UTC is left out, then the formatter localizes to the browser timezone, so it needs to stay in.
Thanks for clarifying the browser timezone. My bad.

> I'm not sure that this needs to be revisited. If the calculations internal to this component are all in UTC, and it accepts and returns datestrings only, then the timezone is essentially factored out.

The time zone is only a display option, not for calculations. If the dates are in UTC but a planting site is in another time zone, we should be using that time zone for generating the display string. Unless I'm missing something that you are trying to convey.

charlie-tf: Sorry that I'm not clear. What I was trying to say was that the component receives through props and returns through its callback dates that are in 'yyyy-mm-dd' format, only. It only ever displays a 'Month YYYY' string, so it shouldn't need to know about the timezone. However, if I'm understanding what you're saying, then the date that is sent to this component might be in UTC rather than localized to the planting site.

Since, as you say, we don't exactly know how the dates are getting produced by observations and sent to this component (UTC? local to planting site?) I will add a TODO to revisit. Apologies for the back and forth on this.

karthikbtf: All good. Thanks, yes, only because we don't know if the date will be in UTC or already accounts for the planting site.
Totally agree, if it's the latter.. no change needed.
Just trying to avoid a corner case (showing next month or previous month where a time zone is before or after UTC) at the beginning/end of months.

PR #1428, opened 2023-05-31

terraware-phrase:


karthikbtf (approved):
lista! mapa! charlie you were righta!

karthikbtf:
maybe we should enable switching locales in storybook

PR #1430, opened 2023-06-01
fix build warnings in terraware-web related to typescript version

karthikbtf:
- warning is
```
TypeError: Cannot set property mark of #<Object> which has only a getter
TypeError: Cannot set property mark of #<Object> which has only a getter
    at Object.connectTypeScriptPerformance
```
- fix offending package version
- see https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/797

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,5 +136,8 @@
     "tslint-eslint-rules": "^5.4.0",
     "tslint-react": "^5.0.0",
     "tslint-react-hooks": "^2.2.2"
+  },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


charlie-tf (approved):
Thanks so much for fixing this! I've been meaning to look into it

karthikbtf:
> Thanks so much for fixing this! I've been meaning to look into it

no worries, we have been mostly doing funder web stuff until now

PR #1432, opened 2023-06-02
SW-3715 Add Layer Selection to Planting Site Map

charlie-tf:
Kapture 2023-06-02 at 10 16 15

TODO: update the planting site map to be able to display monitoring plots

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -175,9 +175,6 @@ export default function Map(props: MapProps): JSX.Element {
 
   useEffect(() => {
     const { sources } = options;
-    if (geoData) {
-      return;
-    }
 
     /**
      * Initialize sources - we want ONE source per data type, eg. one source containing all geometries for zones.
@@ -267,7 +264,7 @@ export default function Map(props: MapProps): JSX.Element {
     if (popupRenderer) {
       setLayerIds(geo.filter((g: any) => g.isInteractive).map((g: any) => g.layer.id));
     }
-  }, [options, geoData, setGeoData, token, popupRenderer]);
+  }, [options, setGeoData, token, popupRenderer]);
 
   useEffect(() => {
     if (entityOptions?.highlight) {
@@ -294,7 +291,7 @@ export default function Map(props: MapProps): JSX.Element {
       return null;
     }
     const sources = (geoData as any[]).map((geo: any, index) => (
-      <Source type='geojson' key={index} data={geo.data} id={geo.id}>
+      <Source type='geojson' key={`${geo.id}-${index}`} data={geo.data} id={geo.id}>
         {geo.textAnnotation && <Layer {...geo.textAnnotation} />}
         {geo.layerOutline && <Layer {...geo.layerOutline} />}
         {geo.layer && <Layer {...geo.layer} />}
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -290,8 +290,8 @@ export default function Map(props: MapProps): JSX.Element {
     if (!geoData) {
       return null;
     }
-    const sources = (geoData as any[]).map((geo: any, index) => (
-      <Source type='geojson' key={`${geo.id}-${index}`} data={geo.data} id={geo.id}>
+    const sources = (geoData as any[]).map((geo: any) => (
+      <Source type='geojson' key={geo.id} data={geo.data} id={geo.id}>
         {geo.textAnnotation && <Layer {...geo.textAnnotation} />}
         {geo.layerOutline && <Layer {...geo.layerOutline} />}
         {geo.layer && <Layer {...geo.layer} />}
--- a/src/components/Map/PlantingSiteMap.tsx
+++ b/src/components/Map/PlantingSiteMap.tsx
@@ -13,9 +13,11 @@ import { makeStyles } from '@mui/styles';
 
 const useStyles = makeStyles((theme: Theme) => ({
   layerSelectContainer: {
+    height: theme.spacing(3),
     position: 'relative',
     right: theme.spacing(5.5),
     top: theme.spacing(2),
+    width: theme.spacing(3),
     zIndex: 1000,
   },
 }));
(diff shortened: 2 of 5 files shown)


src/components/Map/Map.tsx
```diff
@@ -293,8 +290,8 @@ export default function Map(props: MapProps): JSX.Element {
     if (!geoData) {
       return null;
     }
-    const sources = (geoData as any[]).map((geo: any, index) => (
-      <Source type='geojson' key={index} data={geo.data} id={geo.id}>
+    const sources = (geoData as any[]).map((geo: any) => (
+      <Source type='geojson' key={geo.id} data={geo.data} id={geo.id}>
```
charlie-tf, line 294: Setting the key here to be the source id was an important and surprising change. Internally, mapbox associates the react key with the source id, and if the source id for a given key changes (e.g. if the sources are reordered, resulting in their index changing) then mapbox crashes the whole page.

src/components/Map/Map.tsx
```diff
@@ -175,9 +175,6 @@ export default function Map(props: MapProps): JSX.Element {
 
   useEffect(() => {
     const { sources } = options;
-    if (geoData) {
-      return;
-    }
```
charlie-tf, line 180: Allow the geoData to be updated

src/components/Map/PlantingSiteMap.tsx
```diff
@@ -104,15 +118,27 @@ export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Elemen
...
+        includedSources.push(zones);
+      }
+      if (includedLayers?.includes('Planting Site')) {
+        includedSources.push(site);
+      }
+
       const newMapOptions = {
         bbox: MapService.getPlantingSiteBoundingBox(plantingSite),
-        sources: [site, subzones, zones],
+        sources: includedSources,
```
karthikbtf, line 141: This is great and works! 
Food for thought if it helps - extend the entity/source props with a visible/boolean flag  - which can then be filtered in the map. This is one way I have seen/done layer based viz in the past.
We could flip on/off specific layers... or on/off specific data.
On one hand it is repopulating data sources, on another it is configuring view options of data sources. Pros/cons to both. Whichever you think scales better for the additional functionalities like search results based viewing etc.
What we have here is simple/correct and works - just wanted to share data/view separation idea if it clicks.

src/components/common/MapLayerSelect/index.tsx
```diff
@@ -2,9 +2,8 @@ import React, { useState } from 'react';
 import { Icon, PopoverMultiSelect } from '@terraware/web-components';
 import { makeStyles } from '@mui/styles';
 import { Theme } from '@mui/material';
-import strings from 'src/strings';
 
-export type MapLayer = 'Planting Site' | 'Zones' | 'Monitoring Plots';
+export type MapLayer = 'Planting Site' | 'Zones' | 'Sub-Zones' | 'Monitoring Plots';
```
karthikbtf, line 6: Nice 🙌  thanks for making this a type.

karthikbtf (commented):
Looks great and clean. I have one comment (a biased suggestion) just want to hear your thoughts.

src/components/Map/PlantingSiteMap.tsx
karthikbtf: The layer versus search are actually orthogonal.. I guess search will always be changes to the data source. My bad.

src/components/common/MapLayerSelect/index.tsx
```diff
@@ -22,32 +21,20 @@ const useStyles = makeStyles((theme: Theme) => ({
 type MapLayerSelectProps = {
   initialSelection: MapLayer[];
   onUpdateSelection: (selection: MapLayer[]) => void;
+  menuSections: { label: string; value: MapLayer }[][];
 };
 
-export default function MapLayerSelect({ initialSelection, onUpdateSelection }: MapLayerSelectProps): JSX.Element {
+export default function MapLayerSelect({
```
karthikbtf, line 27: suggestion, add a `cursor: pointer` style to this component?

karthikbtf (commented):
The planting sites page tested great in vercel.. one css suggestion.
The plants dashboard seems to be broken, if you get a chance to look.

src/components/Map/PlantingSiteMap.tsx
```diff
@@ -104,15 +118,27 @@ export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Elemen
...
   // fetch polygons and boundaries
+  const [includedLayers, setIncludedLayers] = useState<MapLayer[] | undefined>(layerOptions);
   useEffect(() => {
     const fetchPlantingSite = () => {
       const site = extractPlantingSite(plantingSite);
       const zones = extractPlantingZones(plantingSite);
       const subzones = extractSubzones(plantingSite);
 
+      const includedSources = new Array<MapSource>();
+      if (includedLayers?.includes('Sub-Zones')) {
```
karthikbtf, line 129: ok i think the plants dashboard fix is to check if includedLayers === undefined.. in which case just push the sources

karthikbtf (approved):
Looks great 🎉 
Just that one css suggestion and the includedLayers fix.
Re: my data versus view-options, we can discuss offline if needed.. no need to make changes

src/components/Map/PlantingSiteMap.tsx
charlie-tf: That seems like a good fix 👍

src/components/Map/PlantingSiteMap.tsx
charlie-tf: I actually like the visible/non-visible switch option. I'll see about implementing that

charlie-tf: Hmm, not immediately getting this to work. I'll revisit it later

karthikbtf: No worries, thanks for giving this a try.

karthikbtf (approved):

PR #1434, opened 2023-06-05
Set map/view selector icon order based on initial/default view

karthikbtf:
- as per Chudo's request
- we have views where the default view is a map, and the selector needs to show map first in those cases

ListMapSelector - Default Map ⋅ Storybook 2023-06-05 09-29-58

ListMapSelector - Default List ⋅ Storybook 2023-06-05 09-29-33

### DIFF
--- a/src/components/common/ListMapSelector.tsx
+++ b/src/components/common/ListMapSelector.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { Box, Theme, Typography, useTheme } from '@mui/material';
 import hexRgb from 'hex-rgb';
 import strings from 'src/strings';
--- a/src/components/common/ListMapSelector.tsx
+++ b/src/components/common/ListMapSelector.tsx
@@ -23,12 +23,12 @@ const useStyles = makeStyles((theme: Theme) => ({
 export type View = 'list' | 'map';
 
 export type ListMapSelectorProps = {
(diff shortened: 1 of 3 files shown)


src/components/common/ListMapSelector.tsx
```diff
@@ -28,11 +28,14 @@ export type ListMapSelectorProps = {
 };
 
 export default function ListMapSelector({ view, onView }: ListMapSelectorProps): JSX.Element {
+  const [initialView] = useState<View>(view);
```
charlie-tf, line 31: The `view` prop feels a bit overloaded here. It's responsible for determining the initial view (whatever it was first set to) and the current view. Would it be cleaner to have a separate `defaultView` prop?

src/components/common/ListMapSelector.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
```
charlie-tf, line 1: Shouldn't need `useState` anymore

charlie-tf (approved):
Thanks for the change. Looks great!

chudo76 (approved):
Looks good. Thank you for making these changes!

karthikbtf:
> Looks good. Thank you for making these changes!

Sure Chudo.. left a comment in Figma.. if you get a chance :)

PR #1437, opened 2023-06-06
SW-3718 add search input to observations page

karthikbtf:
- search only searches zone names (as per design annotation)
- added a search selector
- search filters will be added later on

Terraware App 2023-06-06 14-21-24

### DIFF
--- a/src/redux/features/observations/observationsSelectors.ts
+++ b/src/redux/features/observations/observationsSelectors.ts
@@ -16,6 +16,7 @@ import {
 } from 'src/types/Observations';
 import { selectSpecies } from 'src/redux/features/species/speciesSelectors';
 import { selectPlantingSites } from 'src/redux/features/tracking/trackingSelectors';
+import { regexMatch } from 'src/utils/search';
 
 export const selectObservationsResults = (state: RootState) => state.observationsResults?.observations;
 export const selectObservationsResultsError = (state: RootState) => state.observationsResults?.error;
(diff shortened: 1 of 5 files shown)


src/redux/features/observations/observationsSelectors.ts
```diff
@@ -55,7 +55,22 @@ export const selectMergedPlantingSiteObservations = createCachedSelector(
...
 
-// add more selectors for drill down views, as needed
+// search observations (search planting zone name only)
+export const searchObservations = createCachedSelector(
+  (state: RootState, plantingSiteId: number, defaultTimeZone: string, search: string) =>
+    selectMergedPlantingSiteObservations(state, plantingSiteId, defaultTimeZone),
+  (state: RootState, plantingSiteId: number, defaultTimeZone: string, search: string) => search,
+  (observations, search) =>
+    observations?.filter((observation: ObservationResults) => {
+      const regexp = new RegExp(search, 'i');
```
charlie-tf, line 65: This should allow regular expressions to be input into the search input and ran unescaped against the zone names. Other searches use the search API, making this behavior different. Perhaps not worth worrying about, but just noting.

charlie-tf: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#escaping has a function that could be used if escaping is desired here.

charlie-tf (approved):
Looks good. A note and suggestion, if wanted, but not necessary to merge.

src/redux/features/observations/observationsSelectors.ts
karthikbtf: I can certainly add the escaping 👍 , fwiw i didn't know regex search was a requirement in our product. you never know, we could have a tech savvy customer..

karthikbtf: btw, I was going to use "string.toLowerCase().includes(pattern.toLowerCase())" but went with regexp purely for the case-insensitive option. I think I understand the reason for the suggestion now.
If it's less confusing, I can switch back to "string.includes" approach.

charlie-tf: Oh, the regexp approach isn't confusing. I just wondered whether allowing full regexp searches was what we wanted here. I'm okay with either approach. Regexp with escaping vs. string.includes should give the same result (I _think_), and either one is clear enough

charlie-tf (approved):
👍

PR #1438, opened 2023-06-06
Planting Site Map Data

charlie-tf:
We change the way that map data flows into the PlantingSiteMap. Instead of
passing in the planting site, we pass in a more generic MapData object, which
contains data for the various map objects (site, zone, subzone, permanentPlot,
temporaryPlot). This will help make the component more reusable where, for
instance, one does not have a planting site object, but observation results
from which the map data object may be created. Behavior for existing maps, such
as the ones in the Planting Site or Plants Dashboard pages, is unchanged.

### DIFF
--- a/src/components/Map/PlantingSiteMap.tsx
+++ b/src/components/Map/PlantingSiteMap.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { Box, CircularProgress, Theme, useTheme } from '@mui/material';
 import hexRgb from 'hex-rgb';
-import { PlantingSite } from 'src/types/Tracking';
 import useSnackbar from 'src/utils/useSnackbar';
 import GenericMap from './GenericMap';
 import { MapEntityId, MapEntityOptions, MapOptions, MapPopupRenderer, MapSource } from 'src/types/Map';
@@ -9,6 +8,7 @@ import { MapService } from 'src/services';
 import _ from 'lodash';
 import { MapLayer } from 'src/components/common/MapLayerSelect';
 import { makeStyles } from '@mui/styles';
+import { MapData, MapObject } from 'src/services/MapService';
 
 const useStyles = makeStyles((theme: Theme) => ({
   bottomLeftControl: {
@@ -30,7 +30,7 @@ const useStyles = makeStyles((theme: Theme) => ({
 }));
 
 export type PlantingSiteMapProps = {
-  plantingSite: PlantingSite;
+  mapData: MapData;
   // style overrides
   style?: object;
   // context on-click renderer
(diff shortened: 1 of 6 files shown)


charlie-tf:
@karthikbtf Let me know what you think of this approach. I'm trying to push some of the complexity out of the PlantingSiteMap component to make it behave more generically.

charlie-tf:
oops, broke a test. fixing....

src/components/Map/PlantingSiteMap.tsx
```diff
@@ -91,74 +90,92 @@ export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Elemen
...
-      const subzones = extractSubzones(plantingSite);
-
-      const includedSources = new Array<MapSource>();
-      if (layers === undefined || layers?.includes('Sub-Zones')) {
-        includedSources.push(subzones);
+      const sources = new Array<MapSource>();
+      if (mapData.temporaryPlot && (layers === undefined || layers?.includes('Monitoring Plots'))) {
+        sources.push({
+          ...mapData.temporaryPlot,
+          isInteractive: true,
```
karthikbtf, line 130: Just calling this out to put it on your radar.
Fwiw, isInteractive tells the Map component which ones to allow click/popups for.
For example, if all layers are off except the planting site, we want the planting site to be clickable and show a popup with plants stats, etc.
I guess dynamically set isInteractive based on what's in the layers. Figma has these various use-cases highlighted.
We can add these features later for sure, just wanted to call it out.

src/services/MapService.ts
```diff
@@ -21,6 +21,9 @@ const MAPBOX_TOKEN_ENDPOINT = '/api/v1/tracking/mapbox/token';
 type MapboxTokenServerResponse =
   paths[typeof MAPBOX_TOKEN_ENDPOINT]['get']['responses'][200]['content']['application/json'];
 
+export type MapObject = 'site' | 'zone' | 'subzone' | 'permanentPlot' | 'temporaryPlot';
+export type MapData = Record<MapObject, MapSourceBaseData | undefined>;
```
karthikbtf, line 25: move these into types/Map?

karthikbtf (approved):
This is cool, I was wondering how you were going to make the observations results data work with existing code. This is some good abstraction 👍 
One comment about defining types under the types folder.

karthikbtf:
Guessing this needs to be merged into the SW-3696 branch

src/components/Map/PlantingSiteMap.tsx
charlie-tf: Cool, I will add those things as I start to work through the interactivity with those map elements 👍

charlie-tf:
> Guessing this needs to be merged into the [SW-3696](https://terraformation.atlassian.net/browse/SW-3696) branch

I actually intentionally based this off main. It doesn't depend on the 3696 branch at all

karthikbtf (approved):

PR #1440, opened 2023-06-07
SW-3724 Make table columns i18n friendly, plus other i18n fixes

karthikbtf:
- table columns were not updating on locale change except for accessions and species tables
- render withdrawal purpose using strings
- withdrawal log filters update with locale
- fix reassign button not showing up in withdrawal table
- see changes in vercel


charlie-tf (approved):
Straightforward changes, works in Vercel

PR #1441, opened 2023-06-07
SW-3720 Routing + hierarchical navigation for observation drill downs

karthikbtf:
- added page frame component which accepts crumbs, title and content for page
- added routes/routing in observation/index
- added empty placeholder content for the detail drill down pages (tested the hierarchical nav with placeholder content)

Terraware App 2023-06-07 15-33-42

Terraware App 2023-06-07 15-34-35

Terraware App 2023-06-07 15-35-01


src/components/Observations/index.tsx
```diff
@@ -57,5 +62,23 @@ export default function Observations(): JSX.Element {
...
+      </Route>
+      <Route exact path={APP_PATHS.OBSERVATION_PLANTING_ZONE_DETAILS}>
+        <ObservationPlantingZoneDetails />
+      </Route>
+      <Route exact path={APP_PATHS.OBSERVATION_DETAILS}>
+        <ObservationDetails />
+      </Route>
+      <Route exact path={APP_PATHS.PLANTING_SITE_OBSERVATIONS}>
+        <ObservationsHome />
+      </Route>
```
charlie-tf, line 78: Is this route needed, given that the next `'*'` route is the same?

src/components/Observations/plot/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+export default function ObservationMonitoringPlotDetails(): JSX.Element {
+  const { plantingSiteId, observationId, plantingZoneId, monitoringPlotId } = useParams<{
+    plantingSiteId: string;
+    observationId: string;
+    plantingZoneId: string;
+    monitoringPlotId: string;
+  }>();
+
+  const urlSite = () =>
+    APP_PATHS.OBSERVATION_PLANTING_ZONE_DETAILS.replace(':plantingSiteId', plantingSiteId?.toString());
```
charlie-tf, line 15: Is this a valid URL? It'll still have the `"/:observationId/:plantingZoneId"` segments if I'm reading correctly

charlie-tf (commented):
Question about URLs constructed for the breadcrumbs

src/components/Observations/plot/index.tsx
karthikbtf: Good catch - yes, this will only work for the leaf level link. Will fix 👍

src/components/Observations/index.tsx
karthikbtf: I should have explained clearly.
Analogy is a switch statement:
```
  switch(arg) {
     case xyz: <code>;
     case def: <code>;
     case rst:  // here both rst and default have same return value, we could skip mentioning rst but it is good practice to express this intent
     default:
           <code>;
  }
```

src/components/Observations/plot/index.tsx
karthikbtf: fixed this, was unable to reuse previously built urls but oh well. correctness is desirable

src/redux/features/observations/utils.ts
```diff
@@ -0,0 +1,202 @@
+import getDateDisplayValue from '@terraware/web-components/utils/date';
```
karthikbtf, line 1: i moved this code in here from observationsSelectors just to keep the selectors code less messy, esp if we need to add more selectors...

src/components/Observations/plot/index.tsx
karthikbtf: actually, i can simplify this using the hierarchical crumbs..

src/components/Observations/index.tsx
charlie-tf: Thanks

src/components/Observations/plot/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+    observationId: string;
+    plantingZoneId: string;
+    monitoringPlotId: string;
+  }>();
+
+  const urlSite = APP_PATHS.OBSERVATIONS_SITE.replace(':plantingSiteId', plantingSiteId?.toString());
+
+  const urlDetails = `/${observationId}`;
+
+  const urlZone = `/${plantingZoneId}`;
```
karthikbtf, line 18: this is simpler now, the bread crumbs component will build the full url (now that `hierarchical` is set to true)

charlie-tf (approved):
Changes look good 👍

karthikbtf:
> Changes look good 👍

Thanks for the review and esp catching the url issue :)

PR #1443, opened 2023-06-08
SW-3732 Observations Map View

charlie-tf:
image
image

## TODO:
- Add interactivity (click, hover) to map elements
- ~Filter~ Highlight zones on search input


src/components/Map/Map.tsx
```diff
@@ -43,17 +44,27 @@ export type MapProps = {
...
   const selectStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
   const highlightStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
+  const [firstVisible, setFirstVisible] = useState(false);
+  const visible = useIsVisible(containerRef);
+
+  useEffect(() => {
+    // `firstVisible` detects when the box containing the map is first visible in the viewport. The map should only be
+    // rendered if `firstVisible` is true. This accounts for cases in which the map is initially rendered hidden, and
+    // is improperly resized when it first becomes visible.
+    setFirstVisible((fv) => fv || visible);
```
charlie-tf, line 66: Here's how the map initially renders when it starts out hidden and is then shown:
image

src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -0,0 +1,122 @@
...
+      return {
+        site: undefined,
+        zone: undefined,
+        subzone: undefined,
+        permanentPlot: undefined,
+        temporaryPlot: undefined,
+      };
+    }
+
+    const selectedObservation = observationsResults?.find((obs) => obs.completedTime === selectedObservationDate)!;
```
karthikbtf, line 48: can there be more than one?

src/hooks/useIsVisible.ts
```diff
@@ -0,0 +1,27 @@
+import React, { useEffect, useState } from 'react';
+
+/**
+ * Determine if the element denoted by the ref parameter is visible. This is done by checking its intersection with
+ * the viewport. For performance, only runs the callback when the ref's visibility changes.
+ * See: https://dev.to/jmalvarez/check-if-an-element-is-visible-with-react-hooks-27h8
+ *
+ * @param ref The ref to the element to detect visibility of.
+ */
+export function useIsVisible(ref: React.MutableRefObject<any>) {
```
karthikbtf, line 10: cool!

src/services/MapService.ts
```diff
@@ -222,13 +227,93 @@ const getMapDataFromPlantingSite = (plantingSite: PlantingSite): MapData => {
   };
 };
 
+/**
+ * Extract Planting Site, Zones, Subzones, and Plots from an ObservationResult
+ */
+const getMapDataFromObservation = (observation: ObservationResults): MapData => {
```
karthikbtf, line 233: since we are using redux for observations.. perhaps leverage redux selectors to fetch this view of data, maybe create a new observationMapDataSelectors under redux/features/observations?

src/utils/color.ts
```diff
@@ -0,0 +1,7 @@
+import hexRgb from 'hex-rgb';
+
+export const getRgbaFromHex = (hex: string, opacity: number) => {
```
karthikbtf, line 3: should we change other usages as well to call this?

karthikbtf (commented):
looks good for a first pass, one suggestion to use redux/selectors instead of services to fetch data views.. cached selectors may also help if the data changes causing re-renders.
btw, search should work out of the box, search input filters the results.

src/services/MapService.ts
karthikbtf: Thinking about this more.. my suggestion doesn't make sense here for the current architecture. The selected data is already passed into the component. Please ignore.
Only way is to rearchitect and have the map component trigger selection of data from the store.. not sure there's much of a win here.

karthikbtf (approved):
Looks good as a first iteration 👍 assuming the TODOs will be follow ups.
One request to see if the hex/rgb util can be used in other places that use hex/rgb.

src/services/MapService.ts
charlie-tf: Good point, I'll take a look at that

src/components/Observations/map/ObservationMapView.tsx
charlie-tf: Hmm, I suppose there technically could be if more than one observation were completed on the same day. Not a completely implausible scenario

src/utils/color.ts
charlie-tf: Hm, there's a fair number of these that could be changed over. I'll do them in another PR

karthikbtf: Create a ticket for tracking?

charlie-tf: https://terraformation.atlassian.net/browse/SW-3736

src/components/Observations/map/ObservationMapView.tsx
karthikbtf: I don't think the designs account for this situation either - maybe it's only theoretically possible. We could leave things as-is for now. And bring it up in any of our meetings as a question.

src/utils/color.ts
karthikbtf: Awesome, thanks !!

src/components/Observations/map/ObservationMapView.tsx
charlie-tf: I'll go ahead and add a question to the figma

karthikbtf: Thanks Charlie!! Saw the comment 👍 and helped me understand the situation a bit more. 
The observation's completion date is the latest date a monitoring plot had an observation submitted - at any point in time. Looks like the chances of last updated monitoring plots having same month/year completion dates across planting sites is likely (ofc we can wait for the PRD response). 
Not sure design needs to capture anything here. Like you mentioned, we make updates to the slider (to dedupe by month/year).. and the map renders data for all observations matching a selected completion date (month/year). Rendering multiple observations in the map is already a requirement anyways. map slider probably has the most changes for sure.

karthikbtf: Basically.. completion time for an observation will change over time as monitoring plots are observed. It will probably stay frozen once all involved planting subzones are `finished planting` (some nuances around undoing `finished planting` but not relevant here).

charlie-tf: >Looks like the chances of last updated monitoring plots having same month/year completion dates across planting sites is likely

That's true, but the map is only going to show a single planting site at a time, so for a single planting site this will be unlikely.

>It will probably stay frozen once all involved planting subzones have fully observed monitoring plots.

Since the selector we're basing the observationResults on is only selecting completed observations, we're probably okay on this front

karthikbtf: Ah yes, good point on the map showing a single planting site (completely missed factoring that in) - perhaps just one running observation at a time. Thanks 🙏 
Probably not relevant anymore but just addressing the last comment. The completed observations selector shows observations with a 'completed time' (since upcoming observations or overdue observations will not have one and we don't want to show those). Completed observations here doesn't imply fully completed.

charlie-tf: Good to know. Thank you

PR #1445, opened 2023-06-08
Redo table columns for i18n in terraware-web

karthikbtf:
- wasn't too thrilled about the last fix
- extracted out changed-locale semantics and moved it into the table wrapper
- now each component supplies a function that returns the table names, function has to be outside scope of the component so it doesn't change on every render
- this will scale more easily without bugs, etc.

### DIFF
--- a/src/components/common/table/index.tsx
+++ b/src/components/common/table/index.tsx
@@ -2,7 +2,7 @@ import { useCallback, useEffect, useState } from 'react';
 import { Table as WebComponentsTable, TableColumnType, TableRowType } from '@terraware/web-components';
 import strings from 'src/strings';
 import { LocalizationProps, Props } from '@terraware/web-components/components/table';
-import { useOrganization } from 'src/providers';
+import { useLocalization, useOrganization } from 'src/providers';
 import { PreferencesService } from 'src/services';
 import _ from 'lodash';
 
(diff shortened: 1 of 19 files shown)


src/components/common/table/index.tsx
```diff
@@ -109,23 +109,29 @@ export function OrderPreserveableTable<T extends TableRowType>(
...
-        columns.find((sourceColumn: TableColumnType) => sourceColumn.key === tableColumn.key)
+        sourceColumns.find((sourceColumn: TableColumnType) => sourceColumn.key === tableColumn.key)
       )
       .filter((tableColumn?: TableColumnType) => !!tableColumn) as TableColumnType[];
 
     if (!_.isEqual(tableColumns, refreshedColumns)) {
       setTableColumns(refreshedColumns);
     }
-  }, [columns, tableColumns]);
+  }, [activeLocale, columns, tableColumns]);
```
karthikbtf, line 134: now the table takes care of locale changes, only gotcha is the the columns callback needs to be in a global scope or within a useCallback (in the calling client)

charlie-tf (approved):
Change makes sense. I agree that centralizing this logic in the table makes more sense and will scale better.

karthikbtf:
> Change makes sense. I agree that centralizing this logic in the table makes more sense and will scale better.

Thanks Charlie, this approach requires that we flag any `getColumns` that is transient within the scope of a component - should be a useCallback or declared outside the component. Hope we don't find that cumbersome. Downside to not doing this is a few extra renders of the table.

charlie-tf:
> > Change makes sense. I agree that centralizing this logic in the table makes more sense and will scale better.
> 
> Thanks Charlie, this approach requires that we flag any `getColumns` that is transient within the scope of a component - should be a useCallback or declared outside the component. Hope we don't find that cumbersome. Downside to not doing this is a few extra renders of the table.

I think that's a reasonable requirement

karthikbtf:
> > > Change makes sense. I agree that centralizing this logic in the table makes more sense and will scale better.
> > 
> > 
> > Thanks Charlie, this approach requires that we flag any `getColumns` that is transient within the scope of a component - should be a useCallback or declared outside the component. Hope we don't find that cumbersome. Downside to not doing this is a few extra renders of the table.
> 
> I think that's a reasonable requirement

cc @constanzauanini for context

PR #1447, opened 2023-06-08
SW-3672 Nav links / titles and table views for observation drill downs

karthikbtf:
- info cards/charts etc will be a follow up

Terraware App 2023-06-08 14-25-11

Terraware App 2023-06-08 17-25-45

Terraware App 2023-06-08 17-55-55


karthikbtf:
converting to draft, will add the monitoring plots table here as well since the changes are trivial/minimal

src/components/Observations/details/index.tsx
```diff
@@ -1,21 +1,79 @@
...
 
 export default function ObservationDetails(): JSX.Element {
   const { plantingSiteId, observationId } = useParams<{
     plantingSiteId: string;
     observationId: string;
   }>();
+  const { activeLocale } = useLocalization();
+  const { isMobile } = useDeviceInfo();
+  const defaultTimeZone = useDefaultTimeZone();
+  const [search, setSearch] = useState<string>('');
```
charlie-tf, line 36: Because it is local to this component, the search term will not persist to other pages. Not a problem I think. Making it common would involve lifting the state up to the `Observations/index.tsx` level

charlie-tf (approved):
Looks good, just a comment about the search term, but not one that requires changes

src/components/Observations/details/index.tsx
```diff
@@ -65,7 +65,7 @@ export default function ObservationDetails(): JSX.Element {
             <Box marginTop={2}>
               <Table
                 id='observation-details-table'
-                columns={columns()}
+                columns={columns}
```
charlie-tf: 👍

src/components/Observations/details/index.tsx
karthikbtf: Thanks for the call out. Drill down views search context is local, I believe. The monitoring plot view for instance will be searching monitoring plot names, not zone names. True that the top level and next level searches are zone names - not sure it is a requirement at this time. We can certainly lift it up for the common use-cases later if needed 👍

karthikbtf:
> converting to draft, will add the monitoring plots table here as well since the changes are trivial/minimal

will add the monitoring plot changes separately, since @charlie-tf has already reviewed this PR

src/components/Observations/details/index.tsx
karthikbtf: I went ahead and lifted it up.. and passed it down to 1) root level data view and 2) one level down into the details drill-down. The search value is being preserved across these views (up and down). For next drill down I will avoid passing it down since we won't be searching zone names.
It just occurred to me, if user reloads the page, the search value is lost - user may expect it to be preserved if it was preserved across views. Wondering if we should capture search in the url params instead. Maybe something we should do across the app at some later time.

karthikbtf:
> > converting to draft, will add the monitoring plots table here as well since the changes are trivial/minimal
> 
> will add the monitoring plot changes separately, since @charlie-tf has already reviewed this PR

will make this a draft and add the monitoring plot changes as well.. sorry about the flip-flopping... lifted up the search state and want to capture the monitoring plot drill down here

karthikbtf:
@charlie-tf reopened for review and re-requested review. Added the other two drill down pages (just the table, titles and links). Moved common code for links/crumbs into DetailsPage since I noticed repeated code.

src/components/Observations/index.tsx
```diff
@@ -71,14 +86,14 @@ export default function Observations(): JSX.Element {
...
-        <ObservationDetails />
+        <ObservationDetails {...searchInputProps} />
       </Route>
       <Route exact path={APP_PATHS.PLANTING_SITE_OBSERVATIONS}>
-        <ObservationsHome />
+        <ObservationsHome {...searchInputProps} />
       </Route>
       <Route path={'*'}>
-        <ObservationsHome />
+        <ObservationsHome {...searchInputProps} />
```
karthikbtf, line 95: search input props are lifted in parent and shared across observations home and details page (both search for zones)

src/components/Observations/common/DetailsPage.tsx
```diff
@@ -1,19 +1,91 @@
...
+      state,
+      {
+        plantingSiteId: Number(plantingSiteId),
+        observationId: Number(observationId),
+      },
+      defaultTimeZone.get()
+    )
+  );
+
+  const crumbs: Crumb[] = useMemo(() => {
```
karthikbtf, line 59: creates crumbs from input ids, common code for drill down views

src/redux/features/observations/observationPlantingZoneSelectors.ts
```diff
@@ -0,0 +1,29 @@
...
+    )
+);
+
+export const searchObservationPlantingZone = createCachedSelector(
+  selectObservationPlantingZone,
+  (state: RootState, params: ZoneSearchParams, defaultTimeZone: string) => params,
+  (plantingZone, params) => searchResultPlots(params.search, plantingZone)
+)(
+  (state: RootState, params: ZoneSearchParams, defaultTimeZone: string) =>
+    `${params.plantingSiteId}_${params.observationId}_${params.plantingZoneId}_${defaultTimeZone}_${params.search}`
```
charlie-tf, line 28: If I understand correctly, this string is constructed of the possible fields we want to search against, and then we regexp match against it?

charlie-tf (approved):
Latest changes look good, just a question to make sure I understand how search works

src/redux/features/observations/observationPlantingZoneSelectors.ts
karthikbtf: Oh, this is the key for the cache. The cached selector returns cached responses so it doesn't have to recalculate.
It recalculates only if the key is missing (new input params such as planting site id, etc.) or if any of the intermediate data changes.
```
  createCachedSelector(
    (state, args) => intermediate selector 1,  // intermediate selectors
     ...
    (state, args) => intermediate selector n,
    (results-1, ... results-n) => return final response, // arguments to this last function are results from above intermediate selectors
  ) (state, args) => { return cache key for this response }
```
This is provided by `re-reselect` an extension of redux's `cachedSelector`.
If we don't used cached selectors, the results generated will have different reference values each time and the component will render a lot.

karthikbtf: Here's a good [article](https://redux.js.org/usage/deriving-data-selectors#writing-memoized-selectors-with-reselect) explaining memoized selectors with mention of  [re-select](https://redux.js.org/usage/deriving-data-selectors#re-reselect) as well.

karthikbtf: @charlie-tf if it helps, there are more cached selector use-cases coming up, happy to assign one to you. Let me know.

charlie-tf: Definitely would love to work on one. Thanks for the article

karthikbtf: Charlie.. assigned https://terraformation.atlassian.net/browse/SW-3748 and https://terraformation.atlassian.net/browse/SW-3746 to you. Whichever you'd like to do, feel free to reassign/un-assign as per your plans/bandwidth.

PR #1448, opened 2023-06-09
Add placeholder content for map view with 'All planting sites

karthikbtf:
- need to get copy from Rachel etc.

Terraware App 2023-06-09 10-44-59

### DIFF
--- a/src/components/Observations/map/ObservationMapView.tsx
+++ b/src/components/Observations/map/ObservationMapView.tsx
@@ -97,7 +97,7 @@ export default function ObservationMapView({ observationsResults }: ObservationM
   );
 
   return (
-    <Box display='flex' flexDirection='column'>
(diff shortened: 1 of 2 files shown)


src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -97,7 +97,7 @@ export default function ObservationMapView({ observationsResults }: ObservationM
   );
 
   return (
-    <Box display='flex' flexDirection='column'>
+    <Box display='flex' flexDirection='column' flexGrow={1}>
```
karthikbtf, line 100: Charlie I added flexGrow so the map gets full height of view.

charlie-tf: Thanks

charlie-tf (approved):
I like this. Thanks for adding!

PR #1451, opened 2023-06-09
SW-3733 Make Map Layers Dynamically Interactive

charlie-tf:
- Make the topmost layer interactive
- Dynamically subscribe and unsubscribe map event handlers
- Show annotations for topmost layer
- Show a tooltip with placeholder text upon click

Kapture 2023-06-09 at 15 21 35

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { Box } from '@mui/material';
 import 'mapbox-gl/dist/mapbox-gl.css';
 import ReactMapGL, { AttributionControl, Layer, NavigationControl, Popup, Source } from 'react-map-gl';
-import { MapSource, MapEntityId, MapEntityOptions, MapOptions, MapPopupRenderer } from 'src/types/Map';
+import { MapSource, MapEntityId, MapEntityOptions, MapOptions, MapPopupRenderer, MapGeometry } from 'src/types/Map';
 import { MapService } from 'src/services';
 
 /**
@@ -49,8 +49,8 @@ export default function Map(props: MapProps): JSX.Element {
   const [geoData, setGeoData] = useState<any[]>();
   const [layerIds, setLayerIds] = useState<string[]>([]);
   const [popupInfo, setPopupInfo] = useState<PopupInfo | null>(null);
-  const [deferredHighlightEntity, setDeferredHighlightEntity] = useState<MapEntityId | undefined>();
-  const [deferredFocusEntity, setDeferredFocusEntity] = useState<MapEntityId | undefined>();
+  const [deferredHighlightEntity, setDeferredHighlightEntity] = useState<MapEntityId[] | undefined>();
+  const [deferredFocusEntity, setDeferredFocusEntity] = useState<MapEntityId[] | undefined>();
   const mapRef = useRef(null);
   const containerRef = useRef(null);
   const hoverStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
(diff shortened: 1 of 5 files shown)


src/types/Map.ts
```diff
@@ -102,8 +102,8 @@ export type MapEntityId = {
  * map entity options
  */
 export type MapEntityOptions = {
-  highlight?: MapEntityId;
-  focus?: MapEntityId;
+  highlight?: MapEntityId[];
+  focus?: MapEntityId[];
```
charlie-tf, line 106: Oops, I started out working on the zoom and highlight feature, but slid into just making the map interactive. I'll still use this. The intent was to allow the map to highlight/focus on more than one entity, which will be useful if, for example, search matches more than one zone.

src/components/Map/Map.tsx
```diff
@@ -121,59 +123,103 @@ export default function Map(props: MapProps): JSX.Element {
...
       }
-      const coordinates = feature.geometry.coordinates;
-      const bbox = MapService.getBoundingBox([[coordinates]]);
-      map.fitBounds([bbox.lowerLeft, bbox.upperRight], { padding: 20 });
     },
     [geoData]
   );
 
-  const onLoad = () => {
+  useEffect(() => {
```
charlie-tf, line 164: This useEffect subscribes and, in its cleanup, unsubscribes event handlers. This allows us to update the fill colors for hover states when the interactive layer changes (this was previously handled only in the onLoad function)

src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -120,8 +125,16 @@ export default function ObservationMapView({ observationsResults }: ObservationM
...
             bottomLeftMapControl={
-              observationsDates && <MapDateSelect dates={observationsDates} onChange={setSelectedObservationDate} />
+              observationsDates &&
+              observationsDates.length > 0 && (
+                <MapDateSelect dates={observationsDates} onChange={setSelectedObservationDate} />
+              )
             }
+            contextRenderer={{
+              render: (properties) => {
+                return <p>{`${properties.type} ${properties.id}: ${properties.name}`}</p>;
```
charlie-tf, line 135: This is just a placeholder tooltip content

src/components/Map/PlantingSiteMap.tsx
```diff
@@ -125,12 +131,13 @@ export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Elemen
   useEffect(() => {
     const fetchPlantingSite = () => {
       const sources = new Array<MapSource>();
+      const isTopmostLayer = () => sources.length === 0;
```
karthikbtf, line 134: This is cool :) .. maybe call it 'leaf level'? (it is top most based on the order of populating sources, but leaf level in the map.. which is what we want selected).
Was thinking about the alternative/deterministic check.
Ex. subzone is interactive if monitoring plots are not selected but subzones are selected - in the layers. True, sources would be empty when populating subzones.. because we skipped populating monitoring plots. If we reordered how sources are populated, this would break.. vs a check of which layers are selected.
I like the simplicity of the current code - no harm as it is.
Just thinking out aloud.

src/components/Map/Map.tsx
karthikbtf: Nice 👍

karthikbtf (approved):
Very cool 🎉

src/components/Map/PlantingSiteMap.tsx
charlie-tf: You're right, having a deterministic check that doesn't depend on ordering would be nicer. I will make this change. Thanks

karthikbtf: Ugh.. sorry, I did like this approach, thought it was clever. I think it should be ok as long as we add a comment in there to maintain the ordering?

charlie-tf: Yeah, the deterministic check is kinda gross looking now that i'm trying to write it. If I changed the name of the function to `isFirstLayerAdded` then it would capture the ordering semantics without needing further explanation, would you agree? I can add a comment explaining the significance of the first layer added (it renders on top of the other layers, and that's where we want to pin the interactivity)

karthikbtf: That sounds good 👍

karthikbtf (approved):
Thanks for the followup changes.

PR #1452, opened 2023-06-12
Fix 'invalid date time' display in observations tables

karthikbtf:
- when a site didn't have a timezone set, code fallsback to the default time zone
- the default time zone was being passed as an object.. leading to incorred date/time generation by Intl
- pass the time zone id from the object instead


charlie-tf (approved):
LGTM!

PR #1454, opened 2023-06-12
redo completedTime vs completedDate for observations in FE

karthikbtf:
- keep completedTime as-is
- introduce completedDate for more correct semantics


karthikbtf:
Charlie - let me know your thoughts on this. For monitoring plots we need to retain the time as well.. I started introducing a 'completedTimeRaw' property for the monitoring plot data.. felt more like a hack and having a completedDate (while preserving completedTime) seemed like the right thing to do.

charlie-tf (approved):
I think this change makes sense 👍

karthikbtf:
> I think this change makes sense 👍

Thanks 👍

PR #1455, opened 2023-06-12
SW-3734 Highlight and Zoom to Zones in Search Results

charlie-tf:
Kapture 2023-06-12 at 15 28 53

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@mui/styled-engine-sc": "^5.12.0",
     "@mui/styles": "^5.11.9",
     "@reduxjs/toolkit": "^1.9.3",
-    "@terraware/web-components": "^2.3.45",
+    "@terraware/web-components": "^2.3.46",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^12.1.5",
     "@testing-library/user-event": "^14.4.3",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@mui/styled-engine-sc": "^5.12.0",
     "@mui/styles": "^5.11.9",
     "@reduxjs/toolkit": "^1.9.3",
-    "@terraware/web-components": "^2.3.46",
+    "@terraware/web-components": "^2.3.47",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^12.1.5",
     "@testing-library/user-event": "^14.4.3",
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -53,9 +53,9 @@ export default function Map(props: MapProps): JSX.Element {
   const [deferredFocusEntity, setDeferredFocusEntity] = useState<MapEntityId[] | undefined>();
   const mapRef = useRef(null);
   const containerRef = useRef(null);
-  const hoverStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
-  const selectStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
-  const highlightStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
+  const hoverStateId: { [key: string]: { [key: string]: number | undefined } } = useMemo(() => ({}), []);
+  const selectStateId: { [key: string]: { [key: string]: number | undefined } } = useMemo(() => ({}), []);
+  const highlightStateId: { [key: string]: { [key: string]: number | undefined } } = useMemo(() => ({}), []);
   const [firstVisible, setFirstVisible] = useState(false);
   const visible = useIsVisible(containerRef);
 
@@ -92,22 +92,39 @@ export default function Map(props: MapProps): JSX.Element {
     [onTokenExpired]
   );
 
+  const clearFeatureVar = (featureVar: any, property?: string) => {
+    const map: any = mapRef && mapRef.current;
+
+    Object.keys(featureVar).forEach((sourceId) => {
+      Object.keys(featureVar[sourceId]).forEach((id) => {
+        if (map && property) {
+          map.setFeatureState({ source: sourceId, id: featureVar[sourceId][id] }, { [property]: false });
+        }
+        delete featureVar[sourceId][id];
+      });
+      delete featureVar[sourceId];
+    });
+  };
+
   const updateFeatureState = useCallback((featureVar: any, property: string, mapEntityId: MapEntityId[]) => {
     const map: any = mapRef && mapRef.current;
     if (!map) {
       return;
     }
+
+    // clear previous features of this property
+    clearFeatureVar(featureVar, property);
+
+    // set new
     mapEntityId.forEach((entity) => {
       const { id, sourceId } = entity;
-      // clear previous
-      if (featureVar[sourceId] && id !== featureVar[sourceId]) {
-        map.setFeatureState({ source: sourceId, id: featureVar[sourceId] }, { [property]: false });
+      if (!featureVar[sourceId]) {
+        featureVar[sourceId] = {};
       }
-      // set new
       if (id) {
         map.setFeatureState({ source: sourceId, id }, { [property]: true });
+        featureVar[sourceId][id] = id;
       }
-      featureVar[sourceId] = id;
     });
   }, []);
 
@@ -168,9 +185,9 @@ export default function Map(props: MapProps): JSX.Element {
     }
     const { sources } = options;
 
-    Object.keys(hoverStateId).forEach((key) => delete hoverStateId[key]);
-    Object.keys(selectStateId).forEach((key) => delete selectStateId[key]);
-    Object.keys(highlightStateId).forEach((key) => delete highlightStateId[key]);
+    clearFeatureVar(hoverStateId);
+    clearFeatureVar(selectStateId);
+    clearFeatureVar(highlightStateId);
 
     const mouseMoveCallbacks = sources
       .filter((source) => source.isInteractive)
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -341,26 +341,6 @@ export default function Map(props: MapProps): JSX.Element {
     }
   }, [options, setGeoData, token, popupRenderer]);
 
-  useEffect(() => {
-    if (entityOptions?.highlight) {
-      if (!mapRef || !mapRef.current) {
-        setDeferredHighlightEntity(entityOptions?.highlight);
-      } else {
-        updateFeatureState(highlightStateId, 'highlight', entityOptions?.highlight);
-      }
-    }
-  }, [entityOptions?.highlight, highlightStateId, updateFeatureState]);
-
-  useEffect(() => {
-    if (entityOptions?.focus) {
-      if (!mapRef || !mapRef.current) {
-        setDeferredFocusEntity(entityOptions?.focus);
-      } else {
-        drawFocusTo(entityOptions?.focus);
-      }
-    }
-  }, [drawFocusTo, entityOptions?.focus]);
-
   const mapSources = useMemo(() => {
     if (!geoData) {
       return null;
@@ -375,6 +355,26 @@ export default function Map(props: MapProps): JSX.Element {
 
     return sources;
   }, [geoData]);
+
+  useEffect(() => {
+    if (entityOptions?.highlight) {
+      if (!mapRef || !mapRef.current) {
+        setDeferredHighlightEntity(entityOptions?.highlight);
+      } else {
+        updateFeatureState(highlightStateId, 'highlight', entityOptions?.highlight);
+      }
+    }
+  }, [entityOptions?.highlight, highlightStateId, updateFeatureState, mapSources]);
+
+  useEffect(() => {
+    if (entityOptions?.focus) {
+      if (!mapRef || !mapRef.current) {
+        setDeferredFocusEntity(entityOptions?.focus);
+      } else {
+        drawFocusTo(entityOptions?.focus);
+      }
+    }
+  }, [drawFocusTo, entityOptions?.focus, geoData]);
 
   const hasEntities = options.sources?.some((source) => {
     return source.entities?.some((entity) => entity?.boundary?.length);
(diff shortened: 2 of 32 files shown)


src/components/Map/Map.tsx
```diff
@@ -53,9 +53,9 @@ export default function Map(props: MapProps): JSX.Element {
   const [deferredFocusEntity, setDeferredFocusEntity] = useState<MapEntityId[] | undefined>();
   const mapRef = useRef(null);
   const containerRef = useRef(null);
-  const hoverStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
-  const selectStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
-  const highlightStateId: { [key: string]: number | undefined } = useMemo(() => ({}), []);
+  const hoverStateId: { [key: string]: { [key: string]: number | undefined } } = useMemo(() => ({}), []);
+  const selectStateId: { [key: string]: { [key: string]: number | undefined } } = useMemo(() => ({}), []);
+  const highlightStateId: { [key: string]: { [key: string]: number | undefined } } = useMemo(() => ({}), []);
```
karthikbtf, line 58: Extract this out as a type? I didn't know about `Record` when I first wrote it.. and was also probably lazy to extract it out earlier.
perhaps
```
  type SomethingState = Record<string, Record<string, numer | undefined>>;
```

src/components/Map/Map.tsx
```diff
@@ -92,22 +92,39 @@ export default function Map(props: MapProps): JSX.Element {
...
 
+  const clearFeatureVar = (featureVar: any, property?: string) => {
+    const map: any = mapRef && mapRef.current;
+
+    Object.keys(featureVar).forEach((sourceId) => {
+      Object.keys(featureVar[sourceId]).forEach((id) => {
+        if (map && property) {
+          map.setFeatureState({ source: sourceId, id: featureVar[sourceId][id] }, { [property]: false });
+        }
+        delete featureVar[sourceId][id];
```
karthikbtf, line 103: not a biggie,, the delete on 105 will take care of this

src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -48,13 +56,31 @@ export default function ObservationMapView({ observationsResults }: ObservationM
...
-  }, [selectedObservationDate, observationsResults, plantingSiteMapData]);
+  }, [selectedObservation, selectedObservationDate, plantingSiteMapData]);
   useEffect(() => {
     if (!plantingSiteMapData && mapData.site) {
       setPlantingSiteMapData(mapData.site);
     }
   }, [mapData, plantingSiteMapData]);
 
+  const defaultTimeZone = useDefaultTimeZone();
+  const details = useAppSelector((state) =>
```
karthikbtf, line 67: Is this app selector mostly to re-apply the search to the planting zones?
Thoughts about doing this at the parent selector level (maybe a boolean option to the selector - set to true only for mapview)? Basically [here](https://github.com/terraware/terraware-web/blob/main/src/redux/features/observations/observationsSelectors.ts#L42), pass the boolean option further down into searchZones?
Will simplify things a bit and will help more when I add search filters.
Otherwise might be good to just separate the search out inside list and map views separately.

karthikbtf (commented):
Looks good overall!
One question/suggestion about searching, would like to hear your thoughts.
A longer term question (for later),  wondering if we should look into extracting out the visual enhancements into a separate layer from the Map, maybe some sort of hook (not sure on feasibility here).

karthikbtf (approved):
We could iterate on the search thing if needed later.. once we have search filters implemented. Not sure what it might look like right now, I'm thinking they will be additional parameters to the search selector. Maybe not worth optimizing things at this time.
Otherwise looks great 🎉

src/components/Observations/map/ObservationMapView.tsx
karthikbtf: Ok, I think the search filters will work here the same way search works. Should be just fine. Ignore my previous suggestion.

src/components/Map/Map.tsx
charlie-tf: That's a good call 👍

src/components/Map/Map.tsx
charlie-tf: Oh, i suppose the GC will get it after the referencing variable has been deleted. Okay

src/components/Observations/map/ObservationMapView.tsx
charlie-tf: The selector was to get the search-filtered set of planting zones. I don't dispatch the thunk here, so I don't think it's causing any additional work, just reading the data from state. But I do see what you mean, and maybe should have created a new selector. Thanks for the suggestions here. I can leave this in place for now, but may replace it in subsequent iterations.

charlie-tf:
>wondering if we should look into extracting out the visual enhancements into a separate layer from the Map, maybe some sort of hook (not sure on feasibility here).

That's a good thought. I did feel like the map components are getting a bit burdened by this visual stuff. I can think about how extracting them might work out. Thank you

src/components/Observations/map/ObservationMapView.tsx
karthikbtf: Been thinking more about this. Figured I'd explain why my original suggestion and then let you make a call.
The map view is a full-depth view of the planting site observations. 
The observation details object passed into the map component has full depth information, figured that might be sufficient to render the map. 
Any search requirements could potentially be applied to the full depth content of the observations, and removes the need to select further data.
Hence my original suggestion. 
It is not a functional thing, more from a data provider/architecture point of view. 
It is not the only way, what we have here is just as good.

karthikbtf:
> > wondering if we should look into extracting out the visual enhancements into a separate layer from the Map, maybe some sort of hook (not sure on feasibility here).
> 
> That's a good thought. I did feel like the map components are getting a bit burdened by this visual stuff. I can think about how extracting them might work out. Thank you

Totally, now that we have more advanced requirements that vary across views  - seems like would be a fun thing to rearchitect.

src/components/Observations/map/ObservationMapView.tsx
charlie-tf: Ohh, gotcha, I understand. So `observationResults` contains all the zones for the matched observations, and I don't want to exclude zones from the map if they don't match search. But i do need to figure out which zones were matched to apply highlights and focus, which is why I added this additional selector. However, since the `searchObservationDetails` selector contains more information than is needed (much of it duplicated in `observationResults`), it might have been better to just select the list of search-matched zones here.

karthikbtf: Between selecting versus doing the search within the component - delegating to the selector (keep rendering separate) like you have in this PR seems better for separation (use redux for data views).
I was thinking along the lines of doing a hierarchical search within the selector code (if the map view is enabled) - but that requires re-selecting when user switches back and forth between list and map views -  seems wasteful, didn't think about that.
Looks like what you have here is probably best middle ground, we could revisit once all the search filters etc are implemented (next on my plate)?

charlie-tf:
ugh, testing a bit more and finding some bugginess in the way the highlight appears and resets. digging in on this further

src/components/Map/Map.tsx
```diff
@@ -341,6 +341,21 @@ export default function Map(props: MapProps): JSX.Element {
     }
   }, [options, setGeoData, token, popupRenderer]);
 
+  const mapSources = useMemo(() => {
```
charlie-tf, line 344: Just moved this up so I could use it in a dependency array.

charlie-tf:
I think I'm mostly happy with how the map highlight and focus is working now.

package.json
```diff
@@ -14,7 +14,7 @@
     "@mui/styled-engine-sc": "^5.12.0",
     "@mui/styles": "^5.11.9",
     "@reduxjs/toolkit": "^1.9.3",
-    "@terraware/web-components": "^2.3.45",
+    "@terraware/web-components": "^2.3.46",
```
karthikbtf, line 17: i have a PR out with version 47, we probably need to coordinate that

src/components/Map/Map.tsx
```diff
@@ -349,7 +364,7 @@ export default function Map(props: MapProps): JSX.Element {
         updateFeatureState(highlightStateId, 'highlight', entityOptions?.highlight);
       }
     }
-  }, [entityOptions?.highlight, highlightStateId, updateFeatureState]);
+  }, [entityOptions?.highlight, highlightStateId, updateFeatureState, mapSources]);
```
karthikbtf, line 367: are you seeing any unnecessary dependency warnings? maybe just force a check like `if (!mapSources) return;` same for the geoData dependency below?

src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -32,7 +30,13 @@ export default function ObservationMapView({ observationsResults, search }: Obse
   const [selectedObservationDate, setSelectedObservationDate] = useState<string | undefined>();
   useEffect(() => {
     if (observationsDates) {
-      setSelectedObservationDate(observationsDates[observationsDates.length - 1]);
+      setSelectedObservationDate((currentDate) => {
+        if (!currentDate || !observationsDates.includes(currentDate)) {
+          return observationsDates[observationsDates.length - 1];
```
karthikbtf, line 35: can observationDates be empty?

src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -63,23 +67,14 @@ export default function ObservationMapView({ observationsResults, search }: Obse
...
-
-  const searchZoneEntities = useMemo(
-    () => details?.plantingZones?.map((zone) => ({ sourceId: 'zones', id: zone.plantingZoneId })) ?? [],
-    [details]
-  );
+  const [searchZoneEntities, setSearchZoneEntities] = useState<MapEntityId[]>([]);
+  useEffect(() => {
+    const entities = (observationsResults ?? [])
+      .flatMap((obs) => obs.plantingZones)
+      .filter((zone) => regexMatch(zone.plantingZoneName, search))
```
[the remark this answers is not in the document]
karthikbtf: this is fine.. we can debug later when we have time to cleanup

src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -163,9 +162,11 @@ export default function ObservationMapView({ observationsResults, search }: Obse
             }}
             highlightEntities={search === '' ? [] : searchZoneEntities}
             focusEntities={
-              search === '' || searchZoneEntities.length === 0
+              search === '' && searchZoneEntities.length === 0
```
karthikbtf, line 165: does it matter if the sites layer is off?

src/components/common/MapDateSelect/DateSlider.tsx
```diff
@@ -3,12 +3,12 @@ import { useMemo } from 'react';
 
 type DateSliderProps = {
   dates: string[]; // date strings in the format 'YYYY-MM-DD'
```
karthikbtf, line 5: did we hear back on the PRD question regarding observation dates with similar month/year?

karthikbtf (approved):
Thanks for the debug, we can definitely iterate later if there are issues.
Some questions/comments otherwise good 👍

src/components/common/MapDateSelect/DateSlider.tsx
charlie-tf: What I was told is that it won't happen in practice, though for test data it's entirely possible (and will break the slider, as I see in the staging dev environment)

src/components/Map/Map.tsx
charlie-tf: Didn't see any warnings or lints about this. I do need this to run the effect on mapSources changing. What happened is that this effect was running before the map updates with new sources. If the search resulted in a new source being added and needing to be highlighted, the highlight would not be applied. Same for focus

package.json
charlie-tf: I guess whichever one goes in first will cause a merge conflict on the other, that's easy to resolve 🤷 How would you like to coordinate further than this?

src/components/Observations/map/ObservationMapView.tsx
charlie-tf: Good question, it shouldn't cause a problem, but I'll test it out real quick to make sure

src/components/Observations/map/ObservationMapView.tsx
charlie-tf: I think it can, and I should make sure I account for that. Thanks

src/components/common/MapDateSelect/DateSlider.tsx
karthikbtf: Got it, the only thing off with test data is having multiple marks with same label. Not sure if that will be an issue. We'd have to use another test org (when we get to QA) with valid observation start dates on planting sites (not like the one I created in the dev org).

src/components/Observations/map/ObservationMapView.tsx
karthikbtf: @charlie-tf one small observation (pun intended).. when we search for zones it zooms into search zone (very cool 👍 ).. when we clear the search it does not zoom back out to the global bounding box.. possible to look into that?

package.json
charlie-tf: Doesn't matter. You can merge yours. I just approved it, and it's simpler and smaller.

src/components/Observations/map/ObservationMapView.tsx
karthikbtf: I had mostly provided the zooming in/out suggestions to Rachel.

charlie-tf: I can switch it back to zooming to encompassing the whole planting site if you think that's a better experience. Let's go with that and get this merged. I can change it back if I get a different answer

karthikbtf: Google maps does it the way you have it - you can probably just clarify/make this the new requirement!

PR #1459, opened 2023-06-13
SW-3394 Deploy using Tailscale instead of bastion

sgrimm:
When deploying to the AWS hosts, connect to the VPC network using Tailscale and
ssh directly to the hosts rather than bouncing through a bastion host.


sgrimm:
See https://github.com/terraware/terraware-server/pull/1083 (this is basically the same change as that)

karthikbtf (approved):

PR #1460, opened 2023-06-13
Tooltips for planting density and total plants cards

karthikbtf:
- Rachel asked to show tooltip instead of '--'
- also the monitoring plot details will have all the data, removed '--' from there and not showing any tooltips there either
- don't know if we want to show tooltips only when data is missing, for now we show it all the time

Screen Shot 2023-06-13 at 3 42 11 PM

Screen Shot 2023-06-13 at 3 41 32 PM


charlie-tf (approved):
nice 👍

PR #1462, opened 2023-06-14
SW-3762 Error handling in drill down url routes

karthikbtf:
- the drill down pages are driven by url path parameters
- add error handling where we redirect to the parent in the drill-down hierarchy, if a parameter is not valid (otherwise the view shows empty data)
- example, if a monitoring plot is not found by id, redirect to the parent planting zone
- if planting zone is not found, redirect to parent observation details
- if observation details not found by id, redirect to parent planting site
- we already have error validation on the planting site
- updated selectors to return a valid value or null (undefined => selector hasn't completed it's fetching)

### DIFF
--- a/src/redux/features/observations/observationMonitoringPlotSelectors.ts
+++ b/src/redux/features/observations/observationMonitoringPlotSelectors.ts
@@ -22,5 +22,5 @@ export const selectObservationMonitoringPlot = createSelector(
           plantingSubzoneName: subzone.plantingSubzoneName,
         }))
       )
-      .find((monitoringPlot: PlotObservations) => monitoringPlot.monitoringPlotId === params.monitoringPlotId)
+      .find((monitoringPlot: PlotObservations) => monitoringPlot.monitoringPlotId === params.monitoringPlotId) ?? null
 );
--- a/src/redux/features/observations/utils.ts
+++ b/src/redux/features/observations/utils.ts
@@ -17,7 +17,7 @@ import { regexMatch } from 'src/utils/search';
 
 // utils
 
-export const searchResultPlots = (search: string, zone?: ObservationPlantingZoneResults) => {
+export const searchResultPlots = (search: string, zone: ObservationPlantingZoneResults | null) => {
   if (!search.trim() || !zone) {
     return zone;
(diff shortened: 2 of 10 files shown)


src/redux/features/observations/observationMonitoringPlotSelectors.ts
```diff
@@ -22,5 +22,5 @@ export const selectObservationMonitoringPlot = createSelector(
           plantingSubzoneName: subzone.plantingSubzoneName,
         }))
       )
-      .find((monitoringPlot: PlotObservations) => monitoringPlot.monitoringPlotId === params.monitoringPlotId)
+      .find((monitoringPlot: PlotObservations) => monitoringPlot.monitoringPlotId === params.monitoringPlotId) ?? null
```
charlie-tf, line 25: Is this necessary? Won't `find` return undefined anyways if its predicate fails to find a match?

src/redux/features/observations/utils.ts
```diff
@@ -17,7 +17,7 @@ import { regexMatch } from 'src/utils/search';
 
 // utils
 
-export const searchResultPlots = (search: string, zone?: ObservationPlantingZoneResults) => {
+export const searchResultPlots = (search: string, zone: ObservationPlantingZoneResults | null) => {
```
charlie-tf, line 20: Is this just to be more explicit about the result here? That is, would it have worked the same before this change?

charlie-tf (approved):
Looks good overall. Just a minor question about the use of `null` vs `undefined`, but no changes requested

charlie-tf:
Sorry, just re-read your description about the undefined vs null semantics. Please disregard my comments above

karthikbtf:
> Sorry, just re-read your description about the undefined vs null semantics. Please disregard my comments above

Thanks for the review @charlie-tf . The undefined vs null was mostly a theoretical decision. It would still work with undefineds, though I don't know about any race conditions. Let me switch back to undefineds, we don't need to make this any more complicated (we can revisit if there are issues).

charlie-tf (approved):
That does simplify things! Thanks!

PR #1464, opened 2023-06-15
SW-3348 Configure Spring Security OAuth2 on server

sgrimm:
Update the local terraware-server configuration to reflect the fact that the
server is switching to a different authentication library. Developers will
have to modify their local `.env` files to move some of the configuration to
different environment variables; `.env.sample` is updated to show what the
configuration needs to look like.


sgrimm:
CI will fail for this one until https://github.com/terraware/terraware-server/pull/1216 is merged; it probably makes sense to rerun the tests on this PR after that to make sure they all pass.

karthikbtf (approved):
Thanks!

.env.sample
```diff
@@ -1,6 +1,5 @@
 REACT_APP_TERRAWARE_API=http://localhost:8080
-KEYCLOAK_RESOURCE=
-KEYCLOAK_CREDENTIALS_SECRET=
-KEYCLOAK_REALM=
-KEYCLOAK_AUTH_SERVER_URL=https://auth.staging.terraware.io
+SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_KEYCLOAK_ISSUERURI=https://auth.staging.terraware.io/realms/terraware
+SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_KEYCLOAK_CLIENTSECRET=
```
karthikbtf, line 3: missed this.. do we need a `SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_CLIENTID`?

sgrimm: The client ID defaults to `dev-terraware-server`.

PR #1465, opened 2023-06-15
SW-3753 Map Tooltip Data

charlie-tf:
## Monitoring Plot
image
image

## Zone
image

## Planting Site
image

### DIFF
--- a/src/components/Observations/map/ObservationMapView.tsx
+++ b/src/components/Observations/map/ObservationMapView.tsx
@@ -1,9 +1,9 @@
 import { ObservationResults } from 'src/types/Observations';
-import { Box, useTheme } from '@mui/material';
+import { Box, Theme, useTheme } from '@mui/material';
 import MapLegend from 'src/components/common/MapLegend';
 import { PlantingSiteMap } from 'src/components/Map';
 import React, { useEffect, useMemo, useState } from 'react';
-import { MapEntityId, MapObject, MapSourceBaseData } from 'src/types/Map';
+import { MapEntityId, MapObject, MapSourceBaseData, MapSourceProperties } from 'src/types/Map';
 import { MapService } from 'src/services';
 import MapLayerSelect, { MapLayer } from 'src/components/common/MapLayerSelect';
 import strings from 'src/strings';
@@ -11,6 +11,17 @@ import MapDateSelect from 'src/components/common/MapDateSelect';
 import { getRgbaFromHex } from 'src/utils/color';
 import { SearchInputProps } from 'src/components/Observations/search';
 import { regexMatch } from 'src/utils/search';
+import TooltipContents from 'src/components/Observations/map/TooltipContents';
+import { makeStyles } from '@mui/styles';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  popover: {
+    '&.mapboxgl-popup': {
+      maxWidth: '324px !important', // !important to override a default mapbox style
+      width: '324px',
+    },
+  },
+}));
 
 type ObservationMapViewProps = SearchInputProps & {
   observationsResults?: ObservationResults[];
@@ -18,6 +29,7 @@ type ObservationMapViewProps = SearchInputProps & {
 
 export default function ObservationMapView({ observationsResults, search }: ObservationMapViewProps): JSX.Element {
   const theme = useTheme();
+  const classes = useStyles();
 
   const observationsDates = useMemo(() => {
     return observationsResults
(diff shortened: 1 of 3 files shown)


charlie-tf:
@rachel-tf-design @chudo76 Can you take a look at this tooltip copy and let me know if there's anything I should change? Thanks!

src/components/Observations/map/TooltipContents.tsx
```diff
@@ -0,0 +1,42 @@
...
+          </Typography>
+          <Typography fontSize='16px' fontWeight={400}>
+            {`${numSpecies ?? strings.UNKNOWN} ${strings.SPECIES}`}
+          </Typography>
+          <Typography fontSize='16px' fontWeight={400}>
+            {`${strings.PLANTING_DENSITY}: ${plantingDensity ?? strings.UNKNOWN} ${strings.PLANTS_PER_HECTARE}`}
+          </Typography>
+          <Typography fontSize='16px' fontWeight={400}>
+            {`${strings.MORTALITY_RATE}: ${percentMortality ?? strings.UNKNOWN}%`}
+          </Typography>
```
karthikbtf, line 37: should we have the quantities on the left hand side? I understand the suffixes plants/ha and % make this confusing. Or maybe have the quantities for the others show up on the right hand side, just to be consistent. Agree, this is a design decision.

src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -1,23 +1,35 @@
...
 import { SearchInputProps } from 'src/components/Observations/search';
 import { regexMatch } from 'src/utils/search';
+import TooltipContents from 'src/components/Observations/map/TooltipContents';
+import { makeStyles } from '@mui/styles';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  popover: {
+    '&.mapboxgl-popup': {
+      maxWidth: '324px !important', // !important to override a default mapbox style
+      width: '324px',
```
karthikbtf, line 21: why 324px? seems to have some extra space.. can this be dynamic based on content?

src/components/Observations/map/ObservationMapView.tsx
```diff
@@ -122,6 +134,32 @@ export default function ObservationMapView({ observationsResults, search }: Obse
...
+    } else if (properties.type === 'zone') {
+      entity = selectedObservation?.plantingZones?.find((z) => z.plantingZoneId === properties.id);
+    } else {
+      // monitoring plot
+      entity = selectedObservation?.plantingZones
+        ?.flatMap((z) => z.plantingSubzones)
+        ?.flatMap((sz) => sz.monitoringPlots)
+        ?.find((p) => p.monitoringPlotId === properties.id);
+    }
+
```
karthikbtf, line 150: Guessing we will always have an entity here.. would we ever need to return null here in case there's no entity? guessing mapbox won't call this code if a source/geometry was not clicked.

karthikbtf (commented):
Code looks good to me 👍 wanted to hear thoughts on having numbers/labels on the same side.

src/components/Observations/map/TooltipContents.tsx
charlie-tf: I think you're right. Maybe it should be "Label: quantity" for all items

src/components/Observations/map/ObservationMapView.tsx
charlie-tf: The values in properties (id, name, etc) _should_ always correspond to an actual entity.

karthikbtf: Got it, makes sense.. the callback only happens on clicked objects.

src/components/Observations/map/ObservationMapView.tsx
karthikbtf: That may be a good option. The widths are narrower in mobile, for example, in the designs.

src/components/Observations/map/TooltipContents.tsx
karthikbtf: That might look cleaner with suffixes/units. Something to iterate on with design input perhaps.

karthikbtf (approved):
This is a good iteration, we can always follow up with design feedback as we go along.

src/components/Observations/map/TooltipContents.tsx
charlie-tf: I'll try that and update the screenshots 👍

rachel-tf-design:
The copy looks great! Thanks @charlie-tf

charlie-tf:
> The copy looks great! Thanks @charlie-tf

@rachel-tf-design Thanks for taking a look! 😊

PR #1466, opened 2023-06-16
Add zone-name filters for observations results, detail view and map view

karthikbtf:
- preserve zone names in drill down if possible
- drill down may lose zone names that don't exist in the context
- updated search in map view
- best to test in vercel

Terraware App 2023-06-15 17-34-48

Terraware App 2023-06-15 17-34-24

Terraware App 2023-06-15 17-34-02


charlie-tf (approved):
Looks good. Thanks for the changes!

PR #1467, opened 2023-06-16
rename observations FE url routes to be more descriptive

karthikbtf:
- /observations/<site-id>/<observation-id>/<zone-id>/<plot-id> is vague
- rename to /observations/<site-id>/results/<observation-id>/zone/<zone-id>/plot/<plot-id>


charlie-tf (approved):
Nice 👍

PR #1470, opened 2023-06-16
Show tooltips for total plants and planting density even when there is data

karthikbtf:
- previously we were showing these tooltips only when data was missing
- Rachel provided feedback that we should show the tooltips for context all the time
- note that total plants tooltip is only relevant for the site level observations


karthikbtf:
in case this was missed

charlie-tf (approved):
👍

PR #1473, opened 2023-06-16

terraware-phrase:

### DIFF
--- a/src/strings/csv/es.csv
+++ b/src/strings/csv/es.csv
@@ -652,6 +652,7 @@ OUNCES,Onzas
 OUT_PLANTING,Replantando en el exterior
 OUTPLANT,Replantar en el exterior
 OUTPLANTS_REQUIRE_READY_SEEDLINGS,"Las replantaciones en el exterior requieren lotes con plántulas que estén ""listas""."
+OUTSTANDING,Destacado
 OVER_WORD_LIMIT,Sobre el límite de palabras
 OVERDUE,Atrasado
 OWNER,Propietario


karthikbtf (approved):

charlie-tf:
I wonder if this is the wrong sense of the word, "outstanding"
image

vs
image

@constanzauanini, your thoughts?

karthikbtf:
> I wonder if this is the wrong sense of the word, "outstanding" image
> 
> vs image
> 
> @constanzauanini, your thoughts?

Good one. Bring this up in the translations slack channel maybe. Perhaps calls this Pending

constanzauanini:
> > I wonder if this is the wrong sense of the word, "outstanding" image
> > vs image
> > @constanzauanini, your thoughts?
> 
> Good one. Bring this up in the translations slack channel maybe. Perhaps calls this Pending

Yes, this seems wrong. If we want to mean "pending" it is definitely not "destacado", I would use "pendiente". I don't know where are we using this word

charlie-tf:
> > > I wonder if this is the wrong sense of the word, "outstanding" 
> > > ...
> > > @constanzauanini, your thoughts?
> > 
> > Good one. Bring this up in the translations slack channel maybe. Perhaps calls this Pending
> 
> Yes, this seems wrong. If we want to mean "pending" it is definitely not "destacado", I would use "pendiente". I don't know where are we using this word

@constanzauanini Oh, I'm sorry for not including the context of where it is used. This is used in the observations table to indicate that an observation is scheduled but not yet started. Thanks for your input

karthikbtf:
@charlie-tf @constanzauanini maybe we could rename the key to OBSERVATION_STATUS_OUTSTANDING to disambiguate (this is a corner case, not sure we need to list other statuses with OBSERVATION_STATUS_ prefix).

karthikbtf:
> @charlie-tf @constanzauanini maybe we could rename the key to OBSERVATION_STATUS_OUTSTANDING to disambiguate (this is a corner case, not sure we need to list other statuses with OBSERVATION_STATUS_ prefix).

any thoughts?

karthikbtf:
> > > @charlie-tf @constanzauanini maybe we could rename the key to OBSERVATION_STATUS_OUTSTANDING to disambiguate (this is a corner case, not sure we need to list other statuses with OBSERVATION_STATUS_ prefix).
> > 
> > 
> > any thoughts?
> 
> Let's do that. The renamed string will most likely get the translation kicked off.

Ok, will do.

PR #1474, opened 2023-06-16
Add a Comment to Note the Sense of the Word "Outstanding" to Translators

charlie-tf:
The word "outstanding" was previously translated as "destacado," which seems to be the wrong sense of the word.

### DIFF
--- a/src/strings/csv/en.csv
+++ b/src/strings/csv/en.csv
@@ -652,7 +652,7 @@ OUNCES,Ounces
 OUT_PLANTING,Out-planting
 OUTPLANT,Outplant
 OUTPLANTS_REQUIRE_READY_SEEDLINGS,"Outplants require batches with seedlings that are ""ready."""
-OUTSTANDING,Outstanding,In the sense of Pending Unresolved or Remaining to be Done
+OUTSTANDING,Outstanding,In the sense of Pending or Unresolved
 OVER_WORD_LIMIT,Over word limit
 OVERDUE,Overdue
 OWNER,Owner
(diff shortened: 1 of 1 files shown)


src/strings/csv/en.csv
```diff
@@ -652,7 +652,7 @@ OUNCES,Ounces
 OUT_PLANTING,Out-planting
 OUTPLANT,Outplant
 OUTPLANTS_REQUIRE_READY_SEEDLINGS,"Outplants require batches with seedlings that are ""ready."""
-OUTSTANDING,Outstanding
+OUTSTANDING,Outstanding,In the sense of Pending or Unresolved
```
karthikbtf, line 655: "Remaining to be done" ?

karthikbtf (approved):
Thanks and good catch 🙏

src/strings/csv/en.csv
charlie-tf: I can add that if it further clarifies

src/strings/csv/en.csv
```diff
@@ -652,7 +652,7 @@ OUNCES,Ounces
 OUT_PLANTING,Out-planting
 OUTPLANT,Outplant
 OUTPLANTS_REQUIRE_READY_SEEDLINGS,"Outplants require batches with seedlings that are ""ready."""
-OUTSTANDING,Outstanding,In the sense of Pending or Unresolved
+OUTSTANDING,Outstanding,In the sense of Pending, Unresolved, or Remaining to be Done
```
karthikbtf: oh you can remove my suggestion, was meant as a replacement.. got that from the interwebz and was also a first hunch for me.. so ran it by you
this is a csv.. those extra commas have some semantics..
revert and check in is fine

karthikbtf (approved):
Any extra comment works, as long as it gives the right hint 👍

PR #1475, opened 2023-06-16
SW-3763 Add HTTPS proxy for remote backends

sgrimm:
The recent switch to Spring Security OAuth2 on the backend broke terraware-web
environments that were configured to send API requests to the staging or
production environments rather than using a local terraware-server instance.

The symptom is that when you try to log in, the server ends up redirecting you
to `https://localhost` instead of `http://localhost:3000`. This is due to the
original URL being lost as the API requests flow through the local Node.js
proxy, and then through a load balancer, before reaching the server; the server
"knows" that the request was using HTTPS (because it came through the load
balancer) and it knows that the hostname was `localhost` (because the hostname
is included in a special HTTP header) which causes it to construct the wrong
redirect URL.

Previously, we were using an ugly server-side hack to work around this problem,
but that hack would have to be even uglier to work with the new OAuth2 setup.
So replace it with a slightly less ugly client-side hack: rather than trying
to get the server to stop redirecting to `https://localhost`, instead run an
HTTPS server on localhost so that `https://localhost` becomes the URL of the
local dev environment.

Do this using Nginx in a Docker container so people don't have to muck around
with local installations. The Docker configuration, as well as instructions,
now live in the `remote-backend` subdirectory.


charlie-tf (approved):
Looks good, thanks for the documentation!

PR #1476, opened 2023-06-16
SW-3748 redux to list observations (not results)

karthikbtf:
- list observations which returns a response payload around the observation (not the results)
- returns end dates, includes observations for all statuses
- added hooks in components where we would do follow up UI work


karthikbtf:
@charlie-tf can you take a look at this, will unblock a few UI PRs for constanza and me

charlie-tf (approved):
Seems straightforward 👍

karthikbtf:
> Seems straightforward 👍

Thanks for the review @charlie-tf

PR #1477, opened 2023-06-19
SW-3689 Old vs New Plants Dashboard

charlie-tf:
- create PlantsDashboardV2 with placeholder
- conditionally render V1 or V2 depending on opt-in


karthikbtf (approved):
Simple 👍

PR #1478, opened 2023-06-19
SW-3767 Show observation status summary message notification

karthikbtf:
- show status summary notification if there are any monitoring plots left to observe

Terraware App 2023-06-19 15-48-31

Terraware App 2023-06-19 15-48-52

### DIFF
--- a/src/components/Observations/details/index.tsx
+++ b/src/components/Observations/details/index.tsx
@@ -1,12 +1,12 @@
 import { useEffect, useMemo } from 'react';
 import { useHistory, useParams } from 'react-router-dom';
 import { Box, Grid } from '@mui/material';
-import { TableColumnType } from '@terraware/web-components';
+import { Message, TableColumnType } from '@terraware/web-components';
 import { FieldOptionsMap } from 'src/types/Search';
 import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
 import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
-import { getShortDate } from 'src/utils/dateFormatter';
+import { getLongDate, getShortDate } from 'src/utils/dateFormatter';
 import { useLocalization } from 'src/providers';
 import { useAppSelector } from 'src/redux/store';
 import {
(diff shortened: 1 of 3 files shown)


chudo76 (approved):
I might need to be added to an org with better tracking dummy data, but I'm seeing the page message correctly on 3 of the 4 observation events `staging` (organization: `dev`). Not sure if that's a data anomaly on staging, but otherwise looks good!

karthikbtf:
> I might need to be added to an org with better tracking dummy data, but I'm seeing the page message correctly on 3 of the 4 observation events `staging` (organization: `dev`). Not sure if that's a data anomaly on staging, but otherwise looks good!

Thanks Chudo, we have a script that generates dummy data - which is how this is set up. Maybe once we start using the mobile app we could generate better data. Do you have a link to the observation that did not show correct info, I can let you know if it is a bug or data issue.

src/components/Observations/details/index.tsx
```diff
@@ -100,16 +128,38 @@ export default function ObservationDetails(props: ObservationDetailsProps): JSX.
...
+                  {
+                    strings.formatString(
+                      strings.OBSERVATIONS_REQUIRED_BY_DATE,
+                      statusSummary.pendingPlots.toString(),
+                      statusSummary.endDate
+                    ) as string
+                  }
+                </Box>
+                {
+                  strings.formatString(
```
charlie-tf, line 151: Should this be inside a `Typography` or `Box` component?

charlie-tf (approved):
Looks good! Just one question

src/components/Observations/details/index.tsx
karthikbtf: Thanks for the review @charlie-tf . I can put this inside a box similar to the first message. The Message component has predefined styles for the text, won't need Typography here.

chudo76:
> Thanks Chudo, we have a script that generates dummy data - which is how this is set up. Maybe once we start using the mobile app we could generate better data. Do you have a link to the observation that did not show correct info, I can let you know if it is a bug or data issue.

staging > org = dev > Planting Site = tw-003-wgs84 > https://staging.terraware.io/observations/56/results/6

For whatever reason, the first one of the four observations (all "June 2023") doesn't have the page message.

karthikbtf:
> > Thanks Chudo, we have a script that generates dummy data - which is how this is set up. Maybe once we start using the mobile app we could generate better data. Do you have a link to the observation that did not show correct info, I can let you know if it is a bug or data issue.
> 
> staging > org = dev > Planting Site = tw-003-wgs84 > https://staging.terraware.io/observations/56/results/6
> 
> For whatever reason, the first one of the four observations (all "June 2023") doesn't have the page message.

Ugh.. yeah, that observation has start/end dates in april and may respectively. We submitted mock observations for that observation in June (hence the june date in the table). Bogus data. We only show notifications for upcoming end dates (inclusive). Do we want to show the message for already passed end dates? In any case, I can follow up when I am back from vacay or cc @charlie-tf

PR #1479, opened 2023-06-20
Add new string for observation status 'Outstanding'

karthikbtf:


src/strings/csv/en.csv
```diff
@@ -655,7 +656,7 @@ OUNCES,Ounces
 OUT_PLANTING,Out-planting
 OUTPLANT,Outplant
 OUTPLANTS_REQUIRE_READY_SEEDLINGS,"Outplants require batches with seedlings that are ""ready."""
-OUTSTANDING,Outstanding,In the sense of Pending or Unresolved
+OUTSTANDING,Outstanding
```
charlie-tf, line 659: Do we need this string anymore?

charlie-tf (approved):

src/strings/csv/en.csv
karthikbtf: we don't have a way to delete strings yet, removing this will only re-add it.. i can remove it

charlie-tf: Oh, no need if it won't have any effect...

charlie-tf (approved):

PR #1486, opened 2023-06-20
SW-3792 / SW-3793 Update es privacy policy, remove i18n opt-in flag

karthikbtf:


charlie-tf (approved):
Simple enough. Is the locale selection still not saved in userPreferences, requiring it to be reset when reloading the page?

karthikbtf:
> Simple enough. Is the locale selection still not saved in userPreferences, requiring it to be reset when reloading the page?

I suspect that WIP comment was super early on in the effort. Locale selection is a first class property on the User object now. Example
```
{
    "id": 2,
    "email": "karthik@terraformation.com",
    "emailNotificationsEnabled": true,
    "firstName": "karthik1254",
    "lastName": "balachandran1345",
    "locale": "en",
    "timeZone": "America/Los_Angeles"
}
```

charlie-tf:
> > Simple enough. Is the locale selection still not saved in userPreferences, requiring it to be reset when reloading the page?
> 
> I suspect that WIP comment was super early on in the effort. Locale selection is a first class property on the User object now. Example
> 
> ```
> {
>     "id": 2,
>     "email": "karthik@terraformation.com",
>     "emailNotificationsEnabled": true,
>     "firstName": "karthik1254",
>     "lastName": "balachandran1345",
>     "locale": "en",
>     "timeZone": "America/Los_Angeles"
> }
> ```

Sweet, thanks.

PR #1487, opened 2023-06-20
SW-3795 react hook to export supported locales for the FE environment (prod vs non-prod)

karthikbtf:
- added an optional inDevelopment flag for locales
- added a hook to fetch supported locales which filters out inDevelopment locales in production
- tested by flipping the inProduction condition locally

### DIFF
--- a/src/components/FeatureNotification/UserNotification.tsx
+++ b/src/components/FeatureNotification/UserNotification.tsx
@@ -3,7 +3,7 @@ import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
 import { Notification } from 'src/types/Notifications';
 import { weightSystemsNames } from 'src/units';
-import { supportedLocales } from 'src/strings/locales';
+import { useSupportedLocales } from 'src/strings/locales';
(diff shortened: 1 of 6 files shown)


sgrimm (approved):
Seems reasonable to me. It's unfortunate we can't completely eliminate all mention of the unsupported locales in prod but it would definitely not be worth the effort.

karthikbtf:
> Seems reasonable to me. It's unfortunate we can't completely eliminate all mention of the unsupported locales in prod but it would definitely not be worth the effort.

Ah, remove from the code distribution? We can certainly look into it - in case it is doable cc @charlie-tf.  For now yes maybe it's ok.

charlie-tf:
> > Seems reasonable to me. It's unfortunate we can't completely eliminate all mention of the unsupported locales in prod but it would definitely not be worth the effort.
> 
> Ah, remove from the code distribution? We can certainly look into it - in case it is doable cc @charlie-tf. For now yes maybe it's ok.

Seems okay for now to me. Thanks for doing this, @karthikbtf!

PR #1488, opened 2023-06-21
SW-3798 translate species growth-form and seed-storage-behavior

karthikbtf:
- the BE returns translated values in search responses
- FE was attempting to translate those values again and it failed
- removed util to get seed storage behavior
- re util to get growth form it is used in reports and in a valid use-case


charlie-tf (approved):
LGTM!

PR #1489, opened 2023-06-21
SW-3802 Render Translated Ecosystems in Species Table

charlie-tf:
- ecosystem type is returned translated from the BE
- do not try to do the lookup for translation on the FE


karthikbtf (approved):
Thanks 👍 
similar to the other translate error

PR #1491, opened 2023-06-21
SW-3799 fix species edit modal clearing data

karthikbtf:
- add/edit modal was auto searching for the scientific name and updating the form when it was open
- this search was only intended when user actually types something
- updated code to have search trigger after has typed something
- also the species details code was incorrect and returning undefined, fixed that to return valid data
- also added conditional rendering on the modal so we don't have cross contamination issues across last opened species states


karthikbtf:
@constanzauanini when you get a chance can you help review this? Lmk if there are any questions.. added info in the PR description.

constanzauanini (approved):
Lgtm! Tested in vercel

karthikbtf:
> Lgtm! Tested in vercel

Thanks @constanzauanini

PR #1497, opened 2023-06-26
SW-3785 Mortality Rate Fill for Map

charlie-tf:
image
Notes:

- the above shows a single zone with 0% mortality rate (30% fill pattern opacity)
- this feature is not live for any map in the app, but will be applied to the new plants dashboard map


src/components/Map/PlantingSiteMap.tsx
```diff
@@ -189,6 +198,19 @@ export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Elemen
                 textSize: 16,
               }
             : undefined,
+          patternFill: showMortalityRateFill
+            ? {
+                imageName: 'mortality-rate-indicator',
+                opacityExpression: [
+                  'case',
```
charlie-tf, line 205: see https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#case

src/components/Map/Map.tsx
```diff
@@ -68,6 +77,29 @@ export default function Map(props: MapProps): JSX.Element {
...
+    (map: MapRef | null) => {
+      if (map !== null) {
+        mapRef.current = map;
+
+        // Load all needed map images here
+        // It is done in the callbackRef to ensure the images are available quickly enough for the map layers to use
+        // See https://github.com/visgl/react-map-gl/issues/1118#issuecomment-1139976172
+        mapImages?.forEach(({ name, url }) => {
+          if (!map.hasImage(name)) {
+            map.loadImage(url, (error, image) => {
```
charlie-tf, line 90: `loadImage` required a callback as an argument, which is the reason for not using `await`

charlie-tf:
> Code looks good to me! Can't test in Vercel yet I guess

Thank you! But not yet, I'm afraid :/

PR #1498, opened 2023-06-27
SW-3127 Replace Endangered with Conservation Category

sgrimm:
Remove the "endangered" flag from the species data and replace it with
the IUCN conservation category.

The change affects a few parts of the species UI:

* The table view now has separate columns for IUCN Conservation Category
  and Rare; previously the two values were combined in one column.

* The filter list also breaks these two fields out into separate settings.

* The add/edit species dialog now has a dropdown for conservation category.
  The checkbox for "Rare" is still present.

* The "conservation status" grouping is gone since it only existed to tie
  Endangered and Rare together.

### DIFF
--- a/src/api/types/generated-schema.ts
+++ b/src/api/types/generated-schema.ts
@@ -2389,8 +2389,17 @@ export interface components {
     SpeciesLookupDetailsResponsePayload: {
       /** @description List of known common names for the species, if any. */
       commonNames?: components["schemas"]["SpeciesLookupCommonNamePayload"][];
-      /** @description True if the species is known to be endangered, false if the species is known to not be endangered. This value will not be present if the server's taxonomic database doesn't indicate whether or not the species is endangered. */
-      endangered?: boolean;
+      /** @description IUCN Red List conservation category code. */
+      conservationCategory?:
+        | "CR"
+        | "DD"
+        | "EN"
+        | "EW"
+        | "EX"
+        | "LC"
+        | "NE"
+        | "NT"
+        | "VU";
       familyName: string;
       /** @description If this is not the accepted name for the species, the type of problem the name has. Currently, this will always be "Name Is Synonym". */
       problemType?: "Name Misspelled" | "Name Not Found" | "Name Is Synonym";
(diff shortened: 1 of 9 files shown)


sgrimm:
CI tests will fail until https://github.com/terraware/terraware-server/pull/1260 is merged.

sgrimm:
Table
Filters
Category Filter
Rare Filter
Edit Species

src/components/Species/index.tsx
```diff
@@ -109,9 +107,13 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
-type SpeciesSearchResultRow = Omit<Species, 'growthForm' | 'seedStorageBehavior' | 'ecosystemTypes'> & {
-  conservationStatus?: string;
+type SpeciesSearchResultRow = Omit<
+  Species,
+  'growthForm' | 'seedStorageBehavior' | 'ecosystemTypes' | 'conservationCategory' | 'rare'
+> & {
+  conservationCategory?: string;
```
constanzauanini, line 114: I don't think we need to omit conservationCategory and add it here as a string, because it's already a string

constanzauanini: Oh, no, I forgot typescript would complain about that, just leave it as it is

src/api/types/generated-schema.ts
```diff
@@ -2389,8 +2389,17 @@ export interface components {
     SpeciesLookupDetailsResponsePayload: {
       /** @description List of known common names for the species, if any. */
       commonNames?: components["schemas"]["SpeciesLookupCommonNamePayload"][];
-      /** @description True if the species is known to be endangered, false if the species is known to not be endangered. This value will not be present if the server's taxonomic database doesn't indicate whether or not the species is endangered. */
-      endangered?: boolean;
+      /** @description IUCN Red List conservation category code. */
+      conservationCategory?:
```
sgrimm, line 2393: I cheated a bit on the schema change. There are also other schema changes related to observation results but they cause one of the observation functions to break, and I didn't want to clutter this PR with changes to that code. So I just cherry-picked the schema changes related to conservation category.

src/components/Species/index.tsx
constanzauanini: Last question, do we want to show on the species table the category code? Or should we show the full label?

sgrimm: I asked the same, and the answer was that we should only show the code: https://terraformation.atlassian.net/browse/SW-3127?focusedCommentId=15190

constanzauanini (approved):

PR #1500, opened 2023-06-27
SW-3818 Add new growth forms

sgrimm:
The forestry team requested some new growth forms; add them.

Start sorting the list of growth forms in drop-down menus and the
Learn More popup to make it easier for users to find the growth form
they're looking for, since the list is now too long to take in at a
glance all at once.

### DIFF
--- a/src/components/TooltipLearnMoreModal/index.tsx
+++ b/src/components/TooltipLearnMoreModal/index.tsx
@@ -87,12 +87,10 @@ export const LearnMoreModalContentGrowthForm = (): JSX.Element => {
     [strings.MULTIPLE_FORMS, strings.LEARN_MORE_GROWTH_FORM_MULTIPLE_FORMS],
   ]
     .sort((a, b) => collator.compare(a[0], b[0]))
-    .map((x) => (
-      <>
-        <p>
-          <strong>{x[0]}:</strong> {x[1]}
-        </p>
-      </>
+    .map(([name, description], index) => (
+      <p key={`growthForm-${index}`}>
+        <strong>{name}:</strong> {description}
+      </p>
     ));
 
   return <>{elements}</>;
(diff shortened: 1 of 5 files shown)


sgrimm:
Learn More

src/components/TooltipLearnMoreModal/index.tsx
```diff
@@ -66,34 +67,36 @@ export const LearnMoreLink = (props: { onClick: () => void }): JSX.Element => {
...
+    [strings.MOSS, strings.LEARN_MORE_GROWTH_FORM_MOSS],
+    [strings.VINE, strings.LEARN_MORE_GROWTH_FORM_VINE],
+    [strings.LIANA, strings.LEARN_MORE_GROWTH_FORM_LIANA],
+    [strings.SHRUB_TREE, strings.LEARN_MORE_GROWTH_FORM_SHRUB_TREE],
+    [strings.SUBSHRUB, strings.LEARN_MORE_GROWTH_FORM_SUBSHRUB],
+    [strings.MULTIPLE_FORMS, strings.LEARN_MORE_GROWTH_FORM_MULTIPLE_FORMS],
+  ]
+    .sort((a, b) => collator.compare(a[0], b[0]))
+    .map((x) => (
+      <>
```
constanzauanini, line 91: if we want the same structure as before we don't need the` < >  < / > ` here.

sgrimm: Without those, I get an error:
```
ERROR: 91:9  jsx-key  Missing "key" prop for element.
```
Is there a better way to solve that?

constanzauanini: Yes, you can add a key prop to the <p> like this:
`<p key={`growthForm-${index}`}>` 
and to have that index available, in line 90, in the map function add the index as parameter:
`    .map((x, index) => (`

constanzauanini (approved):

PR #1502, opened 2023-06-27
SW-3819 Add new seed storage behaviors

sgrimm:
The forestry team requested some additional seed storage behavior
options for cases where the behavior hasn't yet been scientifically
confirmed.


charlie-tf (approved):
👍

sgrimm:
This needs https://github.com/terraware/terraware-server/pull/1264 as well.

PR #1504, opened 2023-06-27
SW-3774 Number of Species Card

charlie-tf:
- add yLimits to the bar chart props
- add barAnnotations and the chartjs-plugin-annotation library

image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "axios": "^0.21.1",
     "chart.js": "^4.2.0",
     "chartjs-adapter-date-fns": "^2.0.0",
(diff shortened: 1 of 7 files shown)


constanzauanini (approved):
🎉 

Just one question, shouldn't 0% value have no bar (it has a very small one now)?

charlie-tf:
> 🎉
> 
> Just one question, shouldn't 0% value have no bar (it has a very small one now)?

Good question. That's just how chartjs is rendering it. I actually find it helpful to have _something_ there, since it allows for having a hover target to get the tooltip.

PR #1507, opened 2023-06-28
Perform Population Summation Calculation Differently

charlie-tf:
- previously the calculation was extremely cumbersome
- it sometimes produced the wrong result
- simplify it

## Very Incorrect Total Plants:
image

## Very Correct Total, Matches V1 Dashboard
image

### DIFF
--- a/src/components/PlantsV2/components/TotalReportedPlantsCard.tsx
+++ b/src/components/PlantsV2/components/TotalReportedPlantsCard.tsx
@@ -15,20 +15,11 @@ export default function TotalReportedPlantsCard({ plantingSiteId }: TotalReporte
   const [totalPlants, setTotalPlants] = useState(0);
   useEffect(() => {
     if (populationSelector) {
-      const sum =
-        populationSelector?.reduce((acc1, zone) => {
-          const sum1 = Number(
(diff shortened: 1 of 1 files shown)


src/components/PlantsV2/components/TotalReportedPlantsCard.tsx
```diff
@@ -15,20 +15,11 @@ export default function TotalReportedPlantsCard({ plantingSiteId }: TotalReporte
...
-              return acc2 + (isNaN(sum2) ? 0 : sum2);
-            }, 0)
-          );
-          return acc1 + (isNaN(sum1) ? 0 : sum1);
-        }, 0) ?? 0;
+      const populations = populationSelector
+        .flatMap((zone) => zone.plantingSubzones)
+        .flatMap((sz) => sz.populations)
+        .filter((pop) => pop !== undefined);
+      const sum = populations.reduce((acc, pop) => +pop.totalPlants + acc, 0);
```
charlie-tf, line 22: hopefully much clearer than the mess i had here before 😅

constanzauanini: This looks much simpler now!

PR #1509, opened 2023-06-29
SW-3647 Accession species only editable if no deliveries

sgrimm:
In preparation for an upcoming change that causes changes to an accession's
species to be propagated to seedling batches that come from that accession,
add logic to prevent editing an accession's species if the accession's seeds
have already been sent to a nursery, and seedlings have already been withdrawn
from the nursery and delivered to a planting site.

This prevents accession species changes from causing inconsistencies between
nursery withdrawals and the species populations at planting sites. In the
future we may remove this restriction and make the species changes propagate to
planting site data too, but doing so would introduce a number of edge cases
that would require more detailed requirements definition.

If the species dropdown is disabled, there is a tooltip explaining why.

### DIFF
--- a/src/api/types/generated-schema.ts
+++ b/src/api/types/generated-schema.ts
@@ -470,6 +470,8 @@ export interface components {
       estimatedWeight?: components["schemas"]["SeedQuantityPayload"];
       /** Format: int64 */
       facilityId: number;
+      /** @description If true, plants from this accession's seeds were delivered to a planting site. */
+      hasDeliveries?: boolean;
       /**
        * Format: int64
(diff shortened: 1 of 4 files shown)


sgrimm:
https://github.com/terraware/terraware-server/pull/1269 is the server side of this, but it should be safe to merge this PR first (the new API field is marked as optional, and if it's missing, the species will be editable as normal).

sgrimm:
Oops, forgot to include a screenshot.
Screenshot 2023-06-29 at 1 58 21 PM

charlie-tf (approved):
Looks good!

PR #1510, opened 2023-06-30
SW-3778 Zone Level Data Map

charlie-tf:
## Without Observation
image

## With Observation
image

TODO (in a followup SW-3839): add tooltips


constanzauanini (approved):
Tested in Vercel and looks good!

PR #1513, opened 2023-07-04
SW-3782 Highest and lowest mortality rate card

constanzauanini:
Screenshot 2023-07-04 at 12 03 17

### DIFF
--- a/src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
+++ b/src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
@@ -1,0 +1,91 @@
+import React from 'react';
+import OverviewItemCard from 'src/components/common/OverviewItemCard';
+import strings from 'src/strings';
+import { Box, Divider, Typography, useTheme } from '@mui/material';
+import { ObservationResults } from 'src/types/Observations';
+import { useAppSelector } from 'src/redux/store';
+import { selectObservationPlantingZone } from 'src/redux/features/observations/observationPlantingZoneSelectors';
+import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
+
+type HighestAndLowestMortalityRateCardProps = {
+  observation?: ObservationResults;
+};
+
+export default function TotalMortalityRateCard({ observation }: HighestAndLowestMortalityRateCardProps): JSX.Element {
+  const theme = useTheme();
+  const defaultTimeZone = useDefaultTimeZone();
+
+  let highestMortalityRate = 0;
+  let highestZoneId: number;
+  observation?.plantingZones.forEach((zone) => {
+    if (zone.mortalityRate > highestMortalityRate) {
+      highestMortalityRate = zone.mortalityRate;
+      highestZoneId = zone.plantingZoneId;
+    }
+  });
+
+  let lowestMortalityRate = 100;
+  let lowestZoneId: number;
+  observation?.plantingZones.forEach((zone) => {
+    if (zone.mortalityRate < lowestMortalityRate) {
+      lowestMortalityRate = zone.mortalityRate;
+      lowestZoneId = zone.plantingZoneId;
+    }
+  });
+
+  const highestPlantingZone = useAppSelector((state) =>
+    selectObservationPlantingZone(
+      state,
+      {
+        plantingSiteId: Number(observation?.plantingSiteId),
+        observationId: Number(observation?.observationId),
+        plantingZoneId: Number(highestZoneId),
+      },
+      defaultTimeZone.get().id
+    )
+  );
+
+  const lowestPlantingZone = useAppSelector((state) =>
+    selectObservationPlantingZone(
+      state,
+      {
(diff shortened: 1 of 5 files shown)


src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const defaultTimeZone = useDefaultTimeZone();
+
+  let highestMortalityRate = 0;
+  let highestZoneId: number;
+  observation?.plantingZones.forEach((zone) => {
+    if (zone.mortalityRate > highestMortalityRate) {
+      highestMortalityRate = zone.mortalityRate;
+      highestZoneId = zone.plantingZoneId;
+    }
+  });
```
charlie-tf, line 25: This would fail to pick any zone if you had all zones with mortality rate of `0`.

src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  let lowestMortalityRate = 100;
+  let lowestZoneId: number;
+  observation?.plantingZones.forEach((zone) => {
+    if (zone.mortalityRate < lowestMortalityRate) {
+      lowestMortalityRate = zone.mortalityRate;
+      lowestZoneId = zone.plantingZoneId;
+    }
+  });
```
charlie-tf, line 34: Likewise, if all zones had mortality rate `100`, then none would get picked.

src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const highestPlantingZone = useAppSelector((state) =>
+    selectObservationPlantingZone(
+      state,
+      {
+        plantingSiteId: Number(observation?.plantingSiteId),
+        observationId: Number(observation?.observationId),
+        plantingZoneId: Number(highestZoneId),
```
charlie-tf, line 42: If no zone is picked, this is `NaN`

src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    )
+  );
+
+  const lowestPlantingZone = useAppSelector((state) =>
+    selectObservationPlantingZone(
+      state,
+      {
+        plantingSiteId: Number(observation?.plantingSiteId),
+        observationId: Number(observation?.observationId),
+        plantingZoneId: Number(lowestZoneId),
```
charlie-tf, line 54: Same here

src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <Typography fontSize='12px' fontWeight={400}>
+            {strings.HIGHEST}
+          </Typography>
+          {highestPlantingZone && (
+            <>
+              <Typography fontSize='24px' fontWeight={600} paddingY={theme.spacing(2)}>
+                {highestPlantingZone.plantingZoneName}
+              </Typography>
+              <Typography fontSize='24px' fontWeight={600}>
+                {highestMortalityRate + '%'}
```
charlie-tf, line 77: Does this every need to be rounded?

charlie-tf (changes requested):
Minor issue with selecting the highest and lowest rate zones, but you should be able to solve using `<=` and `>=`.

src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
constanzauanini: Thank you!

src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
constanzauanini: If no zone is picked, nothing shows because the render has this validation` lowestPlantingZone && `

src/components/PlantsV2/components/HighestAndLowestMortalityRateCard.tsx
charlie-tf: But we want to show _something_, correct? Something would be picked and displayed in the case of a tie that was greater than zero, so it seems like we want to display something if the tie is _at_ zero.

constanzauanini: Oooh, that's why you were saying to use `>=` . Now I gotcha. Will fix it

charlie-tf: Thanks!

charlie-tf (approved):
Thanks for the update!

PR #1514, opened 2023-07-04
SW-3783 Highest and lowest mortality species rate card

constanzauanini:
Screenshot 2023-07-04 at 15 06 41


src/components/PlantsV2/components/HighestAndLowestMortalityRateSpeciesCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const theme = useTheme();
+  const species = useAppSelector(selectSpecies);
+
+  let highestMortalityRate = 0;
+  let highestSpecies = '';
+
+  observation?.species.forEach((sp) => {
+    if (sp.mortalityRate && sp.mortalityRate > highestMortalityRate) {
+      highestMortalityRate = sp.mortalityRate;
+      // highestSpecies = sp.speciesName || '';
```
constanzauanini, line 25: Not sure why this speciesName is always null

charlie-tf: According to the api this field is:
```
If certainty is Other, the user-supplied name of the species. Null if certainty is Known or Unknown.
```
So it shouldn't be relied upon for a known species, i.e. one with an ID. The way your look it up below is correct.

charlie-tf (approved):
👍

PR #1515, opened 2023-07-05
SW-3784 Add live/dead plants per species chart card

constanzauanini:
Screenshot 2023-07-05 at 10 33 30


src/components/PlantsV2/components/LiveDeadPlantsPerSpeciesCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  useEffect(() => {
+    if (selectedSpecies) {
+      setLabels([strings.LIVE, strings.DEAD]);
+      const selectedObservationSpecies = observation?.species.find(
+        (sp) => sp.speciesId?.toString() === selectedSpecies
+      );
+      if (selectedObservationSpecies) {
+        const totalPlants = selectedObservationSpecies.totalPlants;
+        const dead = selectedObservationSpecies.mortalityRate || (0 * totalPlants) / 100;
```
constanzauanini, line 43: I used the latest observation and the mortality rate and total plants of each species to calculate this. Not sure if this is correct or not

src/components/common/Chart/BarChart.tsx
```diff
@@ -102,7 +108,7 @@ function BarChartContent(props: BarChartContentProps): JSX.Element {
       if (ctx) {
         const colors = generateTerrawareRandomColors(theme, chartLabels?.length || 0);
         chartRef.current = await newChart(locale, ctx, {
-          type: 'bar',
+          type: type || 'bar',
```
karthikbtf, line 111: This is a BarChart component.. we should probably create a generic Chart component instead of supporting new types in the BarChart?
My suggestion:
1. Create a new Chart.tsx in this directory which is essentially a copy of this BarChart.tsx
2. In this BarChart.tsx simply call Chart.tsx with type='bar' so we don't have to update all the other chart calls
3. you can optionally also create a PieChart.tsx in this directory.. and set showLegend=true

src/components/PlantsV2/components/LiveDeadPlantsPerSpeciesCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  useEffect(() => {
+    if (selectedSpecies) {
+      setLabels([strings.LIVE, strings.DEAD]);
+      const selectedObservationSpecies = observation?.species.find(
+        (sp) => sp.speciesId?.toString() === selectedSpecies
+      );
+      if (selectedObservationSpecies) {
+        const totalPlants = selectedObservationSpecies.totalPlants;
+        const dead = ((selectedObservationSpecies.mortalityRate || 0) * totalPlants) / 100;
```
karthikbtf, line 42: yeah.. i noticed we have some null mortality rates.. ideally have the redux framework set the moratlity rate to 0 in the reducer or selector.. all other components will get this adjustment for free

src/components/PlantsV2/components/LiveDeadPlantsPerSpeciesCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+            options={allSpecies}
+            selectedValue={selectedSpecies}
+          />
+          <Box marginTop={theme.spacing(3)}>
+            <PieChart
+              chartId='liveDeadplantsBySpecies'
+              chartLabels={labels}
+              chartValues={values}
+              maxWidth='100%'
+              showLegend={true}
```
karthikbtf, line 72: remove this?

src/components/common/Chart/Chart.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+const useStyles = makeStyles<Theme, StyleProps>(() => ({
+  chart: {
+    minHeight: (props) => props.minHeight ?? '200px',
+    maxWidth: (props) => props.maxWidth ?? undefined,
+  },
+}));
+
+export interface ChartProps {
+  type?: keyof ChartTypeRegistry;
```
karthikbtf, line 26: make this required?

src/components/common/Chart/Chart.tsx
```diff
@@ -0,0 +1,170 @@
...
+      locale={locale}
+      chartLabels={chartLabels}
+      chartValues={chartValues}
+      key={`${locale}_${chartLabels.length}_${chartValues.length}`}
+    />
+  );
+}
+
+interface ChartContentProps {
+  type?: keyof ChartTypeRegistry;
```
karthikbtf, line 64: required

src/components/common/Chart/Chart.tsx
```diff
@@ -0,0 +1,170 @@
...
+      if (chartRef.current) {
+        chartRef.current.destroy();
+        chartRef.current = null;
+      }
+
+      const ctx = canvasRef?.current?.getContext('2d');
+      if (ctx) {
+        const colors = generateTerrawareRandomColors(theme, chartLabels?.length || 0);
+        chartRef.current = await newChart(locale, ctx, {
+          type: type || 'bar',
```
karthikbtf, line 111: remove `|| 'bar'`?

src/components/common/Chart/Chart.tsx
```diff
@@ -0,0 +1,170 @@
...
+          data: {
+            labels: chartLabels,
+            datasets: [
+              {
+                data: chartValues,
+                barThickness,
+                backgroundColor: colors,
+                minBarLength: 3,
```
karthikbtf, line 119: interesting. didn't think about this.. does the barThickness and bar length affect pie chart or are those options ignored?
Not sure how the datasets should be populated based on chart type.. if pie and bar datasets should be different.. maybe they should come from BarChart and PieChart components.
If it works as-is.. we can maybe revisit later.

src/types/Observations.ts
```diff
@@ -56,7 +56,8 @@ export type ObservationMonitoringPlotPhoto = components['schemas']['ObservationM
 
 // species related observation statistics
 export type ObservationSpeciesResultsPayload = components['schemas']['ObservationSpeciesResultsPayload'];
-export type ObservationSpeciesResults = ObservationSpeciesResultsPayload & {
+export type ObservationSpeciesResults = Omit<ObservationSpeciesResultsPayload, 'moralityRate'> & {
```
karthikbtf, line 59: thanks, this is great... sorry for the flip-flopping.. should we maybe show '--' for null mortality rates? maybe 0 is misleading.. don't know. we could check with Steve/design team maybe? Once again, sorry for backtracking here.

karthikbtf (commented):
Thanks for the follow up. Looks good, a few comments/questions.

karthikbtf (approved):
Looks good otherwise. If you can follow up on what to do for null mortality rate - that might help.

src/types/Observations.ts
constanzauanini: Yes, I wasn't sure either. I will ask design what to do

src/components/common/Chart/Chart.tsx
constanzauanini: Datasets are the same, array of values and labels. barThickness and barLength are specific to bar chart, they don't affect pie chart though

src/types/Observations.ts
constanzauanini: I will leave it as undefined, and not show the chart in case we have a mortality rate undefined

karthikbtf: Ok sounds good thanks!!

src/components/PlantsV2/components/HighestAndLowestMortalityRateSpeciesCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const species = useAppSelector(selectSpecies);
+
+  let highestMortalityRate = 0;
+  let highestSpecies = '';
+
+  observation?.species.forEach((sp) => {
+    if (sp.mortalityRate && sp.mortalityRate > highestMortalityRate) {
+      highestMortalityRate = sp.mortalityRate;
+      // highestSpecies = sp.speciesName || '';
+      highestSpecies = species?.find((iSpecies) => iSpecies.id === sp.speciesId)?.scientificName || '';
```
karthikbtf, line 26: we could use `??` instead of `||`,,, same result but just to clarify intent

src/components/PlantsV2/components/HighestAndLowestMortalityRateSpeciesCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+    }
+  });
+
+  let lowestMortalityRate = 100;
+  let lowestSpecies = '';
+
+  observation?.species.forEach((sp) => {
+    if (sp.mortalityRate && sp.mortalityRate < lowestMortalityRate) {
```
karthikbtf, line 34: if `sp.mortalityRate` is '0', the comparison will be ignored.. we should probably look for `!isNaN(mortalityRate)`

src/components/PlantsV2/components/HighestAndLowestMortalityRateZonesCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    if (zone.mortalityRate > highestMortalityRate) {
+      highestMortalityRate = zone.mortalityRate;
+      highestZoneId = zone.plantingZoneId;
+    }
+  });
+
+  let lowestMortalityRate = 100;
+  let lowestZoneId: number;
+  observation?.plantingZones.forEach((zone) => {
+    if (zone.mortalityRate < lowestMortalityRate) {
```
karthikbtf, line 30: null < 100 is true

src/components/PlantsV2/components/LiveDeadPlantsPerSpeciesCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }, [observation]);
+
+  useEffect(() => {
+    if (selectedSpecies) {
+      setLabels([strings.LIVE, strings.DEAD]);
+      const selectedObservationSpecies = observation?.species.find(
+        (sp) => sp.speciesId?.toString() === selectedSpecies
+      );
+      if (selectedObservationSpecies && selectedObservationSpecies.mortalityRate) {
```
karthikbtf, line 41: mortalityRate can be '0' but the check here will fail

karthikbtf (commented):
Looked at this again. A few questions/comments. Suggestions to consider mortality rates of '0' and 'null/undefined' as well in the various components.

src/components/PlantsV2/components/HighestAndLowestMortalityRateSpeciesCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  observation,
+}: HighestAndLowestMortalityRateSpeciesCardProps): JSX.Element {
+  const theme = useTheme();
+  const species = useAppSelector(selectSpecies);
+
+  let highestMortalityRate = 0;
+  let highestSpecies = '';
+
+  observation?.species.forEach((sp) => {
+    if (sp.mortalityRate && sp.mortalityRate > highestMortalityRate) {
```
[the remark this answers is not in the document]
constanzauanini: The highest will be 0, but there will no be any "highestSpecies", so the value doesn't show. There is a check in the render for highestSpecies (and lowestSpecies)

src/components/PlantsV2/components/HighestAndLowestMortalityRateSpeciesCard.tsx
```diff
@@ -20,19 +20,19 @@ export default function HighestAndLowestMortalityRateSpeciesCard({
   let highestSpecies = '';
 
   observation?.species.forEach((sp) => {
-    if (sp.mortalityRate && sp.mortalityRate > highestMortalityRate) {
+    if (sp.mortalityRate !== undefined && sp.mortalityRate >= highestMortalityRate) {
```
karthikbtf, line 23: Thanks for the changes @constanzauanini 
Just curious... I see a lot of `null` mortalityRates as well... any idea what we should do about those?
Do we need a !== undefined and !== null check? A bit confused for my own purposes.

constanzauanini: I think value != null will be true for null or undefined, but I don't think the linter will allow the double equals (instead of triple)

karthikbtf: sounds good, maybe just 2 separate checks..
here's what I see in staging  for https://staging.terraware.io/observations/56/results/5/zone/107 (dev org)
DevTools - staging terraware io:observations:56:results:5:zone:107 2023-07-06 08-07-50

PR #1522, opened 2023-07-06
SW-3787 Adding dashboard header text

constanzauanini:
Screenshot 2023-07-06 at 16 45 04

### DIFF
--- a/src/components/PlantsV2/index.tsx
+++ b/src/components/PlantsV2/index.tsx
@@ -170,7 +170,7 @@ export default function PlantsDashboardV2(): JSX.Element {
         latestObservation?.completedTime
           ? (strings.formatString(
               strings.DASHBOARD_HEADER_TEXT,
-              <b>{strings.formatString(strings.SAMPLE_OF, 2000)}</b>,
+              <b>2000</b>,
               <>{getShortDate(latestObservation.completedTime, locale.activeLocale)}</>
             ) as string)
           : undefined
--- a/src/components/PlantsV2/index.tsx
+++ b/src/components/PlantsV2/index.tsx
@@ -164,14 +164,11 @@ export default function PlantsDashboardV2(): JSX.Element {
     !!plantingSiteResult && !!plantingSiteResult.boundary && plantingSiteResult.boundary.coordinates?.length > 0;
 
   const getObservationHectares = () => {
-    let observationMonitoringPlots = 0;
-    latestObservation?.plantingZones.forEach((pz) => {
-      pz.plantingSubzones.forEach((psz) => {
-        observationMonitoringPlots += psz.monitoringPlots.length;
-      });
-    });
+    const numMonitoringPlots =
+      latestObservation?.plantingZones.flatMap((pz) => pz.plantingSubzones.flatMap((psz) => psz.monitoringPlots))
+        ?.length ?? 0;
     const monitoringPlotHa = 0.0625;
-    const totalHa = observationMonitoringPlots * monitoringPlotHa;
+    const totalHa = numMonitoringPlots * monitoringPlotHa;
     return totalHa;
   };
 
--- a/src/components/PlantsV2/index.tsx
+++ b/src/components/PlantsV2/index.tsx
@@ -163,6 +163,18 @@ export default function PlantsDashboardV2(): JSX.Element {
   const hasPolygons =
     !!plantingSiteResult && !!plantingSiteResult.boundary && plantingSiteResult.boundary.coordinates?.length > 0;
 
+  const getObservationHectares = () => {
+    let observationMonitoringPlots = 0;
+    latestObservation?.plantingZones.forEach((pz) => {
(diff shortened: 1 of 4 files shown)


src/components/PlantsV2/index.tsx
```diff
@@ -164,6 +166,15 @@ export default function PlantsDashboardV2(): JSX.Element {
...
     <PlantsPrimaryPage
       title={strings.DASHBOARD}
+      text={
+        latestObservation?.completedTime
+          ? (strings.formatString(
+              strings.DASHBOARD_HEADER_TEXT,
+              <b>{strings.formatString(strings.SAMPLE_OF, 2000)}</b>,
+              <>{getShortDate(latestObservation.completedTime, locale.activeLocale)}</>
+            ) as string)
+          : undefined
```
karthikbtf, line 176: guessing `2000` is only for testing...
This construction of sentences pattern is not going to be localization friendly. It will break (joining two parts of translated sentences is not guaranteed to make sense as a full sentence in another language). 
I have brought this up with design before. We don't want to be supporting varying styles across text. Maybe just bolden the number or make the entire sentence bold.

src/strings/csv/en.csv
```diff
@@ -199,6 +199,7 @@ CUSTOMIZE_TABLE_COLUMNS,Customize table columns
 CUSTOMIZE_TABLE_COLUMNS_DESCRIPTION,Select columns you want to add. Deselect columns you want to remove.
 CUT_TEST,Cut Test
 DASHBOARD,Dashboard
+DASHBOARD_HEADER_TEXT,The data on this dashboard is based on a {0} from the {1} observation.
```
karthikbtf, line 202: Add a third field describing what {1} is or maybe use an example value.

karthikbtf (commented):
I know this is only a draft but wanted to capture something early on.

src/components/PlantsV2/index.tsx
constanzauanini: Yes, not sure where to get that number yet. 
That's totally true. Thanks for catching this up early. Ok, I will bold just the number for now

karthikbtf: You can certainly ask design for suggestions :)

src/components/PlantsV2/index.tsx
```diff
@@ -170,7 +170,7 @@ export default function PlantsDashboardV2(): JSX.Element {
         latestObservation?.completedTime
           ? (strings.formatString(
               strings.DASHBOARD_HEADER_TEXT,
-              <b>{strings.formatString(strings.SAMPLE_OF, 2000)}</b>,
+              <b>2000</b>,
```
karthikbtf, line 173: were you able to find out where to get this info from? is that why this is still in draft more?

constanzauanini: I still don't know. The planting site hectares doesn't make sense here, because the text says "a sample of ... based on last observation", maybe I should sum up all plantingZones' areas of the lastObservation, but still waiting confirmation. And yes, that's why I left it as draft

karthikbtf: Got it, this could be the area of all monitoring plots that were sampled in this observation round. Maybe sum the number of monitoring plots in the observation results.. and multiply that by the area of a monitoring plot? I think the area is standard (get it from Alex or Steve maybe).

constanzauanini: Thanks Karthik! Should I wait for a confirmation on how to calculate it or just do it that way?

karthikbtf: You can mention this approach and confirm it. My understanding is as such.

src/components/PlantsV2/index.tsx
```diff
@@ -161,9 +163,30 @@ export default function PlantsDashboardV2(): JSX.Element {
   const hasPolygons =
     !!plantingSiteResult && !!plantingSiteResult.boundary && plantingSiteResult.boundary.coordinates?.length > 0;
 
+  const getObservationHectares = () => {
+    let observationMonitoringPlots = 0;
+    latestObservation?.plantingZones.forEach((pz) => {
+      pz.plantingSubzones.forEach((psz) => {
+        observationMonitoringPlots += psz.monitoringPlots.length;
+      });
+    });
```
karthikbtf, line 172: Nice!! we can certainly get this in and iterate as we get confirmation.
Re: code.. another suggestion if you want something shorter/functional
```
const numMonitoringPlots = latestObservation?.plantingZones.flatMap(pz => pz.plantingSubzones.flatMap(psz => psz.monitoringPlots))?.length ?? 0;
```

karthikbtf (approved):

src/components/PlantsV2/index.tsx
constanzauanini: Thanks!! Will use this nicer way. And yes, when I get a confirmation I will iterate if needed

PR #1524, opened 2023-07-06
Chart Grouping

charlie-tf:
- Allow multiple datasets per series
- For bar charts, this allows multiple bars per x-label

## Example:
- Each datum is grouped with a second bar that is half the height and the same color
image

### DIFF
--- a/src/components/Observations/common/SpeciesMortalityRateChart.tsx
+++ b/src/components/Observations/common/SpeciesMortalityRateChart.tsx
@@ -32,7 +32,7 @@ export default function SpeciesTotalPlantsChart({ minHeight, species }: SpeciesT
 
   return (
     <BarChart
-      chartId='observationsTotalPlantsBySpecies'
+      chartId='observationsMortalityRateBySpecies'
       chartData={{
         labels: mortalityRates.labels,
         datasets: [
--- a/src/components/Observations/common/SpeciesMortalityRateChart.tsx
+++ b/src/components/Observations/common/SpeciesMortalityRateChart.tsx
@@ -33,8 +33,14 @@ export default function SpeciesTotalPlantsChart({ minHeight, species }: SpeciesT
   return (
     <BarChart
       chartId='observationsTotalPlantsBySpecies'
-      chartLabels={mortalityRates.labels}
-      chartValues={mortalityRates.values}
+      chartData={{
+        labels: mortalityRates.labels,
+        datasets: [
+          {
+            values: mortalityRates.values,
+          },
+        ],
+      }}
       barWidth={0}
       minHeight={minHeight}
     />
--- a/src/components/Observations/common/SpeciesMortalityRateChart.tsx
+++ b/src/components/Observations/common/SpeciesMortalityRateChart.tsx
@@ -30,16 +30,17 @@ export default function SpeciesTotalPlantsChart({ minHeight, species }: SpeciesT
     return data;
   }, [species]);
 
-  const chartData = useMemo(() => {
-    return {
+  const chartData = useMemo(
+    () => ({
       labels: mortalityRates.labels,
       datasets: [
         {
           values: mortalityRates.values,
         },
       ],
-    };
-  }, [mortalityRates]);
+    }),
+    [mortalityRates]
+  );
 
   return (
     <BarChart chartId='observationsMortalityRateBySpecies' chartData={chartData} barWidth={0} minHeight={minHeight} />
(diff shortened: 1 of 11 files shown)


karthikbtf (commented):
Looks good, a few questions. Also, just curious on use-case here.

src/components/common/Chart/Chart.tsx
```diff
@@ -27,6 +38,7 @@ export interface ChartProps {
   chartId: string;
   chartLabels?: string[];
   chartValues?: number[];
+  chartData?: ChartData;
```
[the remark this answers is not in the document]
charlie-tf: Yeah, I thought to keep them both, to have a simpler option, but I think it clutters the interface. I'll remove it

src/components/common/Chart/Chart.tsx
```diff
@@ -10,6 +10,17 @@ import { AnnotationPluginOptions } from 'chartjs-plugin-annotation/types/options
 
 ChartJS.register(annotationPlugin);
 
+type ChartDataset = {
+  color?: string;
+  values: number[];
+  label?: string;
```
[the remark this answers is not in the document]
charlie-tf: This is the label for the dataset and shows up in the legend and tooltip. See the below example from the chartjs docs ("Dataset 1" is the label):
image

I'll add a comment to try to explain it, because yes, it is somewhat confusing with the x-axis label there as well.

charlie-tf:
Hmm, getting an error for the pie chart now. Will look into further tomorrow

charlie-tf:
>just curious on use-case here

@karthikbtf I'm adding this to be able to display grouped bars for the planting zone density chart:
image

src/components/common/Chart/Chart.tsx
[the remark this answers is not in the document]
karthikbtf: Will leave it to you. Theoretically will be good to have just one way of supplying the data - otherwise gets ambiguous. Whatever makes sense practically is totally fine.

src/components/common/Chart/Chart.tsx
```diff
@@ -121,7 +121,6 @@ function ChartContent(props: ChartContentProps): JSX.Element {
               barThickness,
               backgroundColor: ds.color ?? colors,
               minBarLength: 3,
-              stack: index.toString(),
```
charlie-tf: Thought this was necessary for the bar groups (from looking at an example), but they work fine without, and it was breaking the pie chart.

src/components/common/Chart/Chart.tsx
```diff
@@ -167,7 +167,7 @@ function ChartContent(props: ChartContentProps): JSX.Element {
     };
     createChart();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [chartLabels, chartValues, locale]);
+  }, [chartData, locale]);
```
karthikbtf, line 170: suggestion: use `useMemo` in all the clients to populate the chartData, they currently set the chartData in the props, which will trigger this useEffect a lot.. you could optionally test whether this is triggered a lot

karthikbtf (commented):
Thanks for cleaning up the interface 🎉 
one suggestion in a comment below

src/components/common/Chart/Chart.tsx
charlie-tf: Yeah, I am seeing quite a bit of that effect running. I'll look into your suggestion. Thanks!

charlie-tf:
@constanzauanini This PR changes up the chart interface a bit. Wanted to make sure you're aware and have a chance to comment

constanzauanini:
> @constanzauanini This PR changes up the chart interface a bit. Wanted to make sure you're aware and have a chance to comment

Yes, I was following the pr. It looks good for me changing the labels and values for chartData objects

karthikbtf (commented):
looks good thanks, one chart was missed (added a comment)
unrelated to this PR, noticed that `truncateToLength` is defined in all the plantsv2 charts - guessing you need to rebase with main since that func was moved into utils?

src/components/Observations/common/SpeciesMortalityRateChart.tsx
```diff
@@ -30,13 +30,18 @@ export default function SpeciesTotalPlantsChart({ minHeight, species }: SpeciesT
     return data;
   }, [species]);
 
+  const chartData = useMemo(() => {
```
karthikbtf, line 33: nit here and in the other chart, use es6 short syntax (skip the return)
```
useMemo(() => ({
  labels: ...,
  datasets: [...],
}), [mortalityRates]);
```

karthikbtf (approved):
Thanks for the revisions.. and your patience.

karthikbtf (approved):

PR #1525, opened 2023-07-07
SW-3857 Initial redux to fetch observations results for a single planting site

karthikbtf:
- Redux to dispatch/select planting-site observations results, advanced selectors to aggregate information will be in subsequent PRs
- This will be used on the location->PlantingSites->PlantingSite page
- We already have redux for the org-wide results, which is a superset of what this PR does
- We could theoretically leverage the org-wide redux for this view but it seems wasteful to fetch all the data when not needed
- Optionally we could trigger the fetching of org-wide data during App initialization and select what we want in this view (and for the observations views), downside is we may have stale data over time unless we periodically refresh the data, another downside is the load on server because app load guarantees we fetch observations results for all sessions
- Downside to this current PR is maintaining two code paths to fetch observations results (org-wide versus planting-site specific)
- Prefer a downside that affects us versus one that affects the end-user, thoughts?

### DIFF
--- a/src/redux/features/observations/observationsSlice.ts
+++ b/src/redux/features/observations/observationsSlice.ts
@@ -48,3 +48,26 @@ export const observationsSlice = createSlice({
 
 export const { setObservationsAction } = observationsSlice.actions;
 export const observationsReducer = observationsSlice.reducer;
+
+type PlantingSitePayload = {
+  plantingSiteId: number;
+  data: ResultsData;
+};
+
+type PlantingSiteData = Record<number, ResultsData>;
+
+const plantingSiteInitialResultsState: PlantingSiteData = {};
+
+export const plantingSiteObservationsResultsSlice = createSlice({
+  name: 'plantingSiteObservationsResultsSlice',
+  initialState: plantingSiteInitialResultsState,
+  reducers: {
+    setPlantingSiteObservationsResultsAction: (state, action: PayloadAction<PlantingSitePayload>) => {
+      const data: ResultsData = action.payload.data;
+      state[action.payload.plantingSiteId] = data;
+    },
+  },
+});
+
+export const { setPlantingSiteObservationsResultsAction } = plantingSiteObservationsResultsSlice.actions;
+export const plantingSiteObservationsResultsReducer = plantingSiteObservationsResultsSlice.reducer;
--- a/src/redux/features/observations/observationsThunks.ts
+++ b/src/redux/features/observations/observationsThunks.ts
@@ -1,7 +1,11 @@
 import { Dispatch } from 'redux';
 import { ObservationsService } from 'src/services';
 import { RootState } from 'src/redux/rootReducer';
-import { setObservationsAction, setObservationsResultsAction } from './observationsSlice';
+import {
+  setObservationsAction,
+  setObservationsResultsAction,
+  setPlantingSiteObservationsResultsAction,
+} from './observationsSlice';
 
 /**
  * Fetch observation results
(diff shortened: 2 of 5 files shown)


constanzauanini (approved):
lgtm!

karthikbtf:
> lgtm!

Thanks!!
When you and @charlie-tf get a chance.. lmk your thoughts re: concerns described in PR description.
I will get this in for now.

karthikbtf:
> > lgtm!
> 
> Thanks!! When you and @charlie-tf get a chance.. lmk your thoughts re: concerns described in PR description. I will get this in for now.

Please ignore. After looking at my refactored routing PR, using the org-wide observations results redux makes more sense. I will revert this commit.

charlie-tf (commented):
I like separating this into its own redux, honestly. I don't think there's much downside to having two separate code paths to fetch—it's like two separate api calls. And the upside of fetching less data less often is worth it, I think.

charlie-tf:
> I like separating this into its own redux, honestly. I don't think there's much downside to having two separate code paths to fetch—it's like two separate api calls. And the upside of fetching less data less often is worth it, I think.

Oh, just saw your comment about reverting. What's the rationale?

karthikbtf:
> I like separating this into its own redux, honestly. I don't think there's much downside to having two separate code paths to fetch—it's like two separate api calls. And the upside of fetching less data less often is worth it, I think.

Thanks, yes - that was my first thinking.

I reverted it because I was looking at it bottom up.. but implementing it top-down (like in the observations view). 
The planting sites list view is similar to the root observations view. 
Selecting individual planting sites from the list view is similar to selecting the planting site in the observations drop down. With that analogy, I figured we can just use the org-wide fetch.. when a user clicks on a single planting site, the selector would fetch that data from org-wide state. Only difference in the observations view is the 'All' planting sites option (which warrants usage of an org-wide fetch).. we don't have that requirement in the planting sites list view.

Happy to re-revert if we prefer. Both approaches give us the data we want.

karthikbtf:
> > I like separating this into its own redux, honestly. I don't think there's much downside to having two separate code paths to fetch—it's like two separate api calls. And the upside of fetching less data less often is worth it, I think.
> 
> Thanks, yes - that was my first thinking.
> 
> I reverted it because I was looking at it bottom up.. but implementing it top-down (like in the observations view). The planting sites list view is similar to the root observations view. Selecting individual planting sites from the list view is similar to selecting the planting site in the observations drop down. With that analogy, I figured we can just use the org-wide fetch.. when a user clicks on a single planting site, the selector would fetch that data from org-wide state. Only difference in the observations view is the 'All' planting sites option (which warrants usage of an org-wide fetch).. we don't have that requirement in the planting sites list view.
> 
> Happy to re-revert if we prefer. Both approaches give us the data we want.

I will re-revert this. We can use the site specific API.

charlie-tf:
> > > I like separating this into its own redux, honestly. I don't think there's much downside to having two separate code paths to fetch—it's like two separate api calls. And the upside of fetching less data less often is worth it, I think.
> > 
> > 
> > Thanks, yes - that was my first thinking.
> > I reverted it because I was looking at it bottom up.. but implementing it top-down (like in the observations view). The planting sites list view is similar to the root observations view. Selecting individual planting sites from the list view is similar to selecting the planting site in the observations drop down. With that analogy, I figured we can just use the org-wide fetch.. when a user clicks on a single planting site, the selector would fetch that data from org-wide state. Only difference in the observations view is the 'All' planting sites option (which warrants usage of an org-wide fetch).. we don't have that requirement in the planting sites list view.
> > Happy to re-revert if we prefer. Both approaches give us the data we want.
> 
> I will re-revert this. We can use the site specific API.

Oh! I wasn't saying you need to do that. I'm okay with this reverted as you said. Just wondered what your thinking was, and thanks for explaining it

charlie-tf:
I'm okay with this either way, just let me know what you finally decide 😅

karthikbtf:
> > > > I like separating this into its own redux, honestly. I don't think there's much downside to having two separate code paths to fetch—it's like two separate api calls. And the upside of fetching less data less often is worth it, I think.
> > > 
> > > 
> > > Thanks, yes - that was my first thinking.
> > > I reverted it because I was looking at it bottom up.. but implementing it top-down (like in the observations view). The planting sites list view is similar to the root observations view. Selecting individual planting sites from the list view is similar to selecting the planting site in the observations drop down. With that analogy, I figured we can just use the org-wide fetch.. when a user clicks on a single planting site, the selector would fetch that data from org-wide state. Only difference in the observations view is the 'All' planting sites option (which warrants usage of an org-wide fetch).. we don't have that requirement in the planting sites list view.
> > > Happy to re-revert if we prefer. Both approaches give us the data we want.
> > 
> > 
> > I will re-revert this. We can use the site specific API.
> 
> Oh! I wasn't saying you need to do that. I'm okay with this reverted as you said. Just wondered what your thinking was, and thanks for explaining it

Oops.. I should have clarified. We both had the same hunch re: using site speicific API - that's confirmation enough to use this approach, hence the re-revert :) It's a learning experience for sure, we can always iterate later.

PR #1526, opened 2023-07-07
SW-3780 Hectares planted card

constanzauanini:
Screenshot 2023-07-07 at 14 34 33

### DIFF
--- a/src/components/PlantsV2/components/HectaresPlantedCard.tsx
+++ b/src/components/PlantsV2/components/HectaresPlantedCard.tsx
@@ -1,0 +1,93 @@
+import OverviewItemCard from 'src/components/common/OverviewItemCard';
+import { Box, LinearProgress, Typography, useTheme } from '@mui/material';
+import strings from 'src/strings';
+import React, { useMemo } from 'react';
+import { useAppSelector } from 'src/redux/store';
+import { selectPlantingSite } from 'src/redux/features/tracking/trackingSelectors';
+import { ObservationResults } from 'src/types/Observations';
+import { getShortDate } from 'src/utils/dateFormatter';
+import { useLocalization } from 'src/providers';
+
+type HectaresPlantedCardProps = {
+  plantingSiteId: number;
+  observation?: ObservationResults;
+};
+
+export default function HectaresPlantedCard({ plantingSiteId, observation }: HectaresPlantedCardProps): JSX.Element {
+  const theme = useTheme();
+  const plantingSite = useAppSelector((state) => selectPlantingSite(state, plantingSiteId));
+  const locale = useLocalization();
+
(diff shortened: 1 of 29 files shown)


src/components/PlantsV2/components/HectaresPlantedCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+            {observation?.completedTime
+              ? strings.formatString(
+                  strings.HECTARES_PLANTED_CARD_TITLE,
+                  getShortDate(observation.completedTime, locale.activeLocale)
+                )
+              : ''}
+          </Typography>
+          <Box display={'flex'} alignItems='baseline'>
+            <Typography fontSize='84px' fontWeight={600} lineHeight={1} marginBottom={theme.spacing(3)}>
+              {totalPlantedArea}
```
karthikbtf, line 46: may need to localize this, create a ticket to follow up later? "20,000" instead of "20000" (and "20.000" for esp, etc.).
Unless the BE returns localized values (my understanding is that search results alone are localized).
There is a number formatter under utils/useNumber.ts
I suppose my suggestion applies to all charts.. cc @charlie-tf 
A one-time refactor later may work.

src/components/PlantsV2/components/HectaresPlantedCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+            variant='determinate'
+            value={totalPlantedArea}
+            valueBuffer={totalArea}
+            sx={{
+              height: '32px',
+              borderRadius: '4px',
+              backgroundColor: theme.palette.TwClrBaseGray100,
+              '& span': { backgroundColor: theme.palette.TwClrBgBrand },
+            }}
+          />
```
karthikbtf, line 71: are these styles used in other places? If so, make this a component

karthikbtf (approved):
Looks good, a few suggestions.

src/components/PlantsV2/components/HectaresPlantedCard.tsx
charlie-tf: Thanks for the heads up!

charlie-tf:
Just pulled main and am seeing
image

Maybe need to add a `Math.round()` somewhere?

charlie-tf:
I just realized that you aren't really able to test this without manually going into the database and setting some subzones as complete. Which I did locally. Sorry, that will be added in the Observation Settings section we're working on next.

constanzauanini:
> Just pulled main and am seeing image
> 
> Maybe need to add a `Math.round()` somewhere?

Oh yes, it was alway 0% for me and didn't test it well, I will open a pr with this fix

PR #1528, opened 2023-07-07
SW-3777 SW-3786 Planting Density by Zone Card

charlie-tf:
## Without Observations (no actuals shown)
image

## With Observations (zones for which there is no actual planting density—which are not marked as completed planting—do not appear)
image


karthikbtf (approved):
Looks great!

charlie-tf:
Oops, just noticed that I have to update the card title when there's an observation. Patch coming....

karthikbtf (approved):

PR #1530, opened 2023-07-07
SW-3868 Handle locales with country codes

sgrimm:
To support variation of number formatting rules in different Spanish-speaking
countries, we're going to keep track of users' countries. If the user has selected
a country, their locale will include a country code as well as a language.

For the most part, locales with country codes work transparently because they're
well supported by browsers. However, our application code currently expects the
locale to only include a language. If the server sends back a locale like `es-MX`,
the client doesn't know which item in the locale selector it should use.

We could just strip off the country code, but in the future, we'll likely need to
support languages where the country code is more significant, e.g., Chinese.
Instead, update the client to pick the longest match from its list of supported
locales.

There will be a separate field in the user API payloads for country code; add it
to the type definitions and make the client code pass it back to the server when
updating user data. A later change will add UI to let users see and edit their
countries, but for now, this lets us test locale selection by manually setting
the country in the server's database.

### DIFF
--- a/src/api/types/generated-schema.ts
+++ b/src/api/types/generated-schema.ts
@@ -2952,6 +2952,7 @@ export interface components {
       preferences: { [key: string]: unknown };
     };
     UpdateUserRequestPayload: {
+      countryCode?: string;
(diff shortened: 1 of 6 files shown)


karthikbtf (approved):
Looks good to me 👍

sgrimm:
I'm going to hold off merging this for now since I'm still wrapping up the server side.

PR #1533, opened 2023-07-10
Adding math round to hectares planted card

constanzauanini:


charlie-tf (approved):
This oughta do. Thanks Constanza!

PR #1535, opened 2023-07-10
Upgrade to latest Tailscale GitHub Action

sgrimm:
Tailscale now supports OAuth for client authentication; use it instead
of API-key-based authentication.


sgrimm:
https://github.com/terraware/terraware-web/actions/runs/5510159567/jobs/10043942764 is a build where I moved the Tailscale step near the top of the workflow to verify that it works.

karthikbtf (approved):
👍

PR #1537, opened 2023-07-10
Refactor planting-sites fetch to use redux + initial map/list view in planting sites details page

karthikbtf:
- was working on the tv2 updates for planting sites and had a need to fetch planting sites list
- realized we are already overfetching (in the App, in the Observations page, etc.)
- update the App to dispatch a redux request
- all components now rely on the dispatched request in the App to select planting sites
- also, when we update (edit/create) planting sites, the app reloads planting sites
- simplifies code
- added initial map/list view for planting site details page

Terraware App 🔊 2023-07-10 13-27-02

Terraware App 🔊 2023-07-10 13-26-44

### DIFF
--- a/src/components/PlantingSites/index.tsx
+++ b/src/components/PlantingSites/index.tsx
@@ -1,8 +1,11 @@
-import { useState } from 'react';
-import { CircularProgress } from '@mui/material';
+import { useEffect } from 'react';
+import { useParams } from 'react-router-dom';
 import { Route, Switch } from 'react-router-dom';
 import { APP_PATHS } from 'src/constants';
 import isEnabled from 'src/features';
+import { useOrganization } from 'src/providers';
+import { useAppDispatch } from 'src/redux/store';
+import { requestPlantingSiteObservationsResults } from 'src/redux/features/observations/observationsThunks';
 import PlantingSiteCreate from './PlantingSiteCreate';
 import PlantingSitesList from './PlantingSitesList';
 import PlantingSiteView from './PlantingSiteView';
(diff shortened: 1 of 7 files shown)


src/components/PlantingSites/index.tsx
```diff
@@ -16,27 +19,34 @@ export type PlantingSitesProps = {
...
-
-  // show spinner while initializing data
-  if (initializingData) {
-    return <CircularProgress sx={{ margin: 'auto' }} />;
-  }
-
-  return <PlantingSitesWrapper {...props} />;
+export default function PlantingSites({ reloadTracking }: PlantingSitesProps): JSX.Element {
+  return (
+    <Switch>
```
charlie-tf, line 24: Child Switches appear to route on the full path? That is, does the app ultimately see a flattened list of Routes, or does the nesting affect how things are routed?

charlie-tf: If the `PlantingSitesWrapper`, for instance, is not rendered, can the app see its routes?

charlie-tf (commented):
Looks good! Thank you for this refactor. Have a couple questions about routing, as the nesting of `Switch`s confuses me a little bit.

src/components/PlantingSites/index.tsx
karthikbtf: That is correct, the routing is on the full path.
The root App.tsx has a route `/planting-sites` (non-exact match, just needs to start with that) - which renders the PlantingSitesWrapper. Once that is rendered, the finer route handling happens.
Not sure if I answered the question?

charlie-tf: I think I understand. So a Switch isn't like a conditional renderer. It renders down its component tree to make sure all the routes are rendered?

karthikbtf: The routing happens at run-time, there is no dictionary of routes to components.

karthikbtf: Switch happens at run-time

charlie-tf (approved):

src/components/PlantingSites/index.tsx
charlie-tf: Thanks Karthik

karthikbtf: I think I may have misunderstood the question completely - let me know if that's the case.

charlie-tf: Oh it's okay. I've worked out what was confusing me.

karthikbtf: No worries. I just matched a lot of your questions to the [docs](https://reactrouter.com/en/main/start/concepts). Those were good questions.
Re: routing with full paths versus relative-links, react has support for both -  we need to look for paths that don't start with `/` for relative links.. we don't do that here.. we could certainly do that though. Our route configs match the entire path starting with leading `/`.
Re: rendering routes (creating a tree, etc.) to match the paths, that happens during the render of the component. When App.tsx is rendered, the route config for App.tsx alone is processed, converted to a tree and matched.
When PlantingSitesWrapper is rendered, the same thing happens for its specific route config.
That is my understanding at least :)

PR #1540, opened 2023-07-11
SW-3874 Localize all numeric displays in charts

constanzauanini:


src/components/PlantsV2/components/TotalReportedPlantsCard.tsx
```diff
@@ -50,7 +55,7 @@ export default function TotalReportedPlantsCard({ plantingSiteId }: TotalReporte
           </Typography>
           <Box display='flex' alignItems='flex-end' flexWrap='wrap' marginBottom={theme.spacing(3)}>
             <Typography fontSize={numberFontSize(totalPlants)} fontWeight={600} lineHeight={1}>
-              {totalPlants}
+              {totalPlants ? numericFormater.format(totalPlants) : ''}
```
charlie-tf, line 58: I don't think `totalPlants` can be undefined here, and this will display `''` for zero, which is not what we want

charlie-tf: I think this should be okay as `{numericFormater.format(totalPlants)}`

charlie-tf (changes requested):
Thanks for doing this! Just on small change needed

src/components/PlantsV2/index.tsx
```diff
@@ -39,6 +40,12 @@ export default function PlantsDashboardV2(): JSX.Element {
   const [plantsDashboardPreferences, setPlantsDashboardPreferences] = useState<Record<string, unknown>>();
   const locale = useLocalization();
   const history = useHistory();
+  const numberFormatter = useNumberFormatter();
+  const user = useUser().user;
+  const numericFormatter = useMemo(
+    () => numberFormatter(user?.locale, user?.countryCode),
+    [user?.locale, numberFormatter, user?.countryCode]
+  );
```
karthikbtf, line 49: Suggestion:
Move this into a standalone `FormattedNumber` component which takes in the value to render:
```
function FormattedNumber({ value }: Props) {
  const user = useUser().user;
  const numericFormatter = useMemo( ....);

  return numericFormatter.format(value);
}

// in the clients just call
<Typography...>
  <FormatterNumber value={value} />
</Typography>
```

simplifies the code, don't need to pass all those props around
i added links to reference code in the jira earlier in case it helps

src/utils/useNumber.ts
```diff
@@ -6,8 +6,11 @@ const getLocaleToUse = (locale?: string) => (locale === 'gx' ? 'fr' : locale ||
  * formatter
  */
 export const useNumberFormatter = (): any => {
-  const formatter = (locale?: string): any => {
-    const localeToUse = getLocaleToUse(locale);
+  const formatter = (locale?: string, countryCode?: string): any => {
+    let localeToUse = getLocaleToUse(locale);
```
karthikbtf, line 10: i think the country code is in the user's locale
the `getlocaletouse` strips off the country code

karthikbtf (commented):
Thanks for doing this sweeping effort.
some suggestions

src/utils/useNumber.ts
karthikbtf: ugh, nvm, i am wrong here.. apologies.

src/utils/useNumber.ts
```diff
@@ -6,8 +6,11 @@ const getLocaleToUse = (locale?: string) => (locale === 'gx' ? 'fr' : locale ||
...
+    if (countryCode) {
+      localeToUse = `${localeToUse}-${countryCode}`;
```
karthikbtf, line 12: can you use this function instead? https://github.com/terraware/terraware-web/blob/main/src/strings/locales.ts#L39
(i think it uses the locales we support, and also fixes any issues where the language-country pair is incorrect or unsupported)

charlie-tf (approved):
👍

karthikbtf (approved):
Thanks!! This looks much cleaner.

PR #1541, opened 2023-07-11
Planting site details table with observations info

karthikbtf:
- add selector for data (aggregates monitoring plots across observations results for planting site)
- aggregated data will be used in other views and potentially map as well
- create table

Terraware App 2023-07-11 13-06-33


karthikbtf:
> Now I get where the data come from. Thanks for explaining each step, that really helped. Looks good to me!

Awesome, thanks @constanzauanini 🙏

src/utils/dateUtils.ts
```diff
@@ -0,0 +1,10 @@
+export const isAfter = (dateString1: string | undefined, dateString2: string | undefined): boolean => {
+  if (dateString1 === undefined || dateString2 === undefined) {
+    return false;
```
charlie-tf, line 3: I wonder if returning false here is correct. It would imply that dateString1 is on or before dateString2, and I wonder whether null or undefined is the correct return value.. Of course, where you used it it won't make a difference, but just worrying over the semantics if this is reused elsewhere.

charlie-tf (commented):
Selector and components make sense. Thanks for the thorough comments. One question about the isAfter util function

src/utils/dateUtils.ts
karthikbtf: Thanks for the feedback. The current behavior should be ok as long as the contract is documented. I can at least do that. There is little value in returning undefined etc. Alternatively I could make the input required.

karthikbtf: I could look at the various combinations of undefined, that might help. Let me do that.

charlie-tf: Nice fix!

charlie-tf (approved):

PR #1196, opened 2023-02-14
SW-2949 Use (raw) fields for inventory quantity based logic

karthikbtf:
- remove parsing logic
- cleaned up a whole bunch of semantics help in components, moved then to services
- updated plants dashboard to use raw fields

PR #1542, opened 2023-07-11
SW-3879 Planting Density Card and Chart Update

charlie-tf:
Added a planting density card, which shows the total observed and target planting density for the planting site. It also get the plantings since the most recent observation and gives an estimate of the updated planting density.

In addition, there were some issues with updating chart values, in particular, when updating the data for a chart it would try to create a new chart on the existing canvas, which is not fully released by the old chart. Instead, keep the same chart and update its data.

image

image


src/components/PlantsV2/components/PlantingSiteDensityCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const theme = useTheme();
+  const plantingSite = useAppSelector((state) => selectPlantingSite(state, plantingSiteId));
+  const plantingsSinceObservation = useAppSelector((state) =>
+    selectPlantingsDateRange(state, observation?.completedDate ? [observation?.completedDate] : [])
+  );
+  const locale = useLocalization();
+
+  const weightedSum =
+    plantingSite?.plantingZones?.map((z) => z.targetPlantingDensity * z.areaHa)?.reduce((a, b) => a + b) ?? 0;
+  const targetPlantingDensity = plantingSite?.areaHa ? weightedSum / plantingSite.areaHa : 0;
```
karthikbtf, line 31: do we need to worry about divide-by-zero? (areaHa should never by 0 but it is a number type and theoretically can)

src/components/PlantsV2/components/PlantingSiteDensityCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const plantingDensity = observation?.plantingDensity ?? 0;
+  const percentageOfTargetDensity = (100 * plantingDensity) / targetPlantingDensity;
+  const numPlantedSinceObs = plantingsSinceObservation.reduce((prev, curr) => +curr.numPlants + prev, 0);
+  const newDensityEstimate = plantingSite?.areaHa
+    ? plantingDensity + numPlantedSinceObs / plantingSite.areaHa
+    : plantingDensity;
```
karthikbtf, line 37: is there an API that returns this info?

src/components/common/Chart/Chart.tsx
```diff
@@ -96,78 +96,90 @@ function ChartContent(props: ChartContentProps): JSX.Element {
...
+          chart?.destroy();
         };
       }
     };
     createChart();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [chartData, locale]);
+  }, [locale]);
+
+  useEffect(() => {
```
karthikbtf, line 150: nice debug/resolution, I don't know much here will let @constanzauanini take a look as well

src/redux/features/Plantings/plantingsThunks.ts
```diff
@@ -0,0 +1,26 @@
+import { Dispatch } from 'redux';
+import { PlantingsService } from 'src/services';
+import { RootState } from 'src/redux/rootReducer';
+import { setPlantingsAction } from './plantingsSlice';
+
+export const requestPlantings = (organizationId: number, plantingSiteId: number) => {
+  return async (dispatch: Dispatch, _getState: () => RootState) => {
+    try {
+      const response = await PlantingsService.listPlantings(organizationId, [
+        {
```
karthikbtf, line 10: thoughts on just passing the org id and planting site id into the listPlantings api? have the service populate the search criteria.. unless we will be using the search criteria and sort order in some upcoming work

karthikbtf (commented):
Looks cool, a few thoughts.

src/components/PlantsV2/components/PlantingSiteDensityCard.tsx
charlie-tf: if `plantingSite?.areaHa` is zero then it should take the else branch, due to 0 being falsey 😄

src/components/PlantsV2/components/PlantingSiteDensityCard.tsx
charlie-tf: No, I don't believe so

src/utils/useNumber.ts
```diff
@@ -16,3 +16,33 @@ export const useNumberFormatter = (): any => {
 
   return useMemo(() => formatter, []);
 };
+
+/**
+ * Parser
+ * ref: https://observablehq.com/@mbostock/localized-number-parsing
+ */
+export const useNumberParser = (locale?: string): any => {
```
charlie-tf, line 24: I added this because when I did a really big outplanting, 5000, to test something, the way we were usually converting the value using `+` was failing on `5,000`

src/components/PlantsV2/components/PlantingSiteDensityCard.tsx
karthikbtf: ugh ofc.. 🤦

src/redux/features/Plantings/plantingsThunks.ts
charlie-tf: I'll look into this, thanks

src/components/PlantsV2/components/PlantingProgressPerZoneCard.tsx
```diff
@@ -26,9 +30,9 @@ export default function PlantingProgressPerZoneCard({ plantingSiteId }: Planting
       plantingSite.plantingZones?.forEach((zone) => {
         let completedPlantingArea = 0;
         zone.plantingSubzones.forEach((sz) => {
-          completedPlantingArea += sz.plantingCompleted ? +sz.areaHa : 0;
+          completedPlantingArea += sz.plantingCompleted ? parse(sz.areaHa) : 0;
```
karthikbtf, line 33: you can avoid the parsing by fetching `(raw)` values in the search.. might be simpler
see example https://github.com/terraware/terraware-web/blob/9a7bd29b56a6ae2d1ab7f4dbd04e4546e20fa4fe/src/services/NurseryBatchService.ts#L23

src/utils/useNumber.ts
charlie-tf: Oh, just seeing that I should have employed `getLocaleToUse` here. I'll make that update

karthikbtf: yes, we used to have this parser approach for nursery withdrawals.. we removed it after Steve added `(raw)` fields in search responses.. returns non-localized values.. we use them in most withdrawals math

charlie-tf: I could have put `useLocalization` inside here instead of forcing it to be passed in. I'll make that change

karthikbtf: see PR when we got rid of parsing https://github.com/terraware/terraware-web/pull/1196/files

charlie-tf: Oh heck, I didn't know about that... I'll look into using `raw` instead. Thanks, Karthik

karthikbtf: no worries, i can relate to this issue you ran into.. a lot 🤣

src/components/common/Chart/Chart.tsx
constanzauanini: I remember destroying the chart when data changed because if not there was an error that more than 1 chart was referencing the same canvas, but I think I was not able to only update it without creating a new one, so this new approach looks correct to me! Not destroying the chart, but just refreshing its data! Thanks @charlie-tf !

src/components/common/Chart/ProgressChart.tsx
```diff
@@ -7,11 +7,13 @@ type ProgressChartProps = {
...
   const theme = useTheme();
 
+  const percentage = (100 * value) / target;
+
   return (
     <LinearProgress
       variant='determinate'
-      value={value}
-      valueBuffer={target}
+      value={percentage}
```
constanzauanini, line 15: I'm using this chart for HectaresPlantedCard. Should it be updated?

charlie-tf: The change should be transparent to the consumer.

constanzauanini: Thank you!

src/services/TrackingService.ts
```diff
@@ -175,7 +175,7 @@ const getTotalPlantsInZones = async (organizationId: number, siteId: number): Pr
       'plantingSubzones.fullName',
       'plantingSubzones.populations.species_scientificName',
       'plantingSubzones.populations.species_organization_id',
-      'plantingSubzones.populations.totalPlants',
+      'plantingSubzones.populations.totalPlants(raw)',
```
karthikbtf, line 178: there might be some v1 code depending on this

karthikbtf (commented):
Thanks for the `(raw)` updates... one suggestion to also look at v1 dependencies on totalPlants.
Looks great otherwise.

src/services/TrackingService.ts
charlie-tf: oh, i'll check into it. Thanks

charlie-tf: I found an instance of this and fixed it. Thanks for the callout

karthikbtf: Awesome, thanks @charlie-tf 👍

karthikbtf (approved):
🎉

PR #1545, opened 2023-07-11
SW-3828 Use import format for species CSV export

sgrimm:
Users want to share species lists between organizations by exporting and importing
them, but currently, the export format isn't the same as the import format so they
have to manually edit the files first.

Change the species list UI so the columns appear in the same order as in the
import template, and change the export code so it omits a couple of ID columns
that are required by the UI code but aren't needed for import.


karthikbtf (approved):
Looks simple enough 👍

PR #1551, opened 2023-07-13
SW-3876 Observed Number of Species Card

charlie-tf:
image

### DIFF
--- a/src/components/PlantsV2/components/ObservedNumberOfSpeciesCard.tsx
+++ b/src/components/PlantsV2/components/ObservedNumberOfSpeciesCard.tsx
@@ -60,7 +60,7 @@ export default function ObservedNumberOfSpeciesCard({ observation }: ObservedNum
             </Typography>
             &nbsp;
             <Typography fontSize='24px' fontWeight={600}>
-              {strings.PLANTS}
+              {strings.SPECIES}
             </Typography>
           </Box>
           <Typography fontSize='12px' fontWeight={400}>
--- a/src/components/PlantsV2/components/ObservedNumberOfSpeciesCard.tsx
+++ b/src/components/PlantsV2/components/ObservedNumberOfSpeciesCard.tsx
@@ -1,0 +1,78 @@
+import React, { useEffect, useState } from 'react';
+import OverviewItemCard from 'src/components/common/OverviewItemCard';
+import strings from 'src/strings';
+import { Box, Typography, useTheme } from '@mui/material';
+import { useAppSelector } from 'src/redux/store';
+import { selectSitePopulation } from 'src/redux/features/tracking/sitePopulationSelector';
+import FormattedNumber from 'src/components/common/FormattedNumber';
+import { ObservationResults } from 'src/types/Observations';
+import { getShortDate } from 'src/utils/dateFormatter';
+import { useLocalization } from 'src/providers';
+
+type ObservedNumberOfSpeciesCardProps = {
+  observation: ObservationResults;
+};
+
+export default function ObservedNumberOfSpeciesCard({ observation }: ObservedNumberOfSpeciesCardProps): JSX.Element {
+  const theme = useTheme();
+  const locale = useLocalization();
+  const populationSelector = useAppSelector((state) => selectSitePopulation(state));
+  const [numReportedSpecies, setNumReportedSpecies] = useState(0);
+  const [numObservedSpecies, setNumObservedSpecies] = useState(0);
+  useEffect(() => {
+    const results = new Set<number>();
+    const species = observation.species.map((s) => s.speciesId).filter((sid) => sid !== undefined);
+    species.forEach((s) => results.add(s!));
+
+    setNumObservedSpecies(results.size);
+  }, [observation]);
+
+  useEffect(() => {
+    if (populationSelector) {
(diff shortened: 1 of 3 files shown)


src/components/PlantsV2/components/ObservedNumberOfSpeciesCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const theme = useTheme();
+  const locale = useLocalization();
+  const populationSelector = useAppSelector((state) => selectSitePopulation(state));
+  const [numReportedSpecies, setNumReportedSpecies] = useState(0);
+  const [numObservedSpecies, setNumObservedSpecies] = useState(0);
+  useEffect(() => {
+    const results = new Set<number>();
+    const species = observation.species.map((s) => s.speciesId).filter((sid) => sid !== undefined);
+    species.forEach((s) => results.add(s!));
+
```
karthikbtf, line 26: The observation species list is already de-duped.
this could be
```
const results = observation.species.filter(s => !!s.id).length;
```
If we want to include uncertain species as well just use the observation.species.length count. There will be species without ids but with a name.

karthikbtf (approved):
Looks good otherwise 👍 
@charlie-tf sorta orthogonal to this.. we have been doing a lot of biz logic within components recently.. we should revisit this perhaps and move them to selectors later? Thoughts?

src/components/PlantsV2/components/ObservedNumberOfSpeciesCard.tsx
charlie-tf: Okay, thanks for that feedback. It seems like including uncertain species might be good? I'll put that in and we'll see how that goes.

charlie-tf:
> Looks good otherwise 👍 @charlie-tf sorta orthogonal to this.. we have been doing a lot of biz logic within components recently.. we should revisit this perhaps and move them to selectors later? Thoughts?

I was thinking that putting it in a component was alright unless we planned to reuse the behavior. But I guess that does sorta go against the thinking of using selectors to slice up the data. I think You're probably right that moving to selectors is the right idea here 👍

karthikbtf:
> > Looks good otherwise 👍 @charlie-tf sorta orthogonal to this.. we have been doing a lot of biz logic within components recently.. we should revisit this perhaps and move them to selectors later? Thoughts?
> 
> I was thinking that putting it in a component was alright unless we planned to reuse the behavior. But I guess that does sorta go against the thinking of using selectors to slice up the data. I think You're probably right that moving to selectors is the right idea here 👍

Got it. My thought was mostly from a theoretical perspective.
We could maybe discuss this with Constanza. 
If moving it to selectors (or services) just adds more overhead than enable our workflows, might as well keep things as they are.

charlie-tf:
> > > Looks good otherwise 👍 @charlie-tf sorta orthogonal to this.. we have been doing a lot of biz logic within components recently.. we should revisit this perhaps and move them to selectors later? Thoughts?
> > 
> > 
> > I was thinking that putting it in a component was alright unless we planned to reuse the behavior. But I guess that does sorta go against the thinking of using selectors to slice up the data. I think You're probably right that moving to selectors is the right idea here 👍
> 
> Got it. My thought was mostly from a theoretical perspective. We could maybe discuss this with Constanza. If moving it to selectors (or services) just adds more overhead than enable our workflows, might as well keep things as they are.

I will go ahead and create a ticket to look at this, at least. There's one spot that I do want to convert some logic to a selector and there may be more.

karthikbtf (approved):
Thanks for the changes!

constanzauanini:
> > > Looks good otherwise 👍 @charlie-tf sorta orthogonal to this.. we have been doing a lot of biz logic within components recently.. we should revisit this perhaps and move them to selectors later? Thoughts?
> > 
> > 
> > I was thinking that putting it in a component was alright unless we planned to reuse the behavior. But I guess that does sorta go against the thinking of using selectors to slice up the data. I think You're probably right that moving to selectors is the right idea here 👍
> 
> Got it. My thought was mostly from a theoretical perspective. We could maybe discuss this with Constanza. If moving it to selectors (or services) just adds more overhead than enable our workflows, might as well keep things as they are.

Yes, I like trying to move all to selectors to avoid having a mix of the two ways

charlie-tf:
> > > > Looks good otherwise 👍 @charlie-tf sorta orthogonal to this.. we have been doing a lot of biz logic within components recently.. we should revisit this perhaps and move them to selectors later? Thoughts?
> > > 
> > > 
> > > I was thinking that putting it in a component was alright unless we planned to reuse the behavior. But I guess that does sorta go against the thinking of using selectors to slice up the data. I think You're probably right that moving to selectors is the right idea here 👍
> > 
> > 
> > Got it. My thought was mostly from a theoretical perspective. We could maybe discuss this with Constanza. If moving it to selectors (or services) just adds more overhead than enable our workflows, might as well keep things as they are.
> 
> Yes, I like trying to move all to selectors to avoid having a mix of the two ways

I don't think there's really a problem with having a mix. It's perfectly natural in many situations to do a one-off calculation within a component that doesn't justify pulling out into a selector. I think selectors especially make sense where a complex calculation is to be reused between components.

PR #1555, opened 2023-07-13
SW-3914 Fix Highlight for Map

charlie-tf:
Previously, the map could get into a state where the highlight from
selection of a map entity was stuck. This was due to the select state
being cleared when the map data changed and a map popup was still open.
In the code for updating callbacks on map data change, only clear the
select map feature if the popup is not open. This allows the popup
close handler to properly reset this feature state.


karthikbtf (approved):
Thanks for the quick followup!!
Code makes sense and works in vercel 👍

PR #1557, opened 2023-07-13
SW-3921 Feedback changes, add missing tooltips + fix tooltip style

karthikbtf:
- demo'd to Chudo, found some missing tooltip properties and style issues
- added missing tooltips
- give more width to tooltip based on content
- added back a field in service which is needed in the previous plants dashboard map tooltip (totalPlants as localized by server)

Terraware App 2023-07-14 07-00-27

Terraware App 2023-07-14 07-00-12

Terraware App 2023-07-14 06-59-52

Terraware App 2023-07-14 06-59-29

### DIFF
--- a/src/services/TrackingService.ts
+++ b/src/services/TrackingService.ts
@@ -176,6 +176,7 @@ const getTotalPlantsInZones = async (organizationId: number, siteId: number): Pr
       'plantingSubzones.populations.species_scientificName',
       'plantingSubzones.populations.species_organization_id',
       'plantingSubzones.populations.totalPlants(raw)',
+      'plantingSubzones.populations.totalPlants',
       'id',
       'name',
     ],
--- a/src/components/Map/MapRenderUtils.tsx
+++ b/src/components/Map/MapRenderUtils.tsx
@@ -72,17 +72,17 @@ export function MapTooltip({ title, properties }: MapTooltipProps): JSX.Element 
     fontSize: '16px',
     color: theme.palette.TwClrBaseBlack as string,
     whiteSpace: 'pre',
+    textAlign: 'left',
   };
 
   const keyStyle = {
     ...textStyle,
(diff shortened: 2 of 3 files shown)


src/services/TrackingService.ts
```diff
@@ -176,6 +176,7 @@ const getTotalPlantsInZones = async (organizationId: number, siteId: number): Pr
       'plantingSubzones.populations.species_scientificName',
       'plantingSubzones.populations.species_organization_id',
       'plantingSubzones.populations.totalPlants(raw)',
+      'plantingSubzones.populations.totalPlants',
```
karthikbtf, line 179: Added this back for the v1 plants dashboard map tooltip which renders totalPlants.

charlie-tf: Oops, I must have missed that one.

charlie-tf (approved):

chudo76 (approved):
Looks great. I think the only thing might be that the tooltip titles (tracking-v2-tw3, Jolishee, Alizes-AB, etc.) are left-aligned instead of centered for other map tooltips so we may want to do the same here.

karthikbtf:
> Looks great. I think the only thing might be that the tooltip titles (tracking-v2-tw3, Jolishee, Alizes-AB, etc.) are left-aligned instead of centered for other map tooltips so we may want to do the same here.

I should probably left alight the first column as well, right now the first column is right aligned and second columns is left aligned.

karthikbtf:
> > Looks great. I think the only thing might be that the tooltip titles (tracking-v2-tw3, Jolishee, Alizes-AB, etc.) are left-aligned instead of centered for other map tooltips so we may want to do the same here.
> 
> I should probably left alight the first column as well, right now the first column is right aligned and second columns is left aligned.

updated screenshots

PR #1560, opened 2023-07-14
SW-3682 Convert Latest Observation to a Selector and use in Cards

charlie-tf:
- Take the latest observation logic out of the dashboard index
- Put into a selector
- Select the latest observation in the card components

### DIFF
--- a/src/components/PlantsV2/index.tsx
+++ b/src/components/PlantsV2/index.tsx
@@ -16,7 +16,7 @@ import NumberOfSpeciesPlantedCard from './components/NumberOfSpeciesPlantedCard'
 import PlantingSiteProgressCard from './components/PlantingSiteProgressCard';
 import PlantingProgressPerZoneCard from './components/PlantingProgressPerZoneCard';
 import ZoneLevelDataMap from './components/ZoneLevelDataMap';
-import { searchObservations } from 'src/redux/features/observations/observationsSelectors';
+import { selectLatestObservation } from 'src/redux/features/observations/observationsSelectors';
 import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
 import TotalMortalityRateCard from './components/TotalMoratlityRateCard';
 import { selectSitePopulation } from 'src/redux/features/tracking/sitePopulationSelector';
@@ -37,7 +37,6 @@ import ObservedNumberOfSpeciesCard from 'src/components/PlantsV2/components/Obse
 export default function PlantsDashboardV2(): JSX.Element {
   const org = useOrganization();
   const { isMobile } = useDeviceInfo();
-  const defaultTimeZone = useDefaultTimeZone();
   const dispatch = useAppDispatch();
   const [selectedPlantingSiteId, setSelectedPlantingSiteId] = useState(-1);
   const [plantsDashboardPreferences, setPlantsDashboardPreferences] = useState<Record<string, unknown>>();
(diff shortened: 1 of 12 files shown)


src/components/PlantsV2/index.tsx
```diff
@@ -50,26 +49,11 @@ export default function PlantsDashboardV2(): JSX.Element {
     [setPlantsDashboardPreferences]
   );
 
-  const observationsResults = useAppSelector((state) =>
-    searchObservations(state, selectedPlantingSiteId, defaultTimeZone.get().id, '', [])
+  const defaultTimeZone = useDefaultTimeZone();
+  const latestObservation = useAppSelector((state) =>
+    selectLatestObservation(state, selectedPlantingSiteId, defaultTimeZone.get().id)
```
karthikbtf, line 54: nice!!

src/components/PlantsV2/index.tsx
```diff
@@ -94,16 +78,16 @@ export default function PlantsDashboardV2(): JSX.Element {
...
-        <HighestAndLowestMortalityRateZonesCard observation={latestObservation} />
+        <HighestAndLowestMortalityRateZonesCard plantingSiteId={selectedPlantingSiteId} />
       </Grid>
       <Grid item xs={isMobile ? 12 : 3}>
-        <HighestAndLowestMortalityRateSpeciesCard observation={latestObservation} />
+        <HighestAndLowestMortalityRateSpeciesCard plantingSiteId={selectedPlantingSiteId} />
       </Grid>
       <Grid item xs={isMobile ? 12 : 3}>
-        <LiveDeadPlantsPerSpeciesCard observation={latestObservation} />
+        <LiveDeadPlantsPerSpeciesCard plantingSiteId={selectedPlantingSiteId} />
```
karthikbtf, line 90: i'm on the fence here.. maybe these smaller widgets are ok receiving the data as props?

src/redux/features/observations/observationsSelectors.ts
```diff
@@ -100,3 +100,24 @@ export const selectPlantingSiteObservations = createCachedSelector(
...
+    searchObservations(state, plantingSiteId, defaultTimeZoneId, '', []),
+  (observationsResults: ObservationResults[] | undefined) => {
+    if (!observationsResults || observationsResults.length === 0) {
+      return undefined;
+    }
+    const result = observationsResults.reduce((prev, curr) => {
+      if (!prev.completedTime) {
+        return curr;
+      }
+      if (!curr.completedTime || prev.completedTime.localeCompare(curr.completedTime) > 0) {
```
karthikbtf, line 116: dateUtils isAfter :) ?

karthikbtf (commented):
Nice upgrade 👍 
One suggestion and one comment on approach.. would like to hear your thoughts...

src/redux/features/observations/observationsSelectors.ts
charlie-tf: Ah, I wrote this before that util and just copypasted. I'll upgrade :)

src/components/PlantsV2/index.tsx
charlie-tf: Hm, yeah it's kind of a tossup. I guess I liked the consistency of selecting the observation result data from state using a selector at the point of use. However, it does create a dependency between the index component and each of the cards, wherein the index is responsible for dispatching the appropriate thunks. Rather than the data flow being explicit in the props interface.

karthikbtf: ideally these smaller widgets should be 'story-able', props would make that easier me thinks.. like you said .. 6 of one, half a dozen of the other

charlie-tf: Okay, most of these cards have other data that they get with selectors internally. I think that sorta makes them all less than story-able without rewriting most of them. So given that, do you think consistency wins out here?

karthikbtf: Got it, yes if we are already using selectors within those components.. this is the right path 👍 
Thanks for sharing that extra context.

charlie-tf: Cool, thanks for the discussion on this. Gives me a better sense of when to use props vs selectors in the future :)

karthikbtf (approved):
Oops, I thought I had already approved this.
Looks great 🎉

PR #1565, opened 2023-07-17
SW-3919 MyAccount language and preferred units selectors should be full width

constanzauanini:
Screenshot 2023-07-17 at 12 03 06

### DIFF
--- a/src/components/LocaleSelector.tsx
+++ b/src/components/LocaleSelector.tsx
@@ -11,6 +11,7 @@ type LocaleSelectorProps = {
   transparent?: boolean;
   onChangeLocale?: (newValue: string) => void;
   localeSelected?: string;
+  fullWidth?: boolean;
 };
 
 const useStyles = makeStyles((theme: Theme) => ({
@@ -25,6 +26,7 @@ export default function LocaleSelector({
   transparent,
   onChangeLocale,
   localeSelected,
+  fullWidth,
 }: LocaleSelectorProps): JSX.Element {
   const { user, reloadUser } = useUser();
   const supportedLocales = useSupportedLocales();
@@ -74,6 +76,7 @@ export default function LocaleSelector({
                   onChange={onChangeLocale}
                   selectedValue={localeDetails.id}
                   options={localeItems}
+                  fullWidth={fullWidth}
                 />
               )
             )}
--- a/src/components/MyAccount/index.tsx
+++ b/src/components/MyAccount/index.tsx
@@ -532,44 +532,38 @@ const MyAccountContent = ({
                 onChange={(value) => onChange('emailNotificationsEnabled', value)}
               />
             </Grid>
-            <Grid item xs={12} />
-            {organizations && organizations.length > 0 ? (
-              <>
-                <Grid item xs={12}>
-                  <Typography fontSize='20px' fontWeight={600}>
-                    {strings.ORGANIZATIONS}
-                  </Typography>
-                </Grid>
-                <Grid item xs={12}>
-                  <div>
-                    <Grid container spacing={4}>
-                      <Grid item xs={12}>
-                        {organizations && (
-                          <Table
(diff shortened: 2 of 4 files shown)


karthikbtf (approved):
Looks great thanks!
For mobile the inputs use up table width. Unrelated to this PR for sure.
Moving the Table out of the top level grid container should fix it.. small change but sure we can do this separately if preferred.
Also, do you think we should get Chudo's feedback?

constanzauanini:
> Looks great thanks! For mobile the inputs use up table width. Unrelated to this PR for sure. Moving the Table out of the top level grid container should fix it.. small change but sure we can do this separately if preferred. Also, do you think we should get Chudo's feedback?

Oh, will fix that now. I think we can merge it when it's ready and if Chudo has any feedback we can follow up later

constanzauanini:
> Looks great thanks! For mobile the inputs use up table width. Unrelated to this PR for sure. Moving the Table out of the top level grid container should fix it.. small change but sure we can do this separately if preferred. Also, do you think we should get Chudo's feedback?

Locally it looks good, that's strange

karthikbtf:
> > Looks great thanks! For mobile the inputs use up table width. Unrelated to this PR for sure. Moving the Table out of the top level grid container should fix it.. small change but sure we can do this separately if preferred. Also, do you think we should get Chudo's feedback?
> 
> Locally it looks good, that's strange

Hmm.. yes for some reason broken in vercel. We can maybe just create a new ticket if you prefer.
Thanks so much for the changes. Yes we can follw up on any @chudo76 to give feedback later.

constanzauanini:
> > > Looks great thanks! For mobile the inputs use up table width. Unrelated to this PR for sure. Moving the Table out of the top level grid container should fix it.. small change but sure we can do this separately if preferred. Also, do you think we should get Chudo's feedback?
> > 
> > 
> > Locally it looks good, that's strange
> 
> Hmm.. yes for some reason broken in vercel. We can maybe just create a new ticket if you prefer. Thanks so much for the changes. Yes we can follow up on any @chudo76 to give feedback later.

But all tables work like that, you have to scroll to the right to see all the data

karthikbtf:
> > > > Looks great thanks! For mobile the inputs use up table width. Unrelated to this PR for sure. Moving the Table out of the top level grid container should fix it.. small change but sure we can do this separately if preferred. Also, do you think we should get Chudo's feedback?
> > > 
> > > 
> > > Locally it looks good, that's strange
> > 
> > 
> > Hmm.. yes for some reason broken in vercel. We can maybe just create a new ticket if you prefer. Thanks so much for the changes. Yes we can follow up on any @chudo76 to give feedback later.
> 
> But all tables work like that, you have to scroll to the right to see all the data

Yes, that is correct. I was referring to the input elements like language selector etc.
We can make that section independent of the table width with Grids. We do that in some other places.

karthikbtf:
> > > > > Looks great thanks! For mobile the inputs use up table width. Unrelated to this PR for sure. Moving the Table out of the top level grid container should fix it.. small change but sure we can do this separately if preferred. Also, do you think we should get Chudo's feedback?
> > > > 
> > > > 
> > > > Locally it looks good, that's strange
> > > 
> > > 
> > > Hmm.. yes for some reason broken in vercel. We can maybe just create a new ticket if you prefer. Thanks so much for the changes. Yes we can follow up on any @chudo76 to give feedback later.
> > 
> > 
> > But all tables work like that, you have to scroll to the right to see all the data
> 
> Yes, that is correct. I was referring to the input elements like language selector etc. We can make that section independent of the table width with Grids. We do that in some other places.

The problem is unrelated to this PR :) I will create a ticket for us later.

chudo76:
Looks a bit odd with a three-column row on top of a four-column row but I don't have a good alternative so I think we're good for now. The form fields look much better at full-width, though. Thanks!

PR #1567, opened 2023-07-17
SW-3894 Add title/description and map/list view components in Planting Progress

karthikbtf:
- added list/map view
- added search props and filter for planting complete
- some style fixes

Terraware App 2023-07-17 10-42-33

### DIFF
--- a/src/components/NurseryWithdrawals/NurseryPlantingsAndWithdrawals.tsx
+++ b/src/components/NurseryWithdrawals/NurseryPlantingsAndWithdrawals.tsx
@@ -8,7 +8,6 @@ import strings from 'src/strings';
 import PageSnackbar from 'src/components/PageSnackbar';
 import PageHeaderWrapper from 'src/components/common/PageHeaderWrapper';
 import TfMain from 'src/components/common/TfMain';
-import Card from 'src/components/common/Card';
 import PlantingProgress from './PlantingProgressTabContent';
 import NurseryWithdrawals from './NurseryWithdrawalsTabContent';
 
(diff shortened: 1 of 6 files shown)


src/components/NurseryWithdrawals/NurseryPlantingsAndWithdrawals.tsx
```diff
@@ -37,14 +36,12 @@ export default function NurseryPlantingsAndWithdrawals(): JSX.Element {
             <PageSnackbar />
           </Grid>
           <Box ref={contentRef} display='flex' flexDirection='column' flexGrow={1}>
-            <Card flushMobile style={{ minWidth: 'fit-content' }}>
```
karthikbtf, line 40: this was not needed as per the design

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
```diff
@@ -0,0 +1,3 @@
+export default function PlantingProgressList(): JSX.Element {
+  return <p>'I am a map'</p>;
```
karthikbtf, line 2: @charlie-tf to replace this with Map code

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -0,0 +1,3 @@
+export default function PlantingProgressList(): JSX.Element {
+  return <p>'I am a list'</p>;
```
karthikbtf, line 2: table code goes here

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
karthikbtf: We can let the map select what data it wants since list versus map views are different. Map needs a single planting site selected for instance.

src/components/NurseryWithdrawals/PlantingProgressList.tsx
karthikbtf: @constanzauanini you and I can share this work for now.. replacing this with table etc.

constanzauanini: Sounds good!

constanzauanini (approved):

PR #1568, opened 2023-07-17
SW-3891 Service for marking planting as complete/incomplete

constanzauanini:


src/services/PlantingsService.ts
```diff
@@ -30,11 +39,21 @@ const listPlantings = async (
   return await SearchService.search(searchParams);
 };
 
+const updatePlantingSubzone = async (subzoneId: number, completed: boolean): Promise<Response> => {
```
karthikbtf, line 42: rename  `completed` to `plantingCompleted`?

src/services/PlantingsService.ts
```diff
@@ -30,11 +39,21 @@ const listPlantings = async (
...
+  return await httpPlantingSubzone.put({
+    urlReplacements: {
+      '{id}': subzoneId.toString(),
+    },
+    entity: { plantingCompleted: completed } as PlantingSubzonePutRequestBody,
```
karthikbtf, line 47:
```
entity: { plantingCompleted } as ..
```
after rename

src/services/PlantingsService.ts
```diff
@@ -30,11 +39,21 @@ const listPlantings = async (
...
+  });
+};
+
 /**
  * Exported functions
  */
 const PlantingsService = {
   listPlantings,
+  updatePlantingSubzone,
```
karthikbtf, line 56: we could call this updatePlantingCompleted for now since that's all it does

karthikbtf (approved):
We need redux for the workflows. Ok to do that in a separate PR (and my request to rename Plantings as plantings in the features directory just to be consistent)

karthikbtf (approved):
Thanks!!

PR #1569, opened 2023-07-17
SW-3891 Adding total plants for site to plantings progress selector

constanzauanini:


karthikbtf (approved):
@constanzauanini looks like plantings in the redux state is specific to a single planting site.
We need to add calls to fetch/store 'all' sites in the redux state. I am working on adding that functionality.
I will have to update your selectors after that.
Approving this for now so you can get it in and I can work on top of all these.

constanzauanini:
> @constanzauanini looks like plantings in the redux state is specific to a single planting site. We need to add calls to fetch/store 'all' sites in the redux state. I am working on adding that functionality. I will have to update your selectors after that. Approving this for now so you can get it in and I can work on top of all these.

Ok, thanks for working on that, I thought it was for all sites

PR #1571, opened 2023-07-17
Refactor plantings redux to fetch all plantings

karthikbtf:
- the scope here is all plantings and we select site using a selector
- used observations pattern here to select all plantings and select a single site off that
- this will also be reused for withdrawals plantings
- updated selectors and calls
- also some code fixes to only listen to org id

### DIFF
--- a/src/components/PlantsV2/index.tsx
+++ b/src/components/PlantsV2/index.tsx
@@ -62,7 +62,9 @@ export default function PlantsDashboardV2(): JSX.Element {
   }, [dispatch, org.selectedOrganization.id]);
 
   useEffect(() => {
-    dispatch(requestSitePopulation(org.selectedOrganization.id, selectedPlantingSiteId));
+    if (selectedPlantingSiteId !== -1) {
+      dispatch(requestSitePopulation(org.selectedOrganization.id, selectedPlantingSiteId));
+    }
   }, [dispatch, org.selectedOrganization.id, selectedPlantingSiteId]);
 
   const sectionHeader = (title: string) => (
--- a/src/components/PlantsV2/index.tsx
+++ b/src/components/PlantsV2/index.tsx
@@ -5,7 +5,7 @@ import { APP_PATHS } from 'src/constants';
 import PlantsPrimaryPage from 'src/components/PlantsPrimaryPage';
 import { Box, Grid, Typography } from '@mui/material';
 import { useAppDispatch, useAppSelector } from 'src/redux/store';
-import { requestPlantingSites, requestSitePopulation } from 'src/redux/features/tracking/trackingThunks';
+import { requestSitePopulation } from 'src/redux/features/tracking/trackingThunks';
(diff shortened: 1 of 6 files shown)


src/components/PlantsV2/index.tsx
```diff
@@ -55,16 +55,15 @@ export default function PlantsDashboardV2(): JSX.Element {
   );
 
   useEffect(() => {
-    dispatch(requestPlantingSites(org.selectedOrganization.id));
```
karthikbtf, line 58: planting sites are now dispatched at the root app level and available to select from within any component

src/components/PlantsV2/index.tsx
```diff
@@ -55,16 +55,15 @@ export default function PlantsDashboardV2(): JSX.Element {
...
     dispatch(requestObservations(org.selectedOrganization.id));
     dispatch(requestObservationsResults(org.selectedOrganization.id));
     dispatch(requestSpecies(org.selectedOrganization.id));
-  }, [dispatch, org]);
+    dispatch(requestPlantings(org.selectedOrganization.id));
```
karthikbtf, line 61: fetch once for all sites in org, we use selectors to select data for specific sites

charlie-tf (approved):
Looks good. Thanks for the update!

src/redux/features/plantings/plantingsThunks.ts
```diff
@@ -3,18 +3,10 @@ import { PlantingsService } from 'src/services';
...
-      const response = await PlantingsService.listPlantings(organizationId, [
-        {
-          field: 'delivery.plantings.plantingSite.id',
-          values: [plantingSiteId.toString()],
-          type: 'Exact',
-          operation: 'field',
-        },
-      ]);
-
+      const response = await PlantingsService.listPlantings(organizationId, []);
```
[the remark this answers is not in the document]
karthikbtf: Guessing the dashboard won't depend on this - once we switch to API that returns plantings since last observation, different redux perhaps.

src/components/PlantsV2/index.tsx
```diff
@@ -55,16 +55,17 @@ export default function PlantsDashboardV2(): JSX.Element {
...
+  }, [dispatch, org.selectedOrganization.id]);
 
   useEffect(() => {
-    dispatch(requestSitePopulation(org.selectedOrganization.id, selectedPlantingSiteId));
-    dispatch(requestPlantings(org.selectedOrganization.id, selectedPlantingSiteId));
-  }, [dispatch, org, selectedPlantingSiteId]);
+    if (selectedPlantingSiteId !== -1) {
```
karthikbtf, line 65: added this one cleanup

src/redux/features/plantings/plantingsThunks.ts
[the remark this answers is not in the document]
charlie-tf: Yes, I think you're right

charlie-tf (approved):
New changes are good

PR #1573, opened 2023-07-18
SW-3903 Display Map for Selected Planting Site

charlie-tf:
image
image

### DIFF
--- a/src/components/NurseryWithdrawals/PlantingProgressMap.tsx
+++ b/src/components/NurseryWithdrawals/PlantingProgressMap.tsx
@@ -1,7 +1,33 @@
+import { useAppSelector } from 'src/redux/store';
+import { selectPlantingSite } from 'src/redux/features/tracking/trackingSelectors';
+import { MapService } from 'src/services';
+import { MapData } from 'src/types/Map';
+import { useEffect, useState } from 'react';
+import { PlantingSiteMap } from 'src/components/Map';
+import strings from 'src/strings';
+import { Typography, useTheme } from '@mui/material';
+
 type PlantingProgressMapProps = {
   plantingSiteId: number;
 };
 
 export default function PlantingProgressMap({ plantingSiteId }: PlantingProgressMapProps): JSX.Element {
-  return <p>{`I am a map of planting site ${plantingSiteId}`}</p>;
+  const theme = useTheme();
+  const plantingSite = useAppSelector((state) => selectPlantingSite(state, plantingSiteId));
+  const [mapData, setMapData] = useState<MapData | undefined>();
+  useEffect(() => {
+    if (plantingSite && plantingSite.boundary) {
+      setMapData(MapService.getMapDataFromPlantingSite(plantingSite));
+    } else {
+      setMapData(undefined);
+    }
+  }, [plantingSite]);
+
+  return mapData ? (
+    <PlantingSiteMap mapData={mapData} focusEntities={[{ sourceId: 'sites', id: plantingSite!.id }]} />
+  ) : (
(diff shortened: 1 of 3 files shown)


src/strings/csv/en.csv
```diff
@@ -602,6 +602,7 @@ NO_BATCHES_TO_WITHDRAW_FROM,There are no batches with withdrawable quantities.
 NO_DATA_YET,No data yet
 NO_FUTURE_DATES,No future dates allowed
 NO_INVENTORY_DESCRIPTION,"It looks like you haven’t added inventory yet. You can either import data from a CSV file, or add inventory using our inventory form."
+NO_MAP_DATA,No map data.
```
karthikbtf, line 605: We could create a ticket for @chudo76 to provide copy?

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
```diff
@@ -1,7 +1,33 @@
...
   plantingSiteId: number;
 };
 
 export default function PlantingProgressMap({ plantingSiteId }: PlantingProgressMapProps): JSX.Element {
-  return <p>{`I am a map of planting site ${plantingSiteId}`}</p>;
+  const theme = useTheme();
+  const plantingSite = useAppSelector((state) => selectPlantingSite(state, plantingSiteId));
+  const [mapData, setMapData] = useState<MapData | undefined>();
+  useEffect(() => {
+    if (plantingSite && plantingSite.boundary) {
```
karthikbtf, line 19: `if (plantingSite?.boundary)` ?

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
```diff
@@ -1,7 +1,33 @@
+import { useAppSelector } from 'src/redux/store';
+import { selectPlantingSite } from 'src/redux/features/tracking/trackingSelectors';
+import { MapService } from 'src/services';
+import { MapData } from 'src/types/Map';
+import { useEffect, useState } from 'react';
```
karthikbtf, line 5: Not requesting any changes. Just something that I hope we can fix soon.

We don't have a strict rule for ordering imports yet (created a ticket for that).. i usually order them as such
1. react
2. external libraries (eg. mui)
3. internal libraries (eg. web-components)
4. internal models/constants
5. internal services/utils/functions
7. internal components

There are node packages that can also order them, we could pick one maybe. Just wanted to put this on everyone's radar in case we have preferences.

karthikbtf (approved):
Looks simple enough 👍

src/strings/csv/en.csv
charlie-tf: I'll make a comment on the figma asking for this

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
charlie-tf: ah yes, that would work too

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
charlie-tf: Good point, I've just let them pile up however. It would be nice to get something standardized, I agree

karthikbtf: Cool,  I can request bumping up that ticket in the tech_investment pile.

src/strings/csv/en.csv
charlie-tf: I'm going to take this out of strings and put a placeholder literal string here, so that we don't request translation for something that might change.

karthikbtf: sounds good 👍

PR #1574, opened 2023-07-18
SW-3935 Store withdrawals tab in url params

karthikbtf:
- allow preserving tabs in url deep links and upon reload of page
- similar to what we do in the accessions view page (code lifted from there)


charlie-tf (approved):
LGTM!

PR #1578, opened 2023-07-18
SW-3938 Redux for getting plantingSites using search

constanzauanini:


src/redux/rootReducer.ts
```diff
@@ -20,6 +20,7 @@ export const reducers = {
   sitePopulation: sitePopulationReducer,
   plantingSiteObservationsResults: plantingSiteObservationsResultsReducer,
   updatePlantingCompleted: updatePlantingCompletedReducer,
+  plantingSitesResults: plantingSitesResultsReducer,
```
karthikbtf, line 23: plantingSitesSearchResults?

src/redux/features/tracking/trackingSlice.ts
```diff
@@ -24,9 +24,31 @@ export const trackingSlice = createSlice({
...
 
+// Define a type for the search slice state
+type SearchData = {
+  sites?: PlantingSiteSearchResult[];
+};
+
+// Define the initial state for search
+const initialSearchState: SearchData = {};
+
+export const plantingSitesResultsSlice = createSlice({
```
karthikbtf, line 37: ...SearchResults..?

karthikbtf (commented):
looks good so far

src/redux/features/plantings/plantingsSelectors.ts
```diff
@@ -22,13 +23,13 @@ export const getTotalPlantsBySubzone = (plantings: PlantingSearchData[]) => {
   }, {});
 };
 
-export const getTotalPlantsBySite = (plantings: PlantingSearchData[]) => {
-  return plantings?.reduce((totalPlantsBySite: { [key: string]: number }, planting) => {
-    const plantingSiteId = planting.plantingSite.id;
+export const getTotalPlantsBySite = (plantingsSites: PlantingSiteSearchResult[]) => {
```
karthikbtf, line 26: @constanzauanini I made some changes to the selectors, there will be conflicts. You can override the conflicts related to this particular function.

src/redux/features/plantings/plantingsSelectors.ts
```diff
@@ -43,11 +44,11 @@ export const selectPlantingsDateRange = (state: RootState, dateRange: string[],
...
 
 export const selectPlantingProgressSubzones = createSelector(
-  [(state: RootState) => selectPlantingSites(state), (state: RootState) => selectPlantings(state)],
+  [(state: RootState) => selectSearchPlantingSites(state), (state: RootState) => selectPlantings(state)],
   (plantingSites, plantings) => {
-    if (plantings) {
+    if (plantingSites && plantings) {
       const plantingsBySubzone = getTotalPlantsBySubzone(plantings);
-      const totalPlantsBySite = getTotalPlantsBySite(plantings);
+      const totalPlantsBySite = getTotalPlantsBySite(plantingSites);
```
karthikbtf, line 51: Noticed on line 56 we are traversing all zones and subzones, this generates a full list of all subzones (slow and incorrect).
The correct result should only have plantings progress in reported.
If you get a chance, can you 
1. return only subzones that are in the plantings data?
2. instead of iterating over all zones and subzones, we could create a map of plantingSite->reportedPlants simply from the plantings data.. and here just use that information
example. iterate over plantings once and populate a map `plantingsProgressMap` of plantingSite -> plantings
here we just set `reported: plantingsProgressMap[plantingSite.id] ?? []`
something like that...

karthikbtf: Ok np, we can revisit later maybe. I already added a check to only populate those entries that have a totalSeedlingsSent - that takes care of one of my requests. 
Previously it was returning a list of all the subzones even if they didn't have plantings progress. Basically this should only return subzones that are in the plantings progress result. The code should be doing that by now. I haven't reviewed the merged response yet let me look :)

karthikbtf: Looks like my changes didn't make it in your merge - can you add them in?

karthikbtf: oops nevermind i see them now, sorry

src/redux/features/tracking/trackingSlice.ts
```diff
@@ -24,9 +24,31 @@ export const trackingSlice = createSlice({
...
+};
+
+// Define the initial state for search
+const initialSearchState: SearchData = {};
+
+export const plantingSitesSearchResultsSlice = createSlice({
+  name: 'plantingSitesResultsSlice',
+  initialState: initialSearchState,
+  reducers: {
+    setSearchPlantingSitesAction: (state, action: PayloadAction<SearchData>) => {
```
karthikbtf, line 41: one request, there's a mix of PlantingSitesSearch and SearchPlantingSites across the code. We could just make this consistent PlantingSitesSearchResults

karthikbtf:
I noticed that the search doesn't work well at times.. i will fix that later.

PR #1586, opened 2023-07-19

terraware-phrase:

### DIFF
--- a/src/strings/csv/es.csv
+++ b/src/strings/csv/es.csv
@@ -950,6 +950,7 @@ SEARCH_BY_NAME_OR_FAMILY,Búsqueda por nombre o familia...
 SEARCH_OR_SELECT,Busque o Seleccione
 SEE_ALL_ACCESSIONS,Ver todos las accesiones
 SEE_ENTRIES,Ver entradas
+SEE_WITHDRAWAL_HISTORY,Ver historial de retirada
 SEED,Semilla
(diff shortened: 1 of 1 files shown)


src/strings/csv/es.csv
```diff
@@ -1222,6 +1225,7 @@ TYPE_TO_SEARCH,Escriba para buscar...
 UNABLE_TO_ADD_PERSON,No se puede añadir una persona
 UNABLE_TO_CONNECT_TO_SENSOR_KIT,No hemos podido conectar con su kit de sensores. En ocasiones esto se debe a que no está conectado al suministro eléctrico o a Internet. Envíe un correo electrónico a {0} para que podamos ayudarle.
 UNABLE_TO_LOAD_NOTIFICATIONS,No se han podido cargar las notificaciones. Vuelva a intentarlo más tarde.
+UNDO_PLANTING_COMPLETE,Deshacer plantación completa
```
karthikbtf, line 1228: @constanzauanini maybe this is not translating well 'undo full planting' versus 'undo planting complete'.. is it ok?

constanzauanini: Yes, it is ok for me. Full planting would be "plantación llena", but we want to mark the planting as complete, so "completa" sounds good to me

constanzauanini (approved):

PR #1588, opened 2023-07-20
SW-3898 Show modal warning if observations stats exist

constanzauanini:
- Added modal when undoing planting complete
- TODO: Verify if selected subzones have statistics

Screenshot 2023-07-20 at 10 52 27

### DIFF
--- a/src/components/NurseryWithdrawals/PlantingProgressList.tsx
+++ b/src/components/NurseryWithdrawals/PlantingProgressList.tsx
@@ -165,9 +165,7 @@ export default function PlantingProgressList({
         <StatsWarninigDialog
           open={showWarningModal}
           onClose={() => setShowWarningModal(false)}
-          onSubmit={() => {
-            onModalSubmit();
-          }}
+          onSubmit={onModalSubmit}
         />
       )}
       <Box>{updatePlantingResult?.status === 'pending' && <BusySpinner withSkrim={true} />}</Box>
--- a/src/components/NurseryWithdrawals/PlantingProgressList.tsx
+++ b/src/components/NurseryWithdrawals/PlantingProgressList.tsx
@@ -116,18 +116,19 @@ export default function PlantingProgressList({
 
   const onModalSubmit = () => {
     setShowWarningModal(false);
-    setPlantingCompleted(false, true);
+    setPlantingCompleted(false);
   };
 
-  const setPlantingCompleted = (complete: boolean, skipModal?: boolean) => {
-    if (!complete && !skipModal) {
-      const haveStatistics = true;
-      if (haveStatistics) {
-        setShowWarningModal(true);
-        return;
-      }
+  const validateUndoPlantingComplete = () => {
+    const haveStatistics = true;
+    if (haveStatistics) {
+      setShowWarningModal(true);
+      return;
     }
+    setPlantingCompleted(false);
+  };
 
+  const setPlantingCompleted = (complete: boolean) => {
     const subzoneIds = selectedRows.map((row) => row.subzoneId);
     const request = dispatch(
       requestUpdatePlantingsCompleted({ subzoneIds, planting: { plantingCompleted: complete } })
@@ -145,7 +146,7 @@ export default function PlantingProgressList({
       topBarButtons.push({
         buttonType: 'passive',
         buttonText: strings.UNDO_PLANTING_COMPLETE,
-        onButtonClick: () => setPlantingCompleted(false),
+        onButtonClick: () => validateUndoPlantingComplete(),
         disabled: !areAllCompleted,
       });
 
--- a/src/components/NurseryWithdrawals/PlantingProgressList.tsx
+++ b/src/components/NurseryWithdrawals/PlantingProgressList.tsx
@@ -16,6 +16,7 @@ import Link from 'src/components/common/Link';
 import { TopBarButton } from '@terraware/web-components/components/table';
 import { requestUpdatePlantingsCompleted } from 'src/redux/features/plantings/plantingsAsyncThunks';
 import useSnackbar from 'src/utils/useSnackbar';
+import StatsWarninigDialog from './StatsWarningModal';
 
 const useStyles = makeStyles(() => ({
   text: {
@@ -93,6 +94,7 @@ export default function PlantingProgressList({
   const [requestId, setRequestId] = useState<string>('');
   const updatePlantingResult = useAppSelector((state) => selectUpdatePlantingsCompleted(state, requestId));
   const snackbar = useSnackbar();
+  const [showWarningModal, setShowWarningModal] = useState(false);
 
   useEffect(() => {
     if (data && hasZones === undefined) {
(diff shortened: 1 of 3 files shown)


src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -112,7 +114,20 @@ export default function PlantingProgressList({
...
 
-  const setPlantingCompleted = (complete: boolean) => {
+  const onModalSubmit = () => {
+    setShowWarningModal(false);
+    setPlantingCompleted(false, true);
+  };
+
+  const setPlantingCompleted = (complete: boolean, skipModal?: boolean) => {
+    if (!complete && !skipModal) {
+      const haveStatistics = true;
```
constanzauanini, line 124: @karthikbtf @charlie-tf I'm missing how to set if the subzone has statistics calculated

src/strings/csv/en.csv
```diff
@@ -224,6 +224,7 @@ DELETE_ORGANIZATION_MSG,Are you sure you want to delete {0}?
 DELETE_SEEDLINGS_BATCHES,Delete Seedlings Batches
 DELETE_SEEDLINGS_BATCHES_MSG,Are you sure you want to delete the seedlings batches?
 DELETE_SPECIES,Delete Species
+DELETE_STATISTICS,Delete Statistics,
```
karthikbtf, line 227: remove trailing comma

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -146,6 +161,15 @@ export default function PlantingProgressList({
 
   return (
     <Box>
+      {showWarningModal && (
+        <StatsWarninigDialog
+          open={showWarningModal}
+          onClose={() => setShowWarningModal(false)}
+          onSubmit={() => {
```
karthikbtf, line 168: onSubmit={onModalSubmit}?

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -112,7 +114,20 @@ export default function PlantingProgressList({
...
   }
 
-  const setPlantingCompleted = (complete: boolean) => {
+  const onModalSubmit = () => {
+    setShowWarningModal(false);
+    setPlantingCompleted(false, true);
+  };
+
+  const setPlantingCompleted = (complete: boolean, skipModal?: boolean) => {
+    if (!complete && !skipModal) {
```
karthikbtf, line 123: can we separate out the 2 paths into separate functions? this approach may leave room for errors ending up in infinite loops

src/components/NurseryWithdrawals/PlantingProgressList.tsx
karthikbtf: Good question. When a planting subzone is marked completed, the BE calculates statistics like planting density etc based on observation results that were completed on or before the plantingCompletedTime.
How do we get this info now? I think we may need to look at the api and selectors we already have perhaps. Will defer to Charlie here since we need this for the map as well.

src/components/NurseryWithdrawals/PlantingProgressList.tsx
karthikbtf: also, i'm assuming we will have a confirmation modal at some point? (with Chudo's feedback ofc)

src/components/NurseryWithdrawals/PlantingProgressList.tsx
karthikbtf: getting this in first and patching up with the checking code later is fine by me

charlie-tf: I think that "statistics calculated" means that all of the subzones for a zone have been marked as complete, thereby triggering the BE to calculate aggregate statistics for the zone. Will have to think about how to detect this...

karthikbtf:
```
    ObservationPlantingZoneResultsPayload: {
      /**
       * Format: int32
       * @description Estimated number of plants in planting zone based on estimated planting density and planting zone area. Only present if all the subzones in the zone have been marked as having completed planting.
       */
      estimatedPlants?: number;
```

karthikbtf: And somehow factor in the plantingCompletedTime as well (I guess skip observation results completed after this time). Ideally there should be only one observation round in motion .. most of my corner cases probably only impact testing.

charlie-tf: Here's selectLatestObservation: https://github.com/terraware/terraware-web/blob/main/src/redux/features/observations/observationsSelectors.ts#L106

karthikbtf: Cool, that would work. 
I would suggest a new selector that takes in a list of zone-ids and checks for this (since list can do batch undos). 
The selector can be shared by list and map components. thoughts?
cc @constanzauanini maybe that clarifies?

charlie-tf: >a new selector that takes in a list of zone-ids and checks for this (since list can do batch undos).
The selector can be shared by list and map components. thoughts?

That sounds good!

karthikbtf: Thanks @charlie-tf 🙏

constanzauanini: Thanks @charlie-tf and @karthikbtf! Yes, this clarifies. I will create the selector to check for this. I will get this in first, and open a follow up pr

src/components/NurseryWithdrawals/PlantingProgressList.tsx
constanzauanini: Do you mean one function for set plating complete and one for undoing planting complete?

karthikbtf: It may end up being that way - suggestion is to look at it this way.
The setPlantingCompleted function was correct before, it finds the subzones and dispatches a call etc. and uses the passed in boolean value.
The logic to show warning modal should ideally be isolated from this function.

constanzauanini: Gotcha!

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -116,18 +116,19 @@ export default function PlantingProgressList({
...
 
-  const setPlantingCompleted = (complete: boolean, skipModal?: boolean) => {
-    if (!complete && !skipModal) {
-      const haveStatistics = true;
-      if (haveStatistics) {
-        setShowWarningModal(true);
-        return;
-      }
+  const validateUndoPlantingComplete = () => {
+    const haveStatistics = true;
```
karthikbtf, line 123: `hasStatistics`?

karthikbtf (approved):
Looks good. Do you want to add the selector etc in a followup PR?

src/components/NurseryWithdrawals/PlantingProgressList.tsx
constanzauanini: This will validate an array of various subzones to see if at least one has statistics, that's why I used plural, but can change it if it's better

karthikbtf: thanks for clarifying.. makes sense now

PR #1589, opened 2023-07-20
SW-3950 Use Site Reported Plants Endpoint

charlie-tf:
- create service
- create redux
- replace plantings date range selector with new selector
- delete extraneous calculations

### DIFF
--- a/src/redux/features/tracking/trackingSlice.ts
+++ b/src/redux/features/tracking/trackingSlice.ts
@@ -77,7 +77,11 @@ export const sitePopulationReducer = sitePopulationSlice.reducer;
 type SiteReportedPlantsData = {
   error?: string;
   site?: PlantingSiteReportedPlants;
+};
+
+type SiteReportedPlantsPayload = {
   plantingSiteId: number;
+  data: SiteReportedPlantsData;
 };
 
 const initialSiteReportedPlantsState: Record<number, SiteReportedPlantsData> = {};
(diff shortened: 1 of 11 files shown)


src/redux/features/tracking/trackingSlice.ts
```diff
@@ -71,3 +72,27 @@ export const sitePopulationSlice = createSlice({
...
 
 export const sitePopulationReducer = sitePopulationSlice.reducer;
+
+// Planting Site Reported Plants Slice
+type SiteReportedPlantsData = {
+  error?: string;
+  site?: PlantingSiteReportedPlants;
+};
+
+const initialSiteReportedPlantsState: SiteReportedPlantsData = {};
```
karthikbtf, line 82: this should ideally be a record of site id to data

src/redux/features/tracking/trackingThunks.ts
```diff
@@ -49,3 +54,17 @@ export const requestPlantingSitesSearchResults = (organizationId: number) => {
     }
   };
 };
+
+export const requestSiteReportedPlants = (plantingSiteId: number) => {
+  return async (dispatch: Dispatch, _getState: () => RootState) => {
+    try {
+      const response = await TrackingService.getReportedPlants(plantingSiteId);
+      const { error, site } = response;
+      dispatch(setSiteReportedPlantsAction({ error, site }));
```
karthikbtf, line 63: pass the site id as well

karthikbtf (commented):
Looks good in general. Thanks so much for doing this.
Ideally for anything by id, we should book-keep by id (redux state and selectors).
Otherwise there will be race conditions where we load a component and end up showing incorrect data while a request is being dispatched.

karthikbtf (approved):
Looks good, just that one change if you can 🙏

karthikbtf (approved):
Sorry for the further suggestions but will be cleaner that way.
Looks great otherwise 🎉

charlie-tf:
> Sorry for the further suggestions but will be cleaner that way. Looks great otherwise 🎉

@karthikbtf thanks for your suggestions in cleaning up this code!

karthikbtf (approved):
Nice 🎉 
Thanks so much!

PR #1590, opened 2023-07-20
SW-3900 Add subzone filter to nursery withdrawals table

karthikbtf:
- parse subzones filter options in service (sort values)
- enable subzones filter option
- this will be used in a follow up PR to set filters based on url params

Terraware App 2023-07-20 10-58-41

Terraware App 2023-07-20 10-57-47

Terraware App 2023-07-20 10-57-22

Terraware App 2023-07-20 10-52-53

### DIFF
--- a/src/components/NurseryWithdrawals/NurseryWithdrawalsTable.tsx
+++ b/src/components/NurseryWithdrawals/NurseryWithdrawalsTable.tsx
@@ -72,6 +72,7 @@ export default function NurseryWithdrawalsTable(): JSX.Element {
             { name: 'purpose', label: strings.PURPOSE, type: 'multiple_selection' },
             { name: 'facility_name', label: strings.FROM_NURSERY, type: 'multiple_selection' },
             { name: 'destinationName', label: strings.DESTINATION, type: 'multiple_selection' },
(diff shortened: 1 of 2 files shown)


charlie-tf (approved):
seems pretty straightforward 👍

constanzauanini (approved):
Working as expected!

karthikbtf:
Thanks @charlie-tf @constanzauanini  -  fwiw, I didn't add a trackingV2 check on adding this filter. Not sure if I need to.. this is feasible with v1 as well. Thoughts?

constanzauanini:
> Thanks @charlie-tf @constanzauanini - fwiw, I didn't add a trackingV2 check on adding this filter. Not sure if I need to.. this is feasible with v1 as well. Thoughts?

Yes, I don't see any problem having it in v1

karthikbtf:
> > Thanks @charlie-tf @constanzauanini - fwiw, I didn't add a trackingV2 check on adding this filter. Not sure if I need to.. this is feasible with v1 as well. Thoughts?
> 
> Yes, I don't see any problem having it in v1

Thanks!

PR #1591, opened 2023-07-20
SW-3898 Create selectZonesHaveStatistics selector

constanzauanini:


src/redux/features/plantings/plantingsSelectors.ts
```diff
@@ -116,3 +118,25 @@ export const selectUpdatePlantingCompleted = (state: RootState, requestId: strin
 
 export const selectUpdatePlantingsCompleted = (state: RootState, requestId: string) =>
   (state.updatePlantingsCompleted as any)[requestId];
+
+export const selectZonesHaveStatistics = createSelector(
+  [
+    (state: RootState, zoneIds?: number[]) => selectLatestObservation(state, -1),
```
karthikbtf, line 124: This will return the latest observation across all planting sites.

You may need to pass in a record of  Record<siteId, list of zone ids>
and then maybe do a select for each of those site Ids.

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -92,7 +93,9 @@ export default function PlantingProgressList({
   const [selectedRows, setSelectedRows] = useState<any[]>([]);
   const dispatch = useAppDispatch();
   const [requestId, setRequestId] = useState<string>('');
+  const [selectedZoneIds, setSelectedZoneIds] = useState<number[]>();
   const updatePlantingResult = useAppSelector((state) => selectUpdatePlantingsCompleted(state, requestId));
+  const subzonesStatisticsResult = useAppSelector((state) => selectZonesHaveStatistics(state, selectedZoneIds));
```
karthikbtf, line 98: To see data here, we need to dispatch a call to request observation results in the parent component, example: https://github.com/terraware/terraware-web/blob/main/src/components/Observations/index.tsx#L52

Also may need to refresh that data if planting complete is updated

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -120,7 +129,7 @@ export default function PlantingProgressList({
   };
 
   const validateUndoPlantingComplete = () => {
-    const haveStatistics = true;
+    const haveStatistics = subzonesStatisticsResult;
     if (haveStatistics) {
```
karthikbtf, line 133: merge into one `if (subzonesStatisticsResults)`

src/redux/features/plantings/plantingsSelectors.ts
```diff
@@ -116,3 +118,25 @@ export const selectUpdatePlantingCompleted = (state: RootState, requestId: strin
...
+    (state: RootState, zoneIds?: number[]) => zoneIds,
+  ],
+  (latestObservations, zoneIds) => {
+    let subzonesHaveStats = false;
```
karthikbtf, line 128: zonesHaveStatistics

src/redux/features/plantings/plantingsSelectors.ts
```diff
@@ -116,3 +118,25 @@ export const selectUpdatePlantingCompleted = (state: RootState, requestId: strin
...
+    zoneIds?.forEach((zoneId) => {
+      // if one subzone has stats, then we return true and don't need to continue
+      if (!subzonesHaveStats) {
+        const plantingZoneFound = latestObservations?.plantingZones.find(
+          (plantingZone) => plantingZone.plantingZoneId === zoneId
+        );
+        if (plantingZoneFound && plantingZoneFound.estimatedPlants) {
```
karthikbtf, line 135: check for null or undefined.. i guess if 0 that is still ok as having stats

karthikbtf (commented):
Thanks for working on this selector. Some changes requested.

src/components/NurseryWithdrawals/PlantingProgressList.tsx
karthikbtf: Here when you make the dispatch call, you can use '-1' as the site id (to fetch all results).

src/redux/features/plantings/plantingsSelectors.ts
```diff
@@ -116,3 +118,25 @@ export const selectUpdatePlantingCompleted = (state: RootState, requestId: strin
...
   (state.updatePlantingsCompleted as any)[requestId];
+
+export const selectZonesHaveStatistics = createSelector(
+  [
+    (state: RootState, zoneIds?: number[]) => selectLatestObservation(state, -1),
+    (state: RootState, zoneIds?: number[]) => zoneIds,
+  ],
+  (latestObservations, zoneIds) => {
+    let subzonesHaveStats = false;
+    zoneIds?.forEach((zoneId) => {
```
karthikbtf, line 129: you can shortcut with `some` - more efficient

```
 const hasStatistics = zoneIds?.some(zoneId => {
  return latestObservations?.plantingZones.some(zone => zone.plantingZoneId === zoneId && zone.estimatedPlants !== null && zone.estimatedPlants !== undefined);
});
```
something like that

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -107,17 +114,31 @@ export default function PlantingProgressList({
...
+      dispatch(requestObservationsResults(selectedOrganization.id));
     } else if (updatePlantingResult?.status === 'error') {
       snackbar.toastError(strings.GENERIC_ERROR);
     }
-  }, [updatePlantingResult, reloadTracking, snackbar]);
+  }, [updatePlantingResult, reloadTracking, snackbar, dispatch, selectedOrganization.id]);
+
+  useEffect(() => {
+    dispatch(requestObservationsResults(selectedOrganization.id));
+  }, [dispatch, selectedOrganization.id]);
```
karthikbtf, line 141: if we do this in the parent component, we can leverage it for maps as well

karthikbtf (commented):
thanks for the changes, this is looking much better..
some more suggestions

src/redux/features/plantings/plantingsSelectors.ts
```diff
@@ -121,22 +121,26 @@ export const selectUpdatePlantingsCompleted = (state: RootState, requestId: stri
 
 export const selectZonesHaveStatistics = createSelector(
   [
-    (state: RootState, zoneIds?: number[]) => selectLatestObservation(state, -1),
-    (state: RootState, zoneIds?: number[]) => zoneIds,
+    (state: RootState, zoneIdsBySiteId?: Record<number, number[]>, defaultTimeZoneId?: string) => state,
```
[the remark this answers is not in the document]
constanzauanini: Not sure, but I can make it cached if we think it would be better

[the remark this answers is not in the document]
karthikbtf: you currently have it as a cached selector (createSelector) but the usage pattern is interesting, hence my question.
ideally
createSelector(
  [ someSelectorfunction,... ],
  (selectedData, ...) {
     // run some expensive operation on selected data
  }
)

here, a lot of the selection is happening in the callback to selected data.
theoretically it still works. i don't know enough to gauge if this pattern is ok but it is different.

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -91,24 +95,46 @@ export default function PlantingProgressList({
...
+        return selectedZoneIdsBySiteIdObj;
+      }, {});
+      setSelectedZoneIdsBySiteId(zoneIds);
+    }
+  }, [selectedRows]);
+
   useEffect(() => {
     if (updatePlantingResult?.status === 'success') {
       reloadTracking();
+      dispatch(requestObservationsResults(selectedOrganization.id));
```
karthikbtf, line 133: move this into the parent, make it part of reload tracking in the parent (a wrapped function) i think i made a comment but it isn't showing up here..

src/components/NurseryWithdrawals/PlantingProgressList.tsx
```diff
@@ -91,24 +95,46 @@ export default function PlantingProgressList({
...
       setHasZones(data.some((d) => d.subzoneName));
     }
   }, [data, hasZones]);
 
+  useEffect(() => {
+    if (selectedRows) {
+      const zoneIds = selectedRows.reduce((selectedZoneIdsBySiteIdObj: Record<number, number[]>, row) => {
+        const siteId = row.siteId;
+        if (selectedZoneIdsBySiteIdObj[siteId]) {
+          selectedZoneIdsBySiteIdObj[siteId].push(row.zoneId);
```
karthikbtf, line 120: missed this.. we will have a list with duplicate zoneIds if multiple subzones belong to the same zone.
make this a set.. otherwise we increase the cost of the selector function

src/redux/features/plantings/plantingsSelectors.ts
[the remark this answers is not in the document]
karthikbtf: we can leave this as-is and revisit later, the end result is still cached for now so that's good maybe.

karthikbtf (approved):
Overall approach is good now and code looks better.
Asking for some changes if you can.

karthikbtf (approved):
Nice 🙌 thanks 🎉

PR #1594, opened 2023-07-21
SW-3953 Show planting site details page when tracking-v2 is disabled

karthikbtf:
- skip observations results status check if tracking v2 is disabled
- we were not showing data while observations data was not requested, this should only happen with trackingv2 enabled


constanzauanini (approved):

karthikbtf:
Thanks for the review 🙏

PR #1596, opened 2023-07-21
SW-3960 Show Success planting complete snackbar

constanzauanini:
The snackbar in design is only shown when subzone is marked as planting complete.

Adding messages for both cases

Screenshot 2023-07-21 at 15 08 54

### DIFF
--- a/src/strings/csv/en.csv
+++ b/src/strings/csv/en.csv
@@ -916,6 +916,7 @@ SAVE,Save
 SAVE_ACCESSION_ERROR,An error occurred when saving the accession.
 SAVE_AND_NEXT,Save and Next
 SAVE_CHANGES,Save Changes
+SAVED,Saved
 SAVING,Saving...
 SCARIFY,Scarify
 SCHEDULE_DATE_INFO,Schedule a date by selecting a future date.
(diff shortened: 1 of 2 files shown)


src/strings/csv/en.csv
```diff
@@ -1093,6 +1094,8 @@ SUBSHRUB,Subshrub
 SUBSTRATE,Substrate
 SUBTITLE_GET_STARTED,"To get started, please create your organization now. You'll also be able to set up and manage information on people and species."
 SUBZONE,Subzone
+SUBZONE_PLANTING_COMPLETED_SUCCESS,Subzone marked as planting complete.
+SUBZONE_PLANTING_INCOMPLETED_SUCCESS,Subzone marked as planting incomplete.
```
karthikbtf, line 1098: UNCOMPLETED should be fine

karthikbtf, line 1098: what if user had selected multiple subzones? should this be. `Subzone(s)` instead or ok as-is?

karthikbtf (approved):

src/strings/csv/en.csv
constanzauanini: Yes, I think suzones(s) would be better

PR #1597, opened 2023-07-21
SW-3908 Warn When Undoing Planting Complete will Delete Stats

charlie-tf:
- determine the zone for a subzone when clicking map
- select whether that zone has statistics
- if there are no statistics then allow update without a warning
- if there are statistics:
  - open warning modal
  - set a callback that is executed if the deletion is confirmed
- after update refresh observation results as well as planting site

image

### DIFF
--- a/src/components/NurseryWithdrawals/PlantingProgressMap.tsx
+++ b/src/components/NurseryWithdrawals/PlantingProgressMap.tsx
@@ -2,17 +2,23 @@ import { useAppDispatch, useAppSelector } from 'src/redux/store';
 import { selectPlantingSite } from 'src/redux/features/tracking/trackingSelectors';
 import { MapService } from 'src/services';
 import { MapData, MapSourceProperties } from 'src/types/Map';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { PlantingSiteMap } from 'src/components/Map';
 import { Typography, useTheme } from '@mui/material';
 import PlantingProgressMapDialog from './PlantingProgressMapDialog';
 import { makeStyles } from '@mui/styles';
 import { requestPlantingSites, requestSitePopulation } from 'src/redux/features/tracking/trackingThunks';
 import { useOrganization } from 'src/providers';
-import { selectUpdatePlantingCompleted } from 'src/redux/features/plantings/plantingsSelectors';
+import {
+  selectUpdatePlantingCompleted,
+  selectZonesHaveStatistics,
+} from 'src/redux/features/plantings/plantingsSelectors';
 import strings from 'src/strings';
 import { requestUpdatePlantingCompleted } from 'src/redux/features/plantings/plantingsAsyncThunks';
 import useSnackbar from 'src/utils/useSnackbar';
+import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
+import StatsWarningDialog from 'src/components/NurseryWithdrawals/StatsWarningModal';
+import { requestObservationsResults } from 'src/redux/features/observations/observationsThunks';
 
 export const useStyles = makeStyles(() => ({
   popup: {
@@ -33,6 +39,7 @@ export default function PlantingProgressMap({ plantingSiteId }: PlantingProgress
   const classes = useStyles();
   const dispatch = useAppDispatch();
   const org = useOrganization();
+  const defaultTimeZone = useDefaultTimeZone();
   const plantingSite = useAppSelector((state) => selectPlantingSite(state, plantingSiteId));
   const [mapData, setMapData] = useState<MapData | undefined>();
   const [dispatching, setDispatching] = useState(false);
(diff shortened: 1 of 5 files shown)


src/components/NurseryWithdrawals/PlantingProgressMap.tsx
```diff
@@ -73,48 +86,74 @@ export default function PlantingProgressMap({ plantingSiteId }: PlantingProgress
       if (updateStatus.status === 'error') {
         snackbar.toastError(strings.GENERIC_ERROR);
       } else if (updateStatus.status === 'success') {
-        // refresh planting site data to get new completed state for subzone
+        // refresh planting site and observations data
         dispatch(requestPlantingSites(org.selectedOrganization.id));
+        dispatch(requestObservationsResults(org.selectedOrganization.id));
```
constanzauanini, line 91: There is a function _reloadTrackingAndObservations_ in the parent (Planting progress tab content) that refreshes this. Maybe we can pass it to the map as we are passing it to the list? It also refreshes plantingSites

charlie-tf: Oh, I'll take a look at using that 👍

charlie-tf: I'm not sure why but it caused the map to glitch like crazy. I'm going to leave as is for now, and we can possibly follow up.

constanzauanini (approved):
Looks good otherwise !

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
karthikbtf: I will take a look at this today.

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
```diff
@@ -92,29 +106,63 @@ export default function PlantingProgressMap({ plantingSiteId }: PlantingProgress
...
+    },
+    [selectedZoneHasStats, completeUpdate]
+  );
+
+  const onOpenMapDialog = useCallback(
+    (id: number) => {
+      const selectedZone = plantingSite?.plantingZones?.find((zone) =>
+        zone.plantingSubzones.map((sz) => sz.id).includes(id)
+      );
+      setZoneIdSelected(selectedZone?.id ?? -1);
```
karthikbtf, line 132: do we need the -1 fallback?

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
```diff
@@ -73,15 +86,16 @@ export default function PlantingProgressMap({ plantingSiteId }: PlantingProgress
...
         dispatch(requestPlantingSites(org.selectedOrganization.id));
+        dispatch(requestObservationsResults(org.selectedOrganization.id));
       }
     }
   }, [updateStatus, dispatch, snackbar, org.selectedOrganization.id]);
 
-  const updatePlantingComplete = useCallback(
-    (id: number, val: boolean) => {
-      // TODO: warn if undoing planting complete will erase statistics
+  const [confirmDeleteCb, setConfirmDeleteCb] = useState<() => void>(() => () => null);
```
karthikbtf, line 96: can we leverage onSubmit to call updatePlantingComplete, without this additional callback?

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
```diff
@@ -92,29 +106,63 @@ export default function PlantingProgressMap({ plantingSiteId }: PlantingProgress
...
     [dispatch]
   );
 
+  const updatePlantingComplete = useCallback(
+    (id: number, val: boolean) => {
+      if (!selectedZoneHasStats) {
+        completeUpdate(id, val)();
+      } else {
+        setConfirmDeleteCb(() => completeUpdate(id, val));
+        setStatsWarningDialogOpen(true);
```
karthikbtf, line 121: suggestion:
remove the callback approach, find the zone id here and set that as a property.
if the zone id is valid then show the modal (pass the zone id to the modal)
modal can send the zone id back as part of the onsubmit.
we can remove a bunch of additional state/message-passing

src/components/NurseryWithdrawals/PlantingProgressMapDialog.tsx
```diff
@@ -44,6 +46,12 @@ export default function PlantingProgressMapDialog({
     return Object.values(species).reduce((prev, curr) => prev + curr, 0);
   }, [species]);
 
+  useEffect(() => {
```
karthikbtf, line 49: i think we can remove this.. we can find the zone id at the time of deciding whether to open the modal.. and pass that to the modal.. which can pass it back onSubmit

karthikbtf (commented):
Might be better to remove the callback and the onOpen.. find the zone id when we decide to show the modal and set that as a property. Modal can pass it back in onSubmit

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
constanzauanini: It could be undefined instead of -1 maybe

karthikbtf:
Merging this in for now.
Charlie when you get a chance review https://github.com/terraware/terraware-web/pull/1601 which has my suggestions implemented.

karthikbtf:
i merged the wrong PR :(

PR #1600, opened 2023-07-24
SW-3955 Limit decimal places in percentage bar charts

constanzauanini:
Screenshot 2023-07-24 at 11 01 26

### DIFF
--- a/src/components/PlantsV2/components/NumberOfSpeciesPlantedCard.tsx
+++ b/src/components/PlantsV2/components/NumberOfSpeciesPlantedCard.tsx
@@ -53,7 +53,11 @@ export default function NumberOfSpeciesPlantedCard({ plantingSiteId }: NumberOfS
       const speciesCount = speciesNames.length;
       setTotalSpecies(speciesCount);
       setLabels(Object.keys(speciesByCategory));
-      setValues(Object.values(speciesByCategory).map((cat) => (speciesCount > 0 ? (cat * 100) / speciesCount : 0)));
+      setValues(
+        Object.values(speciesByCategory).map((cat) =>
+          speciesCount > 0 ? Number(((cat * 100) / speciesCount).toFixed(2)) : 0
+        )
+      );
     } else {
       setLabels([]);
       setValues([]);


src/components/PlantsV2/components/NumberOfSpeciesPlantedCard.tsx
```diff
@@ -53,7 +53,11 @@ export default function NumberOfSpeciesPlantedCard({ plantingSiteId }: NumberOfS
       const speciesCount = speciesNames.length;
       setTotalSpecies(speciesCount);
       setLabels(Object.keys(speciesByCategory));
-      setValues(Object.values(speciesByCategory).map((cat) => (speciesCount > 0 ? (cat * 100) / speciesCount : 0)));
+      setValues(
+        Object.values(speciesByCategory).map((cat) =>
+          speciesCount > 0 ? Number(((cat * 100) / speciesCount).toFixed(2)) : 0
```
karthikbtf, line 58: should we do this only if there are decimal values? if the number is '2', this will end up with a 2.00
Also curious if the chart is localization friendly (. vs , for decimal delimiter)

unrelated question, are there other places? wondering if we need to enhance our FormattedNumber component as well.

constanzauanini: I just tested on your org, on plantsVsZombies planting site, and I'm not seeing the '.00' in the percentage, but I don't understand why, because you are right, toFixed() should add it

constanzauanini: This is not using the FormattedNumber component, maybe we need to format it before passing it to the chart (for the . vs , )  but don't think we need to update it

karthikbtf: > I just tested on your org, on plantsVsZombies planting site, and I'm not seeing the '.00' in the percentage, but I don't understand why, because you are right, toFixed() should add it

Ok no worries, we can see if we run into it later.

karthikbtf: > This is not using the FormattedNumber component, maybe we need to format it before passing it to the chart (for the . vs , ) but don't think we need to update it

I think the chart locale should handle it automatically. Re: FormattedNumber, I was wondering if we have use-cases that need the toFixed solution applied to that component as well - unrelated to charts.

karthikbtf: This PR addresses the specific bug, we can follow up as we run into issues maybe :)

PR #1602, opened 2023-07-24
Add locale to chart label

constanzauanini:
This allows showing ' , or . ' in chart label
Screenshot 2023-07-24 at 14 17 12

### DIFF
--- a/src/components/PlantsV2/components/NumberOfSpeciesPlantedCard.tsx
+++ b/src/components/PlantsV2/components/NumberOfSpeciesPlantedCard.tsx
@@ -8,7 +8,6 @@ import BarChart from 'src/components/common/Chart/BarChart';
 import { selectSpecies } from 'src/redux/features/species/speciesSelectors';
 import { useUser } from 'src/providers';
 import { useNumberFormatter } from 'src/utils/useNumber';
-import { useSupportedLocales } from 'src/strings/locales';
 
 type NumberOfSpeciesPlantedCardProps = {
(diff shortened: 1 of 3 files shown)


src/components/PlantsV2/components/NumberOfSpeciesPlantedCard.tsx
```diff
@@ -18,6 +21,14 @@ export default function NumberOfSpeciesPlantedCard({ plantingSiteId }: NumberOfS
   const [totalSpecies, setTotalSpecies] = useState<number>();
   const [labels, setLabels] = useState<string[]>();
   const [values, setValues] = useState<number[]>();
+  const user = useUser().user;
+  const numberFormatter = useNumberFormatter();
+  const supportedLocales = useSupportedLocales();
+  const numericFormatter = useMemo(
+    () => numberFormatter(user?.locale, supportedLocales),
+    [user?.locale, numberFormatter, supportedLocales]
```
karthikbtf, line 29: suggestion, we should probably expose a wrapper that does this. ideally we don't want to pass supportedLocales everywhere we use numberFormatter. I think useNumberFormatter should handle the supportedLocales automatically

karthikbtf (commented):
Looks good, I have a suggestion if you can follow up.

karthikbtf (commented):
thanks for the changes

karthikbtf (approved):
Thanks 🎉

PR #1609, opened 2023-07-25
SW-3967 site planting density is shown even when no observations have occurred

constanzauanini:
Hide planting density chart if site has no observations


karthikbtf (approved):
Looks harmless 👍 
@constanzauanini add a PR description explaining the changes (even if simple)?

PR #1610, opened 2023-07-25
SW-3969 Seed bank, Nursery, and Planting Sites selection lists should be alphabetized

constanzauanini:


src/services/TrackingService.ts
```diff
@@ -75,7 +75,17 @@ const listPlantingSites = async (organizationId: number, full?: boolean): Promis
         full: (full || false).toString(),
       },
     },
-    (data) => ({ sites: data?.sites })
+    (data) => ({
+      sites: data?.sites.sort((a, b) => {
```
karthikbtf, line 79: sort((a, b) = a.localeCompare(b, <user-locale>)

src/utils/organization.tsx
```diff
@@ -11,7 +11,15 @@ export const getFacilitiesByType = (organization: Organization, type: FacilityTy
 };
 
 export const getAllSeedBanks = (organization: Organization): Facility[] => {
-  return getFacilitiesByType(organization, 'Seed Bank');
+  return getFacilitiesByType(organization, 'Seed Bank').sort((a, b) => {
```
karthikbtf, line 14: same.. might as well add a string utils hook which automatically fetches user locale and does a localecompare
useStringCompare

src/utils/organization.tsx
```diff
@@ -27,7 +35,15 @@ export const isContributor = (roleHolder: Organization | OrganizationUser | unde
...
-  return getFacilitiesByType(organization, 'Nursery');
+  return getFacilitiesByType(organization, 'Nursery').sort((a, b) => {
+    if (a.name < b.name) {
+      return -1;
+    }
+    if (a.name > b.name) {
+      return 1;
+    }
+    return 0;
+  });
```
karthikbtf, line 46: usually repeated code should be extracted out

karthikbtf (commented):
looks like the build is failing.. and missing PR description

src/utils/organization.tsx
karthikbtf: also you can move the sorting inside getFacilitiesByType so we don't need to call it in the calling-code

src/utils/organization.tsx
constanzauanini: I was trying to do this, but a hook can't be called here

karthikbtf: thanks for trying.. ok looks like we will have to pass the user locale into all these functions then, sigh

src/utils/organization.tsx
karthikbtf: another option is to leave the sorting up to the rendering component

karthikbtf: oops wrong place

src/utils/organization.tsx
karthikbtf: another option is to leave the sorting up to the rendering component, or however you think works best

constanzauanini: I'll pass the locale into these functions

src/App.tsx
```diff
@@ -177,6 +177,7 @@ function AppContent() {
   const [showNavBar, setShowNavBar] = useState(true);
   const trackingV2 = isEnabled('TrackingV2');
   const dispatch = useAppDispatch();
+  const { user } = useUser();
```
karthikbtf, line 180: i believe we use `useLocalization` and pick `activeLocale` from that

karthikbtf (commented):
thanks for these changes.. just one more request

src/App.tsx
constanzauanini: Should I change it on every component? What is the difference with the user locale?

karthikbtf: I believe most components already uses useLocalization. If you just change it here in your PR that should be ok.
Objective was that localization information is provided by the useLocalization provider. This allows us to not depend on internal semantics such as user.locale.

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -117,13 +117,13 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
     if (plantingSites) {
       return;
     }
-    const response = await TrackingService.listPlantingSites(selectedOrganization.id, true);
+    const response = await TrackingService.listPlantingSites(selectedOrganization.id, true, user?.locale);
```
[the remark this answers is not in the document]
karthikbtf: We use user.locale only in MyAccount page and language selectors to get/set the value and in the LocalizationProvider to set it in the provider context.

karthikbtf (approved):
Feel free to merge after that change 👍

src/components/Inventory/withdraw/flow/SelectPurposeForm.tsx
```diff
@@ -117,13 +118,13 @@ export default function SelectPurposeForm(props: SelectPurposeFormProps): JSX.El
     if (plantingSites) {
       return;
     }
-    const response = await TrackingService.listPlantingSites(selectedOrganization.id, true, user?.locale);
+    const response = await TrackingService.listPlantingSites(selectedOrganization.id, true, activeLocale || undefined);
```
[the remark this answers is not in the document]
karthikbtf: do we need the || undefined here and in other places? because of null? if so, might be cleaner to update listPlantingSites to accept null as well

karthikbtf (commented):
one question, otherwise lgtm thanks

PR #1611, opened 2023-07-25
SW-3961 Fix SerializableStateInvariantMiddleware warning in app

karthikbtf:
- the default middleware checks for immutability/serializability of redux state / responses
- this happens in dev mode only!
- can be slow for some of our API such as observations which can take time
- middleware complains that it takes too much time
- disabled these checks (reference https://github.com/reduxjs/redux-toolkit/issues/805)
- not sure on implications of disabling these checks yet


constanzauanini (approved):
I don't know the implications either but seems ok if its taking too much time

PR #1613, opened 2023-07-25
SW-3908 Changes to PR 1597 (warning dialog on update planting complete on map)

karthikbtf:
- refactored passing of id/val in a more constrained context
- the earlier effort was accidentally merged onto Charlie's branch and reverted again, that PR is no longer valid... sending this out again for Charlie to review

### DIFF
--- a/src/components/NurseryWithdrawals/PlantingProgressMap.tsx
+++ b/src/components/NurseryWithdrawals/PlantingProgressMap.tsx
@@ -112,7 +112,7 @@ export default function PlantingProgressMap({ plantingSiteId, reloadTracking }: 
 
   const updatePlantingComplete = useCallback(
     (id: number, val: boolean) => {
-      if (val === true || !selectedZoneHasStats) {
+      if (!selectedZoneHasStats) {
         completeUpdate(id, val);
       } else {
         setStatsWarningDialogProps({ id, val });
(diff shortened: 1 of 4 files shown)


karthikbtf:
@charlie-tf I left this one for you to review - if you think it makes sense we can get it in otherwise no harm, this doesn't change any functionality.

charlie-tf (changes requested):
Couple small cleanup changes, but otherwise looks good. Thanks for the update!

karthikbtf:
> Couple small cleanup changes, but otherwise looks good. Thanks for the update!

Thanks for the review Charlie, this one should be ready to go as well.

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
```diff
@@ -112,7 +112,7 @@ export default function PlantingProgressMap({ plantingSiteId, reloadTracking }:
 
   const updatePlantingComplete = useCallback(
     (id: number, val: boolean) => {
-      if (!selectedZoneHasStats) {
+      if (val === true || !selectedZoneHasStats) {
```
charlie-tf, line 115: I think that in practice this should not happen—there won't be statistics if the subzones aren't already all completed. I don't object to having this check in there though.

charlie-tf (approved):
👍 Thanks for the changes

src/components/NurseryWithdrawals/PlantingProgressMap.tsx
karthikbtf: Ah yes, that is true. Hmm.. I can remove it since it doesn't make sense.

charlie-tf (approved):

PR #1614, opened 2023-07-27
SW-3845 Add "zoom to fit" button to maps

constanzauanini:
Adding "zoom to fit" button to all maps.

Icon still needs to be updated

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { Box } from '@mui/material';
+import { Box, Button, useTheme } from '@mui/material';
 import 'mapbox-gl/dist/mapbox-gl.css';
 import ReactMapGL, { AttributionControl, Layer, MapRef, NavigationControl, Popup, Source } from 'react-map-gl';
 import { MapSource, MapEntityId, MapEntityOptions, MapOptions, MapPopupRenderer, MapGeometry } from 'src/types/Map';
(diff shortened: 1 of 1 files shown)


src/components/Map/Map.tsx
```diff
@@ -211,6 +213,11 @@ export default function Map(props: MapProps): JSX.Element {
     [geoData]
   );
 
+  const zoomToFit = () => {
+    const map: any = mapRef?.current;
+    map.fitBounds([options.bbox.lowerLeft, options.bbox.upperRight], { padding: 20 });
```
karthikbtf, line 218: `map?.fitBounds(..)` ? since map can be undefined

src/components/Map/Map.tsx
```diff
@@ -456,6 +463,21 @@ export default function Map(props: MapProps): JSX.Element {
           {mapSources}
           <NavigationControl showCompass={false} style={navControlStyle} position='bottom-right' />
           <AttributionControl compact={true} style={{ marginRight: '5px' }} />
+          <Box
+            style={{
+              width: 27,
+              backgroundColor: `${theme.palette.TwClrBaseWhite}`,
+              position: 'absolute',
+              bottom: '30%',
```
karthikbtf, line 471: Curious.. do we know how the mobile app does this?

you might be able to use absolute pixels here, maybe? see what the height is up to the info button perhaps?

30% is not going to scale well in large screens (see image)

Terraware App 2023-07-27 16-49-49

karthikbtf (commented):
yay for feature \o/
some comments

src/components/Map/Map.tsx
constanzauanini: True, I just checked in mobile and it is done same way (with absolute positioning). Will replace it with pixels

karthikbtf (approved):
Looks decent to me.
There are references online about `useControl` - which is a hook to add custom map controls. Was unsuccessful in finding enough references/examples.
Perhaps for now this is ok, consistent with mobile app behavior.

constanzauanini:
> Looks decent to me. There are references online about `useControl` - which is a hook to add custom map controls. Was unsuccessful in finding enough references/examples. Perhaps for now this is ok, consistent with mobile app behavior.

Yes, I saw the useControl hook and tested last week, but couldn't find enough data to add the custom icon or execute the function. Will get this in like this for now

PR #1615, opened 2023-07-31
SW-3957 Fix odd zoom-to-fit behavior on map when switching between list and map views

karthikbtf:
- See Jira for video of behavior
- onRender hook was resizing on every render, which happens a lot and leads to some undesirable behavior
- added state in onRender to only resize the first time
- also added more height for the map in desktop mode, fill up remaining height in browser

Terraware App 2023-07-31 14-34-14

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -67,7 +67,8 @@ export default function Map(props: MapProps): JSX.Element {
   const hoverStateId: FeatureStateId = useMemo(() => ({}), []);
   const selectStateId: FeatureStateId = useMemo(() => ({}), []);
   const highlightStateId: FeatureStateId = useMemo(() => ({}), []);
-  const [firstVisible, setFirstVisible] = useState(false);
+  const [firstVisible, setFirstVisible] = useState<boolean>(false);
+  const [resized, setResized] = useState<boolean>(false);
   const visible = useIsVisible(containerRef);
   const snackbar = useSnackbar();
   const theme = useTheme();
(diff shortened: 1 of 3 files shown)


src/components/Map/Map.tsx
```diff
@@ -458,7 +459,12 @@ export default function Map(props: MapProps): JSX.Element {
             destroying = true;
           }}
           ref={mapRefCb}
-          onRender={(event) => event.target.resize()}
+          onRender={(event) => {
+            if (!resized) {
```
karthikbtf, line 463: This is the fix. @charlie-tf if this somehow breaks what you had originally intended we can follow up.

src/components/NurseryWithdrawals/NurseryPlantingsAndWithdrawals.tsx
```diff
@@ -18,11 +19,30 @@ import PlantingProgress from './PlantingProgressTabContent';
...
+const useStyles = makeStyles(() => ({
+  tabs: {
+    '& .MuiTabPanel-root[hidden]': {
+      flexGrow: 0,
+    },
+    '& .MuiTabPanel-root': {
+      display: 'flex',
+      flexDirection: 'column',
+      flexGrow: 1,
+    },
```
karthikbtf, line 31: this is to allow the tab content to use full height

alexTerraformation:
Increasing the height of the map should also improve the behavior noted in SW-3968 cc: @charlie-tf

src/components/Map/Map.tsx
charlie-tf: Just checked it out, and it appears to still work as expected 👍

charlie-tf (approved):
Looks good. Thank you for the fix!

src/components/Map/Map.tsx
karthikbtf: Great, thanks!!

PR #1618, opened 2023-08-01
SW-3996 Fix positioning of empty state message in Observations view

karthikbtf:
- fix distance from page title
- fix border radius
- fix height of content
- the layout here is a bit different from our older pages, hence different way to style

Terraware App 2023-08-01 11-34-12

Terraware App 2023-08-01 11-33-53


constanzauanini (approved):
Looks good!

PR #1621, opened 2023-08-01
SW-4003 Map planting site season start/end months to string keys

karthikbtf:
- instead of using date utils


charlie-tf (approved):
👍 lgtm

PR #1622, opened 2023-08-02
SW-4005 Fix plantings selectors math (skip the math)

karthikbtf:
- the total seedlings are already calculated by subzone

### DIFF
--- a/src/redux/features/plantings/plantingsSlice.ts
+++ b/src/redux/features/plantings/plantingsSlice.ts
@@ -8,7 +8,7 @@ export type PlantingSearchData = {
   delivery: { withdrawal_id: string };
   'numPlants(raw)': string;
   plantingSite: { id: string };
-  plantingSubzone?: { id: string; 'totalPlants(raw)': number };
+  plantingSubzone?: { id: string; totalPlants: string };
   species: { id: string };
   type: string;
 };
--- a/src/redux/features/plantings/plantingsSlice.ts
+++ b/src/redux/features/plantings/plantingsSlice.ts
@@ -8,7 +8,7 @@ export type PlantingSearchData = {
   delivery: { withdrawal_id: string };
   'numPlants(raw)': string;
   plantingSite: { id: string };
-  plantingSubzone?: { id: string; totalPlants: string };
+  plantingSubzone?: { id: string; 'totalPlants(raw)': number };
   species: { id: string };
   type: string;
 };
--- a/src/services/PlantingsService.ts
+++ b/src/services/PlantingsService.ts
@@ -21,7 +21,7 @@ const listPlantings = async (
       'delivery.plantings.id',
       'delivery.plantings.plantingSite.id',
       'delivery.plantings.plantingSubzone.id',
-      'delivery.plantings.plantingSubzone.totalPlants',
+      'delivery.plantings.plantingSubzone.totalPlants(raw)',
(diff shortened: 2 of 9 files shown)


src/redux/features/plantings/plantingsSlice.ts
```diff
@@ -8,7 +8,7 @@ export type PlantingSearchData = {
   delivery: { withdrawal_id: string };
   'numPlants(raw)': string;
   plantingSite: { id: string };
-  plantingSubzone?: { id: string; 'totalPlants(raw)': number };
+  plantingSubzone?: { id: string; totalPlants: string };
```
karthikbtf, line 11: using string value so it comes pre-formatted/localized.. we won't be using raw numbers for math.
Also, these types should maybe be moved into src/types?

src/services/PlantingsService.ts
```diff
@@ -21,7 +21,7 @@ const listPlantings = async (
       'delivery.plantings.id',
       'delivery.plantings.plantingSite.id',
       'delivery.plantings.plantingSubzone.id',
-      'delivery.plantings.plantingSubzone.totalPlants(raw)',
+      'delivery.plantings.plantingSubzone.totalPlants',
```
karthikbtf, line 24: using string value so it comes pre-formatted/localized.. we won't be using raw numbers for math

src/services/SearchService.ts
```diff
@@ -60,60 +59,6 @@ async function search(entity: SearchRequestPayload): Promise<SearchResponseEleme
   }
 }
 
-type SearchFieldType = {
```
karthikbtf, line 63: moved all this into TrackingService, SearchService was to be for more generic functions

src/types/Tracking.ts
```diff
@@ -10,7 +10,7 @@ export type PlantingSiteSearchResult = {
   name: string;
   numPlantingZones: number;
   numPlantingSubzones: number;
-  'totalPlants(raw)': number;
+  totalPlants: string;
```
karthikbtf, line 13: using string value so it comes pre-formatted/localized.. we won't be using raw numbers for math

src/services/PlantingsService.ts
karthikbtf: going to switch back to (raw) since the string values don't sort well in the table.. will use the FormatterNumber in the renderer instead

src/services/TrackingService.ts
```diff
@@ -230,6 +241,53 @@ const getReportedPlants = async (plantingSiteId: number): Promise<SiteReportedPl
...
 
+/**
+ * Search planting sites
+ */
+async function searchPlantingSites(
+  organizationId: number,
+  searchField?: SearchFieldType,
+  sortOrder?: SearchSortOrder
+): Promise<PlantingSiteSearchResult[] | null> {
+  const defaultSortOder = {
```
charlie-tf, line 252: Spelling: `defaultSortOrder`

charlie-tf (approved):
Looks great. Checked vercel and it fixes the issue I saw. One tiny spelling comment, though it was there before and you didn't introduce it.

src/services/TrackingService.ts
karthikbtf: good catch :) will fix it..

charlie-tf (approved):

PR #1625, opened 2023-08-02
SW-3998 Get exact matches for selected planting site on withdrawals page

karthikbtf:
- BE returns planting sites whose names contain the selected planting site name, eg. if selected site is 'site', responses will also include 'site1' if they exist
- [see](https://terraformation.atlassian.net/browse/SW-3070) for more context
- added another layer of filtering out non-exact names within client code

### DIFF
--- a/src/components/NurseryWithdrawals/NurseryReassignment.tsx
+++ b/src/components/NurseryWithdrawals/NurseryReassignment.tsx
@@ -132,7 +132,11 @@ export default function NurseryReassignment(): JSX.Element {
 
   const goToWithdrawals = () => {
     const withdrawalId = query.has('fromWithdrawal') ? delivery?.withdrawalId : undefined;
-    history.push({ pathname: APP_PATHS.NURSERY_WITHDRAWALS + (withdrawalId ? `/${withdrawalId}` : '') });
+    const search = trackingV2 && !withdrawalId ? 'tab=withdrawal_history' : '';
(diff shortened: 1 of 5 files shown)


constanzauanini (approved):
Looks good!

charlie-tf (approved):
Looks good. Tested on withdrawals history page and works as expected

karthikbtf:
> Looks good. Tested on withdrawals history page and works as expected

Thanks @charlie-tf 
The client side double filter is only because 'Exact' doesn't return exact matches, but also 'contained' matches. PR description has a link to a Jira that triggered the exact/contains behavior.
For now this should be good, unless the BE can return exact matches some other way.
There is a `destinationFacilityId`field in the table view but not sure that is actively populated, searching on that would be ideal. Perhaps a separate debug effort.

karthikbtf:
i have one more change to make related to nursery withdrawals details url generation in reassignment flow (since it was changed to `/withdrawals/details`).. coming up shortly

karthikbtf:
> i have one more change to make related to nursery withdrawals details url generation in reassignment flow (since it was changed to `/withdrawals/details`).. coming up shortly

this change is in now

karthikbtf:
Converting to draft since the selector approach may not be correct.

constanzauanini:
I think we can close this one @karthikbtf

PR #1627, opened 2023-08-04
SW-3932 Map view for simple planting sites

sgrimm:
For planting sites with site boundaries but without planting zones or subzones
(which we refer to as "simple" planting sites), show a map view that doesn't
mention zones and subzones.

There is currently no way to create a simple planting site from the UI; that
will be added in a later change.


sgrimm (commented):
I created a simple planting site on staging; it's in the "dev" org and is called "Simple Site" (ID 101).

When reviewing, it might be wise to pay special attention to my use of MUI components. My knowledge of them is still extremely basic and I may have misused them. In particular, I'm not sure if I'm doing padding/margins the right way.

charlie-tf (approved):
This looks reasonable to me. Is the SimplePlantingSiteMap meant to display user drawn maps? EDIT: read the PRD, so you can disregard this question.

PR #1629, opened 2023-08-07
SW-4010 Try Changing Dispatch to Improve Initial Map Load

charlie-tf:
In some cases users were seeing map tooltips with no data populated for plant counts on initial map load. This change moves the dispatch for that data to the containing component to see whether that causes the data to be in place at the correct time for those instances.


charlie-tf:
@alexTerraformation would you mind seeing whether you can reproduce https://terraformation.atlassian.net/browse/SW-4010 in the deploy here when you get a chance? Thanks.

alexTerraformation:
Yes, I can still reproduce the issue with that deploy.  It seems isolated to the withdrawals planting progress map.

charlie-tf:
> Yes, I can still reproduce the issue with that deploy. It seems isolated to the withdrawals planting progress map.

Dang, okay. Thanks for checking

PR #1630, opened 2023-08-07
Improve planting progress search

constanzauanini:
For org with no shape files, we were allowing to write the siteName in the search field to do a fuzzy search, but if a siteName is selected in filter, then just return the exacts coincidence with the siteName selected


charlie-tf (approved):
So if i understand correctly, this ignores the search term for siteName if there is a filter selected for the planting site?

constanzauanini:
> So if i understand correctly, this ignores the search term for siteName if there is a filter selected for the planting site?

Exactly, because if not it was searching also for sites with similar names when the filter was selected for a specific planting site

PR #1631, opened 2023-08-07
SW-4016 Do Not Add Actual Density Series if All Values are Null

charlie-tf:
This will avoid the case where an observation has been completed, but no zone has had all its subzones marked as complete, resulting in a series of `null`s in the actual planting density dataset.


constanzauanini (approved):
Sounds reasonable

PR #1632, opened 2023-08-07
SW-4017 Auto-Select First Species for Live/Dead Plants Chart

charlie-tf:
Before (on initial dashboard load):

After:
image


constanzauanini (approved):
💯

PR #1635, opened 2023-08-07
SW-4020 Whole Numbers for Live/Dead Plants per Species

charlie-tf:
- add Math.round() because only whole numbers make sense here


constanzauanini (approved):
Make sense 👍

PR #1638, opened 2023-08-08
SW-4009 Add % and plants/ha labels to per-zone progress/density charts

karthikbtf:
- also cleaned up the props a bit
- this is sorta wip but could be finished (waiting on comments in jira)

Terraware App 2023-08-08 11-27-36


constanzauanini (approved):
Looks good! Thanks for cleaning up the types

PR #1640, opened 2023-08-08
SW-3998 Filter down withdrawals results based on matching destination names

karthikbtf:
- 'Exact' search for destination names will also return any destination that contains the filter name, ex. 'test' will also return 'test1'
- Add another layer of filtering on destination names to avoid showing non-matching destinations

### Before

Terraware App 2023-08-08 12-27-40

### After

Terraware App 2023-08-08 12-27-12


constanzauanini (approved):
Oops, I think I missed this when I did it! Thanks!

PR #1647, opened 2023-08-09
SW-4045 Add support for InProgress and Overdue observations in the Observations page

karthikbtf:
- to satisfy PRD requirement allowing users to see which observations are not completed, check out zones, monitoring plots info, etc.

Terraware App 2023-08-09 11-05-37

Terraware App 2023-08-09 11-06-05

Terraware App 2023-08-09 11-06-45

### DIFF
--- a/src/components/Observations/ObservationsDataView.tsx
+++ b/src/components/Observations/ObservationsDataView.tsx
@@ -4,6 +4,7 @@ import strings from 'src/strings';
 import { useLocalization } from 'src/providers';
 import { FieldOptionsMap } from 'src/types/Search';
 import { ObservationState } from 'src/types/Observations';
+import { PlantingSite } from 'src/types/Tracking';
 import { useAppSelector } from 'src/redux/store';
 import { useDefaultTimeZone } from 'src/utils/useTimeZoneUtils';
 import { searchObservations, selectObservationsZoneNames } from 'src/redux/features/observations/observationsSelectors';
@@ -15,10 +16,11 @@ import ObservationMapView from './map/ObservationMapView';
 export type ObservationsDataViewProps = SearchProps & {
   setFilterOptions: (value: FieldOptionsMap) => void;
   selectedPlantingSiteId: number;
+  selectedPlantingSite?: PlantingSite;
 };
 
 export default function ObservationsDataView(props: ObservationsDataViewProps): JSX.Element {
-  const { selectedPlantingSiteId, setFilterOptions } = props;
+  const { selectedPlantingSiteId, selectedPlantingSite, setFilterOptions } = props;
   const { ...searchProps }: SearchProps = props;
   const defaultTimeZone = useDefaultTimeZone();
   const { activeLocale } = useLocalization();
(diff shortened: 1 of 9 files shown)


karthikbtf:
> Looks pretty good! Needs a couple fixes for the map view when there are no observations in the filter.



> Looks pretty good! Needs a couple fixes for the map view when there are no observations in the filter.



> Looks pretty good! Needs a couple fixes for the map view when there are no observations in the filter.

Thanks, made a change to show site boundary (minimum data view) when there are no observations to show.

karthikbtf:
@constanzauanini this one is ready to go if you have time

src/components/Observations/ObservationsDataView.tsx
```diff
@@ -13,44 +16,80 @@ import ObservationMapView from './map/ObservationMapView';
 export type ObservationsDataViewProps = SearchProps & {
   setFilterOptions: (value: FieldOptionsMap) => void;
   selectedPlantingSiteId: number;
```
constanzauanini, line 18: If we are now passing the selectedPlantingSite, we can get rid of this prop and get it from selectedPlantingSite

karthikbtf: The selected planting site may be undefined. The site id also captures the -1 “All” value

constanzauanini: I realized that after writing the comment haha

constanzauanini (approved):
Looks good to me!

karthikbtf:
> Looks good to me!

Thanks Constanza! I'll verify I didn't break anything in staging before I merge.

karthikbtf:
> > Looks good to me!
> 
> Thanks Constanza! I'll verify I didn't break anything in staging before I merge.

I made a couple changes to fix the 'latest observation'.

karthikbtf:
Merging this revert as it breaks withdrawals page if there are can't-tell/other species

karthikbtf:
> Merging this revert as it breaks withdrawals page if there are can't-tell/other species

oops wrong PR

src/redux/features/observations/observationsSelectors.ts
```diff
@@ -30,7 +30,7 @@ export const selectPlantingSiteObservationsResults = createCachedSelector(
       const matchesSite = plantingSiteId === -1 || observationResults.plantingSiteId === plantingSiteId;
       const matchesState =
         (!status?.length && observationResults.state === 'Completed') ||
-        status?.indexOf(observationResults.state) !== -1;
+        (status && status.indexOf(observationResults.state) !== -1);
```
constanzauanini, line 33: The "fix" here is the parentheses, right? Removing the '?' and doing a separate validation for status should not change anything I guess

karthikbtf: so if status is undefined, the check previously returns true (since undefined is !== -1)
that was wrong, we want to check for the indexOf result to be a number (not undefined), hence the `status && status.indexOf` fix

PR #1648, opened 2023-08-09
Revert "SW-4034 Show other/can't-tell species in observations bar charts (#1646)"

karthikbtf:
This reverts commit 96a6c2f05ea57191f5edfa61ed989895856323df.

Reverting this change, have a cleaner fix in another PR.


karthikbtf:
merging this revert as it breaks withdrawals when there are Other/Can't-tell species

PR #1650, opened 2023-08-09
SW-4052 Include critically endangered species in endangered tally bar chart (plants v2 dashboard)

karthikbtf:
- also check for 'CR' category


charlie-tf (approved):
Seems reasonable

PR #1653, opened 2023-08-11
SW-4056 Nursery inventory add species dropdown should be alphabetized

constanzauanini:
Sort species list alphabetically  

Screenshot 2023-08-11 at 14 22 38

### DIFF
--- a/src/components/accession2/properties/Species2Dropdown.tsx
+++ b/src/components/accession2/properties/Species2Dropdown.tsx
@@ -40,7 +40,7 @@ export default function Species2Dropdown<T extends AccessionPostRequestBody>(
             );
           })
         : response.species;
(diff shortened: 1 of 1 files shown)


charlie-tf (approved):
Approving. For very large lists, the `localeCompare` method can be costly for performance. If it becomes an issue we might need to revisit

joshbuchea:
I'm not sure if anyone has had to do this on the web, but on mobile we also ran into perf issues with `localeCompare` and we resolved it there by creating a new `Intl.Collator` instance and using that rather than `string.localeCompare`:

```js
const collator = new Intl.Collator(locales, options);
const sorted = unsorted.map.sort((a, b) => collator.compare(a.someProperty, b.someProperty));
```

PR #1654, opened 2023-08-11
SW-3993 Quantity not accurate with nursery filter

constanzauanini:
Allow searching inventories by different nurseries and getting its corresponding quantities. Previously we were getting the totals for all nurseries, not considering nursery filter.

Query executed to the backend changes depending on if nurseries are selected or not. From backend, response comes by species id, so sum all same species id quantities to show the number of each quantity for selected nurseries (and not total as before)

This PR requires a backend change: https://github.com/terraware/terraware-server/pull/1326

### DIFF
--- a/src/components/Inventory/index.tsx
+++ b/src/components/Inventory/index.tsx
@@ -1,6 +1,6 @@
 import { Box, CircularProgress, Container, Grid, Theme, Typography, useTheme } from '@mui/material';
 import { makeStyles } from '@mui/styles';
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import strings from 'src/strings';
 import EmptyMessage from 'src/components/common/EmptyMessage';
@@ -21,8 +21,9 @@ import NurseryInventoryService, { BE_SORTED_FIELDS } from 'src/services/NurseryI
 import ImportInventoryModal from './ImportInventoryModal';
 import PageHeaderWrapper from 'src/components/common/PageHeaderWrapper';
 import { Button, DropdownItem } from '@terraware/web-components';
-import { useOrganization } from 'src/providers/hooks';
+import { useOrganization, useUser } from 'src/providers/hooks';
 import OptionsMenu from 'src/components/common/OptionsMenu';
+import { useNumberFormatter } from 'src/utils/useNumber';
 
 interface StyleProps {
   isMobile: boolean;
@@ -67,6 +68,18 @@ type InventoryResult = {
   facilityInventories: FacilityName[];
 };
 
+type InventoryResultWithFacilityNames = Omit<InventoryResult, 'facilityInventories'> & { facilityInventories: string };
+
+type FacilityInventoryResult = {
+  facility_name: string;
+  species_id: string;
+  species_scientificName: string;
+  'germinatingQuantity(raw)': string;
+  'readyQuantity(raw)': string;
+  'notReadyQuantity(raw)': string;
+  'totalQuantity(raw)': string;
+};
+
 export default function Inventory(props: InventoryProps): JSX.Element {
   const { selectedOrganization } = useOrganization();
   const { isMobile } = useDeviceInfo();
@@ -85,6 +98,10 @@ export default function Inventory(props: InventoryProps): JSX.Element {
     direction: 'Ascending',
   });
   const contentRef = useRef(null);
+
+  const { user } = useUser();
+  const numberFormatter = useNumberFormatter();
+  const numericFormatter = useMemo(() => numberFormatter(user?.locale), [user?.locale, numberFormatter]);
 
   const goTo = (appPath: string) => {
     const appPathLocation = {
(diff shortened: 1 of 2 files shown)


src/components/Inventory/index.tsx
```diff
@@ -146,11 +163,64 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
+              Number(existingSpecies.notReadyQuantity) + Number(resultTyped['notReadyQuantity(raw)'])
+            ).toString(),
+            readyQuantity: (
+              Number(existingSpecies.readyQuantity) + Number(resultTyped['readyQuantity(raw)'])
+            ).toString(),
+            totalQuantity: (
+              Number(existingSpecies.totalQuantity) + Number(resultTyped['totalQuantity(raw)'])
+            ).toString(),
+            facilityInventories: `${existingSpecies.facilityInventories}, ${resultTyped.facility_name}`,
+          };
```
charlie-tf, line 191: In this branch, the `nextResults` aren't added to `updatedResult`. Is that a problem?

constanzauanini: Yes, we need the results in `updatedResult` to show results in the table. In this branch we add the `nextResults` that have the results when searching with facilityId filters, and in the other (the else) we add the apiSearchResults that have results when searching without facilityId filter applied

src/services/NurseryInventoryService.ts
```diff
@@ -131,22 +131,52 @@ const searchInventory = async ({
...
+          {
+            operation: 'field',
+            field: 'organization_id',
+            values: [organizationId],
+          },
+        ],
+      },
+      count: 0,
+    };
+  }
```
karthikbtf, line 179: suggestion, since the only difference between those params are the prefix and fields, we can maybe do something like this

```
  const params: SearchNodePayload = {
    prefix: forSpecificFacilities ? 'inventories.facilityInventories' : 'inventories',
    fields: forSpecificFacilities ? FACILITY_SPECIFIC_FIELDS : INVENTORY_FIELDS, (maybe extract them into a constant)
    sortOrder: searchSortOrder ? [searchSortOrder] : undefined,
    search: {
        operation: 'and',
        children: [
          {
            operation: 'field',
            field: 'organization_id',
            values: [organizationId],
          },
        ],
    },
    count: 0,
  };
```

Thoughts?

src/components/Inventory/index.tsx
```diff
@@ -146,11 +163,64 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
-    const updatedResult = apiSearchResults?.map((result) => {
-      const resultTyped = result as InventoryResult;
-      const facilityInventoriesNames = resultTyped.facilityInventories.map((nursery) => nursery.facility_name);
-      return { ...result, facilityInventories: facilityInventoriesNames.join('\r') };
-    });
+
+    let updatedResult: InventoryResultWithFacilityNames[] | undefined = [];
+    if (filters.facilityIds && filters.facilityIds.length) {
+      const nextResults: InventoryResultWithFacilityNames[] = [];
+      apiSearchResults?.forEach((result) => {
```
karthikbtf, line 170: this could be a `reduce`

src/components/Inventory/index.tsx
```diff
@@ -146,11 +163,64 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
+          nextResults.push(transformedResult);
+
+          // format results
+          updatedResult = nextResults.map((uR) => {
+            return {
+              ...uR,
+              germinatingQuantity: numericFormatter.format(uR.germinatingQuantity),
+              notReadyQuantity: numericFormatter.format(uR.notReadyQuantity),
+              readyQuantity: numericFormatter.format(uR.readyQuantity),
+              totalQuantity: numericFormatter.format(uR.totalQuantity),
```
karthikbtf, line 212: Thoughts on having the rendering component use the NumberFormatter component? Will simplify this file and keep semantics of number formatting in the right place. If this is to re-use existing component without further changes, this is fine as-is. Maybe use a reduce().map()

src/components/Inventory/index.tsx
```diff
@@ -146,11 +163,64 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
+            notReadyQuantity: (
+              Number(existingSpecies.notReadyQuantity) + Number(resultTyped['notReadyQuantity(raw)'])
+            ).toString(),
+            readyQuantity: (
+              Number(existingSpecies.readyQuantity) + Number(resultTyped['readyQuantity(raw)'])
+            ).toString(),
+            totalQuantity: (
+              Number(existingSpecies.totalQuantity) + Number(resultTyped['totalQuantity(raw)'])
+            ).toString(),
+            facilityInventories: `${existingSpecies.facilityInventories}, ${resultTyped.facility_name}`,
```
karthikbtf, line 190: Just curious, can a facility show up multiple times here?
Suggestion, keep this as an array (or Set) of facility names and do a `join(',')` in the final updatedResults.map.

karthikbtf (commented):
looks good, some comments/thoughts

src/components/Inventory/index.tsx
constanzauanini: Each result comes by species and facilities, so if the species already exists, it's because that facility was already added, and we are in other combination of species-facility. So this would not happen, but I will create a set so it's clearer

karthikbtf: Ok, if it is guaranteed to be unique don’t worry. Should we sort the array?

src/components/Inventory/index.tsx
karthikbtf: Got it, leave it as-is 👍

src/components/Inventory/index.tsx
karthikbtf: Np, let’s keep them the same and maybe revisit later.

src/components/Inventory/index.tsx
```diff
@@ -146,11 +163,65 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
+            notReadyQuantity: resultTyped['notReadyQuantity(raw)'],
+            readyQuantity: resultTyped['readyQuantity(raw)'],
+            totalQuantity: resultTyped['totalQuantity(raw)'],
+            facilityInventories: resultTyped.facility_name,
+          };
+
+          acc.push(transformedResult);
+        }
+        return acc;
+      }, nextResults);
```
karthikbtf, line 206: Thanks for the changes Constanza. Any chance you can revisit the reduce?
Ideally reduce pattern is 
const xyz =  arr.reduce(..., initialValue);

What you have here may still work but is not ideal, the acc and initial value are references to the original constant in your case, not passed around by value.

src/services/NurseryInventoryService.ts
```diff
@@ -131,9 +144,11 @@ const searchInventory = async ({
...
   query,
 }: SearchInventoryParams): Promise<SearchResponseElement[] | null> => {
-  const params: SearchNodePayload = {
-    prefix: 'inventories',
-    fields: [...BE_SORTED_FIELDS, 'species_commonName', 'totalQuantity(raw)'],
+  let params: SearchNodePayload;
+  const forSpecificFacilities = !!facilityIds && !!facilityIds.length;
+  params = {
+    prefix: forSpecificFacilities ? 'inventories.facilityInventories' : 'inventories',
+    fields: forSpecificFacilities ? FACILITY_SPECIFIC_FIELDS : INVENTORY_FIELDS,
```
karthikbtf, line 151: thanks much cleaner

karthikbtf (approved):
Almost there, just one nit fix request.
Feel free to merge after that.

src/components/Inventory/index.tsx
constanzauanini: Thanks, I was replacing the for each with the reduce and didn't realize I was doing that!

karthikbtf (approved):
🎉

PR #1664, opened 2023-08-15
SW-4050 Fix overlapping bars in reported plants by species chart

karthikbtf:
- use 'flex' bar width
- bar charts are not updating with data changes, using a key property to re-create the bar chart when we have update values, fixes the y-values race condition of creating bar chart before we have all the data

Terraware App 2023-08-15 16-48-34

### DIFF
--- a/src/components/PlantsV2/components/PlantsReportedPerSpeciesCard.tsx
+++ b/src/components/PlantsV2/components/PlantsReportedPerSpeciesCard.tsx
@@ -81,18 +81,17 @@ export default function PlantsReportedPerSpeciesCard({
             {strings.REPORTED_PLANTS_PER_SPECIES_CARD_TITLE}
           </Typography>
           <Box>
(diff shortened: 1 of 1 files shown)


src/components/PlantsV2/components/PlantsReportedPerSpeciesCard.tsx
```diff
@@ -81,13 +82,15 @@ export default function PlantsReportedPerSpeciesCard({
           </Typography>
           <Box>
             <BarChart
+              key={values?.length}
```
constanzauanini, line 85: Could it happen that data changes but not the length?

karthikbtf: Not the way it is setup today. It's a one time read, data is not refreshed or updated after that.

constanzauanini (approved):

PR #1671, opened 2023-08-17
SW-3932 Expand simple planting site map

sgrimm:
Make the simple planting site map grow to fill the viewport if
the window is tall, rather than using a fixed height.


constanzauanini (approved):
Maps expands correctly now! Thanks!!

PR #1673, opened 2023-08-17
SW-3932 Use simple/detailed site types in dialog

sgrimm:
Update the wording of the "select a planting site type" dialog to refer to
planting sites as "simple" and "detailed" rather than as "without map" and
"with map" to reflect the fact that we will soon allow simple planting
sites to have maps.


constanzauanini (approved):
Working correctly!

PR #1677, opened 2023-08-22
SW-4092 Show mortality rate only when there are observed permanent plots

karthikbtf:
- capture this state in the model
- leverage that in components

Terraware App 2023-08-22 15-13-52


constanzauanini (approved):
Nice!

PR #1678, opened 2023-08-22
SW-4096 Use latest observations data in plants dashboard

karthikbtf:
- we were sorting by `completedDate` (yyyy-mm-dd format), there were 2 observations on that date, the earlier one was used
- sort by `completedTime` instead which has millisecond level granularity

### DIFF
--- a/src/redux/features/observations/observationsSelectors.ts
+++ b/src/redux/features/observations/observationsSelectors.ts
@@ -9,7 +9,6 @@ import { RootState } from 'src/redux/rootReducer';
 import { selectSpecies } from 'src/redux/features/species/speciesSelectors';
 import { selectPlantingSites } from 'src/redux/features/tracking/trackingSelectors';
 import { mergeObservations, searchZones } from './utils';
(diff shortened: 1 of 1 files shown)


sgrimm:
The server always returns observation results in reverse completed-time order, so you should be able to always use the first one from the API response. (The order is part of the API contract, not just an implementation detail.)

sgrimm:
If all you want is the latest results, you can include `limit=1` in the query string and it will give you back the results of the most recently completed observation (so you don't have to download the details of older ones).

karthikbtf:
> If all you want is the latest results, you can include `limit=1` in the query string and it will give you back the results of the most recently completed observation (so you don't have to download the details of older ones).

Thanks for the info, Steve.
The app fetches observation results for the entire org and keeps that in the app redux store, after a bit of post-processing (merging planting zone, subzone names, etc) - does not use the search API.
Code in this PR deals with data at the other end of post-processing/searching of in-memory redux observations state. I suppose the server generated order is still maintained, but could break if something is changed in between.

constanzauanini (approved):

PR #1680, opened 2023-08-22
SW-4080 Use consistent styling for photos thumbnail viewer

karthikbtf:
- Port styles from photo uploader in web-components
- Updated PhotosList component in terraware web
- Reused PhotosList in withdrawals and accessions (already being used in observations)
- Updated Reports photos thumbnails to be consistent (could not use PhotosList here due to custom behavior, editing)
- Also fixed the back-to link from withdrawals details to go back to withdrawal history

Terraware App 2023-08-22 16-20-21

Terraware App 2023-08-22 16-05-01

Terraware App 2023-08-22 15-49-41

Terraware App 2023-08-22 15-48-02

### DIFF
--- a/src/components/NurseryWithdrawals/NurseryReassignment.tsx
+++ b/src/components/NurseryWithdrawals/NurseryReassignment.tsx
@@ -222,7 +222,7 @@ export default function NurseryReassignment(): JSX.Element {
         <Box>
           <BackToLink
             id='back'
-            to={APP_PATHS.NURSERY_WITHDRAWALS}
+            to={`${APP_PATHS.NURSERY_WITHDRAWALS}?tab=withdrawal_history`}
             className={classes.backToWithdrawals}
             name={strings.WITHDRAWAL_LOG}
           />
--- a/src/components/Reports/ViewPhotos.tsx
+++ b/src/components/Reports/ViewPhotos.tsx
@@ -6,6 +6,7 @@ import { Button, ViewPhotosDialog } from '@terraware/web-components';
 import { makeStyles } from '@mui/styles';
 import { ReportPhoto } from 'src/types/Report';
 import strings from 'src/strings';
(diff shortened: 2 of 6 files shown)


chudo76 (approved):
Looks great, thanks!

src/components/NurseryWithdrawals/NurseryReassignment.tsx
```diff
@@ -222,7 +222,7 @@ export default function NurseryReassignment(): JSX.Element {
         <Box>
           <BackToLink
             id='back'
-            to={APP_PATHS.NURSERY_WITHDRAWALS}
+            to={`${APP_PATHS.NURSERY_WITHDRAWALS}?tab=withdrawal_history`}
```
constanzauanini, line 225: Should we add it only if tracking enabled? I think it doesn't break adding it though

src/components/Reports/ViewPhotos.tsx
```diff
@@ -80,15 +89,16 @@ export default function ViewPhotos({ reportId, onPhotoRemove, editable }: Photos
...
+      <Box display='flex' flexWrap='wrap' flexDirection='row'>
         {photos.map((photo, index) => (
           <Box
             key={index}
+            display='flex'
             position='relative'
-            marginRight={theme.spacing(3)}
-            marginTop={theme.spacing(2)}
-            maxWidth='400px'
-            sx={{ cursor: 'pointer' }}
```
constanzauanini, line 91: I think we should keep the cursor pointer to know the photo is clickable

constanzauanini (approved):
Some comments but looks perfect!

src/components/Reports/ViewPhotos.tsx
karthikbtf: oops,, will add it back, had added it into the thumbnail classes instead but that only covers the image and not the enclosing box.

constanzauanini (approved):

PR #1683, opened 2023-08-25
SW-3740 Remove recoil usage from species list page

constanzauanini:
Recoil was used to show the review errors modal in species list when coming from the empty state.
Remove recoil usage and replace it using a query param to show the modal in those cases.

### DIFF
--- a/src/components/Species/index.tsx
+++ b/src/components/Species/index.tsx
@@ -1,12 +1,10 @@
 import { Container, Grid, Popover, Theme } from '@mui/material';
 import { makeStyles } from '@mui/styles';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { useRecoilState } from 'recoil';
 import Button from 'src/components/common/button/Button';
 import EmptyMessage from 'src/components/common/EmptyMessage';
 import { OrderPreserveableTable as Table } from 'src/components/common/table';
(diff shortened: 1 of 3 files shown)


src/components/Species/index.tsx
```diff
@@ -498,11 +500,13 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
   }, [columns, setSelectedColumns, activeLocale]);
 
   useEffect(() => {
-    if (speciesState?.checkData) {
-      setSpeciesState({ checkData: false });
+    const shouldCheckData = query.get('checkData');
```
karthikbtf, line 503: if `checkData` is `false` we will still check data. My suggestion is to use a query param with no value.
Here's an example: https://github.com/terraware/terraware-web/blob/main/src/components/NurseryWithdrawals/NurseryReassignment.tsx#L134

karthikbtf (approved):
nice, good to get rid of recoil.. one suggestion but up to you

src/components/Species/index.tsx
constanzauanini: Why will we still check data? If shouldCheckData = false ?

karthikbtf: false is a string not boolean

karthikbtf (approved):
Thanks!

PR #1684, opened 2023-08-25
SW-3740 Remove Recoil Usage for Snackbars

constanzauanini:
- Replace recoil usage with redux implementation.
- Copy snackbar component/selectors/slices/utils from funder-web.


charlie-tf (approved):
Nice work. Seems to work great in Vercel. Thanks for doing this :)

PR #1685, opened 2023-08-25
SW-3740 Remove recoil usage from database

constanzauanini:
In accessions list, recoil was being used when importing accessions, to store the facility selected on modal.
Remove recoil and just save it in local state


karthikbtf (approved):
Looks safe enough I think 👍

PR #1686, opened 2023-08-28
SW-3740 Remove recoil usage from user provider

constanzauanini:


src/providers/UserProvider.tsx
```diff
@@ -40,8 +39,8 @@ export default function UserProvider({ children }: UserProviderProps): JSX.Eleme
 
       if (response.requestSucceeded) {
         setUser(response.user!);
-        if (response.user && !userState?.gtmInstrumented && (window as any).INIT_GTAG) {
-          setUserState({ gtmInstrumented: true });
+        if (response.user && !store.getState().user?.gtmInstrumented && (window as any).INIT_GTAG) {
```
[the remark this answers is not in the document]
karthikbtf: useAppSelector to select the state of the store. Create a selector under redux/features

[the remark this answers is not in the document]
constanzauanini: I tried doing that, but I was getting an error of using outside provider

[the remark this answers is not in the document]
karthikbtf: Can you paste the error here? Not sure I understand the error.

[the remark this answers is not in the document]
karthikbtf: this is why https://github.com/terraware/terraware-web/blob/main/src/App.tsx#L564
the store provider is not visible to user provider (which is inside app bootstrap), we may need to switch the order of AppBootstrap and store provider here, or not use redux for this (maybe a global variable?)

src/App.tsx
```diff
@@ -561,10 +561,10 @@ function AppContent() {
 
 export default function App(): JSX.Element {
   return (
-    <AppBootstrap>
-      <Provider store={store}>
+    <Provider store={store}>
```
karthikbtf, line 564: Let's test and make sure there are no side effects from this?

src/redux/features/user/userSelectors.ts
```diff
@@ -0,0 +1,3 @@
+import { RootState } from 'src/redux/rootReducer';
+
+export const selectUser = (state: RootState) => state.user;
```
karthikbtf, line 3: can we call this something other than selectUser? We have a concept of a user in the app and it is not this. Call it maybe selectUserAnalytics? and state.userAnalytics?

src/redux/features/user/userSlice.ts
```diff
@@ -0,0 +1,24 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+type UserState = {
```
karthikbtf, line 3: UserAnalyticsState

src/redux/features/user/userSlice.ts
```diff
@@ -0,0 +1,24 @@
...
+  gtmInstrumented: boolean;
+};
+
+const initialState: UserState = {
+  gtmInstrumented: false,
+};
+
+export const userSlice = createSlice({
+  name: 'user',
```
karthikbtf, line 12: userAnalytics

src/redux/features/user/userSlice.ts
```diff
@@ -0,0 +1,24 @@
...
+
+type UserState = {
+  gtmInstrumented: boolean;
+};
+
+const initialState: UserState = {
+  gtmInstrumented: false,
+};
+
+export const userSlice = createSlice({
```
karthikbtf, line 11: userAnalyticsSlice

src/redux/features/user/userSlice.ts
```diff
@@ -0,0 +1,24 @@
...
+  name: 'user',
+  initialState,
+  reducers: {
+    updateGtmInstrumented: (state, action: PayloadAction<UserState>) => {
+      const userState = action.payload;
```
karthikbtf, line 16: userAnalyticsState

src/redux/features/user/userSlice.ts
```diff
@@ -0,0 +1,24 @@
...
+      state.gtmInstrumented = userState.gtmInstrumented;
+    },
+  },
+});
+
+export const { updateGtmInstrumented } = userSlice.actions;
+
+export const userReducer = userSlice.reducer;
```
karthikbtf, line 24: userAnalyticsReducer

src/redux/rootReducer.ts
```diff
@@ -36,6 +37,7 @@ export const reducers = {
   siteReportedPlantsResults: siteReportedPlantsReducer,
   snackbar: snackbarReducer,
   message: messageReducer,
+  user: userReducer,
```
karthikbtf, line 40: userAnalytics: ...

karthikbtf (commented):
Looks good, I missed some naming issues earlier. We should be good after some renaming.

src/App.tsx
constanzauanini: Looks like there are not

karthikbtf (approved):
Nice 🎉

src/providers/UserProvider.tsx
```diff
@@ -40,8 +41,8 @@ export default function UserProvider({ children }: UserProviderProps): JSX.Eleme
 
       if (response.requestSucceeded) {
         setUser(response.user!);
-        if (response.user && !userState?.gtmInstrumented && (window as any).INIT_GTAG) {
-          setUserState({ gtmInstrumented: true });
+        if (response.user && !userAnalyticsState?.gtmInstrumented && (window as any).INIT_GTAG) {
+          store.dispatch(updateGtmInstrumented({ gtmInstrumented: true }));
```
karthikbtf, line 45: oops missed this, useAppDispatch here

constanzauanini: Oh, yes! I should be able to do it now with the <Provider> change. Thanks

src/providers/UserProvider.tsx
```diff
@@ -40,8 +42,8 @@ export default function UserProvider({ children }: UserProviderProps): JSX.Eleme
 
       if (response.requestSucceeded) {
         setUser(response.user!);
-        if (response.user && !userState?.gtmInstrumented && (window as any).INIT_GTAG) {
-          setUserState({ gtmInstrumented: true });
+        if (response.user && !userAnalyticsState?.gtmInstrumented && (window as any).INIT_GTAG) {
+          dispatch(updateGtmInstrumented({ gtmInstrumented: true }));
```
karthikbtf, line 46: one last request, can we do an `await` on this? I'm guessing the previous recoil operation was not async. Here we don't want to initialize Google Analytics twice, if dispatch is async the code can theoretically get called twice (though it may not in the real world).

karthikbtf (approved):
LGTM! One last request.
Btw, might be good to test these changes against google analytics as well, just to be thorough. This can be done locally using a GA chrome tool or in staging after we merge the PR. We want to make sure the code of initializing GA is not called twice.

karthikbtf (approved):

PR #1687, opened 2023-08-29
SW-3740 Remove Recoil dependency

constanzauanini:


src/index.tsx
```diff
@@ -13,24 +12,22 @@ import { ThemeProvider } from '@mui/material';
 
 ReactDOM.render(
   <React.StrictMode>
-    <RecoilRoot>
```
constanzauanini, line 16: Here, I just removed this

karthikbtf:
Nice!! I think we need to remove Recoil from storybook setup as well.

PR #1689, opened 2023-08-30
Add renovate.json

constanzauanini:


karthikbtf (approved):
I think we already have something setup for terraware web in Github, this may give us double upgrade PRs?

renovate.json
```diff
@@ -0,0 +1,4 @@
+{
```
karthikbtf, line 1: see web-components, it has a different structure that @charlie-tf came up with

constanzauanini:
> I think we already have something setup for terraware web in Github, this may give us double upgrade PRs?

I don't think so, renovate doc says to create the renovate.json file to enable it if the first "automatic" pr created by renovate was not merged

karthikbtf:
> > I think we already have something setup for terraware web in Github, this may give us double upgrade PRs?
> 
> I don't think so, renovate doc says to create the renovate.json file to enable it if the first "automatic" pr created by renovate was not merged

We have something other than renovate setup. We used to get upgrade PRs for tw. We can disable that layer maybe.

karthikbtf (approved):

PR #1690, opened 2023-08-31
SW-4124 Move Mapbox info pop-up to the top

constanzauanini:
When screen is small (some mobiles), the info text wraps into two lines and it also shows under the buttons. To fix this, move the info botton to the top

Screenshot 2023-08-31 at 12 45 50


charlie-tf (approved):
Tested in vercel. Looks great 🗺️

PR #1713, opened 2023-09-20
SW-4240 Add 'Terraformation Contact' semantics in FE

karthikbtf:
- show TF contact role string value in people table if there is such a user
- disallow deletion/edit of TF contact
- adding TF Contact as a role is disallowed by default since we hard-code the roles and TF Contact was not there
- note: these capture the basic requirements for TF contact jira, there may be a follow up task around UX (jira created for design)
- this PR is dependent on BE PRs https://github.com/terraware/terraware-server/pull/1370 & https://github.com/terraware/terraware-server/pull/1372
- another outstanding issue is how delete org works when there is a Tf contact, this may be unrelated to FE and completely handled with BE but currently not done
- fixed another issue where we show a warning when all removable users are selected (the org's total users includes non humans, such as devices, the comparison failed in most cases, fixed that)

Terraware App 2023-09-20 15-36-30

Terraware App 2023-09-20 15-18-30

Terraware App 2023-09-20 15-16-35

### DIFF
--- a/src/components/People/index.tsx
+++ b/src/components/People/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import Button from 'src/components/common/button/Button';
 import Table from 'src/components/common/table';
@@ -25,7 +25,8 @@ import useDebounce from 'src/utils/useDebounce';
 import { SearchNodePayload } from 'src/types/Search';
 import { SearchService } from 'src/services';
 import { getRequestId, setRequestId } from 'src/utils/requestsId';
-import { useUser, useOrganization, useLocalization } from '../../providers/hooks';
+import { useUser, useOrganization, useLocalization } from 'src/providers/hooks';
+import { isTfContact } from 'src/utils/organization';
 
 const useStyles = makeStyles((theme: Theme) => ({
   title: {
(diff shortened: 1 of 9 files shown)


src/components/People/index.tsx
```diff
@@ -155,7 +177,8 @@ export default function PeopleList(): JSX.Element {
 
   const removeSelectedPeopleFromOrg = async () => {
     if (selectedOrganization) {
-      if (selectedPeopleRows.length === selectedOrganization.totalUsers) {
+      const removableUsers = selectedPeopleRows.filter((row) => !isTfContact(row.role));
+      if (removableUsers.length === totalUsers) {
```
constanzauanini, line 181: Does totalUsers include TfContacts? This will never be the same is removableUsers remove the TfContacts?

src/components/People/index.tsx
```diff
@@ -263,6 +286,11 @@ export default function PeopleList(): JSX.Element {
     setTemporalSearchValue(value as string);
   };
 
+  const hasRemovableUsers = useMemo(
+    () => selectedPeopleRows.some((row) => row.role !== 'Terraformation Contact'),
```
constanzauanini, line 290: Could use !isTfContact(row.role) ?

src/components/People/index.tsx
karthikbtf: does not include tfContacts (will comment)

src/components/People/index.tsx
```diff
@@ -117,11 +119,12 @@ export default function PeopleList(): JSX.Element {
         params.search.children.push(searchField);
       }
 
-      const requestId = Math.random().toString();
-      setRequestId('searchUsers', requestId);
       const searchResults = await SearchService.search(params);
       const usersResults: OrganizationUser[] = [];
       searchResults?.forEach((result) => {
+        if (skipTfContact && isTfContact(result.roleName as OrganizationRole)) {
```
karthikbtf, line 125: @constanzauanini here we ignore tf contacts

src/components/People/index.tsx
```diff
@@ -139,7 +152,16 @@ export default function PeopleList(): JSX.Element {
     if (activeLocale) {
       refreshSearch();
     }
-  }, [debouncedSearchTerm, selectedOrganization, activeLocale]);
+  }, [debouncedSearchTerm, search, activeLocale]);
+
+  useEffect(() => {
+    const findTotalUsers = async () => {
+      const users = await search('', true);
```
karthikbtf, line 159: @constanzauanini here we fetch all users and skip tf contacts, to get total users

src/components/People/index.tsx
karthikbtf: good catch, will update

src/components/People/index.tsx
constanzauanini: Oh, ok, I missed that "true". Thanks!

karthikbtf:
This one is ready to go. I may have a followup PR later (if design accepts the disabled checkbox web-components PR)

constanzauanini (approved):

PR #1714, opened 2023-09-21
SW-4239 Disable 'Terraformation Contact' user selection in org users table

karthikbtf:
- also show tooltip that user is readonly
- undo some earlier code that assumed TF Contact user could be selected and show up in the selected rows array

Terraware App 2023-09-21 16-43-25

Screen Shot 2023-09-21 at 4 43 08 PM


karthikbtf:
@chudo76 lmk thoughts here

constanzauanini (approved):

karthikbtf:
redoing with different approach

PR #1715, opened 2023-09-22
SW-4249 Add plant monitoring details to quarterly partner reports

constanzauanini:
Screenshot 2023-09-25 at 12 07 07

### DIFF
--- a/src/components/Reports/LocationSection.tsx
+++ b/src/components/Reports/LocationSection.tsx
@@ -220,8 +220,8 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
       return (
         (
           <Box>
-            {zoneNameWithDensities?.map((zd) => (
-              <Box>{zd}</Box>
+            {zoneNameWithDensities?.map((zd, index) => (
+              <Box key={`zone-${index}`}>{zd}</Box>
             ))}
           </Box>
         ) || ''
--- a/src/components/Reports/LocationSection.tsx
+++ b/src/components/Reports/LocationSection.tsx
@@ -1,6 +1,6 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
-import { Grid, Theme, Typography, useTheme } from '@mui/material';
+import { Box, Grid, Theme, Typography, useTheme } from '@mui/material';
 import { DatePicker, TableColumnType, Textfield } from '@terraware/web-components';
 import strings from 'src/strings';
 import OverviewItemCard from 'src/components/common/OverviewItemCard';
@@ -113,14 +113,14 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
   const plantingSite = useAppSelector((state) => selectPlantingSite(state, location.id));
 
   useEffect(() => {
-    if (selectedOrganization) {
+    if (selectedOrganization && trackingV2) {
       dispatch(requestObservations(selectedOrganization.id));
       dispatch(requestObservationsResults(selectedOrganization.id));
       dispatch(requestSpecies(selectedOrganization.id));
       dispatch(requestPlantings(selectedOrganization.id));
       dispatch(requestPlantingSitesSearchResults(selectedOrganization.id));
     }
-  }, [dispatch, selectedOrganization]);
+  }, [dispatch, selectedOrganization, trackingV2]);
 
   useEffect(() => {
     if (plantingSite) {
@@ -185,21 +185,21 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
     }
   };
 
-  const getLivePlants = () => {
+  const getLivePlants = useMemo(() => {
     return latestObservation?.species.reduce((acc, sp) => (acc = acc + sp.permanentLive), 0);
-  };
+  }, [latestObservation]);
 
-  const getDeadPlants = () => {
+  const getDeadPlants = useMemo(() => {
     return latestObservation?.species.reduce((acc, sp) => (acc = acc + sp.cumulativeDead), 0);
-  };
+  }, [latestObservation]);
 
-  const getNumberOfPlots = () => {
+  const getNumberOfPlots = useMemo(() => {
     return latestObservation?.plantingZones.flatMap((pz) =>
       pz.plantingSubzones.flatMap((subzone) => subzone.monitoringPlots)
     ).length;
-  };
+  }, [latestObservation]);
 
-  const getMarkedAsComplete = () => {
+  const getMarkedAsComplete = useMemo(() => {
     if (plantingSite) {
       const totalArea = plantingSite.areaHa ?? 0;
       const totalPlantedArea =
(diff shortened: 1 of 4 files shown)


src/components/Reports/LocationSection.tsx
```diff
@@ -355,6 +417,108 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
...
+                  title={strings.PLANTING_PROGRESS_PERCENT}
+                  contents={reportedPlants?.progressPercent || ''}
+                  className={classes.infoCardStyle}
+                />
+              </Grid>
+              <Grid item xs={smallItemGridWidth()}>
+                <OverviewItemCard
+                  isEditable={false}
+                  title={strings.PLANTING_DENSITY_OF_PLANTED_ZONES}
+                  contents={''}
```
constanzauanini, line 508: Missing this value, not sure what we want to display

src/strings/csv/en.csv
```diff
@@ -311,6 +313,7 @@ ENTER_SIX_DIGIT_KEY,Enter 6-digit key
 ENVIRONMENTAL_NOTES,Environmental Notes
 ENVIRONMENTAL_NOTES_PLACEHOLDER,"Important notes about landscape, climate, and environmental conditions"
 EST_READY_DATE,Est. Ready Date
+EST_TOTAL_PLANTS_PLANTING_DENSITY_AREA,Est. Total Plants (Planting Density × Area)
```
karthikbtf, line 316: not sure if translation will know what `Est.` is, should we expand this fully `Estimated`? or add a hint after a comma
`Est. Total Plants (Planting Density x Area), Estimated Total Plants`

src/components/Reports/LocationSection.tsx
karthikbtf: Observation results payload has planting progress and planting density. Not sure what `reportedPlants` is in this change. Should it be the observation results instead?

constanzauanini: I don't see the plantingProgress in the observationResults payload, that's why I got it from reportedPlants, as we do in the PlantingSiteProgressCard in the dashboard

constanzauanini: And for density we want to show the value per zone, so I'm adding same logic as we do in the PlantingDensityPerZoneCard

karthikbtf: > And for density we want to show the value per zone, so I'm adding same logic as we do in the PlantingDensityPerZoneCard

Possible I don't understand the requirements. Density will be different for different zones. The density per zone card shows barchart for various zones. Here it is just one number, for the most recent observation result perhaps. Not sure it's the same as planting density per zone use-case? There's a density for the observation result, which is populated only when all subzones are marked as completed - if that's what you want.

karthikbtf: > I don't see the plantingProgress in the observationResults payload, that's why I got it from reportedPlants, as we do in the PlantingSiteProgressCard in the dashboard

Got it, I see it in the generated types for reported plants api.

constanzauanini: > > And for density we want to show the value per zone, so I'm adding same logic as we do in the PlantingDensityPerZoneCard
> 
> Possible I don't understand the requirements. Density will be different for different zones. The density per zone card shows barchart for various zones. Here it is just one number, for the most recent observation result perhaps. Not sure it's the same as planting density per zone use-case? There's a density for the observation result, which is populated only when all subzones are marked as completed - if that's what you want.

This is not one value, is a list of <zone: value> (you can check in design) . Going to update the code

karthikbtf: Ah ok got it. I didn't see the designs. Update the screenshot here if you can?

src/components/Reports/LocationSection.tsx
```diff
@@ -133,6 +185,43 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
...
+  const getPlantingDensityForZones = () => {
+    if (plantingSite && plantingDensity) {
+      const zoneNameWithDensities = plantingSite.plantingZones?.map(
+        (zone) => `${zone.name}: ${plantingDensity[zone.name]}`
+      );
+      return zoneNameWithDensities?.join('\r\n') || '';
+    }
+    return '';
+  };
+
```
karthikbtf, line 224: if it makes sense, use `useMemo` for each of these? Will make the page less busy, it is a busy page.

src/components/Reports/LocationSection.tsx
```diff
@@ -82,6 +96,44 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
...
+    selectNextObservation(state, location.id, defaultTimeZone.get().id)
+  );
+  const latestObservation = useAppSelector((state) =>
+    selectLatestObservation(state, location.id, defaultTimeZone.get().id)
+  );
+  const reportedPlants = useAppSelector((state) => selectSiteReportedPlants(state, location.id));
+  const plantingSite = useAppSelector((state) => selectPlantingSite(state, location.id));
+
+  useEffect(() => {
+    if (selectedOrganization) {
```
karthikbtf, line 116: do we need a trackingV2 flag here?

src/components/Reports/LocationSection.tsx
```diff
@@ -133,6 +185,43 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
...
+    }
+    return '0%';
+  };
+
+  const getPlantingDensityForZones = () => {
+    if (plantingSite && plantingDensity) {
+      const zoneNameWithDensities = plantingSite.plantingZones?.map(
+        (zone) => `${zone.name}: ${plantingDensity[zone.name]}`
+      );
+      return zoneNameWithDensities?.join('\r\n') || '';
```
karthikbtf, line 220: curious why both \r and \n or should we do this as html `zoneNameWithDensities.map(x => <span>{x}</span>)` ?

karthikbtf (approved):
Some comments / suggestions, otherwise lgtm. If possible get Charlie's review as well.

src/components/Reports/LocationSection.tsx
constanzauanini: I read that '\r\n is often used in preference to \n as it displays properly on both unix and Windows. Carriage Return ( \r ) just returns the cursor to the beginning of the same line (without advancing to the next line) whereas a Line Feed ( \n ) feeds a new line. ' It could be also a span, I can update it to that

src/components/Reports/LocationSection.tsx
```diff
@@ -133,6 +185,51 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
...
+          <Box>
+            {zoneNameWithDensities?.map((zd, index) => (
+              <Box key={`zone-${index}`}>{zd}</Box>
+            ))}
+          </Box>
+        ) || ''
+      );
+    }
+    return '';
+  }, [plantingSite, plantingDensity]);
```
karthikbtf, line 231: Thanks for the useMemo change 🙌 , one last suggestion, rename these by removing `get` prefix since they are no longer functions. Should be good to merge after that.

karthikbtf (approved):
🎉

karthikbtf (approved):
Nice 👍

PR #1716, opened 2023-09-22
SW-4250 Fix plant dashboard cards for small widths

constanzauanini:


src/components/PlantsV2/components/HectaresPlantedCard.tsx
```diff
@@ -38,10 +38,10 @@ export default function HectaresPlantedCard({ plantingSiteId }: HectaresPlantedC
       isEditable={false}
       contents={
         <Box display='flex' flexDirection='column'>
-          <Typography fontSize='16px' fontWeight={600} marginBottom={theme.spacing(5)}>
+          <Typography fontSize='16px' fontWeight={600} marginBottom={theme.spacing(5)} sx={{ flexFlow: 'row wrap' }}>
```
karthikbtf, line 41: I think you should be able to remove these flex properties from the Typography component since it is not a flex box, vercel shows these properties as invalid. Seems to be working without this property. Or.. use word-break/white-space/width properties to get what you want. Removing for now may be ok?

karthikbtf (approved):
lgtm! just one comment about flex properties on typography components

PR #1719, opened 2023-09-26
Service/Redux for Schedule/Reschedule Observations

karthikbtf:
- added service/types/redux to schedule/reschedule observations
- refactored existing code to enable reuse of async reducers

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -97,7 +97,7 @@ jobs:
           path: cypress/screenshots
 
       - name: build
-        run: NODE_OPTIONS="--max-old-space-size=2048" yarn build
+        run: yarn build
         env:
           REACT_APP_TERRAWARE_API: ''
           REACT_APP_TERRAWARE_FE_BUILD_VERSION: ${{ env.APP_VERSION }}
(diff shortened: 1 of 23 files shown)


constanzauanini (approved):

.github/workflows/workflow.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           path: cypress/screenshots
 
       - name: build
-        run: yarn build
+        run: NODE_OPTIONS="--max-old-space-size=2048" yarn build
```
karthikbtf, line 100: @constanzauanini I added this for now to see if it helps (see [stackoverflow](https://stackoverflow.com/questions/53230823/fatal-error-ineffective-mark-compacts-near-heap-limit-allocation-failed-javas)) we could still debug the issue later

constanzauanini: I almost at the same time created a PR to increase memory

constanzauanini: https://github.com/terraware/terraware-web/pull/1721

karthikbtf: Saw that. Should we just do this for github workflows (where the problem exists) versus for everyone that runs the build?

PR #1721, opened 2023-09-26
SW-4263 Add more memory when building

constanzauanini:


karthikbtf (approved):
This is fine as a workaround, might be good to understand why we got here?

PR #1722, opened 2023-09-28
SW-4250 Fix planting density card not taking the whole space

constanzauanini:
Before
Screenshot 2023-09-28 at 17 32 50

After
Screenshot 2023-09-28 at 17 44 54


charlie-tf (approved):
Code change looks good 👍

PR #1723, opened 2023-10-02
SW-4274 Opt-in flag to show/hide schedule/reschedule observations

karthikbtf:
- this in case we release Tracking V2 before user scheduled observations is feature complete


constanzauanini (commented):
Looks good. Just one question, should we add the isEnabledScheduleObservations to the path in Observations/index.tsx too?

karthikbtf:
> Looks good. Just one question, should we add the isEnabledScheduleObservations to the path in Observations/index.tsx too?

I can add it, thanks 👍

PR #1724, opened 2023-10-02
SW-4249 - Improvements to plant monitoring details in reports

constanzauanini:
Fixes in this PR include: 
- Planting progress is not populating at all
- Planting progress should not be shown when we have planting density for the whole site
- Planting density per zone should not be shown if we don’t have any values
- Planting density should be labeled with the units (Plants/ha)

### DIFF
--- a/src/components/Reports/LocationSection.tsx
+++ b/src/components/Reports/LocationSection.tsx
@@ -194,6 +194,10 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
     }
   };
 
+  const estimatedPlants = useMemo(() => {
+    return latestObservation?.estimatedPlants?.toString();
+  }, [latestObservation]);
+
   const livePlants = useMemo(() => {
     return latestObservation?.species.reduce((acc, sp) => (acc = acc + sp.permanentLive), 0);
   }, [latestObservation]);
@@ -537,7 +541,7 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
                   className={classes.infoCardStyle}
                 />
               </Grid>
-              {!latestObservation.estimatedPlants?.toString() && (
+              {!estimatedPlants && (
                 <Grid item xs={smallItemGridWidth()}>
                   <OverviewItemCard
                     isEditable={false}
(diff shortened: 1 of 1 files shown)


src/components/Reports/LocationSection.tsx
```diff
@@ -122,6 +125,12 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
     }
   }, [dispatch, selectedOrganization, trackingV2]);
 
+  useEffect(() => {
+    if (plantingSite) {
+      dispatch(requestSiteReportedPlants(plantingSite.id));
+    }
+  }, [plantingSite]);
```
karthikbtf, line 132: `plantingSite.id`?

src/components/Reports/LocationSection.tsx
```diff
@@ -528,27 +537,35 @@ export default function LocationSection(props: LocationSectionProps): JSX.Elemen
...
-              </Grid>
-              <Grid item xs={smallItemGridWidth()}>
-                <OverviewItemCard
-                  isEditable={false}
-                  title={strings.PLANTING_DENSITY_OF_PLANTED_ZONES}
-                  contents={plantingDensityForZones}
-                  className={classes.infoCardStyle}
-                />
-              </Grid>
+              {!latestObservation.estimatedPlants?.toString() && (
```
karthikbtf, line 540: this is used in many places, maybe extract it out into a `useMemo`?

karthikbtf (approved):
Makes sense I think. A few comments and should be good to merge after that 👍

src/components/Reports/LocationSection.tsx
karthikbtf: Can you validate plantingSite?.id ? I’m probably missing something?

constanzauanini: Do you mean making it 
`dispatch(requestSiteReportedPlants(plantingSite?.id));`.  ?

I can't do that because I get `Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.ts(2345`

karthikbtf:
```
 useEffect(() => {
    if (plantingSite?.id) {
      dispatch(requestSiteReportedPlants(plantingSite.id));
    }
  }, [plantingSite?.id, dispatch]);
```

karthikbtf (approved):
lgtm otherwise 👍

PR #1726, opened 2023-10-10
Remove TrackingV2 Feature Flags

charlie-tf:
Release the tracking feature!

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,7 +52,6 @@ import {
   BatchBulkWithdrawWrapperComponent,
   SpeciesBulkWithdrawWrapperComponent,
 } from './components/Inventory/withdraw';
-import PlantsDashboard from './components/Plants';
 import { NurseryWithdrawals, NurseryWithdrawalsDetails, NurseryReassignment } from './components/NurseryWithdrawals';
 import { SpeciesService } from 'src/services';
 import { PlantingSite } from 'src/types/Tracking';
@@ -63,7 +62,6 @@ import { Provider } from 'react-redux';
 import { store } from './redux/store';
 import { useAppVersion } from './hooks/useAppVersion';
 import { ReportList, ReportView, ReportEdit } from './components/Reports';
-import isEnabled from 'src/features';
 import Observations from 'src/components/Observations';
 import { getRgbaFromHex } from 'src/utils/color';
 import PlantsDashboardV2 from 'src/components/PlantsV2';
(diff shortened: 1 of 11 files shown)


src/components/NavBar.tsx
```diff
@@ -189,7 +187,7 @@ export default function NavBar({
             }}
             id='plants-dashboard'
           />
-          {trackingV2 && hasPlantingSites === true && (
+          {hasPlantingSites === true ? (
```
charlie-tf, line 190: This one's weird, it won't typecheck correctly without this change. I'm not sure why it was okay with the previous implementation

karthikbtf (approved):
Thanks, excited about this release!!
Some comments in there, mostly suggestions.

src/components/NavBar.tsx
charlie-tf: I think maybe the type of the `SubNavBar` `children` is the culprit

src/App.tsx
```diff
@@ -435,10 +432,10 @@ function AppContent() {
               <NewNursery />
             </Route>
             <Route exact path={APP_PATHS.PLANTS_DASHBOARD}>
-              {trackingV2 ? <PlantsDashboardV2 /> : <PlantsDashboard />}
```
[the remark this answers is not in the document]
charlie-tf: I'll take it out now. Thanks!

src/components/NavBar.tsx
karthikbtf: Got it, thanks for trying out the suggestion. We can follow up later maybe.  Looks good for now.

src/components/Plants/index.tsx
```diff
@@ -1,36 +1,303 @@
-import { useCallback, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import strings from 'src/strings';
 import { PlantingSite } from 'src/types/Tracking';
 import { APP_PATHS } from 'src/constants';
-import PlantingSiteDetails from './PlantingSiteDetails';
 import PlantsPrimaryPage from 'src/components/PlantsPrimaryPage';
+import { Box, Grid, Typography, useTheme } from '@mui/material';
```
karthikbtf, line 6: Guessing all these differences are between the previous Plants/index and PlantsV2/index.. but this file is really just PlantsV2/index

karthikbtf (approved):
Thanks for removing the old implementation 🙌

src/components/Plants/index.tsx
charlie-tf: Yup, I think that is the case here

PR #1728, opened 2023-10-11
SW-3157 Convert to new seed/weight calculator and withdraw modal logic

alexTerraformation:
If a user has both a subset weight and subset count allow them to withdraw in either seeds or weight.  For nursery transfers and viability test convert all weight withdrawals to seed count.  Otherwise perform withdrawal in the units selected.
Screen Shot 2023-10-11 at 12 12 15 PM
Screen Shot 2023-10-11 at 12 12 23 PM
Screen Shot 2023-10-11 at 12 12 31 PM
Screen Shot 2023-10-11 at 12 12 58 PM
Screen Shot 2023-10-11 at 12 13 13 PM
Screen Shot 2023-10-11 at 12 17 45 PM

### DIFF
--- a/src/components/accession2/edit/QuantityModal.tsx
+++ b/src/components/accession2/edit/QuantityModal.tsx
@@ -2,14 +2,15 @@ import React, { useEffect, useState } from 'react';
 import strings from 'src/strings';
 import Button from 'src/components/common/button/Button';
 import DialogBox from 'src/components/common/DialogBox/DialogBox';
-import { Box, Grid, useTheme } from '@mui/material';
+import { Box, Grid, useTheme, Radio, RadioGroup, FormControlLabel, Typography, Theme } from '@mui/material';
+
+import { makeStyles } from '@mui/styles';
 import { Textfield } from '@terraware/web-components';
 import { Accession } from 'src/types/Accession';
 import AccessionService from 'src/services/AccessionService';
 import useForm from 'src/utils/useForm';
 import { isUnitInPreferredSystem, Unit, usePreferredWeightUnits } from 'src/units';
 import useSnackbar from 'src/utils/useSnackbar';
-import CalculatorModal from './CalculatorModal';
 import { Dropdown } from '@terraware/web-components';
 import Link from 'src/components/common/Link';
 import EditState from './EditState';
@@ -17,6 +18,22 @@ import _ from 'lodash';
 import { useUser } from 'src/providers';
 import ConvertedValue from 'src/components/ConvertedValue';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
+import Icon from 'src/components/common/icon/Icon';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  addIcon: {
+    fill: theme.palette.TwClrIcnBrand,
+    height: '20px',
+    width: '20px',
+  },
+  units: {
+    marginLeft: theme.spacing(0.5),
+  },
+  subset: {
+    borderTop: `1px solid ${theme.palette.TwClrBrdrTertiary}`,
+    paddingTop: theme.spacing(2),
+  },
+}));
 
 export interface QuantityModalProps {
   open: boolean;
@@ -31,14 +48,21 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
   const { onClose, open, accession, reload, statusEdit } = props;
 
   const [record, setRecord, onChange] = useForm(accession);
-  const [isCalculatorOpened, setIsCalculatorOpened] = useState(false);
+  const [isSubsetOpen, setIsSubsetOpen] = useState(false);
   const [quantityError, setQuantityError] = useState(false);
   const theme = useTheme();
   const snackbar = useSnackbar();
   const preferredUnits = usePreferredWeightUnits();
   const { userPreferences } = useUser();
   const { isMobile } = useDeviceInfo();
+  const [isByWeight, setIsByWeight] = useState(!(accession.remainingQuantity?.units === 'Seeds'));
 
+  const [subsetError, setSubsetError] = useState('');
+  const [subsetWeightError, setSubsetWeightError] = useState('');
+
+  const [subsetCountError, setSubsetCountError] = useState('');
+  const [totalWeightError, setTotalWeightError] = useState('');
+  const classes = useStyles();
   const validate = () => {
     const quantity = parseFloat(record.remainingQuantity?.quantity as unknown as string);
     if (isNaN(quantity) || quantity < 0) {
@@ -88,17 +112,6 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
     }
   };
 
-  const onChangeUnit = (newValue: string) => {
-    if (record) {
-      setRecord({
-        ...record,
-        remainingQuantity: {
-          quantity: record.remainingQuantity?.quantity || 0,
-          units: newValue as Unit['value'],
-        },
-      });
-    }
-  };
   const gridSize = () => (isMobile ? 12 : 6);
 
   const onChangeStatus = (id: string, value: unknown) => {
@@ -107,33 +120,70 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
 
   const onCloseHandler = () => {
     setRecord(accession);
+    setIsSubsetOpen(false);
+    setIsByWeight(!(record.remainingQuantity?.units === 'Seeds'));
     onClose();
   };
 
-  const openCalculator = () => {
-    setIsCalculatorOpened(true);
+  const showSubset = () => {
+    setIsSubsetOpen(true);
+  };
+
+  const onChangeQuantityBy = (_: React.ChangeEvent<HTMLInputElement>, value: string) => {
+    if (value === 'count') {
+      setIsByWeight(false);
+    } else if (value === 'weight') {
+      setIsByWeight(true);
+    }
+  };
+
+  const onChangeRemainingQuantityUnit = (newValue: string) => {
+    setTotalWeightError('');
+    if (record) {
+      setRecord({
+        ...record,
+        remainingQuantity: { quantity: record.remainingQuantity?.quantity || 0, units: newValue as Unit['value'] },
+      });
+    }
+  };
+
+  const onChangeSubsetWeight = (value: any) => {
+    setSubsetWeightError('');
+    setSubsetError('');
+    setRecord({
+      ...record,
+      subsetWeight: { quantity: value, units: record.subsetWeight?.units || 'Grams' },
+    });
+  };
+
+  const onChangeSubsetUnit = (newValue: string) => {
+    setSubsetWeightError('');
+    setSubsetError('');
+    setRecord({
+      ...record,
+      subsetWeight: { quantity: record.subsetWeight?.quantity || 0, units: newValue as Unit['value'] },
+    });
+  };
+
+  const onChangeSubsetCount = (value: number) => {
+    setSubsetCountError('');
+    onChange('subsetCount', value);
   };
 
   const hasChanged =
     (!statusEdit || accession.state !== record.state) &&
     (!_.isEqual(accession.remainingQuantity, record.remainingQuantity) ||
-      !_.isEqual(accession.estimatedCount, record.estimatedCount));
+      !_.isEqual(accession.estimatedCount, record.estimatedCount) ||
+      !_.isEqual(accession.subsetCount, record.subsetCount) ||
+      !_.isEqual(accession.subsetWeight?.units, record.subsetWeight?.units) ||
+      !_.isEqual(accession.subsetWeight?.quantity, record.subsetWeight?.quantity));
 
   return (
     <>
-      <CalculatorModal
-        open={isCalculatorOpened}
-        onClose={() => setIsCalculatorOpened(false)}
-        record={record}
-        setRecord={setRecord}
-        onChange={onChange}
-        reload={reload}
-        onPrevious={() => setIsCalculatorOpened(false)}
-      />
       <DialogBox
         scrolled={true}
         onClose={onCloseHandler}
-        open={open && !isCalculatorOpened}
+        open={open}
         title={props.title}
         size='medium'
         middleButtons={[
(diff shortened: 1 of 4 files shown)


src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -173,29 +192,42 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
 
+    if (isByWeight && accession.subsetCount && accession.subsetWeight) {
+      if (
+        accession.remainingQuantity?.units &&
+        accession.remainingQuantity?.units === 'Seeds' &&
+        accession.estimatedWeight?.units
+      ) {
+        estimated =
+          convertUnits(value, accession.estimatedWeight?.units, accession.subsetWeight.units) *
+          (accession.subsetCount / accession.subsetWeight.quantity);
```
karthikbtf, line 203: extract out as a helper function (within this context, maybe above line 196) 
```
  const calculateEstimated = (unit: string) => convertUnits(value, unit, accession.subsetWeight.units) * (accession.subsetCount / accession.subsetWeight.quantity);
```

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -327,6 +347,32 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
+    if (value === 'count') {
+      setIsByWeight(false);
+      setRecord({
+        ...record,
+        withdrawnQuantity: {
+          quantity: record.withdrawnQuantity?.quantity ? record.withdrawnQuantity?.quantity : 0,
+          units: 'Seeds',
+        },
+      });
+      setIndividualError('withdrawnQuantity', '');
```
karthikbtf, line 360: Clear this at the beginning of the function (applying to both value cases)? Line 370 sets the error again if relevant.

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -464,49 +518,113 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
-              isNurseryTransfer ? (
-                <Box>{strings.CT}</Box>
-              ) : (
-                <Dropdown
-                  options={preferredUnits}
-                  placeholder={strings.SELECT}
-                  onChange={onChangeUnit}
-                  selectedValue={record.withdrawnQuantity?.units}
-                  fullWidth={true}
+        <Box className={classes.withdraw}>
```
karthikbtf, line 521: Here again we may benefit from having flavors of components based on semantics. It may be a bit tricky but helps avoid the various if/else logic. The common stuff could be a separate shared component. This is mostly a theoretical recommendation.

Example:
```
Withdraw Modal
  .. all the withdraw stuff
  - withdraw by selector
  - ByWeightComponent (if selected by weight)
     - weight specific log
     - use a shared sub component for the input/withdraw-all elements
  - ByQuantityComponent (if selected by quantity)
     - quantity specific logic
     - use a shared sub component for the input/withdraw-all elements
  ... rest of the withdraw stuff
```
Here again there might be some properties passing to be done.
Again, my suggestion is mostly theoretical to help separate context and less centralized if/else logic. It may end up being a waste of time on its own but maybe worth trying if we think this component is a tad complex.

src/units.ts
```diff
@@ -97,6 +98,110 @@ export function convertValue(value: number, unit: string) {
...
+        }
+      }
+    }
+    case 'Milligrams': {
+      switch (outputUnit) {
+        case 'Ounces': {
+          return value * 0.035274;
+        }
+        case 'Pounds': {
+          return value * 2.20462e-6;
```
karthikbtf, line 148: Not sure if we may run into very large number representations at this time. If/when the time comes, we may need to look into something like [big.js](https://www.npmjs.com/package/big.js?activeTab=readme)

karthikbtf (commented):
I understand this is WIP.
Code looks good so far, happy to continue looking with iterations.
Made some simple suggestions and other refactor suggestions, the latter is mostly theoretical and may help or end up being a waste. If the reasoning makes sense, might be worth the consideration.

src/components/accession2/edit/QuantityModal.tsx
```diff
@@ -161,81 +209,144 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
...
+                    <Icon name='iconAdd' className={classes.addIcon} />
+                    &nbsp;{`${strings.ADD_SUBSET_WEIGHT_AND_COUNT}`}
+                  </Box>
                 </Link>
               </Box>
-            ) : null}
+            ) : (
+              <Box className={classes.subset}>
+                <Grid container item justifyContent='space-between'>
+                  <Grid container item xs={isMobile ? 12 : 7} spacing={1}>
```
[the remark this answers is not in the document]
alexTerraformation: Yeah just to have some separation between Subset weight units and Subset count fields.  I had asked about placeholder text getting cutoff before and design was totally fine with it.

src/components/accession2/edit/QuantityModal.tsx
```diff
@@ -161,81 +209,144 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
...
+                        type='number'
+                        min={0}
+                        disabledCharacters={['-']}
+                        value={record.subsetWeight?.quantity}
+                        errorText={subsetError || subsetWeightError}
+                        tooltipTitle={strings.ENTER_SUBSET_WEIGHT_AND_COUNT}
+                      />
+                    </Grid>
+                    <Grid item xs={4}>
+                      <Box height={subsetWeightError ? '85px' : subsetError ? '65px' : 'auto'} paddingTop='24px'>
```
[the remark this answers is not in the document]
alexTerraformation: I'm fine with either of those approaches.  Worth noting that the existing production calculator has the same cutoff units selector, so I don't think it's worth spending too much time trying to fix.

src/strings/csv/en.csv
```diff
@@ -1165,6 +1170,7 @@ TEST_METHOD,Test Method
 TEST_METHOD_REQUIRED,Test Method *
 TEST_TYPE,Test Type
 TESTING_STAFF,Testing Staff
+TILDE,"~"
```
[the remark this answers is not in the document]
alexTerraformation: Good call!

src/units.ts
```diff
@@ -97,6 +98,110 @@ export function convertValue(value: number, unit: string) {
...
+          return value * 0.035274;
+        }
+        case 'Grams': {
+          return value * 0.035274;
+        }
+        case 'Milligrams': {
+          return value * 0.035274;
+        }
+        case 'Kilograms': {
+          return value * 0.035274;
```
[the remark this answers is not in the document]
alexTerraformation: Oops, don't know how I missed these.  Thanks!

src/components/accession2/withdraw/WithdrawModal.tsx
alexTerraformation: I like this idea, but even within the context of subsetWeight and subsetCount being defined Typescript thinks they might possibly be undefined within the function.

alexTerraformation: My TypescriptFu is severly lacking though so maybe I'm missing the proper keyword to make this work.

karthikbtf: Share what you have? I can take a look.

src/components/accession2/withdraw/WithdrawModal.tsx
alexTerraformation: I ended up separating out the components, but didn't try to share too much logic between the two.  The simplification in the logic was well worth it.  Would be interested to hear if you think trying to re-combine some functionality would be a worthwhile exercise.

alexTerraformation:
@karthikbtf can you take another pass through this?  Should be much cleaner now.

karthikbtf:
> @karthikbtf can you take another pass through this? Should be much cleaner now.

Sounds good Alex, will take a look today.

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -320,68 +236,29 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
   };
 
+  const onChangeWithdrawBy = (_: React.ChangeEvent<HTMLInputElement>, value: string) => {
+    setIndividualError('withdrawnQuantity', '');
+
+    if (value === 'count') {
+      setIsByWeight(false);
+    } else if (value === 'weight') {
+      setIsByWeight(true);
+    }
```
karthikbtf, line 252: same here, setIsByWeight(value === 'weight')

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -54,6 +63,13 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
   const [selectedSeedBank, setSelectedSeedBank] = useState<Facility>();
   const tz = useLocationTimeZone().get(selectedSeedBank);
   const [timeZone, setTimeZone] = useState(tz.id);
+  const [isByWeight, setIsByWeight] = useState(accession.remainingQuantity?.units === 'Seeds' ? false : true);
+  const classes = useStyles();
+  const [withdrawalQty, setWithdrawalQty] = useState<number>(0);
+  const [withdrawalUnits, setWithdrawalUnits] = useState<
+    'Seeds' | 'Grams' | 'Milligrams' | 'Kilograms' | 'Ounces' | 'Pounds'
```
karthikbtf, line 70: this could be `Units['value']`

src/components/accession2/withdraw/WithdrawModal.tsx
karthikbtf: This is so much cleaner and manageable. I don't see anything worth re-combining, they are very different. Withdraw-all is a common feature but underlying logic is different enough that a refactor won't really help.

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -464,49 +347,32 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
-                  ? nurseryTransferRecord.germinatingQuantity
-                  : record.withdrawnQuantity?.quantity.toString()
-              }
-              errorText={fieldsErrors.withdrawnQuantity}
+        <Box className={classes.withdraw}>
+          <Grid item xs={12} textAlign='left'>
+            <Typography color={theme.palette.TwClrTxtSecondary} display='flex' fontSize={14}>
+              {strings.WITHDRAW_BY}
+            </Typography>
+            <RadioGroup
```
karthikbtf, line 355: Just a comment for future work, looks like a good candidate for web-components, there's some common layout across radio group usage.

src/components/accession2/withdraw/WithdrawModal.tsx
```diff
@@ -464,49 +347,32 @@ export default function WithdrawDialog(props: WithdrawDialogProps): JSX.Element
...
-            id='withdrawAll'
-            name=''
-            label={strings.WITHDRAW_ALL}
-            onChange={(value) => onSelectAll('withdrawAll', value)}
-            value={withdrawAllSelected}
-          />
-        </Grid>
+          ) : (
+            <CountWithdrawal accession={accession} onWithdrawCtUpdate={onWithdrawCtUpdate} />
+          )}
```
karthikbtf, line 374: Thanks for this exercise, definitely a win.

src/components/accession2/edit/QuantityModal.tsx
```diff
@@ -161,81 +223,134 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
...
                   unit={record.remainingQuantity.units}
                   showTooltip={true}
                 />
-              )}
-          </Grid>
-          <Grid item xs={12} sx={{ marginTop: theme.spacing(2) }}>
-            {!isCalculatorOpened ? (
+              </Grid>
+            )}
+          <Grid item xs={12}>
```
karthikbtf, line 306: Missed this before, wondering if this has enough standalone functionality worth separating out. Curious to get your thoughts.

karthikbtf (commented):
This is looking so much better.
The refactor separating out weight/count withdrawal components was a success.
Added a few comments.

src/components/accession2/edit/QuantityModal.tsx
alexTerraformation: I ended up leaving it there since we don't use it anywhere else.  I didn't think there would be a lot of value in making it a separate component that's still just local to this file.

alexTerraformation:
This was a significant change to the core functionality of seed withdrawals.  @karthikbtf reviewed and suggested improvements while in draft, but I think it would be good to get another review here.

src/components/accession2/withdraw/WeightWithdrawal.tsx
```diff
@@ -0,0 +1,241 @@
...
+        onChangeWithdrawnQuantity(accession.remainingQuantity?.quantity);
+      }
+    } else {
+      setWithdrawnQuantity(undefined);
+    }
+
+    setWithdrawAllSelected(withdrawAll);
+  };
+
+  const validateAmount = (estimated: number, withdrawnQty: number, withdrawnUnits: string) => {
```
nickgraz, line 120: This function could benefit from "never nesting". Not really necessary, but would help to improve the readability. Basically, try to only have 1 level of nesting for each validation case. For example:

```
if (!withdrawnQty) {
  setWithdrawnQtyError(strings.REQUIRED_FIELD);
  return false;
}

if (isNaN(withdrawnQty) || Number(withdrawnQty) <= 0) {
  setWithdrawnQtyError(strings.INVALID_VALUE);
  return false;
}
...
```

src/components/accession2/withdraw/WeightWithdrawal.tsx
```diff
@@ -0,0 +1,241 @@
...
+    if (withdrawnQty) {
+      if (isNaN(withdrawnQty)) {
+        setWithdrawnQtyError(strings.INVALID_VALUE);
+        return false;
+      }
+      if (Number(withdrawnQty) <= 0) {
+        setWithdrawnQtyError(strings.INVALID_VALUE);
+        return false;
+      }
+      if (accession.remainingQuantity?.units && Number(withdrawnUnits === accession.remainingQuantity.units)) {
```
nickgraz, line 130: Another option is to make the validators a bit more isolated. For example 

```
if (!validateWithdrawnQuantity(...)) {
  setWithdrawnQtyError(strings.WITHDRAWN_QUANTITY_ERROR);
  return false;
}
...
```

src/components/accession2/withdraw/WeightWithdrawal.tsx
```diff
@@ -0,0 +1,241 @@
...
+              display={true}
+            />
+          </Grid>
+        </Grid>
+      ) : null}
+      <Grid container direction='row' justifyContent='space-between'>
+        <Grid item xs={accession.subsetWeight?.quantity && accession.subsetCount ? 6 : 12} paddingBottom={2}>
+          <Box display='flex' alignItems={withdrawnQtyError ? 'center' : 'end'}>
+            <Textfield
+              label={strings
```
nickgraz, line 188: This section might benefit from being turned into its own function.

src/components/accession2/withdraw/WeightWithdrawal.tsx
joshbuchea: Love this! I also aim to limit nesting to one level in code to improve readability.

src/components/accession2/withdraw/CountWithdrawal.tsx
```diff
@@ -0,0 +1,155 @@
...
+      }
+      if (isNaN(withdrawnQty)) {
+        setWithdrawnQtyError(strings.INVALID_VALUE);
+        return false;
+      }
+      if (Number(withdrawnQty) <= 0) {
+        setWithdrawnQtyError(strings.INVALID_VALUE);
+        return false;
+      }
+      if (accession.remainingQuantity?.units && Number(accession.remainingQuantity.units === 'Seeds')) {
```
nickgraz, line 97:
```suggestion
      if (accession.remainingQuantity?.units === 'Seeds') {
```

nickgraz (approved):
Looks good to me, a couple nit picks in there but none are necessary. I also think it might be helpful to make some util functions like `isAccessionUnitSeeds(record)` since there is a lot of `record.remainingQuantity?.units === 'Seeds'` out there. Again, not necessary but should reduce the "magic string" count.

src/components/accession2/withdraw/WeightWithdrawal.tsx
alexTerraformation: This was a good exercise.  Makes for some duplicate code, but definitely more readable.

src/components/accession2/withdraw/WeightWithdrawal.tsx
alexTerraformation: Opted against this.  I didn't feel like it was worth the extra information passing and duplicated code.

src/components/accession2/withdraw/CountWithdrawal.tsx
alexTerraformation: Need to keep the check for units because it can be undefined.

PR #1731, opened 2023-10-18
SW-4316 Replace observation plots from map view, intermediate implementation

karthikbtf:
- similar to list view, added replace observation plot in observations map view
- moved the modal into a separate directory for reuse
- refactored the map tooltip to accept the entire plot (to access more properties)
- introduced react Portal in this PR since we can't efficiently render a modal within a modal, rendering it in a Portal instead (end result is similar to rendering a modal from the main app, just not within a modal)
- processing the response from BE etc will be done separately
- only plots that haven't been monitored (no completed time) will be eligible for reassignment

Terraware App 2023-10-18 13-39-38

Terraware App 2023-10-18 13-39-55


src/components/Observations/map/TooltipContents.tsx
```diff
@@ -1,48 +1,92 @@
...
+              className={classes.button}
+              id='reassignPlot'
+              label={`${strings.REQUEST_REASSIGNMENT}...`}
+              type='passive'
+              onClick={() => setShowReplacePlotModal(true)}
+              priority='secondary'
+            />
+          </Box>
+        )}
+      </Box>
```
karthikbtf, line 89: indented 2 spaces, no other change

src/components/Observations/map/TooltipContents.tsx
```diff
@@ -1,48 +1,92 @@
...
-        </>
+    <>
+      {showReplacePlotModal && observationId && monitoringPlot && (
+        <Portal>
+          <ReplaceObservationPlotModal
+            onClose={() => setShowReplacePlotModal(false)}
+            observationId={observationId}
+            monitoringPlot={monitoringPlot}
+          />
+        </Portal>
```
karthikbtf, line 49: Using a react portal to render this modal outside the current tooltip modal's DOM hierarchy, so it has the entire screen estate.. otherwise it is super squished inside the tiny tooltip modal

PR #1735, opened 2023-10-19
SW-3968 / SW-4330 Add full screen control to maps and option to display 'outdoors' style map

karthikbtf:
- added fullscreen control to top-left (default is top-right but we have map layers there)
- using default styles but we can make this a tad smaller if we want
- introduced a hide fullscreen prop
- extracted out zoom to fit as a component while my hands were in there
- Added switcher on top left for Outdoors/Satellite
- Hide map style switcher and layers menus in full screen mode

Terraware App 2023-10-19 11-08-36

Screen Shot 2023-10-19 at 11 09 01 AM 2023-10-19 11-09-31


chudo76 (approved):
All maps have fullscreen button now. Thanks!

PR #1736, opened 2023-10-19
SW-4330 Add 'Outdoors' map view style option

karthikbtf:
- Added switcher on top left for Outdoors/Satellite, not using an IControl at this point, being consistent with how we added the layers and zoom-to-fit controls
- update map view style based on selection
- currently storing selection in localStorage, not distinguished by site id or any other key
- we should discuss best approach for storing this preference (site wide option vs user specific option vs user/site option, etc.)
- outstanding issue is the selector is not clickable in full screen mode because the popover is not viewable in full screen mode (not part of the dom hierarchy).. i could switch this to a radio button

Terraware App 2023-10-19 12-31-57

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -1,5 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { Box, useTheme } from '@mui/material';
+import { Box, Theme, useTheme } from '@mui/material';
+import { makeStyles } from '@mui/styles';
 import 'mapbox-gl/dist/mapbox-gl.css';
 import ReactMapGL, {
   AttributionControl,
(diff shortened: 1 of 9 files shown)


karthikbtf:
@chudo76 @alexTerraformation take this for a spin, has both full screen and the mapbox view-layer switching.
We can iterate on feedback.
A known limiting factor with fullscreen [mode](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) is that only those elements part of the map dom hierarchy will be visible. Anything involving modals (parent dom hierarchy) will not be visible.
Stuff that might be affected are 'request reassignment' modal, any popovers (layers etc.).  We could explore rendering the modals through a [react portal](https://mui.com/base-ui/react-portal/) tied to the full screen parent (only in full screen mode) and see if that works. Theoretical at this point. Will run some experiments next week.

chudo76 (approved):
Nice! We may want to do a UI reassessment for all the map controls now that we have a mix of ours and Mapbox's but this looks good to me.

karthikbtf:
> Nice! We may want to do a UI reassessment for all the map controls now that we have a mix of ours and Mapbox's but this looks good to me.

Thanks @chudo76 😁  yes it is getting a bit crowded, do revisit the controls later.
btw the map button looks smudgy, we can make it smaller if we want. or leave it as-is and revisit later.
See my comments about the modals not being shown in full screen mode, I will report here next week on any findings.

constanzauanini (approved):
Excellent!

PR #1741, opened 2023-10-23
SW-4330-continued Allow reassign-request modal to show in full screen

karthikbtf:
- Created a custom hook that informs when a map is in fullscreen and which container to use for modal portals
- Integrated this hook with map tooltip so reassign request modal shows in full screen mode as well (uses the full screen container as the parent for the modal)
- popovers (such as layers and outdoors-vs-satellite) are still not functional, added a change to not render these in fullscreen mode for now (since they won't work)
- will follow up later with a web-component change if we can get popovers to work
- refactored the top-right/bottom-left control components to be rendered inside of the map hierarchy otherwise they get hidden in fullscreen mode, top-right for now is not rendered until popovers work, the bottom-left observation date slider now works in fullscreen mode

Screen Shot 2023-10-23 at 11 32 22 AM 2023-10-23 11-34-12

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -517,6 +517,13 @@ export default function Map(props: MapProps): JSX.Element {
     return source.entities?.some((entity) => entity?.boundary?.length);
   });
 
+  const renderedPopup = useMemo<JSX.Element | null>(() => {
+    if (!popupInfo || !popupRenderer) {
+      return null;
+    }
+    return popupRenderer.render(popupInfo.properties);
+  }, [popupInfo, popupRenderer]);
+
   // keep track of map being destroyed, this is not bound by react event loop
   let destroying = false;
 
@@ -555,7 +562,7 @@ export default function Map(props: MapProps): JSX.Element {
           <AttributionControl compact={true} style={{ marginRight: '5px' }} position='top-left' />
           <ZoomToFitControl onClick={zoomToFit} />
           <MapViewControl mapViewStyle={mapViewStyle} onChangeMapViewStyle={onChangeMapViewStyle} />
-          {popupInfo && popupRenderer && (
+          {popupInfo && popupRenderer && renderedPopup && (
             <Popup
               anchor={popupRenderer.anchor ?? 'top'}
               longitude={Number(popupInfo.lng)}
@@ -570,7 +577,7 @@ export default function Map(props: MapProps): JSX.Element {
               style={popupRenderer.style}
               className={popupRenderer.className}
             >
-              {popupRenderer.render(popupInfo.properties)}
+              {renderedPopup}
             </Popup>
           )}
           {topRightMapControl && <div className={classes.topRightControl}>{topRightMapControl}</div>}
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -22,6 +22,7 @@ import {
   MapViewStyle,
   MapViewStyles,
 } from 'src/types/Map';
+import { useMapPortalContainer } from './MapRenderUtils';
 import { MapService } from 'src/services';
 import { useLocalization } from 'src/providers';
 import strings from 'src/strings';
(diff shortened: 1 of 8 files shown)


src/components/Map/Map.tsx
```diff
@@ -57,6 +58,22 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+  },
+  topRightControl: {
+    height: 'max-content',
+    position: 'absolute',
+    right: theme.spacing(2),
+    top: theme.spacing(2),
+    width: 'max-content',
+    zIndex: 1000,
+  },
 }));
```
karthikbtf, line 77: moved here from planting site map

src/components/Map/Map.tsx
```diff
@@ -553,6 +573,8 @@ export default function Map(props: MapProps): JSX.Element {
               {popupRenderer.render(popupInfo.properties)}
             </Popup>
           )}
+          {topRightMapControl && <div className={classes.topRightControl}>{topRightMapControl}</div>}
+          {bottomLeftMapControl && <div className={classes.bottomLeftControl}>{bottomLeftMapControl}</div>}
```
karthikbtf, line 577: moved here from planting site map

src/components/Map/Map.tsx
```diff
@@ -613,6 +636,10 @@ const MapViewControl = ({ mapViewStyle, onChangeMapViewStyle }: MapViewControlPr
     ];
   }, [activeLocale]);
 
+  if (mapPortalContainer) {
+    return null;
+  }
```
karthikbtf, line 641: hide in fullscreen mode until we have a popover fix

src/components/Map/PlantingSiteMap.tsx
```diff
@@ -25,25 +24,6 @@ const mapImages = [
...
-  topRightControl: {
-    height: 'max-content',
-    position: 'absolute',
-    right: theme.spacing(2),
-    top: theme.spacing(2),
-    width: 'max-content',
-    zIndex: 1000,
-  },
-}));
-
```
karthikbtf, line 46: moved to Map

src/components/Map/PlantingSiteMap.tsx
```diff
@@ -274,8 +241,6 @@ export default function PlantingSiteMap(props: PlantingSiteMapProps): JSX.Elemen
         mapImages={mapImages}
         {...controlProps}
       />
-      {topRightMapControl && <div className={classes.topRightControl}>{topRightMapControl}</div>}
-      {bottomLeftMapControl && <div className={classes.bottomLeftControl}>{bottomLeftMapControl}</div>}
```
karthikbtf, line 278: moved to map

src/components/common/MapLayerSelect/index.tsx
```diff
@@ -29,13 +30,18 @@ export default function MapLayerSelect({
...
+  const mapPortalContainer = useMapPortalContainer();
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
   const handleClick = (event?: React.MouseEvent<HTMLButtonElement, MouseEvent> | undefined) => {
     setAnchorEl(event?.currentTarget ?? null);
   };
 
+  if (mapPortalContainer) {
+    return null;
+  }
+
```
karthikbtf, line 44: don't show in full screen mode

src/types/Map.ts
```diff
@@ -131,6 +131,10 @@ export type MapData = Record<MapObject, MapSourceBaseData | undefined>;
 export type MapControl = {
   // hide the full screen control
   hideFullScreen?: boolean;
+
+  // custom map controls
+  topRightMapControl?: React.ReactNode;
+  bottomLeftMapControl?: React.ReactNode;
```
karthikbtf, line 137: add the controls to MapControl properties so we don't have to re-define them across map component properties passing

karthikbtf:
friendly ping

constanzauanini (approved):
Looks good!!

karthikbtf:
> Looks good!!

Thanks for the review @constanzauanini I pushed one change while the review was in progress.

src/components/Map/Map.tsx
```diff
@@ -517,6 +517,13 @@ export default function Map(props: MapProps): JSX.Element {
     return source.entities?.some((entity) => entity?.boundary?.length);
   });
 
+  const renderedPopup = useMemo<JSX.Element | null>(() => {
+    if (!popupInfo || !popupRenderer) {
```
constanzauanini, line 521: I don't really get when this will be undefined if you are validating that they exist before calling this function

karthikbtf: This is a `useMemo` it is automatically called when the properties change, the properties can be undefined.

constanzauanini: So the validation in the JSX `{ popupInfo && popupRenderer && ... }` would not be necessary?

constanzauanini (approved):
Just one question, but looks good!

src/components/Map/Map.tsx
karthikbtf: Those properties are dereferenced in the `Popup` component and need to be validated. example, see lines 568 and 578. Typescript will not assume the properties are valid even if we validated them in some other useMemo component.

constanzauanini: Got it! Thanks!

PR #1742, opened 2023-10-23
SW-4334 Make nusery withdrawals photos optional

karthikbtf:
- See [jira](https://terraformation.atlassian.net/browse/SW-4334) for reason
- Did not remove any strings since they will only get added back right now


karthikbtf:
friendly ping

constanzauanini (approved):

PR #1744, opened 2023-10-24
SW-3271 Show all errors on species list upload failure

sgrimm:
Currently, if the user uploads a species list CSV that has validation failures, we
show an error dialog that only includes the first failure and doesn't say where in
the file the offending row is.

Update the dialog so it shows all the errors and says which row each one is on.

### DIFF
--- a/src/components/common/ImportModal.tsx
+++ b/src/components/common/ImportModal.tsx
@@ -121,7 +121,7 @@ export default function ImportSpeciesModal(props: ImportSpeciesModalProps): JSX.
   const [file, setFile] = useState<File>();
   const inputRef = useRef<HTMLInputElement>(null);
   const divRef = useRef<HTMLDivElement>(null);
-  const [error, setError] = useState<string>();
+  const [error, setError] = useState<JSX.Element>();
   const [loading, setLoading] = useState(false);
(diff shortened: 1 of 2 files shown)


sgrimm:
Screenshot 2023-10-24 at 12 46 28 PM

sgrimm:
If there are zillions of errors, the dialog will be quite tall, but a tall dialog is still a better user experience than the current behavior.

src/components/common/ImportModal.tsx
```diff
@@ -156,11 +156,22 @@ export default function ImportSpeciesModal(props: ImportSpeciesModalProps): JSX.
 
   useEffect(() => {
     const getErrors = () => {
-      let errors = strings.DATA_IMPORT_FAILED;
-      if (fileStatus?.details.errors && fileStatus?.details.errors[0]) {
-        errors += ': ' + fileStatus?.details.errors[0].message;
-      }
-      return errors;
+      return (
+        <div className={classes.warningContent} key='import-error-1'>
```
sgrimm, line 160: Not sure if I should have defined a new class for this. `warningContent` is just a left-align rule.

charlie-tf: This is fine

charlie-tf (approved):
Simple enough. This looks good to me

PR #1745, opened 2023-10-25
SW-4350 Use sub-location APIs

sgrimm:
In preparation for adding sub-locations to nurseries, replace usages of the old
"storage location" APIs with the more general "sub-location" ones.

- Rename code symbols and files that included the term "storage location" to
  instead use "sub-location."
- `StorageLocationSelector` had a misleading name because it actually selected
  facilities, not storage locations; rename it to `FacilitySelector` for clarity.
- Use the sub-location API endpoints instead of the seed-bank-specific storage
  location endpoints.
- Use the new sub-location accession and facility payload fields.

This is purely an internal change with no user-visible effect.

Also covers SW-4347.


sgrimm:
This depends on the corresponding server-side change https://github.com/terraware/terraware-server/pull/1430 and the CI build of this PR is expected to fail until that PR is merged.

charlie-tf (approved):
Looks good. Thanks for doing this!

charlie-tf (approved):

PR #1747, opened 2023-10-27
SW-4372 Create batch details modal without keeping state of previous batch selection

karthikbtf:
- don't use the `open` modal parameter, which required us to clean the modal state across batch selections
- conditionally show/render the batch details modal when user clicks on a batch
- this approach is cleaner and we won't have to add a bunch of logic in the modal to update the fields


karthikbtf:
@constanzauanini when you get a chance since you are more familiar with this code?

constanzauanini (approved):

PR #1750, opened 2023-10-31
SW-4376 Distinguish status changes from quantity edits

sgrimm:
Calculating germination and loss rates will require that we distinguish between
the user overwriting the quantities of a batch and the user moving seedlings from
one status to another.

Update the "change status" dialog so it calls a newly-added API endpoint for
changing seedling statuses.


sgrimm:
Depends on https://github.com/terraware/terraware-server/pull/1440

karthikbtf (approved):
👍

PR #1753, opened 2023-11-01
SW-4402 Allow upcoming observations to be rescheduled 

karthikbtf:
- also change criteria for which observations in the table can be rescheduled (in-progress/overdue and no observed plots)
- added a hasObservedTemporaryPlots to capture whether any type of plot was observed (in tandem with hasObservedPermanentPlots)
- refactor layout of page message that shows upcoming observations for reschedule, the site names have permalinks now and had to be constructed programmatically (removed from the string representation)

Terraware App 2023-11-01 08-46-48

Terraware App 2023-11-01 09-12-29

### DIFF
--- a/src/components/Observations/ObservationsEventsNotification.tsx
+++ b/src/components/Observations/ObservationsEventsNotification.tsx
@@ -1,12 +1,15 @@
 import { useMemo } from 'react';
-import { Box } from '@mui/material';
+import { Box, Typography } from '@mui/material';
 import { Button, Message } from '@terraware/web-components';
+import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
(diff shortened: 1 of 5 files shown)


karthikbtf:
This is the last task before I can mark the epic for QA, if anyone has a chance to look..

constanzauanini (approved):
Lgtm!

karthikbtf:
> Lgtm!

Thanks @constanzauanini !!

chudo76:
@karthikbtf Looks like font-size is slightly smaller for the "Reschedule" link (should be 16px instead of 14px?) but otherwise LGTM!

karthikbtf:
> @karthikbtf Looks like font-size is slightly smaller for the "Reschedule" link (should be 16px instead of 14px?) but otherwise LGTM!

I did notice that but could only find 14 when inspecting in figma. Wasn't sure. Will fix it, thanks for the feedback 👍

PR #1754, opened 2023-11-01
SW-2223 Keep state of current org in the url

karthikbtf:
- allows us to share urls maintaining org context
- centralized handling so devs can create links in code without worrying about populating the org id
- some special handling for the welcome page alone as it switched between routing contexts
- also updated the plants primary page (used by observations, dashboard) to avoid redundant calls to get/set preferences, noticed this while debugging

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -316,7 +316,7 @@ function AppContent() {
               <OptInFeatures refresh={reloadPreferences} />
             </Route>
           )}
-          <Route path='/'>
+          <Route path='*'>
             <Redirect to={APP_PATHS.WELCOME} />
           </Route>
         </Switch>
@@ -529,7 +529,7 @@ function AppContent() {
             <Route path={APP_PATHS.SPECIES + '/'}>
               <Redirect to={APP_PATHS.SPECIES} />
             </Route>
-            <Route path='/'>
+            <Route path='*'>
               <Redirect to={APP_PATHS.HOME} />
             </Route>
           </Switch>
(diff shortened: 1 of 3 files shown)


karthikbtf:
QA'd this locally a bunch. If anyone has time to give this a whirl.. will be super helpful.

constanzauanini:
@karthikbtf I'm trying to go to https://tree-location-web-git-karthik-sw-2223-terraformation.vercel.app/planting-sites/70?organizationId=23  from other org that I have selected (id=7), and the link get the org id changed to the one that I have selected (7). Is that the expected behavior? So I don't see the planting site

karthikbtf:
> @karthikbtf I'm trying to go to https://tree-location-web-git-karthik-sw-2223-terraformation.vercel.app/planting-sites/70?organizationId=23 from other org that I have selected (id=7), and the link get the org id changed to the one that I have selected (7). Is that the expected behavior? So I don't see the planting site

Thanks for taking a look @constanzauanini . No that's the wrong behavior. Let me take a look.

karthikbtf:
> @karthikbtf I'm trying to go to https://tree-location-web-git-karthik-sw-2223-terraformation.vercel.app/planting-sites/70?organizationId=23 from other org that I have selected (id=7), and the link get the org id changed to the one that I have selected (7). Is that the expected behavior? So I don't see the planting site

This should be fixed now, thanks for testing. Btw, the behavior you observed will happen - only if you don't have access to the org in the url (23). If you have access, it should switch to that org. Give it a try after the preview builds.

constanzauanini (approved):
Code looks correct to me. I did some testing in Vercel and working correctly!

karthikbtf:
> Code looks correct to me. I did some testing in Vercel and working correctly!

Awesome \o/ thanks! I will wait until next week before merging. In case others can QA.

PR #1755, opened 2023-11-01
Fix font size in Reschedule link

karthikbtf:
- use font size 16 for Reschedule link

Terraware App 2023-11-01 13-32-49


chudo76 (approved):
👍

PR #1756, opened 2023-11-01
SW-4053 Show planting site name and end date in observations table, remove subzones

karthikbtf:
- See [JIRA](https://terraformation.atlassian.net/browse/SW-3999)

Terraware App 2023-11-01 15-10-11

### DIFF
--- a/src/components/Observations/ObservationsDataView.tsx
+++ b/src/components/Observations/ObservationsDataView.tsx
@@ -80,7 +80,9 @@ export default function ObservationsDataView(props: ObservationsDataViewProps): 
     <ListMapView
       initialView='list'
       search={<Search {...searchProps} />}
-      list={<OrgObservationsListView observationsResults={observationsResults} />}
+      list={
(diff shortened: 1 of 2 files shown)


karthikbtf:
@chudo76 let me know thoughts on column default locations - yes they can be shuffled around and we save the layout.

chudo76 (approved):
Having the planting site column is super helpful! End date is useful, too. Column positions look good. 🙏

karthikbtf:
> Having the planting site column is super helpful! End date is useful, too. Column positions look good. 🙏

Awesome, thanks 👍

karthikbtf:
Looking for one more +1

constanzauanini (approved):
Looks good! Why do we want to remove sub-zone column? Is it not helpful?

karthikbtf:
> Looks good! Why do we want to remove sub-zone column? Is it not helpful?

We show monitoring plots for the zone on the next page (not by subzones), thought subzones is not super helpful and uses up space. We will add it back if users want it I suppose.
Thanks for the review 🙏

PR #1757, opened 2023-11-03
SW-4394 Create batch details page with empty components for tabs

constanzauanini:
This PR doesn't have the details tab or history tab. It's the details page with the summary on top and the breadcrumb

Screenshot 2023-11-03 at 12 38 32

### DIFF
--- a/src/components/InventoryV2/BatchSummary.tsx
+++ b/src/components/InventoryV2/BatchSummary.tsx
@@ -39,23 +39,19 @@ export default function BatchSummary(props: BatchSummaryProps): JSX.Element {
         <OverviewItemCard isEditable={false} title={strings.SUB_LOCATION} contents={batchSubLocations.join(',')} />
       </Grid>
       <Grid item xs={2}>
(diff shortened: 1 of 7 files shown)


src/components/InventoryV2/BatchSummary.tsx
```diff
@@ -0,0 +1,62 @@
...
+          isEditable={false}
+          title={strings.GERMINATION_RATE}
+          contents={batch?.germinationRate || '%'}
+        />
+      </Grid>
+      <Grid item xs={2}>
+        <OverviewItemCard isEditable={false} title={strings.LOSS_RATE} contents={batch?.lossRate || '%'} />
+      </Grid>
+      <Grid item xs={2}>
+        <OverviewItemCard isEditable={false} title={strings.TOTAL_WITHDRAWN} contents={0} />
```
constanzauanini, line 52: Should TotalWithdrawn value be calculated here? Or should it come from the api?

charlie-tf (approved):
Looks good!

PR #1761, opened 2023-11-06
SW-4394 Batch Details tab

constanzauanini:
- Show all batch's details in Details tab, including photos
- Open Edit Details modal when clicking edit button
- Allow editing batch details in modal
- Allow adding/removing photos from details modal

Screenshot 2023-11-06 at 15 54 23

Screenshot 2023-11-06 at 15 58 31

### DIFF
--- a/src/components/InventoryV2/BatchDetails.tsx
+++ b/src/components/InventoryV2/BatchDetails.tsx
@@ -1,15 +1,133 @@
-import { Typography, useTheme } from '@mui/material';
+import { Box, Grid, Typography, useTheme } from '@mui/material';
 import strings from 'src/strings';
 import Card from 'src/components/common/Card';
+import { Batch } from 'src/types/Batch';
+import OverviewItemCard from '../common/OverviewItemCard';
+import { useDeviceInfo } from '@terraware/web-components/utils';
+import { useEffect, useState } from 'react';
+import { NurseryBatchService } from 'src/services';
+import PhotosList from '../common/PhotosList';
+import useSnackbar from 'src/utils/useSnackbar';
+import { Button } from '@terraware/web-components';
+import BatchDetailsModal from './BatchDetailsModal';
+import { BATCH_PHOTO_ENDPOINT } from 'src/services/NurseryBatchService';
 
-export default function BatchDetails(): JSX.Element {
+interface BatchDetailsProps {
+  batch: Batch;
+  onUpdate: () => void;
+}
+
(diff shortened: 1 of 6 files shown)


src/components/InventoryV2/BatchDetailsModal.tsx
```diff
@@ -0,0 +1,402 @@
+import { useEffect, useState, useMemo } from 'react';
```
constanzauanini, line 1: This modal is based on v1, but adding some new fields and all photos management

src/components/InventoryV2/BatchDetails.tsx
```diff
@@ -1,15 +1,133 @@
...
+
+  const overviewCardData = [
+    {
+      title: strings.GERMINATING_QUANTITY,
+      data: batch.germinatingQuantity,
+    },
+    {
+      title: '',
+      data: '',
+    },
```
charlie-tf, line 33: Just for space? Will the space be okay when in mobile?

src/components/InventoryV2/BatchDetailsModal.tsx
```diff
@@ -0,0 +1,396 @@
...
+    onChange(id, date);
+  };
+
+  const marginTop = {
+    marginTop: theme.spacing(2),
+  };
+
+  return (
+    <>
+      {record && (
```
charlie-tf, line 213: This seems okay. An alternative would be to do
```typescript
return (
  {record
    ? <DialogBox>
      ...
      </DialogBox>
    : null}
);
```
Which would avoid returning an empty fragment. You'd have to change the return type to `JSX.Element | null` though.

charlie-tf (commented):
Looks good! Main question is about the empty `overviewCardData` member

src/components/InventoryV2/BatchDetails.tsx
constanzauanini: Oh, you are right, I didn't test it in mobile. Let me check

src/components/InventoryV2/BatchDetails.tsx
```diff
@@ -26,10 +26,7 @@ export default function BatchDetails({ batch, onUpdate }: BatchDetailsProps): JS
     {
       title: strings.GERMINATING_QUANTITY,
       data: batch.germinatingQuantity,
-    },
-    {
-      title: '',
-      data: '',
+      fullWidth: true,
```
charlie-tf, line 29: Nice update!

charlie-tf (approved):
LGTM

PR #1762, opened 2023-11-07
SW-4430 Remove opt-in flag for user scheduled observations

karthikbtf:
- remove the flag!!


src/components/Observations/ObservationsHome.tsx
```diff
@@ -44,7 +43,6 @@ export default function ObservationsHome(props: ObservationsHomeProps): JSX.Elem
   const upcomingObservations = useAppSelector(selectUpcomingObservations);
   // get observation schedulable sites
   const newObservationsSchedulable = useAppSelector(selectObservationSchedulableSites).length;
-  const scheduleObservationsEnabled = isEnabled('Schedule Observations');
```
constanzauanini, line 47: Should we keep the isAdmin in this flag as the rest?

karthikbtf: the `isAdmin` check is captured on line 65.. but, you are correct, I should move that up here for consistency.. will update

constanzauanini (approved):

PR #1768, opened 2023-11-08
Clean up search API payload types

sgrimm:
To avoid a circular type reference, we replaced the definition of
`SearchNodePayload` in `generated-schema.ts` with "any".

Unfortunately, "any" effectively causes type checking to be disabled, so there
was some search-related code that used the wrong types and TypeScript couldn't
detect it.

- Update the OpenAPI code generator so it replaces `SearchNodePayload` with a
  real type, though one that still avoids the circular reference.
- Fix the incorrect types that become visible to the type checker once we've
  stopped using "any".
- Introduce a stricter version of `SearchRequestPayload` that requires a search
  query (it is optional in the API).

This shouldn't change any behavior.


karthikbtf (approved):
Thanks for the changes and cleanup 👍

PR #1769, opened 2023-11-08
SW-4432 Upgrade OpenAPI code generator

sgrimm:
The version of openapi-typescript we've been using doesn't have great support for
polymorphic payloads. Upgrade to the latest version, which does a better job.

With the new version, we no longer need to override the generated types for the
geometry payloads or specify the discriminator values for the search payloads;
the code generator handles both properly.


sgrimm:
This depends on https://github.com/terraware/terraware-server/pull/1463

sgrimm:
Seemingly-unrelated formatting changes here are because openapi-typescript depends on a newer version of Prettier than the one we were using, and apparently the rules have been tweaked.

charlie-tf (approved):
👍

PR #1774, opened 2023-11-10
SW-4389 Show empty batches filter option in "batches by seedlings" view

nickgraz:
- Add filter tooltip and popover to `InventorySeedlingsTable` with 'Show Empty Batches' boolean filter option
- Implement boolean filter type into `FilterGroup`
- Render filtered batches instead of searched batches in `InventorySeedlingsTable`.

### DIFF
--- a/src/components/common/FilterGroup/index.tsx
+++ b/src/components/common/FilterGroup/index.tsx
@@ -12,6 +12,7 @@ import { Option } from '@terraware/web-components/components/table/types';
 import { makeStyles } from '@mui/styles';
 import { Button } from '@terraware/web-components';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
+import FilterBoolean from './filters/FilterBoolean';
 
 interface StyleProps {
   isMobile?: boolean;
@@ -32,6 +33,7 @@ const useStyles = makeStyles((theme: Theme) => ({
 export type FilterField = {
   name: string;
   label: string;
+  showLabel?: boolean;
   type:
     | 'multiple_selection'
     | 'single_selection'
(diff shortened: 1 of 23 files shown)


src/components/common/FilterGroup/index.tsx
```diff
@@ -32,20 +33,22 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-    | 'hidden';
+    | 'hidden'
+    | 'boolean';
 };
 
 export type FilterGroupProps = {
   initialFilters: Record<string, SearchNodePayload>;
   fields: FilterField[];
-  values: FieldValuesPayload;
+  values?: FieldValuesPayload;
```
nickgraz, line 51: This was only being used by `multiple_select` and `single_select` so I made it optional.

src/components/common/FilterGroup/index.tsx
```diff
@@ -148,9 +153,18 @@ export default function FilterGroup(props: FilterGroupProps): JSX.Element {
                 payloads={filters[f.name]?.children ?? []}
               />
             )}
+            {f.type === 'boolean' && (
```
nickgraz, line 156: Let me know if you'd rather I call this type `'checkbox'` and the component `FilterCheckbox`

charlie-tf: `boolean` works for me

charlie-tf (commented):
So Constanza and I have been putting the Inventory updates work in `src/components/InventoryV2`. I believe that `Search.tsx`, `FilterUtils.ts`, and `InventoryFiltersPopover.tsx` have been copied over to that directory, but are otherwise unchanged. How would it be to move your changes over there? Your `view/InventorySeedlingsTable.tsx` changes would have to be ported over there, too, and I'm not quite sure what routing changes might need to be effected to get this to work right under the Nursery Updates switch.

charlie-tf:
Besides that, what you have here looks good!

nickgraz:
Ah my bad, I can move it over to the V2 area

charlie-tf (approved):
Thanks for the rearrangement. Looks good and tested working in Vercel.

PR #1783, opened 2023-11-14
SW-4406 Create withdrawal history table by batch

constanzauanini:
- Withdrawal history rows of the table were create by species, and we want to create them by batches now, so each row can link to the batch details views.

- This change is applied only if nurseryV2 is enabled 

Now:
Screenshot 2023-11-14 at 14 30 19

Before:
Screenshot 2023-11-14 at 14 30 35

### DIFF
--- a/src/components/NurseryWithdrawals/WithdrawalDetails/sections/NonOutplantWithdrawalTable.tsx
+++ b/src/components/NurseryWithdrawals/WithdrawalDetails/sections/NonOutplantWithdrawalTable.tsx
@@ -30,6 +30,7 @@ type NonOutplantWithdrawalTableProps = {
 };
 
 const columns = (): TableColumnType[] => [
+  { key: 'batchNumber', name: strings.BATCH, type: 'string' },
   { key: 'name', name: strings.SPECIES, type: 'string' },
   { key: 'notReady', name: strings.NOT_READY, type: 'number' },
   { key: 'ready', name: strings.READY, type: 'number' },
--- a/package.json
+++ b/package.json
@@ -90,7 +90,8 @@
     "server:prepare": "./scripts/preparedb.sh",
     "server:reset": "./scripts/resetdb.sh",
     "storybook": "storybook dev -p 6006",
-    "build-storybook": "storybook build"
+    "build-storybook": "storybook build",
+    "pre-commit": "yarn format && yarn lint && yarn ts && yarn test && yarn build"
   },
   "eslintConfig": {
     "extends": [
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -49,6 +49,7 @@ import InventoryV2 from './components/InventoryV2';
 import NurseryDetails from './components/Nursery';
 import InventoryCreate from './components/Inventory/InventoryCreate';
 import InventoryView from './components/Inventory/InventoryView';
+import InventoryViewV2 from './components/InventoryV2/InventoryView';
 import InventoryViewForNursery from './components/InventoryV2/InventoryViewForNursery';
 import InventoryBatch from './components/InventoryV2/InventoryBatch';
 import {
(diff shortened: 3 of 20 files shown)


src/components/NurseryWithdrawals/WithdrawalDetails/sections/NonOutplantWithdrawalTable.tsx
```diff
@@ -34,6 +36,14 @@ const columns = (): TableColumnType[] => [
   { key: 'total', name: strings.TOTAL, type: 'number' },
 ];
 
+const nurseryV2Columns = (): TableColumnType[] => [
```
karthikbtf, line 39: should this have a germinating quantity column?

src/utils/batch.tsx
```diff
@@ -0,0 +1,13 @@
+import { Batch } from 'src/types/Batch';
```
karthikbtf, line 1: this should be a .ts extension

src/components/NurseryWithdrawals/WithdrawalDetails/sections/OutplantWithdrawalTable.tsx
```diff
@@ -23,49 +36,93 @@ export default function OutplantWithdrawalTable({
...
+            species: species?.find((x) => x?.id === sp)?.scientificName ?? '',
+            to_subzone: subzone > -1 ? subzoneNames[subzone] ?? subzone?.toString() : '',
+            quantity: numericFormatter.format(speciesSubzoneMap[sp][subzone]),
+          });
+        }
+      }
+
+      const batchesMap: BatchesRow[] = [];
+
+      if (Object.keys(batchToSpeciesMap).length > 0) {
```
karthikbtf, line 95: maybe we can add a nurseryV2 check here as well so we aren't doing this in production until released

src/components/NurseryWithdrawals/WithdrawalDetails/sections/NonOutplantWithdrawalTable.tsx
```diff
@@ -42,76 +52,89 @@ export default function NonOutplantWithdrawalTable({
...
-        const ready = getFormattedValue('ready');
-        const total = getFormattedValue('total');
+      if (nurseryV2) {
+        setRowData(batchesMap);
+      } else {
+        setRowData(
+          Object.values(speciesBatchMap).map((speciesInfo: any) => {
+            const getFormattedValue = (key: string) => {
+              const value = speciesInfo[key];
+              return isNaN(value) ? value : numericFormatter.format(value);
```
karthikbtf, line 115: if NaN should we return ''?

karthikbtf (commented):
looks good, some comments..
question - do we need to consider germinating quantities in these tables for nursery v2? I don't see any such references..

src/components/NurseryWithdrawals/WithdrawalDetails/sections/NonOutplantWithdrawalTable.tsx
constanzauanini: I missed it on designs! Looks like for "Nursery Transfer" is not there, but it is for "Other/dead". But I saw you added the input for Nursery Transfer too. Should I add it to all of them?

karthikbtf: I just x-referenced the PRD. outplant doesn't have germinating, all other purposes do. Tagged you in a comment in prd

constanzauanini: Thanks!

karthikbtf (approved):
Nice 🎉

PR #1784, opened 2023-11-14
SW-4448 Species selector dropdown uses BE search endpoint for querying

karthikbtf:
- previous was doing a client-side in-memory search based on client input
- now pass the client input to the BE, similar to species table, for a consistent search fuzzy match response
- request fewer properties for faster query response feedback
- moved Species2Dropdown to src/components/common and renamed it as SpeciesSelector
- updated callees
- added a function in SpeciesService to suggest species for the dropdown
- removed the 'removeDiacritics' utils since it is no longer used, can add it back if it helps

### DIFF
--- a/src/services/SpeciesService.ts
+++ b/src/services/SpeciesService.ts
@@ -1,7 +1,9 @@
 import { paths } from 'src/api/types/generated-schema';
 import HttpService, { Response, ServerData } from './HttpService';
-import { Species, SpeciesDetails } from 'src/types/Species';
+import { Species, SpeciesDetails, SuggestedSpecies } from 'src/types/Species';
 import { GetUploadStatusResponsePayload, UploadFileResponse } from 'src/types/File';
+import SearchService from './SearchService';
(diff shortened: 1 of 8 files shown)


src/services/SpeciesService.ts
```diff
@@ -279,6 +281,63 @@ const getSpeciesNames = async (search: string): Promise<SpeciesNamesResponse> =>
...
+
+    const commonNameNode: FieldNodePayload = {
+      operation: 'field',
+      field: 'commonName',
+      type: 'Fuzzy',
+      values: [query],
+    };
+    searchValueChildren.push(commonNameNode);
+
+    const familyNode: FieldNodePayload = {
```
constanzauanini, line 324: Why are we adding family? If a species name is totally different from its family (not sure if that happens) it will still suggest that species name when writing a totally different thing in the input

constanzauanini (approved):
🎉

karthikbtf:
> 🎉

Thanks for the review!

PR #1785, opened 2023-11-14
SW-4438 Nursery updates for multiple batch withdrawal flow

karthikbtf:
- added germinating quantity for non-outplant withdrawals, updated logic etc.
- fixed a layout width on source nursery input on select purpose form

Terraware App 2023-11-14 14-56-57

### DIFF
--- a/src/components/Inventory/withdraw/flow/WithdrawalBatchesCellRenderer.tsx
+++ b/src/components/Inventory/withdraw/flow/WithdrawalBatchesCellRenderer.tsx
@@ -99,6 +99,18 @@ export default function WithdrawalBatchesCellRenderer(props: RendererProps<Table
     );
   }
 
+  if (column.key === 'germinatingQuantityWithdrawn') {
+    return (
+      <CellRenderer
+        index={index}
+        column={column}
+        value={createQuantityInput('germinatingQuantityWithdrawn', 'germinatingQuantity')}
+        row={row}
+        className={`${classes.text} ${classes.cell}`}
+      />
+    );
+  }
+
   if (column.key === 'readyQuantityWithdrawn') {
     return (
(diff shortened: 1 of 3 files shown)


src/components/Inventory/withdraw/flow/WithdrawalBatchesCellRenderer.tsx
```diff
@@ -132,7 +144,9 @@ export default function WithdrawalBatchesCellRenderer(props: RendererProps<Table
       <CellRenderer
         index={index}
         column={column}
-        value={numberFormatter.format(+row.readyQuantityWithdrawn + +row.notReadyQuantityWithdrawn)}
+        value={numberFormatter.format(
+          +row.readyQuantityWithdrawn + +row.notReadyQuantityWithdrawn + +row.germinatingQuantityWithdrawn
```
constanzauanini, line 148: Should validate nurseryV2 enabled before adding germinatingQuantityWidthrawn?

karthikbtf: I started out that way, but it will be 0 when nurseryV2 is disabled - we don't allow adding to the input. Also tested it.

constanzauanini (approved):

src/components/Inventory/withdraw/flow/WithdrawalBatchesCellRenderer.tsx
karthikbtf: i went ahead and added it anyways 😃

constanzauanini: Excellent! :)

PR #1786, opened 2023-11-15
SW-4450 Disable withdrawal purposes that aren't relevant

karthikbtf:
- disable outplant if there are no planting sites
- disable nursery transfer if there are no nurseries to transfer to
- update initial purpose value such that it's not pointing to a disabled purpose
- refactor code to use planting sites in redux store (App already dispatches the request)
- refactor code to use memoized nurseries options so it's not calculated on every render

Terraware App 2023-11-14 16-52-03


constanzauanini (approved):
Nice!

PR #1791, opened 2023-11-16
SW-4459 Update types and helper functions  in FE to capture organization type

karthikbtf:
- also update the service functions to create/update with new properties
- this is pending BE PR approvals, just a head-start for FE


constanzauanini (approved):
I know this requieres BE, but FE changes looks good

nickgraz (approved):

PR #1793, opened 2023-11-17
SW-4383 List Inventory by Nursery

constanzauanini:
Use same InventoryTable from inventoryBySpecies but change columns and filter results according to what is needed

Screenshot 2023-11-17 at 10 17 04

### DIFF
--- a/src/components/InventoryV2/InventoryListByNursery.tsx
+++ b/src/components/InventoryV2/InventoryListByNursery.tsx
@@ -10,10 +10,12 @@ import { useNumberFormatter } from 'src/utils/useNumber';
 import useDebounce from 'src/utils/useDebounce';
 import useForm from 'src/utils/useForm';
 import { InventoryFiltersType } from 'src/components/Inventory/InventoryFiltersPopover';
-import { InventoryResult, FacilityInventoryResult, InventoryResultWithFacilityNames } from 'src/components/InventoryV2';
+import { FacilityInventoryResult, InventoryResultWithSpeciesNames } from 'src/components/InventoryV2';
 import { makeStyles } from '@mui/styles';
-import { TableColumnType } from '@terraware/web-components';
 import strings from 'src/strings';
+import { TableColumnType } from '@terraware/web-components';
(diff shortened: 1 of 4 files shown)


src/components/InventoryV2/InventoryListByNursery.tsx
```diff
@@ -0,0 +1,193 @@
...
+    const nextResults = apiSearchResults?.reduce((acc, result) => {
+      const resultTyped = result as FacilityInventoryResult;
+      const indexFound = acc.findIndex((savedResult) => savedResult.facility_id === resultTyped.facility_id);
+
+      if (indexFound !== undefined && indexFound !== -1) {
+        const existingFacility = acc[indexFound];
+        acc[indexFound] = {
+          ...existingFacility,
+          germinatingQuantity: (
+            Number(existingFacility.germinatingQuantity) + Number(resultTyped['germinatingQuantity(raw)'])
```
nickgraz, line 105: Are these safe to do math on? `existingFacility.germinatingQuantity`

src/components/InventoryV2/InventoryListByNursery.tsx
```diff
@@ -0,0 +1,193 @@
...
+      searchSortOrder,
+    });
+
+    let updatedResult: InventoryResultWithSpeciesNames[] | undefined = [];
+
+    const nextResults = apiSearchResults?.reduce((acc, result) => {
+      const resultTyped = result as FacilityInventoryResult;
+      const indexFound = acc.findIndex((savedResult) => savedResult.facility_id === resultTyped.facility_id);
+
+      if (indexFound !== undefined && indexFound !== -1) {
```
nickgraz, line 100: This can also be `if (indexFound > -1) {`. Works as expected for `undefined` and `-1`

nickgraz (approved):
Looks good to me! One small question that is probably inconsequential

src/components/InventoryV2/InventoryListByNursery.tsx
constanzauanini: This page is only shown for nursery2, so I guess it is

PR #1797, opened 2023-11-20
SW-4468 Remove planting season months from UI

sgrimm:
We're going to be replacing the admin-specified planting season start and end
months with a new user-editable planting season schedule. Since we won't be
bothering to set the start/end months given the new feature is coming up, stop
showing them in the UI.


sgrimm:
Marked this as a draft while awaiting confirmation that it doesn't need to be gated by an opt-in feature flag.

charlie-tf (approved):
Checking in Vercel, and looks good there 👍

PR #1799, opened 2023-11-20
SW-4392 Batch modal updates for 'inventory for nursery' view

nickgraz:
- [Created](https://github.com/terraware/terraware-web/pull/1799/files#diff-7018a7b1c4244089b6c2a1c54fe6812f396bd237e2182b0c72a3ed1c4464d634) `BatchDetailsModalForNursery` for batch creation / viewing / updating from 'inventory for nursery' view
- [Added](https://github.com/terraware/terraware-web/pull/1799/files#diff-c203aedddfb6d12fcfaf0943168e9a88c78c4ac561ee86126563c32e21c903d9R123) `post2` function within HttpService that passes types down


src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
```diff
@@ -0,0 +1,544 @@
...
+      if (hasErrors()) {
+        setValidateFields(true);
+        return;
+      }
+
+      let response: (Response & BatchData) | (Response & BatchId) | undefined;
+      let responseQuantities: Partial<Response> = { requestSucceeded: true, error: undefined };
+
+      if (record.id === -1) {
+        if (record.accessionId && isBatch(record)) {
```
nickgraz, line 240: If there is an accession ID, we need to create a transfer (which will automatically create the batch)

src/components/InventoryV2/InventoryViewForNursery.tsx
```diff
@@ -21,10 +26,16 @@ export default function InventoryViewForNursery(): JSX.Element {
...
-  // TODO in SW-4392
-  // tslint:disable-next-line:no-empty
-  const onUpdateOpenBatch = () => {};
+  const setBatchNumber = (batchNum: string | null) => {
+    if (batchNum === null) {
+      query.delete('batch');
+    } else {
+      query.set('batch', batchNum);
+    }
+    history.replace(getLocation(location.pathname, location, query.toString()));
```
karthikbtf, line 37: does this open up the batch modal for a specific batch?

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
```diff
@@ -0,0 +1,544 @@
...
+          const nurseryTransferRecord: NurseryTransfer = {
+            date: getTodaysDateFormatted(timeZone),
+            destinationFacilityId: record.facilityId,
+            germinatingQuantity: record.germinatingQuantity,
+            notReadyQuantity: record.notReadyQuantity,
+            notes: record.notes,
+            readyByDate: record.readyByDate,
+            readyQuantity: record.readyQuantity,
+            // TODO figure out if I need this or if is auto populated by auth
+            // withdrawnByUserId: 0
```
karthikbtf, line 250: you can populate this from the active user, assuming this refers to the logged in user..
```
import { useUser } from 'src/providers';

...
const { user } = useUser();
...
 withdrawnByUserId: user.id,
```

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
```diff
@@ -0,0 +1,544 @@
...
+          };
+          const accessionResponse = await AccessionService.transferToNursery(nurseryTransferRecord, record.accessionId);
+          if (!accessionResponse.requestSucceeded || !accessionResponse.data) {
+            snackbar.toastError(accessionResponse.error || strings.GENERIC_ERROR);
```
karthikbtf, line 254: maybe generic error alone,, server-side errors are not localized yet

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
```diff
@@ -0,0 +1,544 @@
...
+      }
+
+      if (response && response.requestSucceeded && responseQuantities.requestSucceeded) {
+        reload();
+        onCloseHandler();
+      } else {
+        snackbar.toastError((response && response.error) || responseQuantities.error || strings.GENERIC_ERROR);
+      }
+    }
+  };
```
karthikbtf, line 289: a tad concerned this could be fragile.. if user closes the browser during the request, etc.. wondering if we need to think about a BE api that does all this..? To be fair we do have other places in the UI code that does some multi api flows such as creating/deleting sublocations as part of a facility create/edit.

karthikbtf (commented):
An initial cursory glance. Some comments.

src/components/InventoryV2/InventoryViewForNursery.tsx
nickgraz: No this just adds it to the history. Within the table, when a batch is selected, the modal opens and calls this passed in handler, which is just responsible for history and URL modification. This was taken from `src/components/InventoryV2/InventoryView.tsx`

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
nickgraz: I _think_ I can actually drop this line because of https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/seedbank/api/TransfersController.kt#L57, which would be the ideal behavior. What do you think?

karthikbtf: Good idea looking up the controller, yes sounds good 👍

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
nickgraz: I asked @sgrimm about it being implemented in the current API but he mentioned I should just call the two APIs consecutively. It wouldn't be bulletproof, but I could add an `onunload` event handler to warn the user about leaving the page while the requests are in flight, what do you think? Not sure how well it will work with a SPA though... Alternatively I could move these to redux, that way they will execute as desired regardless of the user's navigation (unless they leave the site obviously)

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
```diff
@@ -0,0 +1,544 @@
...
+                        ...previousValue,
+                        accessionId: Number(accessionId),
+                      }))
+                    }
+                    fullWidth
+                  />
+                </Grid>
+
+                <Grid item xs={12} padding={theme.spacing(1, 0, 1, 2)}>
+                  <Dropdown
```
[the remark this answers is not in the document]
nickgraz: Ah, yes, in the original designs it was singular but you're right, it needs to be updated.

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
karthikbtf: Gotcha. Maybe something to discuss as a team at some point. The `onunload` at a minimum sounds good to me. I can update other flows later.

nickgraz: Alright I added a note to bring it up at the next round table. How's this for now? 

I had a 5 second timeout between server calls in this example below:

karthikbtf: This is great!! Thanks so much 🎉

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
```diff
@@ -210,6 +211,25 @@ export default function BatchDetailsModal(props: BatchDetailsModalProps): JSX.El
...
   }, [selectedBatch, nurseryId, setRecord, selectedOrganization]);
 
+  useEffect(() => {
+    const onWindowBeforeUnload = (event: BeforeUnloadEvent) => {
+      event.preventDefault();
+
+      if (multiStepRequestInFlight) {
+        // This message does not seem to make it to the alert
+        event.returnValue = window.confirm(
+          'There is a request in flight, are you sure you want to navigate away from this page?'
```
karthikbtf: add this to strings for i18n?

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
```diff
@@ -215,17 +215,15 @@ export default function BatchDetailsModal(props: BatchDetailsModalProps): JSX.El
...
-        // This message does not seem to make it to the alert
-        event.returnValue = window.confirm(
-          'There is a request in flight, are you sure you want to navigate away from this page?'
-        );
-      } else {
-        event.returnValue = true;
-      }
+      // This message does not seem to make it to the alert
+      event.returnValue = window.confirm(
+        'There is a request in flight, are you sure you want to navigate away from this page?'
```
karthikbtf, line 220: add this to strings for i18n?

src/components/InventoryV2/view/BatchDetailsModalForNursery.tsx
```diff
@@ -442,21 +440,28 @@ export default function BatchDetailsModal(props: BatchDetailsModalProps): JSX.El
...
+                    }}
+                    onRemove={(subLocationId: string) => {
+                      setRecord((previousValue) => ({
+                        ...previousValue,
+                        subLocationIds:
+                          previousValue?.subLocationIds?.filter((id: number) => `${id}` !== subLocationId) || [],
+                      }));
+                    }}
+                    options={
+                      new Map(availableSubLocations.map((subLocation) => [`${subLocation.id}`, subLocation.name]))
```
karthikbtf, line 460: the value can stay as a number, no need to string-ify it i think, multiselect component supports generics for key and value here

karthikbtf (approved):
lgtm! a few suggestions/comments

karthikbtf (approved):
Nice 🎉

PR #1801, opened 2023-11-21
SW-4459 Additional questions in the Create Org modal

karthikbtf:
- added 3 new questions, one mandatory
- for organization type, if user picks 'Other', we show an additional details text input
- the modal is getting taller and we want it scrolled on smaller form factors, created a shell scrollable dialogbox component
- we can't scroll the dialog body as long as we have dropdowns whose entries need to be scrolled outside the parent box

Terraware App 2023-11-21 11-17-57

Terraware App 2023-11-21 11-17-36

Terraware App 2023-11-21 11-17-25


nickgraz (approved):
Looks good to me!

PR #1804, opened 2023-11-21
I accidentally committed this! Reversing it back to how it was

nickgraz:
I probably should have asked a while ago... how do I enable this locally without modifying this file?


charlie-tf (approved):
So the switch is within "Opt-In Features" here:
image

nickgraz:
Thanks!

karthikbtf:
A lot of the options like `enabled` etc were looking ahead in how we might use the capability.. I should probably remove these unused options to avoid confusion.
We `enable` a feature today by removing the feature flag throughout the codebase, instead of setting `enabled` to true.

PR #1808, opened 2023-11-22
SW-4388 Export from inventory page

constanzauanini:
Allow downloading report form "By species", "By nursery" and "By batch" tabs.


src/components/InventoryV2/DownloadReportModal.tsx
```diff
@@ -0,0 +1,20 @@
+import ExportCsvModal from 'src/components/common/ExportCsvModal';
+import NurseryInventoryService, { SearchInventoryParams } from 'src/services/NurseryInventoryService';
+
+interface DownloadReportModalProps {
```
karthikbtf, line 4: we have a mixed usage of `type` vs `interface`.. maybe we should stick to one?

src/components/InventoryV2/index.tsx
```diff
@@ -226,12 +238,12 @@ export default function Inventory(props: InventoryProps): JSX.Element {
...
                 id: 'batches_by_species',
                 label: strings.BY_SPECIES,
-                children: <InventoryListBySpecies />,
+                children: <InventoryListBySpecies setReportData={setReportData} />,
               },
               {
                 id: 'batches_by_nursery',
                 label: strings.BY_NURSERY,
-                children: <InventoryListByNursery />,
+                children: <InventoryListByNursery setReportData={setReportData} />,
```
karthikbtf, line 246: interesting, this doesn't error out? setReportData type is different here and in the component props

src/components/InventoryV2/DownloadReportModal.tsx
```diff
@@ -0,0 +1,20 @@
...
+  reportData?: SearchInventoryParams;
```
karthikbtf, line 5: should this be a required prop?

src/components/InventoryV2/DownloadReportModal.tsx
```diff
@@ -0,0 +1,20 @@
...
+  open: boolean;
+  onClose: () => void;
+}
+
+export default function DownloadReportModal(props: DownloadReportModalProps): JSX.Element {
+  const { reportData, open, onClose } = props;
+
+  const onExport = async () => {
+    if (reportData) {
```
karthikbtf, line 14: this check should probably be in the calling code.. the purpose of this modal is to download csv and should be ok to require the reportData prop

karthikbtf (commented):
Mostly good, a few comments.

constanzauanini:
> > Allow downloading report form "By species" and "By nursery" tabs. "By batch" is not ready yet, waiting table to be merged.
> > For the "by nursery" tab, the downloaded csv is not exactly the same as what we get in the table, because in the FE we are grouping by nursery, but the backend don't do that, so we get different rows for each nursery in the csv. Should we add BE changes to allow that? Or is it ok having different rows?
> 
> Interesting, maybe we should bring this up with the team - also check whether this is possible with a BE search instead of client doing the grouping?

I will ask in slack

src/components/InventoryV2/index.tsx
constanzauanini: It doesn't error out, I see we do that in other places. Should I change it to be React.Dispatch ... ?

src/components/InventoryV2/InventoryListByNursery.tsx
```diff
@@ -52,7 +52,11 @@ const columns = (): TableColumnType[] => [
   { key: 'totalQuantity', name: strings.TOTAL, type: 'string', tooltipTitle: strings.TOOLTIP_TOTAL_QUANTITY },
 ];
 
-export default function InventoryListByNursery() {
+type InventoryListByNurseryProps = {
```
karthikbtf, line 55: here's a `type` :) 
i have mostly been using `type` myself... we should decide on one for sure. Not sure if there's a preference among us.

src/components/InventoryV2/index.tsx
karthikbtf: My suggestion is to not make it a React.Dispatch, that's a very internal semantic to this component. Leave it as is maybe if we are doing similar stuff elsewhere.

src/components/InventoryV2/InventoryListByNursery.tsx
constanzauanini: Oh, I searched in the whole project without the "_=_", that's why I only got interfaces in my search result. But you are right, we have a mix of both. I can change the interface to be type if that's our preference

constanzauanini:
> > > Allow downloading report form "By species" and "By nursery" tabs. "By batch" is not ready yet, waiting table to be merged.
> > > For the "by nursery" tab, the downloaded csv is not exactly the same as what we get in the table, because in the FE we are grouping by nursery, but the backend don't do that, so we get different rows for each nursery in the csv. Should we add BE changes to allow that? Or is it ok having different rows?
> > 
> > 
> > Interesting, maybe we should bring this up with the team - also check whether this is possible with a BE search instead of client doing the grouping?
> 
> I will ask in slack

Solved, grouped data come from BE now

src/components/InventoryV2/InventoryListByNursery.tsx
```diff
@@ -84,65 +76,25 @@ export default function InventoryListByNursery() {
...
+
+    setReportData({
       organizationId: selectedOrganization.id,
       query: debouncedSearchTerm,
-      facilityIds: filters.facilityIds || nurseries.map((nr) => nr.id),
+      facilityIds: filters.facilityIds,
       searchSortOrder,
     });
 
-    let updatedResult: InventoryResultWithSpeciesNames[] | undefined = [];
```
constanzauanini, line 94: Removed all this front-end managment of groupping, since it comes from BE now

src/components/InventoryV2/DownloadReportModal.tsx
nickgraz: I think `type` is supposed to be used for more "transient" data types and `interface` is supposed to be used for "public facing" (or I guess exported and used in several places?) data types. This theory could be outdated! I agree we should agree on some convention... admittedly I think I just use `type` everywhere right now 😓

src/components/InventoryV2/InventoryListByNursery.tsx
```diff
@@ -84,65 +76,25 @@ export default function InventoryListByNursery() {
...
-        acc.push(transformedResult);
-      }
-      return acc;
-    }, [] as InventoryResultWithSpeciesNames[]);
+    const apiSearchResults = await NurseryInventoryService.searchInventoryByNursery({
+      organizationId: selectedOrganization.id,
+      query: debouncedSearchTerm,
+      facilityIds: filters.facilityIds,
+      searchSortOrder,
+    });
```
nickgraz, line 92: 🔥

src/services/NurseryInventoryService.ts
```diff
@@ -67,6 +71,7 @@ export type SearchInventoryParams = {
   searchSortOrder?: SearchSortOrder;
   query?: string;
   facilityIds?: number[];
+  isDowloading?: boolean;
```
karthikbtf, line 74: typo isDowloading, or maybe call it isCsvExport?

karthikbtf (approved):
One suggestion, otherwise thanks for following up on the BE changes in slack and here 🎉

nickgraz (approved):
Looks good!

constanzauanini:
> Looks good!


Thanks for the review! I just added the download report for the "By batch" tab

src/services/NurseryBatchService.ts
```diff
@@ -218,7 +230,9 @@ const getAllBatches = async (
         },
       ],
     },
-    fields: [...DEFAULT_BATCH_FIELDS, 'species_id', 'species_scientificName', 'species_commonName'],
+    fields: isCsvExport
+      ? [...REPORT_BATCH_FIELDS]
+      : [...DEFAULT_BATCH_FIELDS, 'species_id', 'species_scientificName', 'species_commonName'],
```
nickgraz, line 235: You might be able to use `NURSERY_BATCHES_FIELDS` here

nickgraz:
> > Looks good!
> 
> Thanks for the review! I just added the download report for the "By batch" tab

One small comment, otherwise looks good!

src/services/NurseryInventoryService.ts
constanzauanini: Fixed!

PR #1809, opened 2023-11-22
Use unique ids for tables, ids are used to generate user preference names

karthikbtf:
- We use the [table ids to generate user preference names](https://github.com/terraware/terraware-web/blob/main/src/components/common/table/index.tsx#L54), to store table column ordering iff user changes the column locations (drag n drop)
- This table id design decision seems to be backfiring a bit now, we don't have any checks in place to ensure unique table ids (sigh)
- For now, fix the table ids to be unique (maybe we can keep an eye out for such changes going forward)
- Open to other ways of ensuring unique ids (maybe leverage some typescript magic?)
- Note that changing the column definitions would require changing the table ids (since preserved columns would no longer match the definition)


nickgraz (approved):
🙏

PR #1816, opened 2023-11-22
SW-4481 Disallow deletion of species that are in-use

karthikbtf:
- in-use => batches, accessions or plantings referring to the species id
- updated species modal to fetch in use species and updated messaging to user (see screenshots)
- messaging may need some rewording (iterate on it separately maybe)
- may need some help in refining the search (currently fetching species ids for all batches, accessions, plantings in the org - only the species ids)

Terraware App 2023-11-28 08-02-03

Terraware App 2023-11-28 08-01-44

Terraware App 2023-11-28 08-01-24

Terraware App 2023-11-28 08-03-29


constanzauanini (approved):
Lgtm! I think we could include the species names (that can/can't be deleted) in the modal, but maybe ask design about that

src/components/Species/index.tsx
```diff
@@ -557,12 +547,10 @@ export default function SpeciesList({ reloadData, species }: SpeciesListProps):
...
-  const deleteSelectedSpecies = async () => {
-    if (selectedSpeciesRows.length > 0) {
-      await Promise.all(
-        selectedSpeciesRows.map(async (iSelectedSpecies) => {
-          await SpeciesService.deleteSpecies(iSelectedSpecies.id, selectedOrganization.id);
-        })
+  const deleteSelectedSpecies = async (speciesIds: number[]) => {
+    if (speciesIds.length > 0) {
+      await Promise.allSettled(
+        speciesIds.map((id: number) => SpeciesService.deleteSpecies(id, selectedOrganization.id))
```
nickgraz, line 553: Should we do something with errors?

nickgraz (approved):
Looks good, one question about delete errors but isn't blocking.

src/services/SpeciesService.ts
```diff
@@ -330,6 +330,62 @@ const suggestSpecies = async (organizationId: number, query: string): Promise<Su
...
+  const params: SearchRequestPayload = {
+    prefix: '',
+    fields: [
+      'facilities.accessions.species_id',
+      'facilities.accessions.species_scientificName',
+      'batches.species_id',
+      'batches.species_scientificName',
+      'plantingSites.populations.species_id',
+      'plantingSites.populations.species_scientificName',
+    ],
```
karthikbtf, line 348: @sgrimm thoughts on this search query - to identify used species in an org?
This function returns a map of species id to scientific name, hence the _scientificName fields.

karthikbtf: Thanks, yes, there are duplicates returned. Code does some de-duping. API sounds good, l will take that approach.

src/components/Species/DeleteSpeciesModal.tsx
```diff
@@ -1,16 +1,95 @@
...
+  speciesToDelete: SpeciesSearchResultRow[];
 }
 
-export default function DeleteSpeciesDialog(props: DeleteSpeciesDialogProps): JSX.Element {
-  const { onClose, open, onSubmit } = props;
+export default function DeleteSpeciesDialog(props: DeleteSpeciesDialogProps): JSX.Element | null {
+  const { onClose, open, onSubmit, speciesToDelete } = props;
+  const theme = useTheme();
+  const { selectedOrganization } = useOrganization();
+  const [inUseSpecies, setInUseSpecies] = useState<Record<number, string>>();
```
sgrimm, line 20: I thought the key type of a `Record` had to be a string (or a union type of string values) since JS objects always have string keys.

karthikbtf: The string/union type of strings is for utilities such as `Pick` or `Omit` where you pick keys from an existing type.
Record supports number in addition to strings, string literals. I suppose this is only for type checking, behind the scenes it is stored as a string.

karthikbtf:
@constanzauanini  updated PR to use the search API (BE PR was merged). Please review again when you can.

constanzauanini (approved):
Looks cleaner now with the change on the BE!

PR #1818, opened 2023-11-24
SW-3933 Fix input elements are as wide as the orgs table

constanzauanini:
I wasn't sure if this is what we wanted here:

Terraware App

### DIFF
--- a/src/components/MyAccount/index.tsx
+++ b/src/components/MyAccount/index.tsx
@@ -391,148 +391,150 @@ const MyAccountContent = ({
             minWidth: 'fit-content',
           }}
         >
-          <Grid container spacing={3}>
-            <Grid item xs={12}>
-              <Typography fontSize='20px' fontWeight={600}>
-                {strings.GENERAL}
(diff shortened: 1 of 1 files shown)


src/components/MyAccount/index.tsx
```diff
@@ -391,148 +391,150 @@ const MyAccountContent = ({
...
-              sx={{ '&.MuiGrid-item': { paddingTop: theme.spacing(isMobile ? 3 : 2) } }}
-            >
-              {edit ? (
-                <LocaleSelector
-                  onChangeLocale={(newValue) => setLocaleSelected(newValue)}
-                  localeSelected={localeSelected}
-                  fullWidth={true}
-                />
-              ) : (
+          <Box sx={isMobile ? { width: 'calc(100vw - 72px)' } : {}}>
```
constanzauanini, line 394: Adding this box is the only change here

nickgraz (approved):
Thanks for fixing that up

PR #1820, opened 2023-11-27
Changes to update react to v18

constanzauanini:
We will need to upgrade web-components to v18 before terraware-web, because if not we have two versions of react and that gives this error:

"You might have mismatching versions of React and the renderer"

React-router-dom was updated to v6 to be compatible with react18 and that's why there are changes on all routers and routes.

Also @mui/styles is the JSS legacy styling solution of MUI and compatibility with React 18 is not guaranteed. Styles are not working correctly in dev, they work correctly in vercel preview, but we need to remove all uses of mui/styles and instead use [@mui/system] (sx prop) to guarantee they work as expected 

Finally, TSLint has been deprecated as of 2019 and was not working correctly with react 18, so I changed to eslint and there are lots of "new" linter errors I fixed (or disabled for now). We can improve this config and fixes later


src/components/Reports/ReportEdit.tsx
```diff
@@ -67,11 +67,13 @@ export default function ReportEdit(): JSX.Element {
 
   useEffect(() => {
     const getReport = async () => {
-      const result = await ReportService.getReport(reportIdInt);
-      if (result.requestSucceeded && result.report) {
-        setReport(result.report);
-      } else {
-        snackbar.toastError(strings.GENERIC_ERROR, strings.REPORT_COULD_NOT_OPEN);
+      if (reportIdInt) {
```
karthikbtf, line 70: I suppose reportIdInt will always be > 0 but we should maybe check for reportIdInt !== undefined or !isNaN(reportIdInt) in all these changes (since that's what we were trying to fix)?

.storybook/preview.tsx
```diff
@@ -21,29 +21,32 @@ export default preview;
...
-          <StyledEngineProvider injectFirst>
-            <MemoryRouter initialEntries={['/']}>
-              <Provider store={store}>
-                <Story />
-              </Provider>
-            </MemoryRouter>
-          </StyledEngineProvider>
-        </LocalizationProvider>
-      </ThemeProvider>
+      <>
```
karthikbtf, line 32: why this change?

.storybook/preview.tsx
```diff
@@ -21,29 +21,32 @@ export default preview;
 import theme from '../src/theme';
 import { useState } from 'react';
 import { LocalizationProvider } from '../src/providers';
+import React from 'react';
```
karthikbtf, line 24: move this to first line?

.storybook/preview.tsx
constanzauanini: Probably not necessary, will remove it

src/components/Reports/ReportEdit.tsx
constanzauanini: Yes, we.need to check !== undefined now, but I thought _if (reportIdInt) {_ would be ok

karthikbtf:
> > can't recall why web-components had a different react version, maybe funder-web? anyways this is a good upgrade.. will need to test the changes for sure 👍
> 
> Web-components had same version as terraware-web (17.0.2) , why do you sya it's different? Yes, this still needs more testing

I thought they were different that's why this PR. My confusion.

src/components/Reports/ReportEdit.tsx
karthikbtf: I see similar code elsewhere, maybe it's ok.. or just check for !isNaN (instead of undefined)

constanzauanini:
> > > can't recall why web-components had a different react version, maybe funder-web? anyways this is a good upgrade.. will need to test the changes for sure 👍
> > 
> > 
> > Web-components had same version as terraware-web (17.0.2) , why do you sya it's different? Yes, this still needs more testing
> 
> I thought they were different that's why this PR. My confusion.

Oh sorry I wasn't clear. We wanted to upgrade react version here, and for doing that, we also needed to do it in web-components. That's why I upgraded both

src/App.tsx
```diff
@@ -90,15 +50,46 @@ function AppContent() {
         <TopBarContent setShowNavBar={setShowNavBar} />
       </TopBar>
 
-      <div className={classes.container}>
+      <Box
+        sx={{
```
karthikbtf, line 54: useMemo for these styles? It is a root level component and maybe worth that optimization.

src/components/Map/Map.tsx
```diff
@@ -447,7 +447,7 @@ export default function Map(props: MapProps): JSX.Element {
           mapStyle={MapViewStyles[mapViewStyle]}
           initialViewState={{
             bounds: hasEntities ? [options.bbox.lowerLeft, options.bbox.upperRight] : undefined,
-            fitBoundsOptions: hasEntities ? { padding: 20, linear: true } : undefined,
+            fitBoundsOptions: hasEntities ? { padding: 20 } : undefined,
```
karthikbtf, line 450: Why this change? Not sure what the behavioral change is.

src/components/Reports/ReportEdit.tsx
```diff
@@ -69,11 +71,13 @@ export default function ReportEdit(): JSX.Element {
...
```
karthikbtf, line 74: if reportIdInt is -1 we shouldn't be fetching the report.

src/components/Reports/ReportEdit.tsx
```diff
@@ -85,22 +89,26 @@ export default function ReportEdit(): JSX.Element {
...
-    await Promise.all(
-      initialReportFiles?.map((f) => {
-        if (!updatedReportFiles?.includes(f)) {
-          return ReportService.deleteReportFile(reportIdInt, f.id);
-        }
-        return undefined;
-      }) ?? []
-    );
-    await Promise.all(newReportFiles?.map((f) => ReportService.uploadReportFile(reportIdInt, f)) ?? []);
+    if (reportIdInt) {
```
karthikbtf, line 92: check for -1

src/components/Reports/ReportEdit.tsx
```diff
@@ -85,22 +89,26 @@ export default function ReportEdit(): JSX.Element {
...
+      await Promise.all(newReportFiles?.map((f) => ReportService.uploadReportFile(reportIdInt, f)) ?? []);
+    }
   };
 
   useEffect(() => {
     const getReport = async () => {
-      const result = await ReportService.getReport(reportIdInt);
-      if (result.requestSucceeded && result.report) {
-        setCurrentUserEditing(result.report.lockedByUserId === user?.id);
+      if (reportIdInt) {
```
karthikbtf, line 107: -1 check

src/components/Reports/ReportEdit.tsx
```diff
@@ -144,15 +152,17 @@ export default function ReportEdit(): JSX.Element {
       setBusyState(false);
     }
 
-    if (!saveResult || saveResult.requestSucceeded) {
+    if ((!saveResult || saveResult.requestSucceeded) && reportIdInt) {
```
karthikbtf, line 155: -1 check and in other places.. only because we have a path that initializes it to -1
Maybe create a function that checks if reportIdInt is valid and use that in all the places?

src/components/Reports/ReportView.tsx
```diff
@@ -51,13 +51,15 @@ export default function ReportView(): JSX.Element {
 
   const confirmEdit = async () => {
     // lock the report
-    const lockResult = await ReportService.forceLockReport(reportIdInt);
+    if (reportIdInt) {
```
karthikbtf, line 54: -1 check

src/components/common/Chart/Chart.tsx
```diff
@@ -98,71 +98,76 @@ function ChartContent(props: ChartContentProps): JSX.Element {
...
-      if (!label) {
-        return {};
-      }
-      return {
-        title: {
-          display: true,
-          align: 'center',
-          text: label,
-        },
+    // used to prevent double render on dev scope (rect 18)
```
karthikbtf, line 106: `react 18`

src/components/common/Chart/Chart.tsx
```diff
@@ -98,71 +98,76 @@ function ChartContent(props: ChartContentProps): JSX.Element {
   const canvasRef = useRef<HTMLCanvasElement>(null);
   const [chart, setChart] = useState<ChartJS | null>(null);
   const theme = useTheme();
+  const initialized = useRef(false);
```
karthikbtf, line 101: Will this be reset to false when the component's props change?

karthikbtf (commented):
An initial pass up until common/Chart. Thanks for this effort.

src/components/Map/Map.tsx
constanzauanini: I think that linear prop was not doing anything because it's not a valid option to pass, but will re-check

src/components/Monitoring/dashboard/PVBatteryChart.tsx
```diff
@@ -214,12 +214,12 @@ export default function PVBatteryChart(props: PVBatteryChartProps): JSX.Element
     const getChartData = async () => {
       if (selectedPVBatteryPeriod) {
         const timePeriodParams = getTimePeriodParams(selectedPVBatteryPeriod, timeZone);
-        const startTime = timePeriodParams.start;
-        const endTime = timePeriodParams.end;
-        if (BMU) {
+        const startTime = timePeriodParams.start.toISO();
+        const endTime = timePeriodParams.end.toISO();
+        if (BMU && startTime && endTime) {
```
karthikbtf, line 219: If it is possible with no errors, can we change this to

```
if (BMU && startTime.isValid && endTime.isValid) {
  const response = await getTimeSeriesHistory(
    startTime.toISO(),
    endTime.toISO(),
   ...
```

And remove toISO from lines 217 and 218.

Same for TemperatureHumidityChart (note that isValid is not a function just a property on DateTime)

src/components/common/Chart/Chart.tsx
constanzauanini: It is not reset, but the only prop that we want to re-trigger the useEffect (where this _"initialized"_ variable is used) is locale, and when changing locale, charts re-renders correctly (I don't know why)

src/scenes/NurseriesRouter/index.tsx
```diff
@@ -1,8 +1,7 @@
 import React, { useCallback } from 'react';
-import { Route, Switch } from 'react-router-dom';
+import { Route, Routes } from 'react-router-dom';
 
 import EmptyStatePage from 'src/components/emptyStatePages/EmptyStatePage';
-import { APP_PATHS } from 'src/constants';
```
karthikbtf, line 5: Why get rid of APP_PATHS and the hardcoded route paths below?

src/scenes/SeedBanksRouter/index.tsx
```diff
@@ -20,20 +19,12 @@ const SeedBanksRouter = () => {
...
-        <SeedBankDetailsView />
-      </Route>
-      <Route exact path={APP_PATHS.SEED_BANKS}>
-        {getSeedBanksView()}
-      </Route>
-    </Switch>
+    <Routes>
+      <Route path={'/new'} element={<SeedBankView />} />
+      <Route path={'/:seedbankId/edit'} element={<SeedBankView />} />
+      <Route path={':/seednankId'} element={<SeedBankDetailsView />} />
```
karthikbtf, line 25: typo `seednankId`

src/scenes/NurseriesRouter/index.tsx
karthikbtf: Ah, because these paths are just fragments of the original url.. makes sense. Kinda sad we can't use the same constants now, but it's a react change.

src/scenes/RedirectsRouter/index.tsx
```diff
@@ -1,41 +1,23 @@
...
-        <Redirect to={APP_PATHS.SEEDS_DASHBOARD} />
-      </Route>
-      <Route path={APP_PATHS.SPECIES + '/'}>
-        <Redirect to={APP_PATHS.SPECIES} />
-      </Route>
-      <Route path='*'>
-        <Redirect to={APP_PATHS.HOME} />
-      </Route>
-    </Switch>
+      {/* <Route path={APP_PATHS.ACCESSIONS + '/*'} element={<Navigate to={APP_PATHS.ACCESSIONS} />} /> */}
```
karthikbtf, line 11: why commented out?

src/scenes/RedirectsRouter/index.tsx
```diff
@@ -1,41 +1,23 @@
...
+      <Route path={APP_PATHS.CHECKIN + '/'} element={<Navigate to={APP_PATHS.CHECKIN} />} />
```
karthikbtf, line 12: do these paths need to be fragments as well? `/checkin` etc. ?

karthikbtf (commented):
@constanzauanini looked through the rest. Thanks again for this detailed effort.
Added some comments/issues.

src/scenes/NurseriesRouter/index.tsx
constanzauanini: Yes, now if a the routers is a subrouter, the first part of the url is already "included"

src/scenes/RedirectsRouter/index.tsx
constanzauanini: We don't need this redirect router anymore. I commented it to test, and looks like it's not necessary, because routes now behaves like we wanted, the '/*' is inside each sub-router

src/scenes/RedirectsRouter/index.tsx
constanzauanini: This is not used anymore

tslint-to-eslint-config.log
```diff
@@ -0,0 +1,31 @@
+6 ESLint rules behave differently from their TSLint counterparts:
```
karthikbtf, line 1: Do we need this checked in?

src/services/ParticipantProjectService.ts
```diff
@@ -168,6 +169,7 @@ const list = async (
 };
 
 const update = async (participantProject: ParticipantProject): Promise<Response> => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
karthikbtf, line 172: there are no unused vars here, interesting

src/services/ParticipantProjectService.ts
```diff
@@ -81,6 +81,7 @@ const download = async ({
   projectId: number;
   projectMeta?: ProjectMeta;
   organization?: AcceleratorOrg;
+  // eslint-disable-next-line @typescript-eslint/require-await
```
karthikbtf, line 84: can we `return await` on line 175 or remove the `async` from the function definition? This warning here is legit.

tslint-to-eslint-config.log
constanzauanini: Not necessary. I left it just in case we wanted to check it later

src/services/ParticipantProjectService.ts
constanzauanini: projectId is not used. Will fix it

karthikbtf (approved):

PR #1822, opened 2023-11-28
Update Cypress session for new Spring Boot version

sgrimm:
We're upgrading Spring Boot on the server side, which means the canned session
object the Cypress test suite uses to log in needs to be updated to use the new
serialized session data format.


sgrimm:
This depends on https://github.com/terraware/terraware-server/pull/1512 -- the Cypress suite will fail until that PR is merged.

karthikbtf (approved):

PR #1823, opened 2023-11-28
SW-4488  Projects List View and Empty state

constanzauanini:
Adding the table with projects and the empty state when no projects.


src/components/Projects/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const history = useHistory();
+  const [temporalSearchValue, setTemporalSearchValue] = useState('');
+  const debouncedSearchTerm = useDebounce(temporalSearchValue, 250);
+  const [results, setResults] = useState<Project[]>();
+  const { isMobile } = useDeviceInfo();
+  const contentRef = useRef(null);
+  const { activeLocale } = useLocalization();
+
+  const search = useCallback(
+    async (searchTerm: string, skipTfContact = false) => {
```
karthikbtf, line 63: can remove the skipTfContact

src/components/Projects/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+          },
+        ],
+        count: 0,
+      };
+
+      if (searchField) {
+        params.search.children.push(searchField);
+      }
+
+      const searchResults = await SearchService.search(params);
```
karthikbtf, line 100: Do we want to use redux here, I understand it's a copy of existing code (which isn't the best to use as a template)... 
either move all this into projects service and just call the function with search term and org id, which returns back a list of Project[]?
Or use redux..?

src/services/ProjectsService.ts
```diff
@@ -0,0 +1,48 @@
...
+  );
+
+  return response;
+};
+
+/**
+ * Exported functions
+ */
+const ProjectsService = {
+  listProjects,
```
karthikbtf, line 45: to be used later?

karthikbtf (commented):
Mostly good, some thoughts.

src/components/Projects/index.tsx
nickgraz: I agree it should be moved. I think in the past I've been putting these sorts of search queries into a service and using redux more for 1:1 entity APIs. Is that correct? What is the general pattern to deciding whether it goes into a service and is used in a component's state versus redux state?

karthikbtf: Redux was recently introduced into the codebase with the idea that all new functionality would 'ideally' use redux.
We didn't consider differentiating between using component state vs redux state for business data - maybe we should be open to that for sure.
As far as what goes into a service, anything involving wrangling types and parsing responses into types, etc. Search is a very good use-case, not sure components need to be aware search request payload semantics.. good candidate for services?

karthikbtf: @constanzauanini , at a minimum move the 'payload creation', search call and 'response parsing' into a service function. Should be good to go after that.

constanzauanini: Yes, I just copied existing code that was part of a component. But it should be part of a service. Will move it

src/services/ProjectsService.ts
constanzauanini: I'm using it to show the empty state if no projects

src/services/ProjectsService.ts
```diff
@@ -38,11 +40,55 @@ const listProjects = async (organizationId: number, locale?: string | null): Pro
   return response;
 };
 
+/**
+ * Search projects
+ */
+const searchProjects = async (organizationId: number, query?: string): Promise<SearchResponseElement[] | null> => {
```
karthikbtf, line 46: can you return Project[] | null

src/components/Projects/index.tsx
```diff
@@ -60,54 +59,9 @@ export default function ProjectsList(): JSX.Element {
...
-          name: result.name as string,
-          description: result.description as string,
-          id: result.id as number,
-          organizationId: result.organizationId as number,
-        });
-      });
-      return projectResults;
+    async (searchTerm: string) => {
+      const searchResults = await ProjectsService.searchProjects(selectedOrganization.id, searchTerm);
+      return searchResults as Project[];
```
karthikbtf, line 64: do the casting as Project[] in the service call instead

karthikbtf (approved):
lgtm, one suggestion and feel free to merge after that

karthikbtf (approved):
🎉

PR #1824, opened 2023-11-28
SW-4385 Change Batch Status

charlie-tf:
- Add table row menu
- Change quantities from menu
- Row click on checkbox only
- Refresh table after changes
image
image


nickgraz (approved):
Nice!

PR #1825, opened 2023-11-28
SW-4516 Refactor InventorySeedlingsTable for use in two views

nickgraz:
- Used in both 'inventory by species' and 'inventory by nursery' views
- Rename a few files for consistency
- Allows us to build more functionality into the tables with less duplication
- Separates the origin-specific logic from the UI


src/components/InventoryV2/view/InventorySeedlingsTableForNursery.tsx
```diff
@@ -0,0 +1,99 @@
...
+  { key: 'totalQuantity', name: strings.TOTAL, type: 'string' },
+  { key: 'totalQuantityWithdrawn', name: strings.WITHDRAWN, type: 'string' },
+  { key: 'readyByDate', name: strings.EST_READY_DATE, type: 'date' },
+  { key: 'addedDate', name: strings.DATE_ADDED, type: 'date' },
+  { key: 'withdraw', name: '', type: 'string' },
+  { key: 'quantitiesMenu', name: '', type: 'string' },
+];
+
+export default function InventorySeedlingsTableForNursery(props: InventorySeedlingsTableForNurseryProps): JSX.Element {
+  const facilityId = props.nurseryId;
```
nickgraz, line 31: This is done with the intention that we might use the `InventorySeedlingsTable` for other facility types in the future.

src/components/InventoryV2/view/InventorySeedlingsTable.tsx
```diff
@@ -414,7 +428,7 @@ export default function InventorySeedlingsTable(props: InventorySeedlingsTablePr
           <Box marginTop={theme.spacing(2)}>
             <Table
               id='inventory-seedlings-table'
```
karthikbtf, line 430: Make this `{`inventory-seedlings-table-${origin.toLowerCase()}`}` so we have unique table ids by use-case.
The id is used as the preference name to store table column ordering, if user shuffles the columns.

karthikbtf (commented):
Nice refactor, one comment so far..

src/components/InventoryV2/view/InventorySeedlingsTableForNursery.tsx
```diff
@@ -0,0 +1,99 @@
...
+    (
+      orgId: number,
+      originId: number,
+      searchFields: FieldNodePayload[],
+      searchSortOrder: SearchSortOrder | undefined
+    ): Promise<SearchResponseElement[] | null> => {
+      // TODO it seems this was not implemented
+      // tslint:disable-next-line:no-console
+      console.log(
+        'TODO - implement batches export for inventory by nursery',
```
constanzauanini, line 73: I don't see this export in designs. For species either. I think it's replaced with the exports of the lists tables

nickgraz: I see, OK I will just implement it as a noop for the nursery view

karthikbtf:
Looks good, nice refactor.
The column names don't seem to be rendering.

Terraware App 2023-11-29 09-54-16

nickgraz:
> Looks good, nice refactor. The column names don't seem to be rendering.
> 
> Terraware App 2023-11-29 09-54-16

Hmm... the `strings` stopped working. Should be an easy fix 
SCR-20231129-kbmx

karthikbtf (approved):
Nice!

PR #1830, opened 2023-11-29
SW-4386 Show/Hide Empty Batches

charlie-tf:
image
image

### DIFF
--- a/src/components/InventoryV2/Search.tsx
+++ b/src/components/InventoryV2/Search.tsx
@@ -1,16 +1,38 @@
-import React, { useCallback, useEffect, useState } from 'react';
-import { Grid, Box } from '@mui/material';
-import { PillListItem, Textfield } from '@terraware/web-components';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { Grid, Box, Popover, useTheme } from '@mui/material';
+import { Button, PillListItem, Textfield, Tooltip } from '@terraware/web-components';
 import { PillList } from '@terraware/web-components';
 import strings from 'src/strings';
-import { useOrganization } from 'src/providers/hooks';
-import theme from 'src/theme';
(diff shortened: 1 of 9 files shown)


src/components/InventoryV2/Search.tsx
```diff
@@ -98,6 +142,45 @@ export default function Search(props: SearchProps): JSX.Element | null {
...
+                onConfirm={(_filterGroupFilters: Record<string, SearchNodePayload>) => {
+                  handleFilterClose();
+                  setFilterGroupFilters(_filterGroupFilters);
+                  setFilters({ ...filters, ...convertFilterGroupToMap(_filterGroupFilters) });
+                }}
+                onCancel={handleFilterClose}
+              />
+            </Popover>
+          </Box>
+        )}
```
nickgraz, line 183: At what point should we consider merging this with https://github.com/terraware/terraware-web/blob/main/src/components/InventoryV2/view/InventorySeedlingsTable.tsx#L391? I suspect we will add more filters eventually, but maybe _that_ is when it makes sense to combine?

nickgraz (approved):
Looks good

src/components/InventoryV2/Search.tsx
charlie-tf: You're probably right that next time we touch this filter would make sense. But to be honest I could have broke it into its own component here. I can take a step back and do it now if you like. What do you think?

nickgraz: Hard to say, I think I am indifferent if it is only written twice. I don't want to be overly forceful  with DRY... I think if we have to write it a third time, or if we have to duplicate a ticket in the future because it has to be implemented in both places, then we should think about it. I'm OK to move on for now

charlie-tf: Okay, I'm with you there. Sounds good

PR #1839, opened 2023-12-01
SW-4484 Update Filters from URL Search Params

charlie-tf:
- Add SubLocations Redux
- In the InventoryListByBatch page get filters from query
- Look up sublocations and apply filters
- When filters change update the URL query
Kapture 2023-12-01 at 16 34 38

### DIFF
--- a/src/components/InventoryV2/InventoryListByBatch.tsx
+++ b/src/components/InventoryV2/InventoryListByBatch.tsx
@@ -124,7 +124,7 @@ export default function InventoryListByBatch({ setReportData }: InventoryListByB
   useEffect(() => {
     const subLocationsIds = subLocationNames
       .map((name) => subLocations?.find((sl) => sl.name === name)?.id)
-      .filter((id) => id !== undefined) as number[];
+      .filter((id): id is number => id !== undefined);
     setFilters((f) => ({ ...f, subLocationsIds }));
   }, [subLocationNames, subLocations, setFilters]);
 
--- a/src/components/InventoryV2/InventoryListByBatch.tsx
+++ b/src/components/InventoryV2/InventoryListByBatch.tsx
@@ -16,6 +16,12 @@ import { TableColumnType } from '@terraware/web-components';
 import { NurseryBatchService } from 'src/services';
 import Card from 'src/components/common/Card';
 import { isBatchEmpty } from 'src/components/InventoryV2/FilterUtils';
+import useQuery from 'src/utils/useQuery';
+import { useAppDispatch, useAppSelector } from 'src/redux/store';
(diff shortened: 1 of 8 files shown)


src/components/InventoryV2/InventoryListByBatch.tsx
```diff
@@ -86,6 +92,56 @@ export default function InventoryListByBatch({ setReportData }: InventoryListByB
...
+  }, [nurseryIds, setFilters]);
+
+  useEffect(() => {
+    dispatch(requestSubLocations(filters.facilityIds ?? []));
+  }, [filters.facilityIds, dispatch]);
+
+  useEffect(() => {
+    const subLocationsIds = subLocationNames
+      .map((name) => subLocations?.find((sl) => sl.name === name)?.id)
+      .filter((id) => id !== undefined) as number[];
```
nickgraz, line 127: You might be able to use a typeguard on the filter which would remove the need to cast.
```suggestion
      .filter((id): id is number => id !== undefined);
```

src/components/InventoryV2/Search.tsx
```diff
@@ -63,25 +63,12 @@ export default function Search(props: SearchProps): JSX.Element | null {
...
-      setSubLocations(results.flatMap((r) => r.subLocations));
-    };
-    if (!filters.facilityIds?.length) {
-      setSubLocations([]);
-    } else {
-      fetchSubLocations();
-    }
-    setFilters({ ...filters, subLocationsIds: [] });
-  }, [filters, setFilters]);
+  const subLocations = useAppSelector(selectSubLocations);
```
nickgraz, line 71: 🔥

nickgraz (approved):
Looks good, one small thing to try to get rid of a cast but not a blocker

src/components/InventoryV2/InventoryListByBatch.tsx
```diff
@@ -86,6 +92,56 @@ export default function InventoryListByBatch({ setReportData }: InventoryListByB
...
+      ?.filter((s) => s !== '');
+    setSubLocationNames(querySubLocationNames ?? []);
+    const queryNurseryIds = query
+      .get('facilityId')
+      ?.trim()
+      ?.split(',')
+      ?.filter((n) => n !== '')
+      ?.map((id) => Number(id));
+    setNurseryIds(queryNurseryIds ?? []);
+  }, [query]);
```
karthikbtf, line 114: Thanks for adding this feature @charlie-tf ❤️ 
Added a comment in the Jira, supporting multiple params is better done this [way](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#duplicate_search_parameters) and using `query.getAll(paramName)`

PR #1849, opened 2023-12-05
SW-4489 Refactor the Create Project flow entity filters 

nickgraz:
- Refactor the entity filters so the filter config for the filters specific to each project entity are passed in


src/components/NewProjectFlow/flow/ProjectEntitySearch.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const data: PillListItemWithEmptyValue[] = [
+      ...projectEntityFilterConfig.pillModifier(filters),
+      ...entitySpecificFilterConfigs.map((config) => config.pillModifier(filters)).flat(),
+    ];
+
+    setFilterPillData(data);
+  }, [entitySpecificFilterConfigs, filters, projectEntityFilterConfig]);
+
+  return (
+    <>
```
karthikbtf, line 78: Was wondering if this component can be leveraged.. https://github.com/terraware/terraware-web/blob/main/src/components/common/SearchFiltersWrapper.tsx .. looks like the use-case here in Projects is a tad more custom.

src/components/NewProjectFlow/flow/SelectAccessions.tsx
```diff
@@ -171,6 +174,34 @@ export default function SelectAccessions(props: SelectAccessionsProps): JSX.Elem
...
   });
 
+  const entitySpecificFilterConfigs: EntitySpecificFilterConfig[] = useMemo(
+    () => [
+      {
+        label: strings.STATUS,
+        initialSelection: filters.statuses || [],
+        filterKey: 'statuses',
+        options: ACCESSION_2_STATES,
+        renderOption: (value: string | number) => `${value}`,
```
karthikbtf, line 184: use `stateName` from https://github.com/terraware/terraware-web/blob/main/src/types/Accession.ts#L33 for i18n purposes

src/components/NewProjectFlow/flow/SelectAccessions.tsx
```diff
@@ -171,6 +174,34 @@ export default function SelectAccessions(props: SelectAccessionsProps): JSX.Elem
...
+              id: 'statuses',
+              label: strings.STATUS,
+              value: statuses.join(','),
+              emptyValue: [],
+            },
+          ];
+        },
+      },
+    ],
+    [filters]
```
karthikbtf, line 202: also add `activeLocale` from useLocalization.. (unless this already works with locale changes)

karthikbtf (approved):
This is really nice! Thanks 🎉 
One comment for follow up. Feel free to merge after that.
git reflog ftw!

src/components/NewProjectFlow/flow/ProjectEntitySearch.tsx
nickgraz: Yea that does look like it could be used. I think there would be some work making it work with the rest of the flow and it looks like I would have to re-work the filter pills because they would be inline next to each filter, instead of at the end. Let me know if you think I should go that route

karthikbtf: Thanks for taking a look at that. What we have here in this PR is better organized for the use-case. Refactoring the generic component may work but could be a forced effort perhaps. My vote is to not go that route. Worth discussing I thought.

src/components/NewProjectFlow/flow/SelectAccessions.tsx
```diff
@@ -14,13 +14,18 @@ import {
   SearchResponseElement,
   SearchSortOrder,
 } from 'src/types/Search';
-import { AccessionState } from 'src/types/Accession';
+import { ACCESSION_2_STATES, AccessionState }stateName,  from 'src/types/Accession';
```
karthikbtf, line 17: misplaced comma :)

src/components/NewProjectFlow/flow/SelectAccessions.tsx
```diff
@@ -14,13 +14,18 @@ import {
...
+  ProjectEntityFilters,
   useProjectEntitySelection,
 } from 'src/components/NewProjectFlow/flow/useProjectEntitySelection';
-import Search from 'src/components/NewProjectFlow/flow/Search';
+import ProjectEntitySearch, {
+  PillListItemWithEmptyValue,
+} from 'src/components/NewProjectFlow/flow/ProjectEntitySearch';
+import { EntitySpecificFilterConfig } from 'src/components/NewProjectFlow/flow/ProjectEntityFilter';
+import { stateName } from '../../../types/Accession';
+import { useLocalization } from '../../../providers';
```
karthikbtf, line 28: `src/types/Accession` and `src/providers`?

PR #1852, opened 2023-12-06
Add single project selector and action to put project into state

nickgraz:

PR #1851, opened 2023-12-06
Add optional border radius prop for Card component that defaults to previous value

nickgraz:

PR #1853, opened 2023-12-06
Add request for single project and redux thunk to get project and put into state

nickgraz:


nickgraz:
* main
  * **PR #1851** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1851?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1852** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1852?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1853** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1853?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/1853?utm_source=stack-comment).

src/redux/features/projects/projectsThunks.ts
```diff
@@ -16,3 +16,17 @@ export const requestProjects = (organizationId: number, locale?: string | null)
...
+
+export const requestProject = (projectId: number) => async (dispatch: Dispatch, _getState: () => RootState) => {
+  try {
+    const response = await ProjectsService.getProject(projectId);
+    const { error, data } = response;
+    if (data) {
+      dispatch(setProjectAction({ error, projects: [data] }));
+    }
+  } catch (e) {
+    // should not happen, the response above captures any http request errors
```
nickgraz, line 28: I copy/pasted this from the request above... is this comment needed? Can we just have an empty catch here? Should we just pipe this error into state somewhere?

karthikbtf: empty catch should be fine i think

karthikbtf (commented):
App.tsx requests all projects today.
This pattern of bulk requesting data and depending on that to view specific data by id, is used across the board. We could change that pattern if needed.. for now the project by id should already exist because of App.tsx call.
Not sure if the batches list page requests the data again.
Thoughts?

nickgraz:
> App.tsx requests all projects today.
> This pattern of bulk requesting data and depending on that to view specific data by id, is used across the board. We could change that pattern if needed.. for now the project by id should already exist because of App.tsx call.
> Not sure if the batches list page requests the data again.
> Thoughts?
> 
> 

Oh right... I forgot about that. I generally prefer to get the data in the view, this guarantees that if the data has changed at some point (let's say outside of the projects area... if that is even possible) it will be accurate by the time you get to the view. I suppose in this case, the project entity wouldn't be modified elsewhere so I can remove it if you want

karthikbtf:
> > App.tsx requests all projects today.
> > This pattern of bulk requesting data and depending on that to view specific data by id, is used across the board. We could change that pattern if needed.. for now the project by id should already exist because of App.tsx call.
> > Not sure if the batches list page requests the data again.
> > Thoughts?
> 
> Oh right... I forgot about that. I generally prefer to get the data in the view, this guarantees that if the data has changed at some point (let's say outside of the projects area... if that is even possible) it will be accurate by the time you get to the view. I suppose in this case, the project entity wouldn't be modified elsewhere so I can remove it if you want

No need to remove it. Since the request by id updates the same state in redux (projects list), data will be consistent across project view and projects list view. This is a cleaner optimal design for sure. We currently fetch all data, edit single views, refetch all data.. to keep things consistent in redux. A per object approach is certainly a better path.

src/redux/features/projects/projectsThunks.ts
```diff
@@ -16,3 +16,17 @@ export const requestProjects = (organizationId: number, locale?: string | null)
     }
   };
 };
+
+export const requestProject = (projectId: number) => async (dispatch: Dispatch, _getState: () => RootState) => {
```
karthikbtf, line 20: When you get a chance, take a look at createAsyncThunk? We have a mixed use-case of async thunks vs non in terraware-web. For the funder-portal we used async thunks across the board. Gave us better request status management.

karthikbtf (approved):
This approach is good 👍

src/redux/features/projects/projectsThunks.ts
karthikbtf: If you'd like, up to you. Or leave it for future requests.

nickgraz:
> 
> 
> If you'd like, up to you. Or leave it for future requests.

Cool, I have to do the edit and delete project requests next so I will do it in that PR

PR #1855, opened 2023-12-06
Stub out project edit view and add to routing, behind feature flag for projects feature

nickgraz:


src/components/ProjectEditView/index.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+import { Typography } from '@mui/material';
+import { useAppDispatch, useAppSelector } from 'src/redux/store';
+import { selectProject } from 'src/redux/features/projects/projectsSelectors';
+import { requestProject } from 'src/redux/features/projects/projectsThunks';
+import TfMain from '../common/TfMain';
```
karthikbtf, line 7: src/components/common?

src/components/ProjectEditView/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  useEffect(() => {
+    if (!project) {
+      void dispatch(requestProject(projectId));
+    }
+  }, [projectId, project, dispatch]);
+
+  return (
+    <TfMain>
+      <Typography variant={'h3'}>{project?.name}</Typography>
```
karthikbtf, line 25: guessing the styles here are temporary

src/App.tsx
```diff
@@ -462,6 +463,9 @@ function AppContent() {
                 <Route exact path={APP_PATHS.PROJECT_VIEW}>
                   <ProjectView />
                 </Route>
+                <Route exact path={APP_PATHS.PROJECT_EDIT}>
+                  <ProjectEditView />
+                </Route>
```
karthikbtf, line 468: Not directly related to this PR,  thoughts on delegating routes downstream?
App.tsx would ideally have something like
```
{featureFlagProjects && (
  <Route path={APP_PATHS.PROJECTS}>
    <projects root component />
  </Route>
)}
```
Projects/index.tsx would have the individual routes. This helps keep App.tsx simpler and higher level.
Tried this approach for planting sites and it seemed cleaner to me.

karthikbtf (approved):
LGTM! Some comments for discussion.

src/App.tsx
```diff
@@ -462,6 +463,9 @@ function AppContent() {
...
               </>
```
charlie-tf, line 469: I found that placing multiple `Route` components within a fragment like this causes problems. The `Switch` component doesn't like any direct children other than `Route`s or `Redirect`s. So you'll need to expand this out and individually wrap each `Route` in a conditional rendering block, like
```typescript
{featureFlagProjects && (
  <Route exact path={APP_PATHS.PROJECTS}>
    {getProjectsView()}
  </Route>
)}
{featureFlagProjects && (
  <Route exact path={APP_PATHS.PROJECTS_NEW}>
    <NewProjectFlow reloadData={reloadProjects} />
  </Route>
)}
{featureFlagProjects && (
  <Route exact path={APP_PATHS.PROJECT_VIEW}>
    <ProjectView />
  </Route>
)}
{featureFlagProjects && (
  <Route exact path={APP_PATHS.PROJECT_EDIT}>
    <ProjectEditView />
  </Route>
)}
```
Which is uglier, but I believe necessary

src/components/ProjectEditView/index.tsx
nickgraz: Right, it is fleshed out a bit more in a later PR

src/App.tsx
nickgraz: Ended up splitting into its own router

src/App.tsx
nickgraz: Great idea!

PR #1859, opened 2023-12-06
Add cell renderer to projects table to display link to project. Add checkboxes (per UI, currently a noop) to the projects table.

nickgraz:


karthikbtf (approved):
Straightforward!

src/components/Projects/ProjectCellRenderer.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+  const createProjectLink = () => {
+    const projectLocation = {
+      pathname: APP_PATHS.PROJECT_VIEW.replace(':projectId', row.id.toString()),
+    };
+    return <Link to={projectLocation.pathname}>{row.name}</Link>;
+  };
+
+  if (column.key === 'name') {
+    return <CellRenderer index={index} column={column} value={createProjectLink()} row={row} />;
```
karthikbtf, line 18: I think this can just be
```
 return <CellRenderer {...props} value={createProjectLink()} />
```

src/components/Projects/index.tsx
```diff
@@ -135,7 +136,19 @@ export default function ProjectsList(): JSX.Element {
...
+                {results && (
+                  <Table
+                    id='projects-table'
+                    columns={columns}
+                    rows={results}
+                    orderBy='name'
+                    Renderer={ProjectCellRenderer}
+                    showCheckbox
+                    selectedRows={[]}
+                    setSelectedRows={() => undefined}
```
karthikbtf, line 148: this to be set later?

karthikbtf (approved):

src/components/Projects/index.tsx
nickgraz: Yea... it should be. The designs don't show any sort of menu or bulk action dropdown to do anything with the selected items, so for now they are just here to make the check boxes. I'll ask in the designs if that is intentional or not.

PR #1898, opened 2023-12-12
SW-4558 Add sublocation edit functionality to batch details view

nickgraz:
- Move SubLocationsDropdown functionality within summary to its own component, save sub locations against the batch when the multi select is blurred - https://github.com/terraware/terraware-web/pull/1898/commits/7c3c04c7f99287e423709f0a728016cc52995785
- [Implement batches state with fetch batch so we can update consumers of batches within state](https://github.com/terraware/terraware-web/pull/1898/commits/7b1ec6c98d0d0f214aacf29898034ab8d3a2b8f2)
- [Use redux for batch within the InventoryBatch view](https://github.com/terraware/terraware-web/pull/1898/commits/94af90eaadc4e5a4b73c66f865558beff7023135)

Depends on https://github.com/terraware/web-components/pull/447

### DIFF
--- a/src/components/InventoryV2/BatchSummary.tsx
+++ b/src/components/InventoryV2/BatchSummary.tsx
@@ -1,5 +1,6 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { Grid, useTheme } from '@mui/material';
+import _ from 'lodash';
 import strings from 'src/strings';
 import { Batch } from 'src/types/Batch';
 import OverviewItemCard from 'src/components/common/OverviewItemCard';
@@ -7,22 +8,27 @@ import Link from 'src/components/common/Link';
 import { APP_PATHS } from 'src/constants';
 import { useSubLocations } from 'src/components/InventoryV2/form/useSubLocations';
 import SubLocationsDropdown from 'src/components/InventoryV2/form/SubLocationsDropdown';
+import { useAppDispatch, useAppSelector } from 'src/redux/store';
+import { requestSaveBatch } from 'src/redux/features/batches/batchesAsyncThunks';
+import { selectBatchesRequest } from 'src/redux/features/batches/batchesSelectors';
+import useSnackbar from 'src/utils/useSnackbar';
 
 interface BatchSummaryProps {
   batch: Batch;
 }
 
 export default function BatchSummary(props: BatchSummaryProps): JSX.Element {
+  const dispatch = useAppDispatch();
   const theme = useTheme();
+  const snackbar = useSnackbar();
 
   const [batch, setBatch] = useState<Batch>(props.batch);
   const { availableSubLocations, selectedSubLocations } = useSubLocations(batch.facilityId, batch);
 
-  const [showSubLocationEdit, setShowSubLocationEdit] = useState<boolean>(false);
+  const [requestId, setRequestId] = useState('');
+  const batchesRequest = useAppSelector(selectBatchesRequest(requestId));
 
-  const handleSubLocationEdit = useCallback(() => {
-    setShowSubLocationEdit(!showSubLocationEdit);
-  }, [showSubLocationEdit]);
+  const [showSubLocationEdit, setShowSubLocationEdit] = useState<boolean>(false);
 
   const handleUpdateSubLocations = useCallback(
     (setFn: (previousBatch: Batch) => Batch) => {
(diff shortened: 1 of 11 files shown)


nickgraz:
* `main`
  * **PR #1898** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1898?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/1898?utm_source=stack-comment).

src/redux/features/batches/batchesAsyncThunks.ts
```diff
@@ -11,7 +11,7 @@ export type SavableBatch = (CreateBatchRequestPayload | UpdateBatchRequestPayloa
 
 export const requestSaveBatch = createAsyncThunk(
   'batches/save',
-  async (request: { batch: SavableBatch; timezone: string }, { rejectWithValue }) => {
+  async (request: { batch: SavableBatch; timezone?: string }, { rejectWithValue }) => {
```
nickgraz, line 14: `timezone` is only used for new batches and defaults to UTC if `undefined`. Is that alright? Or should I add validation to ensure that `timezone` is provided if this is a "new batch" save?

karthikbtf: There's a timezone fallback.
For batches, timezone fallback is nursery-tz or org-tz (for display purposes)
For edit/save, fallback is nursery-tz or org-tz or user-tz.
You can use this util https://github.com/terraware/terraware-web/blob/main/src/utils/useTimeZoneUtils.ts#L84 to get the timezone for the nursery/facility

src/components/InventoryV2/BatchSummary.tsx
```diff
@@ -1,44 +1,87 @@
...
+    },
+    [batch]
+  );
+
+  const toggleSubLocationEdit = useCallback(() => {
+    const nextShowSubLocationEdit = !showSubLocationEdit;
+
+    // If we're "turning off" the edit mode, and the sub locations aren't the same, we need to update the batch
+    if (!nextShowSubLocationEdit && !_.isEqual(props.batch.subLocationIds, batch.subLocationIds)) {
+      const request = dispatch(requestSaveBatch({ batch }));
```
karthikbtf, line 47: thoughts on a bullk update/dispatch for batches that need updating?

karthikbtf (commented):
looks great, a few comments

src/components/InventoryV2/BatchSummary.tsx
nickgraz: Not sure I follow... there is only a single batch in this view.

nickgraz:
> There's a timezone fallback.
> For batches, timezone fallback is nursery-tz or org-tz (for display purposes)
> For edit/save, fallback is nursery-tz or org-tz or user-tz.
> You can use this util https://github.com/terraware/terraware-web/blob/main/src/utils/useTimeZoneUtils.ts#L84 to get the > timezone for the nursery/facility

The caller is already using that util to get the tz for the nursery - https://github.com/terraware/terraware-web/blob/main/src/components/InventoryV2/form/BatchDetailsForm.tsx#L143-L144

src/components/InventoryV2/BatchSummary.tsx
karthikbtf: My bad, I thought the sublocations was part of a species summary - and applied to a list of batches. That's the older view. !Please ignore!

karthikbtf:
> > There's a timezone fallback.
> > For batches, timezone fallback is nursery-tz or org-tz (for display purposes)
> > For edit/save, fallback is nursery-tz or org-tz or user-tz.
> > You can use this util https://github.com/terraware/terraware-web/blob/main/src/utils/useTimeZoneUtils.ts#L84 to get the > timezone for the nursery/facility
> 
> The caller is already using that util to get the tz for the nursery - https://github.com/terraware/terraware-web/blob/main/src/components/InventoryV2/form/BatchDetailsForm.tsx#L143-L144

That's awesome. Timezone won't be undefined in that case. Does default to UTC if org or user have no timezone set (which should be unlikely these days anyways).

src/redux/features/batches/batchesAsyncThunks.ts
karthikbtf: I understand the question better now.. feels like create/update requests should be split to handle the required timezone use-case for new batches. A runtime check is probably not required. We could leave it as-is for now, splitting the func is beyond scope of this PR.

karthikbtf (approved):
🎉

nickgraz:
> I understand the question better now.. feels like create/update requests should be split to handle the required timezone use-case for new batches. A runtime check is probably not required. We could leave it as-is for now, splitting the func is beyond scope of this PR.

OK cool, if I have to mess with it more I will split it up. I refactored it into redux from another component, which is why it is kind of a two-for-one at the moment.

nickgraz:
Since this ticket has some UI that we just kinda made up, I'm going to keep this open until @alexTerraformation has a chance to review it.

alexTerraformation:
Looks good and works well!  Only request I'd have (if it's simple enough) would be to save changes on all clicks that cause a loss of focus, rather than requiring a second click on the pencil icon.

nickgraz:
@alexTerraformation I made it so it saves on blur of the multi select, and when the pencil is clicked (if modifications have been made)

nickgraz:
@charlie-tf I re-worked this to use redux in the main view, so the batch can be updated from a child component if updates are made. Would you mind taking another look?

charlie-tf (approved):
Makes sense to me. Thanks!

alexTerraformation:
The behavior seems the same to me.  Am I missing something?

https://github.com/terraware/terraware-web/assets/104874529/bdfe0398-99b6-4b20-8703-c1c6ec5823c2

alexTerraformation:
Works well!  My preference would be to enable the same behavior when removing sub-locations (really any time that sub-locations have changed).

alexTerraformation:
Good enough!  Let's ship it!  Thanks @nickgraz

PR #1899, opened 2023-12-13
SW-4570 react hook to manage data with undo/redo capability

karthikbtf:
- a simple hook similar to useState that allows undo/redo functionality
- for now this is super simple and can be extended later as needed, current functionality supports what we need for user drawn detailed sites
- added some tests


charlie-tf (approved):
This is really cool! Thanks for the test cases as well

PR #1907, opened 2023-12-14
SW-4575 Remember Last Inventory Tab

charlie-tf:
Use user preferences to store and lookup the tab that was last visited.


nickgraz (approved):
Nice!

PR #1910, opened 2023-12-14
Add project assign request to redux

nickgraz:

PR #1911, opened 2023-12-14
SW-4494 Create ProjectAssignModal which allows for assigning entity to project

nickgraz:
Create ProjectAssignModal that allows any project-assignable-entity to easily add a link and modal to assign an entity to a project from the available projects for an organization.

'Add to Project' string

### DIFF
--- a/src/components/accession2/view/ProjectSelectorModal.tsx
+++ b/src/components/accession2/view/ProjectSelectorModal.tsx
@@ -1,0 +1,116 @@
+import React, { useCallback, useEffect, useState } from 'react';
+import { Icon } from '@terraware/web-components';
+import { Box, Link as LinkMUI, Theme, useTheme } from '@mui/material';
+import { makeStyles } from '@mui/styles';
+import strings from 'src/strings';
+import Button from 'src/components/common/button/Button';
+import DialogBox from 'src/components/common/DialogBox/DialogBox';
+import { useProjects } from 'src/components/InventoryV2/form/useProjects';
+import { Accession } from 'src/types/Accession';
+import ProjectsDropdown from 'src/components/InventoryV2/form/ProjectsDropdown';
+import { useAppDispatch, useAppSelector } from 'src/redux/store';
+import { selectProjectRequest } from 'src/redux/features/projects/projectsSelectors';
+import useSnackbar from 'src/utils/useSnackbar';
+import { requestProjectAssign } from 'src/redux/features/projects/projectsAsyncThunks';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  addIcon: {
+    fill: theme.palette.TwClrIcnBrand,
+    height: '20px',
+    width: '20px',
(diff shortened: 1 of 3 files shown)


nickgraz:
* `main`
  * **PR #1910** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1911** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1911?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #1912** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1912?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/1911?utm_source=stack-comment).

src/components/accession2/view/ProjectSelectorModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  return (
+    <>
+      <LinkMUI sx={linkStyle} onClick={() => setIsOpen(true)}>
+        <Box display='flex' alignItems='center'>
+          <Icon name='iconAdd' className={classes.addIcon} />
+          {strings.ADD_TO_PROJECT}
+        </Box>
+      </LinkMUI>
```
charlie-tf, line 91: Would it make sense to put the modal trigger (and open/close) state in the parent component? To decouple the access pattern from the modal itself (maybe doesn't make sense, if we're not expecting to use this modal elsewhere or access it differently)? It's a tradeoff, making it self-contained here vs. spreading its "state" across this and the parent component. So I'd be okay either way.

nickgraz: I could go either way. I did this intentionally so the component can be re-used for different parents. Since there are 3 entities that can be assigned to a project, presumably we will implement this functionality elsewhere. If we bake it into the parent, we will have more duplicate code when we inevitably implement this modal elsewhere or we will need to refactor more stuff. 

I will make this a bit more abstract to illuminate what I'm proposing, then we can decide.

nickgraz: @charlie-tf take a look now and let me know what you think, along with https://github.com/terraware/terraware-web/pull/1912. Hopefully it makes a bit more sense what I was going for now that it is more generic.

charlie-tf: Okay, yeah, this is good. Thanks

charlie-tf (approved):

nickgraz:
### Merge activity

* **Dec 15, 4:19 PM**: @@nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/1911).

PR #1913, opened 2023-12-15
SW-4548 / SW-4556 Fix spacing of filter pills and batch details cards

alexTerraformation:
Style updates to match designs.

### DIFF
--- a/src/components/InventoryV2/BatchDetails.tsx
+++ b/src/components/InventoryV2/BatchDetails.tsx
@@ -113,7 +113,7 @@ export default function BatchDetails({ batch, onUpdate }: BatchDetailsProps): JS
       <Grid container>
         {overviewCardData.map((item) => (
           <Grid item xs={isMobile || item.fullWidth ? 12 : 6} key={item.title}>
-            <OverviewItemCard isEditable={false} title={item.title} contents={item.data} />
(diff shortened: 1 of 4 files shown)


src/components/InventoryV2/BatchDetails.tsx
```diff
@@ -113,7 +113,7 @@ export default function BatchDetails({ batch, onUpdate }: BatchDetailsProps): JS
       <Grid container>
         {overviewCardData.map((item) => (
           <Grid item xs={isMobile || item.fullWidth ? 12 : 6} key={item.title}>
-            <OverviewItemCard isEditable={false} title={item.title} contents={item.data} />
+            <OverviewItemCard isEditable={false} title={item.title} contents={item.data} grid={true}/>
```
charlie-tf, line 116: Rather than this, I think it would be more appropriate for the `Grid` component here to handle the spacing. That way `OverviewItemCard` doesn't have to "know" how it is being used and isn't responsible for its own layout.

That being said, you would then need to zero out the `padding` style within the `OverviewItemCard` itself, which would then require you to go and find every usage of it and adjust the parent component. Given that that is going to be a pain, I'm okay with this compromise.

charlie-tf (approved):
A bit of me thinking out loud, but looks fine overall

PR #1914, opened 2023-12-15
SW-4581 Restrict Accessions Species if Adding Batch from Species Page

charlie-tf:
- Add a `speciesId` parameter to `useAccessions`
- In `BatchDetailsForm`, if origin is Species, send `speciesId` to
  `useAccessions`
- Send selected species id, if selected

### DIFF
--- a/src/components/InventoryV2/form/useAccessions.ts
+++ b/src/components/InventoryV2/form/useAccessions.ts
@@ -59,10 +59,8 @@ export const useAccessions = (record?: { accessionId?: number }, speciesId?: num
   }, [availableAccessions, record?.accessionId]);
 
   useEffect(() => {
-    if (!availableAccessions) {
-      void initAccessions();
-    }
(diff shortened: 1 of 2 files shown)


src/components/InventoryV2/form/useAccessions.ts
```diff
@@ -11,37 +11,46 @@ export type SearchResponseAccession = {
   speciesName: string;
 };
 
-export const useAccessions = (record?: { accessionId?: number }, excludeUsedUp?: boolean) => {
+export const useAccessions = (record?: { accessionId?: number }, speciesId?: number, excludeUsedUp?: boolean) => {
```
nickgraz, line 14: Can we just use the `speciesId` from the record, if available?

charlie-tf: I changed this to also use the selected species id if available, so it needs to stay here

nickgraz (approved):

PR #1916, opened 2023-12-15
SW-4415 Show planting seasons in site details

sgrimm:
Detailed planting sites can have a list of planting seasons. Display the list on
the details page.

PR #1915, opened 2023-12-15
Expose effective time zone from selector

sgrimm:
`LocationTimeZoneSelector` calculates the effective time zone of the location
(inheriting from the organization if needed) but doesn't expose that value;
callers have to calculate it themselves. Add an optional callback to expose it.

PR #1917, opened 2023-12-15
SW-4416 Add planting season editor

sgrimm:
Add an editable list of planting seasons to the planting site edit page for
detailed planting sites.

### DIFF
--- a/src/components/PlantingSites/PlantingSeasonsEdit.tsx
+++ b/src/components/PlantingSites/PlantingSeasonsEdit.tsx
@@ -222,7 +222,7 @@ export default function PlantingSeasonsEdit(props: Props): JSX.Element {
 
   const onDeletePlantingSeason = useCallback(
     (index: number) => {
-      setRows(rows.slice(0, index).concat(rows.slice(index + 1)));
+      setRows(rows.filter((_row, _index) => _index !== index));
     },
     [rows, setRows]
   );
--- a/src/components/PlantingSites/PlantingSeasonsEdit.tsx
+++ b/src/components/PlantingSites/PlantingSeasonsEdit.tsx
@@ -1,0 +1,292 @@
+import { Fragment, useCallback, useEffect, useMemo, useState } from 'react';
+import { Box, Grid } from '@mui/material';
+import { DateTime, DurationLike } from 'luxon';
+import strings from 'src/strings';
+import { TimeZoneDescription } from 'src/types/TimeZones';
+import AddLink from 'src/components/common/AddLink';
+import DatePicker from 'src/components/common/DatePicker';
+import Icon from 'src/components/common/icon/Icon';
+import Link from 'src/components/common/Link';
+
+// Minimum length of a planting season.
+const minDuration: DurationLike = { weeks: 4 };
+// Maximum length of a planting season.
+const maxDuration: DurationLike = { years: 1 };
+// Maximum amount of time in the future we allow scheduling a planting season.
+const maxTimeUntilStart: DurationLike = { years: 1 };
+
+/** YYYY-MM-DD string representation of a date. This type is just for code clarity. */
+type IsoDate = string;
+
+/** Data from a row in the edit UI. May or may not be valid. */
+type EditableRow = {
+  /**
+   * If this row is a season that already exists on the server, its ID. Undefined if this row was
+   * added in the UI and hasn't been saved yet.
+   */
+  id?: number;
+  startDate?: IsoDate;
+  endDate?: IsoDate;
+};
+
+type ValidPlantingSeason = {
+  id?: number;
+  startDate: IsoDate;
+  endDate: IsoDate;
+};
+
+type Props = {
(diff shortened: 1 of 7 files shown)


sgrimm:
* `main`
  * **PR #1915** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1916** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1917** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1917?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/1917?utm_source=stack-comment).

src/components/PlantingSites/PlantingSeasonsEdit.tsx
```diff
@@ -0,0 +1,292 @@
...
+  /** Called whenever the list of planting seasons becomes either valid or invalid. */
+  setPlantingSeasonsValid: (valid: boolean) => void;
+
+  /**
+   * If true, show any validation errors that should only be shown just after the user tries to save
+   * the site. The component can clear this flag if the user does something such as adding a new row
+   * to the planting seasons list such that it no longer makes sense to show these errors.
+   */
+  showSaveValidationErrors: boolean;
+  setShowSaveValidationErrors: (showErrors: boolean) => void;
```
sgrimm, line 58: This feels a little clunky to me but the idea is that we don't want the UI to show a "Required field" validation error the moment you click the "Add" button to add a planting season, only if you try to save without filling in a value. In the parent component, we set error messages like that (e.g., on the Name field) directly in the save function, but obviously that's not an option here since this is a separate component. Maybe there's a better approach for this kind of thing?

sgrimm:
Screenshot 2023-12-15 at 10 34 40 AM

src/components/PlantingSites/PlantingSeasonsEdit.tsx
```diff
@@ -0,0 +1,292 @@
...
+  }, [rows, allSeasonsValid, setPlantingSeasons, setPlantingSeasonsValid]);
+
+  const onAddPlantingSeason = () => {
+    setRows(rows.concat([{}]));
+    setShowSaveValidationErrors(false);
+  };
+
+  const onDeletePlantingSeason = useCallback(
+    (index: number) => {
+      setRows(rows.slice(0, index).concat(rows.slice(index + 1)));
```
nickgraz, line 225: A lightly more obvious way to write this in JS is 
```
// Filter out the given row by index
setRows(rows.filter((row, _index) => _index !== index))
```

src/components/PlantingSites/PlantingSeasonsEdit.tsx
nickgraz: In the past I've implemented this as "isFormDirty" - if the form is "dirty" (ie - the form has been modified since it was either created or validated) then you hide errors. Although I think in this case it might be better to only consider if a row is dirty. For example, if there are 5 newly added seasons, and 3 of them come back with errors when the user attempts to save, I think we would only want to stop showing errors for each row as it is modified, instead of clearing all errors when any one row is modified. Maybe that is already the behavior though?

sgrimm: One thing that might inform this discussion: at the moment, the only error this pertains to is the "Required field" message if there's no date selected at all.

Current behavior is:

1. Hit "Add". No errors shown.
2. Hit "Save". "Required field" error shown under the newly-added blank inputs.
3. Hit "Add" again. The errors from step 2 disappear (this is the behavior that differs from what you're suggesting).
4. Hit "Save" again. Errors appear under the fields in both of the blank rows.

I did think a little about how to avoid the errors disappearing but it seemed like tracking the additional per-row state would add more complexity than the problem was worth. Totally possible that was the wrong call, though.

nickgraz: I see. I think this could be solved by extracting the individual row and row validation to their own component, this way you can track the "form dirty" and "row dirty" separately. I think it is fine as is though, perhaps if we decide we need that in the future we can look at splitting it out further.

nickgraz (approved):

src/components/PlantingSites/PlantingSeasonsEdit.tsx
sgrimm: One challenge with per-row validation is that one of the rules is, "Seasons can't overlap with each other," which you can't detect on an individual row in isolation.

sgrimm:
### Merge activity

* **Dec 15, 2:55 PM**: @@sgrimm started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/1917).

PR #1926, opened 2023-12-19
Fix planting site type selected button background color on hover/active states

karthikbtf:
- as per Chudo's feedback

screencast 2023-12-19 09-01-25


nickgraz (approved):

chudo76 (commented):
Thank you for the fix!

PR #1928, opened 2023-12-19
Trying to get playwright to work as expected with e2e tests in gh actions pointing to staging

nickgraz:


playwright/e2e/suites/test.spec.js
```diff
@@ -15,67 +15,69 @@ test.beforeEach(async ({ context }, testInfo) => {
...
-  await page.waitForTimeout(5000);
+  const newSpeciesName = `Acacia koa-${new Date().getTime()}`;
+
   await page.getByRole('button', { name: 'Species' }).click();
-  await page.waitForTimeout(5000);
-  await page.getByRole('button', { name: 'Add Species' }).click();
-  await page.waitForTimeout(5000);
+
+  while (!(await page.locator('.dialog-box--opened').isVisible())) {
+    await sleep(1000);
```
alexTerraformation, line 33: Why do we need this?  The default timeout on the locator should already handle waiting right?

nickgraz: I added this because the container seemed to be rerendering after the button is clicked immediately. This is an indication of a rerender breaking the test flow, so this just makes sure that the dialog gets opened if it gets closed from the rerender

nickgraz: Same story with the `while` checking for the `.options-container` below

PR #1929, opened 2023-12-19
SW-4496 Implement project edit from accession view project overview card

nickgraz:

### DIFF
--- a/src/components/ProjectOverviewItemCard/index.tsx
+++ b/src/components/ProjectOverviewItemCard/index.tsx
@@ -1,0 +1,84 @@
+import React, { useState } from 'react';
+import strings from 'src/strings';
+import { useAppSelector } from 'src/redux/store';
+import { selectProjects } from 'src/redux/features/projects/projectsSelectors';
+import { useOrganization } from 'src/providers';
+import { isContributor } from 'src/utils/organization';
+import { AssignProjectRequestPayload } from 'src/services/ProjectsService';
+import ProjectAssignModal from 'src/components/ProjectAssignModal';
+import OverviewItemCard from 'src/components/common/OverviewItemCard';
+import { Link as LinkMUI, Theme, useTheme, Box } from '@mui/material';
+import { Icon } from '@terraware/web-components';
+import { makeStyles } from '@mui/styles';
+
+interface OverviewItemCardProjectProps<T extends { id: number; projectId?: number }> {
+  entity: T;
+  reloadData: () => void;
+  projectAssignPayloadCreator: () => AssignProjectRequestPayload;
+}
+
+const useStyles = makeStyles((theme: Theme) => ({
+  addIcon: {
+    fill: theme.palette.TwClrIcnBrand,
+    height: '20px',
+    width: '20px',
+    marginRight: '5px',
+  },
+}));
+
+const ProjectOverviewItemCard = <T extends { id: number; projectId?: number }>({
+  entity,
+  reloadData,
+  projectAssignPayloadCreator,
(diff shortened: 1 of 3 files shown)


nickgraz:
* `main`
  * **PR #1929** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1929?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/1929?utm_source=stack-comment).

src/components/ProjectOverviewItemCard/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  };
+
+  return (
+    <OverviewItemCard
+      isEditable={userCanEdit}
+      handleEdit={() => setIsProjectAssignModalOpen(true)}
+      title={strings.PROJECT}
+      contents={
+        <>
+          {entityProject?.name ?? (
```
nickgraz, line 59: @charlie-tf you were right, needed to pull the display out of the modal component

src/components/ProjectOverviewItemCard/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+            <LinkMUI sx={linkStyle} onClick={() => setIsProjectAssignModalOpen(true)}>
```
karthikbtf, line 60: There's a Link component under common

karthikbtf (approved):
Nice!

src/components/ProjectOverviewItemCard/index.tsx
nickgraz: Yea... I was wondering about that. Just copied what was already in https://github.com/terraware/terraware-web/blob/ng/SW-4496-accession-project-edit/src/components/accession2/view/Accession2View.tsx#L4

I'll change it over though

nickgraz: Ah, looks like it was that way because the `Link` in common doesn't take `sx` prop.

karthikbtf: Got it, that could have been older code that wasn't updated.

nickgraz: I guess it isn't really a link, it is more of a button. Going to leave as is since we don't actually want to do any routing with this click

karthikbtf: The linkStyle should be standard across all links, that's one reason the Link component was created. Shouldn't need to use the sx prop in this scenario.

karthikbtf: > I guess it isn't really a link, it is more of a button. Going to leave as is since we don't actually want to do any routing with this click

Right, my mistake. Thought this was a text link. Perhaps we don't need the link here at all. Just a cursor pointer on the Box and an onClick. Ok fine to leave as-is.

nickgraz: I switched it over, everything looks / works the same. Looks like the common `Link` component only does routing stuff if there's a `to` prop.

karthikbtf: > I switched it over, everything looks / works the same. Looks like the common `Link` component only does routing stuff if there's a `to` prop.

Thanks for the change. I was looking at the pencil icon, not 'add project' - ignore my last comment about not needing a link.

karthikbtf (approved):

PR #1930, opened 2023-12-19
SW-4588 Show 'empty batches' info modal if a withdrawal creates empty batches

karthikbtf:
- check withdrawal payload against source batches to determine if we need to show the info modal
- extracted out post-withdraw actions into a function for reuse in synchronous (no empty batches info modal) and async (after confirming in modal) paths

Terraware App 2023-12-19 11-30-47


nickgraz (approved):
Looks good to me

PR #1940, opened 2023-12-20
SW-4539 Add instructions modal with placeholder content

karthikbtf:
- implementation assumes a youtube video, this can be changed later depending on content type
- moved files into a createSite directory as there will be a lot more content coming up
- the video modal is currently specific to planting site create and not generic, we can extract things out once we have multiple use-cases
- added a story

Terraware App 2023-12-20 10-37-00

Terraware App 2023-12-20 10-37-19


src/components/PlantingSites/createSite/PlantingSiteDetailedCreate.tsx
```diff
@@ -1,13 +1,22 @@
...
 import TfMain from 'src/components/common/TfMain';
 import { Box, Container, Typography } from '@mui/material';
+import InstructionsModal from './InstructionsModal';
 
 type PlantingSiteDetailedCreateProps = {
   reloadPlantingSites: () => void;
 };
 
 export default function PlantingSiteDetailedCreate(props: PlantingSiteDetailedCreateProps): JSX.Element {
+  // this is a placeholder for the instructions modal trigger
```
nickgraz, line 11: 👍

src/components/PlantingSites/createSite/InstructionsModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+  },
+  icon: {
+    width: '150px',
+    height: '150px',
+  },
+}));
+
+export type InstructionsModalProps = {
+  open: boolean;
+  onClose: (dontShowAgain: boolean) => void;
```
nickgraz, line 19: Should `dontShowAgain` be optional?

nickgraz (approved):
One comment, probably related to this being WIP, otherwise looks great

src/components/PlantingSites/createSite/InstructionsModal.tsx
karthikbtf: Good point, yes will make this optional.

PR #1945, opened 2023-12-21
SW-4586 Route to Batch Withdraw Flow from Nursery and Batches Lists

charlie-tf:
- Fetch batch IDs in the searchInventoryByNursery call
- From Nursery and Batches list view, when withdrawing, route to
  `batch/withdraw` and include all batchIds in query parameters.
- Disable withdraw button if selection spans multiple nurseries
- Show tooltip when button is disabled


charlie-tf:
converted to draft. Couple TODOs I noticed from the ticket that I should get in here

nickgraz (approved):
Looks good! Gonna miss ya Charlie!

PR #1976, opened 2024-01-02
Playwright POC

alexTerraformation:


playwright/e2e/suites/test.spec.js
```diff
@@ -0,0 +1,71 @@
...
+  await expect(page.getByText(newSpeciesName)).toBeVisible();
+  await expect(page.locator(`tr:has(> td[title="${newSpeciesName}"]) td[title="Koa"]`)).toBeVisible();
+
+  await page.getByRole('row', { name: newSpeciesName }).getByRole('checkbox').check();
+
+  await page.getByRole('button', { name: 'Delete' }).click();
+  await page.waitForTimeout(1000); //Wait for modal to load and be hydrated before interacting
+  await page.locator('button.destructive-primary').click();
+
+  await page.waitForTimeout(1000); //Test is slow here for some reason???
```
alexTerraformation, line 69: The Chromium test is flaky here and even with the 5 second wait we need additional wait times here.  The test always passes the second time through.

PR #1982, opened 2024-01-03
SW-4616 Remove extra margin during scroll in batch details view

karthikbtf:
- contentRef was being set on the tabs container, most likely a mistake since the contentRef was also being set correctly on the content container of tabs and info cards correctly
- removed the extra contentRef (which adds a marginTop 200px during scroll up which triggers a static page header)


karthikbtf:
@nickgraz when you get a chance, a quick/simple change

nickgraz (approved):

PR #1983, opened 2024-01-03
SW-4578 / SW-4646 Hide sublocations when there are no selections

karthikbtf:
- also update sublocations when the selected facility id changes
- clear selected sublocations when sublocations change

screencast 2024-01-03 10-21-28

### DIFF
--- a/src/components/InventoryV2/form/BatchDetailsForm.tsx
+++ b/src/components/InventoryV2/form/BatchDetailsForm.tsx
@@ -232,6 +232,10 @@ export default function BatchDetailsForm(props: BatchDetailsFormProps): JSX.Elem
 
   // when sublocations change, clear current sublocations in the record
   useEffect(() => {
+    // if this is a read-only view of an existing batch, don't update the sublocations
+    if (record?.id !== -1) {
+      return;
+    }
     setRecord((previousRecord: FormRecord): FormRecord => {
       if (!previousRecord) {
         return previousRecord;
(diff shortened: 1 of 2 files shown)


src/components/InventoryV2/form/useSubLocations.ts
```diff
@@ -37,10 +37,8 @@ export const useSubLocations = (nurseryId?: number, record?: { subLocationIds?:
   }, [availableSubLocations, record?.subLocationIds]);
 
   useEffect(() => {
-    if (!availableSubLocations) {
-      void initSubLocations();
-    }
-  }, [availableSubLocations, initSubLocations]);
+    void initSubLocations();
```
karthikbtf, line 40: initSublocations callback changes when nursery id or record's facility id changes, we need to refetch sublocations when either of those change

src/components/InventoryV2/form/BatchDetailsForm.tsx
```diff
@@ -230,6 +230,24 @@ export default function BatchDetailsForm(props: BatchDetailsModalProps): JSX.Ele
...
+      return {
+        ...previousRecord,
+        subLocationIds: [],
+      };
+    });
+  }, [availableSubLocations, record?.id, setRecord]);
+
   return (
     <>
       {record && (
```
karthikbtf, line 253: @nickgraz would you know how I can trigger this read-only view? can't seem to find the use-case in the inventory views.

nickgraz: It used to be the main way to view a batch's details, but now they are forwarded to the batch details page. I think you can still trigger it if you have a q

nickgraz: It used to be the main way to view a batch's details, but now they are forwarded to the batch details page. I think you can still trigger it if you have a query param in the parent view for the open batch. I suspect there are still other views out there that might link to it that way.

nickgraz (approved):
Looks good to me

PR #1984, opened 2024-01-03
SW-4652 Only export visible inventory-by-nursery fields

sgrimm:
CSV export of the "by nursery" inventory view was requesting the same set of
fields that are used to populate the UI, including IDs and raw numeric values.
The IDs weren't useful to users, and the raw numeric values were causing the
export to fail because CSVs need to be localized and thus can't include
non-localized fields.

Update the request construction code to only include the user-visible fields
when exporting the view to a CSV.

### DIFF
--- a/src/services/NurseryInventoryService.ts
+++ b/src/services/NurseryInventoryService.ts
@@ -254,21 +254,26 @@ const searchInventoryByNursery = async ({
   query,
   isCsvExport,
 }: SearchInventoryParams): Promise<SearchResponseElement[] | null> => {
+  const exportedFields = [
+    'facility_name',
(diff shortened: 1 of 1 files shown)


sgrimm:
* `main`
  * **PR #1984** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/1984?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/1984?utm_source=stack-comment).

nickgraz (approved):

PR #1990, opened 2024-01-04
SW-4657 Reuse interactive species dropdown in inventory create form

karthikbtf:
- allows for search
- renders species scientific in italic and common in parens
- fixed types to enable reuse


nickgraz (approved):
Looks good

PR #1993, opened 2024-01-04
SW-4666 Add spinners to show busy state when creating nursery inventory

karthikbtf:
- added spinners

### DIFF
--- a/src/components/common/PageForm.tsx
+++ b/src/components/common/PageForm.tsx
@@ -1,17 +1,24 @@
+import { BusySpinner } from '@terraware/web-components';
 import PageForm, { PageFormProps as FormProps } from '@terraware/web-components/components/PageForm';
 import strings from 'src/strings';
 
 export type PageFormProps = Omit<FormProps, 'saveButtonText' | 'cancelButtonText'> & {
+  busy?: boolean;
   cancelButtonText?: string;
   saveButtonText?: string;
 };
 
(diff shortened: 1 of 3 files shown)


src/components/common/PageForm.tsx
```diff
@@ -1,17 +1,24 @@
...
+  const { busy, cancelButtonText, saveButtonText, ...formProps } = props;
 
-  return PageForm({
-    ...formProps,
-    saveButtonText: saveButtonText || strings.SAVE,
-    cancelButtonText: cancelButtonText || strings.CANCEL,
-  });
+  return (
+    <>
+      {busy && <BusySpinner withSkrim={true} />}
```
karthikbtf, line 16: i will move this busy prop into the PageForm and DialogBox in a follow up PR

nickgraz (approved):
Looks good, what does "skrim" mean? I see it a lot and I see that it adds some classes to dialogs and other things... but what does that word mean? (Google turned up a bunch of things that are probably unrelated 😅 )

karthikbtf:
> Looks good, what does "skrim" mean? I see it a lot and I see that it adds some classes to dialogs and other things... but what does that word mean? (Google turned up a bunch of things that are probably unrelated 😅 )

skrim is the translucent backdrop, I can't recall how/when we came up with that name or if it was in some design 🤣

PR #1999, opened 2024-01-04
Fix link font size in app notifications

karthikbtf:
- was large at 16px, fixed to 14px now


karthikbtf:
@nickgraz in case this was missed.

nickgraz (approved):

PR #2000, opened 2024-01-04
SW-4647 Fix display versus comparsion values for accession remaining quantity in inventory create validation

karthikbtf:
- was using the raw value for display and localized value for comparison, switched that around now


nickgraz (approved):

nickgraz:
No problem, looks good!

PR #2008, opened 2024-01-05
SW-4521 When projects feature is on, add org name for non-project reports in report table. Show project name and org name in report table

nickgraz:

### DIFF
--- a/src/components/Reports/ReportsView.tsx
+++ b/src/components/Reports/ReportsView.tsx
@@ -19,6 +19,7 @@ import { requestReportsSettings } from 'src/redux/features/reportsSettings/repor
 import PreSetupView from 'src/components/Reports/PreSetupView';
 import { APP_PATHS } from 'src/constants';
 import ReportSettingsEditFormFields from './ReportSettingsEditFormFields';
+import isEnabled from '../../features';
 
 const columns = (): TableColumnType[] => [
   { key: 'name', name: strings.REPORT, type: 'string' },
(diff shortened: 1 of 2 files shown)


nickgraz:
* `main`
  * **PR #2008** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2008?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2008?utm_source=stack-comment).

src/components/Reports/TableCellRenderer.tsx
```diff
@@ -13,7 +13,17 @@ export default function ReportsCellRenderer(props: RendererProps<TableRowType>):
     const reportLocation = {
       pathname: APP_PATHS.REPORTS_VIEW.replace(':reportId', row.id.toString()),
     };
-    return <Link to={reportLocation.pathname}>{`${row.year}-Q${row.quarter}` as React.ReactNode}</Link>;
+
+    let reportName = `${row.year}-Q${row.quarter}`;
+    if (row.projectName) {
```
nickgraz, line 18: No feature flag check here since the reports can only be created for a project if someone has the ability to create projects in the first place.

src/components/Reports/ReportsView.tsx
```diff
@@ -57,11 +57,19 @@ export default function ReportsView(props: ReportsViewProps): JSX.Element {
   useEffect(() => {
     const refreshSearch = async () => {
       const reportsResults = await ReportService.getReports(selectedOrganization.id);
-      setResults(reportsResults.reports || []);
+      setResults(
+        (reportsResults.reports || []).map((report) => {
+          if (report.projectName) {
+            return report;
+          }
+          // Reports without a project name are for the organization
```
karthikbtf, line 65: can a project name be an empty string? wondering if we should key off the project id !== undefined

karthikbtf (approved):
lgtm, one comment but could be a non-issue

src/components/Reports/ReportsView.tsx
nickgraz: The server doesn't allow it, but it looks like the project creation is not correctly validating that. I will fix that up in another PR.

PR #2018, opened 2024-01-09
SW-4676 Turn the query filters hook into a session filters hook that stores the selected filters into the browser session

nickgraz:
- The "saved" filters are now additionally backed out to the browser session. 
- The filters are based on a particular view, so they don't transfer to other views.

The way it works is that the query will take precedence over the session. If a user navigates to a view from a URL that is sent to them, for example, it will load those filters into the view and write them to their session. If a user navigates to the view without query params, and there are filters present in their session, they will be applied to the view and written to the address bar.

**Navigate to a view with session filters only**

**Navigate to a view with query parameter filters only**


nickgraz:
* `main`
  * **PR #2018** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2018?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2018?utm_source=stack-comment).

alexTerraformation:
Vercel preview works great

karthikbtf (approved):
Nice 🎉 
Accessions table filters uses an in-memory approach, we should maybe redo that as well.. or capture in a ticket.

src/utils/filterHooks/index.ts
```diff
@@ -0,0 +1,121 @@
+import _ from 'lodash';
+import { isArray } from 'src/types/utils';
+
+type FilterValue = (string | number | null)[] | boolean | null;
+export type FiltersType = Record<string, FilterValue>;
+
+// Convert an array to a comma separated string and other values to a string
+// Array values are sorted, so we can detect equality easier elsewhere
+export const normalizeValue = (value: FilterValue): string => {
```
karthikbtf, line 9: just curious.. would a better pattern be
`?filter_facility_id=123&filter_facility_id=345` and use `query.getAll('filter_facility_id')`
over
`?filter_facility_ids=123,345` 
I understand it's a bigger code change but just for discussion.

nickgraz: Yea I was thinking about that, wouldn't mind making the change. I think the way you propose is probably better in the long run because if the filter values end up having a comma things will have to get rewritten anyways.

karthikbtf: Got it, if the rewrite is bigger than anticipated.. we can certainly track it for future tech investment. The pros to keeping it as-is --> already validated and no comma in value use-case yet.

PR #2020, opened 2024-01-09
Updating all material and emotion packages to their latest compatible versions

nickgraz:


nickgraz:
* `main`
  * **PR #2020** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2020?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2020?utm_source=stack-comment).

src/components/common/Link.tsx
```diff
@@ -50,7 +50,6 @@ export default function Link(props: LinkProps): JSX.Element {
       component='button'
       className={classNameToUse}
       onClick={onClick}
-      target={target}
```
nickgraz, line 53: This seems to have been mistakenly added, this prop didn't exist in Material UI 5.11.9 even though it wasn't being reported by TS. It is failing now, and from what I can tell, we aren't passing `target` into this `Link` component anywhere within this app, so it _should_ be safe to remove.

karthikbtf: Good find. I think target is only relevant for the router link.

karthikbtf (approved):
🙏

PR #2021, opened 2024-01-09
SW-4650 Add the ability to change status of plants from batch details

alexTerraformation:
Screenshot 2024-01-09 at 12 56 56 PM
Screenshot 2024-01-09 at 12 57 04 PM


nickgraz (approved):
Nice!

PR #2024, opened 2024-01-09
SW-4680 Handle StatusChanged batch history items

alexTerraformation:
StatusChanged are handled the same as QuantityEdited events


nickgraz (approved):
This could be cleaned up with something like `const isChangeEvent = (type: BatchHistoryItemForTable['type']) => ['StatusChanged', 'QuantityEdited'].includes(type);`. Not necessary but would help to consolidate the checks. You can export it from a relevant `types.ts` file

alexTerraformation:
> This could be cleaned up with something like `const isChangeEvent = (type: BatchHistoryItemForTable['type']) => ['StatusChanged', 'QuantityEdited'].includes(type);`. Not necessary but would help to consolidate the checks. You can export it from a relevant `types.ts` file

I actually tried this and ran into some issues convincing Typescript that the history item would actually have quantities (which are not required) without the explicit check.  I'm sure there is a simple solution that I'm not aware of.  Would be happy to see it if you can show me.

PR #2028, opened 2024-01-10
SW-4639 Misc text and style updates for the planting site create stepper flow

karthikbtf:
- updated text with icons
- added border radius to map
- fixed text to reflect interactive state on map
- we don't have slicing tool icon yet, using a placeholder for now

Terraware App 2024-01-10 12-04-49


src/components/Map/useMapIcons.tsx
```diff
@@ -0,0 +1,32 @@
...
+    width: '20px',
+    height: '20px',
+  },
+}));
+
+export type MapIconType = 'polygon' | 'slice' | 'trash';
+
+type MapIcons = Record<MapIconType, JSX.Element>;
+
+export default function useMapIcons(): MapIcons {
```
nickgraz, line 20: Is there a reason you went for a hook instead of a component?

karthikbtf: I need to switch this to components, I recently got some icon assets from Chudo. Will do that in a follow up PR for https://terraformation.atlassian.net/browse/SW-4697

nickgraz (approved):

PR #2031, opened 2024-01-11
SW-4547 Add interactions to allow cutting zones in planting site editor

karthikbtf:
- added zone cutting interactions, editing the zone name etc will be follow up PRs
- renamed files can components for simplicity
- renamed a bunch of variables for clarity
- some of this code will need to be refactored again as I figure out upcoming capabilities

screencast 2024-01-11 11-11-472


src/components/Map/utils.ts
```diff
@@ -111,3 +115,37 @@ export const getMapDrawingLayer = (source: MapSourceRenderProperties, sourceId:
...
+ * Cut source polygons into smaller polygons
+ * @param source
+ *  The source polygons that need to be split into smaller polygons
+ * @param cutWith
+ *  The polygon to use as the overlap check in order to cut polygons
+ * @returns
+ *  List of polygons that were cut as a result of overlaps with cutWith
+ *  Null if there was no overlap
+ */
+export const cutPolygons = (source: GeometryFeature[], cutWith: Geometry): GeometryFeature[] | null => {
```
nickgraz, line 129: Can you add a test for this?

karthikbtf: This code is not done yet, it is still evolving. I created a [JIRA](https://terraformation.atlassian.net/browse/SW-4715) and will add a test when it is ready. That ok?

nickgraz (approved):
Nice work!

PR #2033, opened 2024-01-12
SW-4698 Move centralized planting site editor logic into relevant components

karthikbtf:
- separation of concern
- make top level editor agnostic of downstream logic
- provide validation hooks upon save
- this is needed to begin implementation of the complete feature set


src/components/Map/MapIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+  },
+}));
+
+export type MapIconType = 'polygon' | 'slice' | 'trash';
+
+export type MapIconProps = {
+  icon: MapIconType;
+};
+
+export default function MapIcon({ icon }: MapIconProps): JSX.Element {
```
nickgraz, line 21: 🙏

nickgraz (approved):
Nice work! Thanks for adding that test

karthikbtf:
> Nice work! Thanks for adding that test

Thanks for the suggestions 🙏 
I need to use graphite.. dealing with merge issues after I merged the parent branch. Will fix that shortly.

PR #2045, opened 2024-01-17
SW-4633 Reuse planting site details input form in user detailed sites flow

karthikbtf:
- import the details input from component
- enable planting seasons selector for detailed sites

Terraware App 2024-01-17 12-10-20


nickgraz (approved):
Looks good to me!

PR #2047, opened 2024-01-17
SW-4697 Use icon-slice asset in user detailed sites flow

karthikbtf:
- see screenshots
- had to add icon-slice.svg as an asset for the map draw button background

Terraware App 2024-01-17 13-48-44


karthikbtf:
> I'm assuming it is impossible to use the icon asset from `web-components` within map, is that right?

Yes, the buttons are provided by the map/polygon-draw tool which is external code.. and comes with background images on the buttons. Not sure of a cleaner way to change this without resetting the bg image.

PR #2049, opened 2024-01-18
SW-4696 Form to update zone names and target planting density on zone creation and selection

karthikbtf:
- added form content to accept zone name and target planting density + interactions/validation
- fixed a race condition between map click and map double-click callbacks that pops up the form
- the styles/font-sizes will be cleaned up towards the end once we have full functionality
- see attached video

screencast 2024-01-17 16-52-42


src/components/Map/EditableMapV2.tsx
```diff
@@ -192,14 +192,18 @@ export default function EditableMap({
...
-          id,
-          lng,
-          lat,
-          properties,
-          sourceId,
-          active: true,
-        });
+        // the timeouts are to avoid a race condition upon click vs double-click both of which have
+        // popup triggers but with a priority (this popup gets overridden by the one from zone creation)
+        setTimeout(() => {
```
nickgraz, line 197: Is there another way to do this? It seems a bit hacky and I suspect that it will make future (hopefully) automated tests fail intermittently.

src/components/Map/EditableMapV2.tsx
```diff
@@ -239,7 +243,7 @@ export default function EditableMap({
       // this is to avoid a racey interaction with double-clicks
       setTimeout(() => {
         initializePopupInfo(overridePopupInfo);
-      }, 0);
+      }, 1);
```
nickgraz, line 246: This also seems a bit suspect

src/components/Map/EditableMapV2.tsx
karthikbtf: Yes, this is sad. We have two click events that do different things - differentiating between the two is getting tricky (click on zone in map vs click to finish new polygon/zone on map) - both popup the info box on the clicked zone. The set timeouts allowed for clicking across zones to popup the info box for that zone, without having to close the previous info box. I can restrict the UX to not popup a new info box as long as one is already open - user will have to close the current info box before opening the new one. Not a big sacrifice.

src/components/Map/EditableMapV2.tsx
```diff
@@ -135,52 +140,44 @@ export default function EditableMap({
...
+  }, [clearPopupInfo, popupInfo, popupRenderer]);
 
   // map click to fetch geometry and show a popup at that location
   const onMapClick = useCallback(
     (event: any) => {
       const { lat, lng } = event.lngLat;
 
       // handle overlap with polygon editor clicks
-      if (editMode === 'CreatingBoundary') {
+      if (editMode === 'CreatingBoundary' || overridePopupInfo) {
```
karthikbtf, line 180: we won't show a popup while the existing popup on new polygon creation is still active, this helps remove race conditions for now.

src/components/Map/EditableMapV2.tsx
nickgraz: I don't want to hold you up or hurt the UX (I think making them close the box before they can open another one sounds less ideal than the current setup), sounds like a necessary evil at the moment.

nickgraz (approved):

src/components/Map/EditableMapV2.tsx
karthikbtf: Gist of the problem:
1. the mapbox polygon draw control has click events that trigger creation of polygon and popping up of the info box
2. the same click is also captured by the map and calls onClose on the popup info box (equivalent to clicking outside the box to close it).. hence all that earlier hacky logic to workaround this
I can't find a way to get a hold of the click event from step 1. --> original plan was to stopPropagation of that event (since it is only relevant to polygon creation).. but there's no hook to this event, sigh. That would have been ideal.

karthikbtf: Created [jira](https://terraformation.atlassian.net/browse/SW-4760) to follow up again later.

PR #2050, opened 2024-01-18
SW-4750 Move accessions database into own view file, implement session / query filters for the accession state

nickgraz:
- Accessions database is now in `AccessionsView`, and I pulled some of the stuff out of the `App` component that was only specific to that view
- Refactored the `populateSpecies` stuff to use the redux dispatch / select that @karthikbtf set up
- Implement `useSessionFilters` hook into the `Database` component, for now it is only syncing the `state` filter since that is the one that is broken (sometimes) in the bug
- Moved a couple things out of `App` since I was also going to use them in the `AccessionsView` like `isPlaceholderOrg` and `selectedOrgHasFacilityType`


nickgraz:
* `main`
  * **PR #2050** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2050?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2050?utm_source=stack-comment).

karthikbtf:
> * Accessions database is now in `AccessionsView`, and I pulled some of the stuff out of the `App` component that was only specific to that view
> * Refactored the `populateSpecies` stuff to use the redux dispatch / select that @karthikbtf set up
> * Implement `useSessionFilters` hook into the `Database` component, for now it is only syncing the `state` filter since that is the one that is broken (sometimes) in the bug
> * Moved a couple things out of `App` since I was also going to use them in the `AccessionsView` like `isPlaceholderOrg` and `selectedOrgHasFacilityType`

Thanks for doing this, we had a lot of state propagated down from App.. removed most of it. Accessions was the last big one, which we avoided touching until since it is a critical feature.

src/App.tsx
```diff
@@ -216,7 +178,7 @@ function AppContent() {
       }
     };
     populatePlantingSites();
-  }, [dispatch, selectedOrganization.id, activeLocale]);
+  }, [selectedOrganization.id, dispatch, activeLocale]);
```
karthikbtf, line 181: is this a preferred order? properties first then objects..?

src/components/AccessionsView.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  useEffect(() => {
+    setDefaults();
+  }, [setDefaults]);
+
+  useEffect(() => {
+    if (!species) {
+      void dispatch(requestSpecies(selectedOrganization.id));
+    }
+  }, [dispatch, selectedOrganization.id, species]);
```
karthikbtf, line 60: Remove this since App is guaranteed to request this data? Worst case we could have two requests in-flight.

src/components/seeds/summary/AccessionByStatus.tsx
```diff
@@ -16,7 +16,7 @@ export default function AccessionByStatus({ label, status, quantity }: Props): J
   return (
     <Link
       id={`update-${status}`}
-      to={`${APP_PATHS.ACCESSIONS}?stage=${status}`}
+      to={`${APP_PATHS.ACCESSIONS}?accessions_filter_state=${status}`}
```
karthikbtf, line 19: Thanks for reusing common mechanism here.

karthikbtf (commented):
Thanks for the refactor, much nicer 👍 
One observation is that we can't clear the state fiters in the accessions view, keeps restoring the previous ones.

src/providers/contexts.ts
```diff
@@ -23,6 +24,18 @@ export const defaultSelectedOrg: Organization = {
   totalUsers: 0,
 };
 
+export const isPlaceholderOrg = (id: number | undefined) => !id || id === defaultSelectedOrg.id;
+
+export const selectedOrgHasFacilityType = (organization: Organization, facilityType: FacilityType): boolean => {
```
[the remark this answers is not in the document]
nickgraz: I'll move em over

src/components/AccessionsView.tsx
nickgraz: Long term I want to remove it from the app component, the main app component shouldn't be requesting a bunch of data that may or may not be looked at in my opinion.

src/App.tsx
nickgraz: That is strange... this was auto populated by my IDE but I was not aware that it had any sort of preference.

nickgraz: I put it back

src/App.tsx
```diff
@@ -65,10 +65,10 @@ import ProjectsRouter from 'src/components/Projects/Router';
...
 import InventoryCreateView from './components/InventoryV2/InventoryCreateView';
 import ReportsRouter from 'src/components/Reports/Router';
-import AccessionsView from './components/AccessionsView';
-import { selectSpecies } from './redux/features/species/speciesSelectors';
-import { requestSpecies } from './redux/features/species/speciesThunks';
-import { isPlaceholderOrg, selectedOrgHasFacilityType } from './providers/contexts';
+import AccessionsView from 'src/components/AccessionsView';
+import { selectSpecies } from 'src/redux/features/species/speciesSelectors';
+import { requestSpecies } from 'src/redux/features/species/speciesThunks';
+import { isPlaceholderOrg, selectedOrgHasFacilityType } from 'src/utils/organization';
```
karthikbtf, line 71: No preference here, `./` is fine too. 
Relative paths with `../` are trickier to grok and update when there are changes.

karthikbtf (commented):
Looks good @nickgraz - just one observation about not being able to clear state filters, they are super sticky.

nickgraz:
> Thanks for the refactor, much nicer 👍 One observation is that we can't clear the state fiters in the accessions view, keeps restoring the previous ones.

Sorry for the delay, this is fixed now.

src/App.tsx
```diff
@@ -416,7 +356,7 @@ function AppContent() {
             </Route>
             <Route exact path={APP_PATHS.SPECIES}>
               {selectedOrgHasSpecies() ? (
-                <SpeciesList reloadData={reloadSpecies} species={species} />
+                <SpeciesList reloadData={reloadSpecies} species={species || []} />
```
karthikbtf, line 359: maybe extract out into a `const nonNullSpecies = useMemo<Species[]>(() => species || [], [species]);` since we use this in multiple places.

src/components/AccessionsView.tsx
karthikbtf: Yeah, that's a fair point.. what are your thoughts on making an assumption - that some types of data are globally available/pre-fetched in the app?

src/App.tsx
```diff
@@ -73,6 +65,10 @@ import ProjectsRouter from 'src/components/Projects/Router';
 import { requestProjects } from './redux/features/projects/projectsThunks';
 import InventoryCreateView from './components/InventoryV2/InventoryCreateView';
 import ReportsRouter from 'src/components/Reports/Router';
+import AccessionsView from 'src/components/AccessionsView';
```
karthikbtf, line 68: will this be moved into scenes?

karthikbtf (approved):
Woot!

src/App.tsx
nickgraz: Yea it is in a higher PR

src/components/AccessionsView.tsx
nickgraz: Things like user and org make sense to fetch at the top level for example

src/App.tsx
nickgraz: In my opinion it doesn't make sense to force the check and jam into React's memory since it isn't an "expensive operation"

PR #2053, opened 2024-01-19
SW-4760 Redo map click handling for popup info from multiple sources

karthikbtf:
- cleaned this up better, leveraging one entry point to initialize popups, this removes race conditions and ambiguity
- disabled interactive layers while polygon draw is active, the interactive layers enable map clicks to return clicked-on geometries, which we don't want while polygon is being drawn
- removed a bunch of unnecessary code
- UX is cleaner now where user can click across geometries to see popup info
- see screen/video attached

screencast 2024-01-18 20-25-30


nickgraz (approved):
Nice work @karthikbtf !

karthikbtf:
> Nice work @karthikbtf !

Thanks for flagging the setTimeout shortcuts!

PR #2054, opened 2024-01-19
Extract out feature selection on map click + validate duplicate zone names on input

karthikbtf:
- selection of feature to highlight on map click was trivial for zone editing (since there are no overlapping zones)
- for subzone editing, there will be overlapping zones and subzones, we need a way to decide which feature to select (both are valid selections depending on the state of the workflow)
- extract out feature selection on click into a callback passed by map client, to customize feature selection logic
- also added validation on zone name editing to prevent duplicate zone names

Terraware App 2024-01-19 08-50-44


nickgraz (approved):
👌

PR #2056, opened 2024-01-19
SW-4637 Add subzone creation in user detailed planting sites (intermediate effort)

karthikbtf:
- extracted out common code (between zone and subzone) into utils
- implemented Subzones.tsx using Zones.tsx as a guiding template and filled out differntiating logic
- added selection of zone to enable scoping the subzone creation to a zone
- added code to highlight a geometry (highlight the zone for which we are creating subzones)
- fixed some copy/text as per Chudo's request for start-over confirmation modal
- there are a bunch of TODOs for follow up, this is not the final effort
- see video/screenshot for basic interactions

screencast 2024-01-19 15-17-28


nickgraz (approved):
Nice work, this is awesome!

karthikbtf:
> Nice work, this is awesome!

Thanks! 
There's plenty more to do here, auto generate names, fix color/highlight styles, minimum size/area validation, save to BE, etc. One PR at a time.

chudo76 (commented):
🔥🔥🔥

PR #2059, opened 2024-01-20
SW-3514 Limit table area to viewport

alexTerraformation:
Every table should now be limited to the width of the viewport minus some area for padding.  Additionally, every table card (with a few exceptions listed below) is now setup to go flush for mobile viewports to give tables extra space.

The tables that I verified for this PR:
- Species
- Seeds database
- Seedling inventory by species
- Seedling inventory by nursery
- Seedling inventory by batch
- Nursery inventory
- Species inventory
- Planting progress
- Withdrawal history
- Observations list
- Observation detail view
- Observation zone view
- People
- Projects
- Nurseries
- Seed Banks
- Planting Sites
- Reports List
- Batch withdrawal quantities
- My account organizations (not mobile flush)
- Nursery sublocations (not mobile flush)
- Seed bank sublocations (not mobile flush)

I may be missing some tables here.  If I am, please let me know and I will get them fixed.


chudo76:
Nice! All the tables look and work good. Obvious improvement on how the batch actions bar works and you can see the pagination UI at the bottom of the table now.

A couple of minor things:
1. The stacked Batch Withdraw tables might have been missed. (I just remembered them so I don't think I called them out.)
2. Reports List table isn't flush on mobile but I suspect this is because there's a "Settings" tab that doesn't contain a table. (I don't think it's an issue, we don't have many cases where we have tabs with table + non-table content.)
3. When you scroll down a long table (staging: dev, Seeds > Accessions) and then you scroll up a bit, the interaction with the header reappearing with the button + more menu has them off-screen. (It fixes itself when you scroll all the way back up, though):
image
image

nickgraz (approved):
Nice work!

alexTerraformation:
> 2. Reports List table isn't flush on mobile but I suspect this is because there's a "Settings" tab that doesn't contain a table. (I don't think it's an issue, we don't have many cases where we have tabs with table + non-table content.)

I just checked the Reports list and it seems flush on mobile to me.  Perhaps you mean the report view/edit screens?

Screenshot 2024-01-23 at 1 15 40 PM

chudo76:
> > 2. Reports List table isn't flush on mobile but I suspect this is because there's a "Settings" tab that doesn't contain a table. (I don't think it's an issue, we don't have many cases where we have tabs with table + non-table content.)
> 
> I just checked the Reports list and it seems flush on mobile to me. Perhaps you mean the report view/edit screens?
> 

Oh is it because I'm logged in as admin? I get the "Settings" tab.
image

alexTerraformation:
> > > 2. Reports List table isn't flush on mobile but I suspect this is because there's a "Settings" tab that doesn't contain a table. (I don't think it's an issue, we don't have many cases where we have tabs with table + non-table content.)
> > 
> > 
> > I just checked the Reports list and it seems flush on mobile to me. Perhaps you mean the report view/edit screens?
> 
> Oh is it because I'm logged in as admin? I get the "Settings" tab. image

Ah okay that's with the Projects feature enabled, which has other issues.  We'll fix those as a part of Projects QA

chudo76:
> Ah okay that's with the Projects feature enabled, which has other issues. We'll fix those as a part of Projects QA

Aaah.. Got it, thanks.

PR #2084, opened 2024-01-24
SW-4777 Refactor zone/subzone edit to use selection styles specific to edit flow

karthikbtf:
- selection on maps will fill the object with the select color
- for creation of planting sites, we don't want to fill with a color, instead just outline the object with a thicker width and color
- also reuse existing MapEntityId/MapEntityOptions types instead of defining that new HighlightData type
- added 'draft-zone', 'draft-subzone' types to allow configuring styles for edit mode
- the actual style values will be updated later with design help

Terraware App 2024-01-24 10-25-23
Terraware App 2024-01-24 10-31-36


src/components/Map/utils.ts
```diff
@@ -83,8 +83,20 @@ export const getMapDrawingLayer = (source: MapSourceRenderProperties, sourceId:
...
       type: 'line',
       paint: {
-        'line-color': source.lineColor,
-        'line-width': source.lineWidth,
+        'line-color': [
+          // if feature is selected, use selectLineColor if defined
+          'case',
+          ['boolean', ['feature-state', 'select'], false],
+          source.selectLineColor ?? source.lineColor,
+          source.lineColor,
```
karthikbtf, line 91: line 90 is active when feature-state 'select' is true on an object, line 91 is the 'else' equivalent
see https://docs.mapbox.com/style-spec/reference/expressions/#case

nickgraz (approved):
Nice refactor

karthikbtf:
> Nice refactor

Thanks 🙏 , there are a few more PRs off this stack, I had to merge/rebase a few times apologies for the email noise.

PR #2088, opened 2024-01-24
SW-4778 Refactor planting site editor undo/redo 

karthikbtf:
- simplified data direction to be top-down
- data changes were previously lateral with the UndoRedoControl component making a copy of all data and listening to changes to update the redo stack
- undo/redo buttons were sending updated data back to the client editor and downstream to the polygon editor
- removed this multiple sources of data and made it top-down with one source
- simplifies the code and need this for the next effort of validating boundary sizes
- no changes to functionality


src/components/Map/UndoRedoControl.tsx
```diff
@@ -0,0 +1,49 @@
+import { Box, useTheme, Theme } from '@mui/material';
```
karthikbtf, line 1: the file was renamed and a bunch of code removed, github flagged the entire file as new

src/components/PlantingSites/editor/SiteBoundary.tsx
```diff
@@ -17,21 +18,21 @@ export type SiteBoundaryProps = {
...
+    ? undefined
+    : {
+        type: 'FeatureCollection',
+        features: [toFeature(boundary, {}, id)],
+      };
 export default function SiteBoundary({ onChange, onValidate, site }: SiteBoundaryProps): JSX.Element {
   const [description, setDescription] = useState<Description[]>([]);
-  const [siteBoundary, setSiteBoundary] = useState<FeatureCollection | undefined>();
+  const [siteBoundary, setSiteBoundary, undo, redo] = useUndoRedoState<FeatureCollection | undefined>(
+    featureSiteBoundary(site.id, site.boundary)
```
karthikbtf, line 31: we want the first state in the undo/redo stack to be the initial boundary state.. hence not using a useEffect which would lead to an initial undefined state (enabling undo up to that point)

src/components/PlantingSites/editor/Subzones.tsx
```diff
@@ -34,31 +35,39 @@ export type SubzonesProps = {
...
+    {} as Record<number, FeatureCollection>
+  );
+
 export default function Subzones({ onValidate, site }: SubzonesProps): JSX.Element {
   const [selectedZone, setSelectedZone] = useState<number | undefined>();
-  const [zones, setZones] = useState<FeatureCollection | undefined>();
   // map of zone id to subzones
-  const [subzones, setSubzones] = useState<Record<number, FeatureCollection>>({});
+  const [subzones, setSubzones, undo, redo] = useUndoRedoState<Record<number, FeatureCollection>>(
+    featureSiteSubzones(site)
```
karthikbtf, line 54: same here, first value on stack should be the initial subzones state, hence not leveraging a useEffect

src/components/PlantingSites/editor/Zones.tsx
```diff
@@ -27,20 +28,22 @@ export type ZonesProps = {
...
+    const features = site.plantingZones.map(plantingZoneToFeature);
+    return { type: 'FeatureCollection', features };
+  } else {
+    return undefined;
+  }
+};
+
 export default function Zones({ onChange, onValidate, site }: ZonesProps): JSX.Element {
-  const [zones, setZones] = useState<FeatureCollection | undefined>();
+  const [zones, setZones, undo, redo] = useUndoRedoState<FeatureCollection | undefined>(featureSiteZones(site));
```
karthikbtf, line 41: same here, first value on stack should be the initial zones

karthikbtf:
@nickgraz when you can, look at this and a few others open as well.

src/hooks/useUndoRedoState.ts
```diff
@@ -19,26 +21,27 @@ export default function useUndoRedoState<T>(initialValue?: T): [T | undefined, S
   const [stack, setStack] = useState<(T | undefined)[]>([initialValue]);
   const [stackIndex, setStackIndex] = useState<number>(0);
 
-  const data = useMemo<T | undefined>(() => stack[stackIndex], [stackIndex, stack]);
+  const data = useMemo<T | undefined>(() => _.cloneDeep(stack[stackIndex]), [stackIndex, stack]);
 
   const setData = useCallback(
-    (value: T) => {
+    (input: Func<T> | T) => {
+      const value = typeof input === 'function' ? (input as Func<T>)(_.cloneDeep(stack[stackIndex])) : input;
```
nickgraz, line 28: Can this be `(input as Func<T>)(data)` since it is memoized above?

nickgraz (approved):
One small question, looks great otherwise, nice refactor!

src/hooks/useUndoRedoState.ts
karthikbtf: Yes it can, I had it that way then decided against it.. will use that. Thanks for all the reviews!

PR #2091, opened 2024-01-25
SW-4783 Update site with edited subzones upon save

karthikbtf:
- populate zones with subzones info and update site with root-level zone property
- the population of subzone is currently a place-holder, which will be updated with new API requirements when it is available
- also fixed start over to update boundary related properties with original site data


src/components/PlantingSites/editor/Subzones.tsx
```diff
@@ -73,9 +73,39 @@ export default function Subzones({ onValidate, site }: SubzonesProps): JSX.Eleme
...
+    const numZones = site.plantingZones?.length ?? 0;
+    let numSubzones = 0;
+    const plantingZones: PlantingZone[] | undefined = site.plantingZones?.map((zone) => {
+      const plantingSubzones: PlantingSubzone[] = (subzones?.[zone.id]?.features ?? [])
+        .map((subzone) => {
+          const { geometry, properties } = subzone;
+          const multiPolygon = toMultiPolygon(geometry);
+          if (multiPolygon && properties) {
+            return {
+              areaHa: 0, // TODO update with api requirements when ready, we shouldn't be calculating this on the client
```
nickgraz, line 91: 👍

src/components/PlantingSites/editor/Subzones.tsx
```diff
@@ -73,9 +73,39 @@ export default function Subzones({ onValidate, site }: SubzonesProps): JSX.Eleme
...
+              id: properties.id!,
+              name: properties.name!,
+              plantingCompleted: false,
+            };
+          } else {
+            return undefined;
+          }
+        })
+        .filter((subzone) => !!subzone) as PlantingSubzone[];
+      numSubzones += plantingSubzones.length;
```
nickgraz, line 103: Might just be a personal preference, but I think mutating a variable in a higher scope within a `.map` isn't great. I know it seems excessive, but I think the `numSubzones` should be derived in a similar way to `numZones`.

Although... saying it out loud, I have been trying to adjust my threshold for "readable code" versus "performant code". What you have here is definitely more performant than subsequently using `reduce` or some other means to derive `numSubzones`. It just feels dirty to do things other than mapping within map (in this case mutating within a map within a map). 

Perhaps some `for` loops would be more semantically correct, and in this case more performant?

nickgraz: Totally reasonable to slap that comment down too, it is a bit nit picky.

karthikbtf: That's a fair suggestion, I will move it out of the .map (whose sole function is to return mapped data).

nickgraz (approved):

PR #2092, opened 2024-01-25
SW-4685 Handle invalid inventory status changes

alexTerraformation:
Went with validation on submit since that is consistent with our design philosophy.  Don't love the way this looks with the error text, but redesigning that is outside the scope of handling this particular error case.

Screenshot 2024-01-25 at 10 03 54 AM


src/strings/csv/en.csv
```diff
@@ -447,6 +447,7 @@ GERMINATING_QUANTITY,Germinating Quantity,
 GERMINATING_QUANTITY_REQUIRED,Germinating Quantity *,
 GERMINATION_QUANTITY_AFTER,Germination Quantity (After),
 GERMINATION_QUANTITY_BEFORE,Germination Quantity (Before),
+GERMINATING_QUANTITY_MUST_BE_GREATER_THAN_ZERO,Germinating quantity must be greater than 0
```
karthikbtf, line 450: thoughts on 'cannot be less than 0'? it can be equal to 0 (can be greater than or equal to 0)... same for the other case

karthikbtf (commented):
lgtm, one comment on text/copy
the style.. yeah this is a super corner case

src/strings/csv/en.csv
alexTerraformation: Good catch thanks!

karthikbtf (approved):
Thanks!

PR #2093, opened 2024-01-26
SW-4788 Support drawing multiple disjoint polygons for site boundary

karthikbtf:
- missed this requirement earlier
- we already have support for drawing multiple polygons
- use that mode in site boundary editor
- remove option 'allowMultiplePolygons' since that's the only mode now
- added autoFocus in details input form, unrelated but was annoying to always click into the input form

Terraware App 2024-01-25 16-41-56


karthikbtf:
reverted 2 commits that were meant to be on another branch

nickgraz (approved):
Nice and simple!

PR #2099, opened 2024-01-27
SW-4792 Use flex/grow for map views

karthikbtf:
- planting site map view
- observations map view
- these were using shorter heights, should have been full height
- will copy Chudo in case this shouldn't be a fix

### Before

Terraware App 2024-01-26 16-52-35

Terraware App 2024-01-26 16-52-18

### Fix

Terraware App 2024-01-26 16-45-54

Terraware App 2024-01-26 16-48-07


chudo76 (approved):
Nice improvement! LGTM!

PR #2104, opened 2024-01-29
SW-4734 Use singular or plural forms of localized strings for project added toast body

nickgraz:
- Added function that renders singular vs plural entity name in project create flow success toast


joshbuchea (approved):
Looks good!

PR #2111, opened 2024-01-31
SW-4762 Show associated project on planting site page

karthikbtf:
- when the project was newly created and associated with a site, navigating to the planting site view shows cached site data (without the associated project)
- fix is for the planting site view to refetch the site (and also update the redux state incrementally)


src/redux/features/tracking/trackingSlice.ts
```diff
@@ -4,28 +4,47 @@ import { PlantingSiteZone } from 'src/types/PlantingSite';
...
+      const data: SiteData = action.payload;
+      const { error, locale, plantingSite } = data;
+      state.error = error;
+      if (plantingSite) {
+        // update state with planting site, resort with new information
+        state.plantingSites = [
+          ...(state.plantingSites ?? []).filter((site) => site.id !== plantingSite.id),
+          plantingSite,
+        ].sort((a, b) => a.name.localeCompare(b.name, locale || undefined));
+      }
```
nickgraz, line 42: 🔥

nickgraz (approved):
Nice work, I like the way you combined the responses from the two requests into state!

PR #2113, opened 2024-01-31
SW-4686 Error handling when user drawn boundary is too large

karthikbtf:
- track bounding area and show error upon save, and on canvas
- use a new error source on the map
- draw error annotations on map
- refactor to use more types
- attached screenshot/video

screencast 2024-01-31 15-04-01

screencast 2024-01-31 15-19-17


karthikbtf:
@chudo76 here's a revised version

chudo76 (approved):
Looks good, thanks!

PR #2114, opened 2024-02-01
SW-4800 Let user know if a site name is in use, when editing the site name

karthikbtf:
- support for duplicate site names within an organization does not exist, backend schema requires unique site names within an organization
- show error when user types a site name that is already in use

Terraware App 2024-01-31 16-03-23


src/scenes/PlantingSitesRouter/DetailsInputForm.tsx
```diff
@@ -33,34 +37,54 @@ export default function DetailsInputForm({
...
   const projectsEnabled = isEnabled('Projects');
+  const { activeLocale } = useLocalization();
+  const { selectedOrganization } = useOrganization();
+  const dispatch = useAppDispatch();
+  const plantingSites = useAppSelector(selectPlantingSites);
+
+  useEffect(() => {
+    if (!plantingSites) {
+      dispatch(requestPlantingSites(selectedOrganization.id, activeLocale));
+    }
```
nickgraz, line 52: Can someone get to this form without the planting sites being force refreshed? For example if a planting site is updated somewhere else, and they navigate to this form, will it have the updated data? I ran into a similar issue with the projects stuff where there was an edge case that made it necessary to always refetch the data for the view to make sure it was up to date since state might have been filled with that data elsewhere.

Another way to solve this (if it is even a problem) is to always dispatch the fetch from any create / update / delete calls so that state is always updated after it needs to be but we don't force the components to re-fetch.

nickgraz (approved):
Looks good, one comment that is probably inconsequential

src/scenes/PlantingSitesRouter/DetailsInputForm.tsx
karthikbtf: It can happen if the site was edited some place else (another user/ another session).
The app currently pre-fetches sites at the root level and doesn't handle cases where edits were made outside of the app session - this is a limitation today. It does capture edits made by the user within the session.
We could revisit this root level fetching (that other components depend on) more holistically to rearchitect later.

PR #2115, opened 2024-02-01
SW-4801 Fix duplicate planting site create/update requests

karthikbtf:
- the useEffect to call the create/update code was being triggered twice
- an unused state variable was being modified within the useEffect retriggering the useEffect
- this was ironic given the state variable was unused (left over from a refactor)
- removed the state variable and retrigger


alexTerraformation (approved):
Thank you!

PR #2116, opened 2024-02-01
SW-4789 Zone name validation

karthikbtf:
- check that zone names are set before saving, reject save otherwise with error (cc @chudo76 )
- for simple sites, a default zone name of 'Zone' is used (since there is no zone editing flow)
- for detailed sites, user sets the zone name
- a few font/color edits (for iteration)

Terraware App 2024-02-01 09-49-26


chudo76 (approved):
Looks good, thanks!

PR #2120, opened 2024-02-02
SW-4789 Zone boundary validation 

karthikbtf:
- Check that zone boundaries are of a minimum size when 'attempting' to create them
- Zones will not be created if they are of an invalid size
- Zone size check is a stopgap measure until we have BE API (jira created)
- Added error annotations to show when there is an error, and to show errors on save
- Fixed/optimized some code to avoid passing around error annotation properties to the draw polygon render (map component can handle it)
- split up some useEffects that should have been separate

Terraware App 2024-02-01 17-45-31


src/scenes/PlantingSitesRouter/editor/Zones.tsx
```diff
@@ -138,45 +162,74 @@ export default function Zones({ onChange, onValidate, site }: ZonesProps): JSX.E
...
+      {
+        text: strings.SITE_ZONE_BOUNDARIES_SIZE,
+        isBold: true,
+      },
     ],
     []
   );
 
-  const onEditableBoundaryChanged = (featureCollection?: FeatureCollection) => {
+  const onEditableBoundaryChanged = (editableBoundary?: FeatureCollection) => {
```
nickgraz, line 173: This function is getting a bit beefy, maybe we should split it out so we can add unit tests?

karthikbtf: Thanks for the suggestion. Refactored into a utility which can now be reused for the next effort - subzone size validation. Added tests, etc.

nickgraz (approved):
Looks great, thanks!

PR #2121, opened 2024-02-02
SW-4815 Check that site boundary can support a minimum zone size

karthikbtf:
- we were previously checking for large boundaries
- also need to check for small boundaries since zones cannot be smaller than 100m x 100m
- add a check for all disjoint site boundary polygons and prevent any of them smaller than 100m x 100m
- create default zone on each of the disjoint polygons (previously we were creating a single zone across polygons that was a bug)

Terraware App 2024-02-01 18-30-10

Terraware App 2024-02-01 18-39-32


nickgraz (approved):
Nice work

PR #2125, opened 2024-02-02
SW-4790 Validate subzone minimum size contraints

karthikbtf:
- visually mark where invalid subzones would be created (without creating them), same flow/utility as in zone validation
- remove the clearOnEdit temporary solution now that zones/subzones have control over this data which flows top down to the draw control
- add option to set the slice tool icon (previously piggy-backed onto clearOnEdit)
- refactor DetailsInputForm to continually validate after first error on save, this was removed during a bugfix last week because it was not separated out into a separate useEffect

Terraware App 2024-02-02 14-52-03


karthikbtf:
Nick, take a look when you get a chance.

nickgraz (approved):
Nice work

PR #2136, opened 2024-02-07
SW-4691 Disallow detailed planting site creation in mobile view

karthikbtf:
- show message asking user to switch to desktop app

Terraware App 2024-02-07 14-28-51


karthikbtf:
@chudo76 thoughts

chudo76 (commented):
I think we need a way for them to go back.. Maybe something like:

image

> **Feature Unavailable** 
> 
> This feature is only available from the desktop application.
>
> [Back to Planting Sites]

karthikbtf:
> I think we need a way for them to go back.. Maybe something like:
> 
> image
> 
> > **Feature Unavailable**
> > This feature is only available from the desktop application.
> > [Back to Planting Sites]

Updated with this feedback, thanks 🙏

nickgraz (approved):
Makes sense

PR #2138, opened 2024-02-07
SW-4643 Add additional tests for locations and accessions

alexTerraformation:
Added a test list which allows me to serialize tests and use outputs from one as inputs to another.  

This is a tradeoff because some tests will be dependent on others and therefore running each test independently is not possible.  The alternative is to build out a lot more test infrastructure that can pre-seed all the scenarios we need.  Would like to get thoughts on this.

Added tests for the following scenarios:

- Adding an accession
- Withdrawing seeds to nursery
- Withdrawing seeds for viability test
- Creating an organization
- Creating a nursery


playwright/e2e/suites/accession.spec.ts
```diff
@@ -0,0 +1,222 @@
+import { test, expect } from '@playwright/test';
+import { waitFor } from '../utils/utils';
+
+test.setTimeout(60000);
+test.beforeEach(async ({ context }, testInfo) => {
+  // Make all requests look like they are associated with an existing login session
+  // so we don't have to depend on a Keycloak server to run the test suite. The
+  // session value here is the base64-encoded session ID from dump/session.sql.
+  await context.addCookies([
```
karthikbtf, line 9: suggestion, move this into utils

const addCookies = async (context) => {
  await context.addCookies(....);
}

each test need only await on this imported addCookies without specifying the cookie/url details

karthikbtf (approved):
Thanks for the new tests 🎉 
This sounds like a good approach for now, and nice find. 
Some thoughts:
Extending your approach - If there's a way to group test suites - run the groups/test-suite in parallel, run tests within the group/test-suite sequentially.. we can use this approach to create data for each test-suite perhaps. 
Pre-seeding the db can help remove extra tests and speed up the test runs for sure. Maybe something for later once we get better at this?

karthikbtf (approved):
Thanks Alex!

karthikbtf (approved):
Nice, thanks for the many new tests 🙏

PR #2143, opened 2024-02-08
SW-4826 Viability test fixes

alexTerraformation:
We were running the validation of the number of seeds tested even when editing an existing viability test which was triggering false failures.  Do not validate number of seeds in viability test after the test has been created. 

Do not allow for editing number of seeds in viability test after the test has been created.  We don't allow editing previously withdrawn quantities elsewhere in the product.


karthikbtf (approved):
Looks good to me. Thanks for the bug discovery and fix.

PR #2146, opened 2024-02-08
Reorg planting sites router code with view/edit subdirectories

karthikbtf:
- the root directory was getting rather large
- moved files under `/view` and `/edit`
- removed `/draft` directory and moved draft create/edit under `/edit` and draft view under `/view`
- will help my next refactor effort for planting site view, potentially introducing new files


nickgraz (approved):
Thanks for the reorg

PR #2147, opened 2024-02-09
SW-4638 Refactor planting site details input form to use a generic type

karthikbtf:
- so we can extend this for both PlantingSite and a planting site draft when it is ready


nickgraz (approved):
🔥

PR #2150, opened 2024-02-09
SW-4838 Add user preference to not show video tutorial again

karthikbtf:
- the properties already existed, just needed to hook them up with preferences

screencast 2024-02-09 10-04-50


nickgraz (approved):
👍

PR #2151, opened 2024-02-09

terraware-phrase:


karthikbtf:
Looks like I need to fix the styles, issue exposed by the translation. Will do that in a follow-up PR.

karthikbtf (approved):

PR #2152, opened 2024-02-09
SW-4631 Refactor planting site view to accept a generic type

karthikbtf:
- generic type is lowest common denominator between sites and draft sites
- most of the planting site specific data like observations is not affected
- next step is to leverage this view to show draft sites which will have similar information

### DIFF
--- a/src/scenes/PlantingSitesRouter/view/GenericSiteView.tsx
+++ b/src/scenes/PlantingSitesRouter/view/GenericSiteView.tsx
@@ -1,0 +1,193 @@
+import { useEffect, useMemo, useState } from 'react';
+import { useHistory } from 'react-router-dom';
+import { Box, Typography, Grid, Theme, useTheme, List, ListItem } from '@mui/material';
+import { makeStyles } from '@mui/styles';
+import { DateTime } from 'luxon';
+import { Button, DropdownItem } from '@terraware/web-components';
+import { useDeviceInfo } from '@terraware/web-components/utils';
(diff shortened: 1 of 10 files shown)


src/scenes/PlantingSitesRouter/view/GenericSiteView.tsx
```diff
@@ -0,0 +1,193 @@
...
+}));
+
+export type GenericSiteViewProps<T extends SiteDetails> = {
+  editDisabled?: boolean;
+  editUrl: string;
+  onDelete: () => void;
+  plantingSite: T;
+};
+
+export default function GenericSiteView<T extends SiteDetails>({
```
karthikbtf, line 40: The bulk of this was extracted out from `PlantingSiteView`, no new functionality.

src/types/Tracking.ts
```diff
@@ -36,3 +36,22 @@ export type MonitoringPlotSearchResult = {
...
+
+export type Location = {
+  timeZone?: string;
+};
+
+export type SitePlantingZone = Omit<PlantingZone, 'areaHa' | 'plantingSubzones'> & {
+  plantingSubzones: Omit<PlantingSubzone, 'areaHa'>[];
+};
+
+export type SiteDetails = Location & {
```
karthikbtf, line 48: This is safe with current state/plans. If PlantingSite and DraftPlantingSite diverge dramatically at some point, we can split things out separately.

nickgraz (approved):
Nice work

PR #2154, opened 2024-02-09
SW-4843 Add types for Draft Planting Sites

karthikbtf:
- integrate with BE generated types
- define client side types
- added tests for conversion to/from draft to BE payload


nickgraz (approved):
Nice work, thanks for the unit test

PR #2162, opened 2024-02-13
SW-4714: FE - Stub out side nav pages

joshbuchea:
This PR includes changes to stub out some pages for the new Accelerator Admin View in Terraware.

I made a couple of assumptions:

- We wanted to replace `/accelerator/admin` with `/accelerator/overview`
- We wanted the same stub UI components from the previous Accelerator Admin View in the new Accelerator Overview View

But am happy to adjust if needed.


nickgraz:
Looking good! Can you combine the `TfMain` and the `Container` components with the `padding: 0` style into an `AcceleratorMain` component or similar?

joshbuchea:
@nickgraz sure, you got it 👍

src/scenes/AcceleratorRouter/NavBar.tsx
```diff
@@ -9,13 +10,18 @@ import LocaleSelector from 'src/components/LocaleSelector';
...
-  setShowNavBar: (value: boolean) => void;
   backgroundTransparent?: boolean;
+  setShowNavBar: (value: boolean) => void;
 };
-export default function NavBar({ setShowNavBar, backgroundTransparent }: NavBarProps): JSX.Element | null {
+export default function NavBar({ backgroundTransparent, setShowNavBar }: NavBarProps): JSX.Element | null {
   const { isDesktop } = useDeviceInfo();
   const history = useHistory();
 
+  const isAcceleratorDeliverablesRoute = useRouteMatch(APP_PATHS.ACCELERATOR_DELIVERABLES + '/');
```
nickgraz, line 20: Are these suffixed `/` needed? Seems like we should be getting sent to the route without a trailing slash. I vaguely remember some redirect code in the regular side of terraware that does some slash to no slash (or the other way around) redirecting.
 
Since we aren't using `exact` in the `Route` components, we should be able to have trailing slash and no slash go to the same place.

nickgraz (approved):
Looks good, one comment / question that may or may not end up changing some code.

src/scenes/AcceleratorRouter/NavBar.tsx
joshbuchea: Good catch! I copied the pattern from `TerrawareRouter/NavBar`, but the trailing slash isn't needed here. Will remove.

PR #2164, opened 2024-02-14
SW-3714 Make species view consistent with other interactions

alexTerraformation:
- Remove bulk action checkbox.
- Add species detail page
- Scientific Name column is a link that goes to detailed view of species.
- There’s an Edit Species button on species detail page.
- There is a Delete action (under “…” kebab) on species detail page.
- Add Species is a page rather than a modal


chudo76:
@alexTerraformation A couple of minor changes if possible:

- Is it possible to have a grayed out checkbox for "Rare" instead of true/false? I think we do this for the Notifications checkbox under My Account:
image
- Can we show a comma delimited list for multiple Ecosystem Types for the Detailed view since we have room?
image

Otherwise looks good. Thank you!

src/scenes/Species/DeleteSpeciesModal.tsx
```diff
@@ -46,60 +44,30 @@ export default function DeleteSpeciesDialog(props: DeleteSpeciesDialogProps): JS
...
     if (inUseSpecies) {
-      const forDeletion = speciesToDelete
-        .filter((species) => !inUseSpecies[species.id.toString()])
-        .map((species) => species.id);
-      const avoidDeletion = speciesToDelete
-        .filter((species) => inUseSpecies[species.id.toString()])
-        .map((species) => species.id);
-      setToDelete(forDeletion);
-      setCannotDelete(avoidDeletion);
+      if (inUseSpecies[speciesToDelete.id.toString()]) {
```
karthikbtf, line 47: Nice refactor. This is fine, another option is disabling the delete button under the options menu with a tooltip on reason.

src/scenes/Species/SpeciesEditView.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { Species } from 'src/types/Species';
+import { useOrganization } from 'src/providers/hooks';
+import { SpeciesService } from 'src/services';
+import TfMain from '../../components/common/TfMain';
+import { APP_PATHS } from 'src/constants';
+import { useHistory, useParams } from 'react-router-dom';
+import useDeviceInfo from 'src/utils/useDeviceInfo';
+import PageSnackbar from 'src/components/PageSnackbar';
+import useSnackbar from 'src/utils/useSnackbar';
+import PageForm from '../../components/common/PageForm';
```
karthikbtf, line 13: src/components

src/scenes/Species/SpeciesEditView.tsx
```diff
@@ -0,0 +1,131 @@
...
+        </Box>
+        <Box
+          sx={{
+            backgroundColor: theme.palette.TwClrBg,
+            borderRadius: '32px',
+            padding: theme.spacing(3),
+            margin: 0,
+          }}
+        >
+          <SpeciesDetailsForm
```
karthikbtf, line 119: nice reuse

src/scenes/Species/index.tsx
```diff
@@ -26,7 +31,22 @@ const SpeciesView = () => {
     return <EmptyStatePage pageName={'Species'} reloadData={reloadSpecies} />;
   }
 
-  return <SpeciesListView reloadData={reloadSpecies} species={species || []} />;
+  return (
```
karthikbtf, line 34: 🎉

karthikbtf (commented):
Great improvement 🙌

alexTerraformation:
> * Can we show a comma delimited list for multiple Ecosystem Types for the Detailed view since we have room?

@chudo76 We have a lot of options for this so I think we should keep the mechanism for shortening the list.  I can extend the cutoff point to the width of the column maybe?

Screenshot 2024-02-14 at 11 27 02 AM

src/scenes/Species/DeleteSpeciesModal.tsx
alexTerraformation: I would be fine with that approach.  Keeps the logic out of the modal.  Do we have a design pattern for a disabled options menu option? @chudo76

karthikbtf: One option is disabling the delete button but showing it under the options menu.
This looks like scope creep, we can do a follow up maybe. I noticed we don't have tooltip options on a button today (unless they are table/top-bar buttons).

alexTerraformation: Good point.  Will enter a Jira to address this later.

karthikbtf (approved):
LGTM 🎉 
I understand there's a Chudo request outstanding.

src/scenes/Species/DeleteSpeciesModal.tsx
chudo76: > I would be fine with that approach. Keeps the logic out of the modal. Do we have a design pattern for a disabled options menu option? @chudo76

For Seed Bank Sub-locations, we just disable the button and have no info for why. For Projects we have a dialog box warning message (we don't prevent). At least with the dialog box we can let people know why they can't delete.

Dialog Box Title: "Delete Species"
Message: "This species is in use and cannot be deleted."
Button: "Got It"

src/scenes/Species/SpeciesDetailView.tsx
```diff
@@ -88,6 +92,7 @@ export default function SpeciesDetailView(): JSX.Element {
       snackbar.toastError(strings.GENERIC_ERROR);
     }
     setDeleteSpeciesModalOpen(false);
+    reloadData();
```
karthikbtf, line 95: reload data only on success?

src/scenes/Species/SpeciesEditView.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const gridSize = () => {
+    if (isMobile) {
+      return 12;
+    }
+    return 4;
+  };
+
+  const goToSpecies = (id?: number) => {
+    const speciesLocation = {
+      pathname: APP_PATHS.SPECIES + (id ? `/${id}` : ''),
```
karthikbtf, line 48: use APP_PATHS.SPECIES_DETAILS similar to L#48 SpeciesAddView.tsx

karthikbtf (approved):
Looks great to me! Thanks so much. A couple comments - ready to merge after that.

PR #2167, opened 2024-02-14
SW-4844 Redux for draft planting site CRUD

karthikbtf:
- search will be a separate effort
- also added an option to use the request arg as key when setting state in redux

### DIFF
--- a/src/redux/features/asyncUtils.ts
+++ b/src/redux/features/asyncUtils.ts
@@ -7,17 +7,17 @@ export type StatusT<T> = { status: Statuses; data?: T };
 export type Status = StatusT<unknown>; // when data is not relevant in the response
 
 export const buildReducers =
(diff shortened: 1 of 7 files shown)


src/redux/features/asyncUtils.ts
```diff
@@ -7,17 +7,17 @@ export type StatusT<T> = { status: Statuses; data?: T };
...
 
 export const setStatus =
-  <T>(status: Statuses) =>
+  <T>(status: Statuses, useArgAsKey?: boolean) =>
   (state: any, action: any) => {
-    const requestId = action.meta.requestId;
+    const key = useArgAsKey ? action.meta.arg : action.meta.requestId;
     const data: T = action.payload as T;
-    state[requestId] = { status, data };
+    state[key] = { status, data };
```
nickgraz, line 22: Nice addition

nickgraz (approved):
Looks good

PR #2168, opened 2024-02-14
SW-4849 Use new draft planting site types in site editor

karthikbtf:
- updated code to use new types
- cleaned up properties that weren't needed
- added a hook to fetch draft site by id (for the edit and view components)
- this is still unfinished work, need to be able to create/save a draft, in a follow up PR


nickgraz (approved):
Looking good!

PR #2173, opened 2024-02-15
SW-4845 Add search for draft sites and list them in planting sites table

karthikbtf:
- added service/redux
- added code to merge/sort sites and draft sites in the table
- added code to fetch draft sites in order to populate set of used site names (so create site flow can flag when user picks a site name that already exists)

Terraware App 2024-02-15 15-09-10

Terraware App 2024-02-15 15-08-46


karthikbtf:
> This all looks good, but I wonder if the `searchData` function is a bit big for this component. It seems like it should be offloaded to a service since it is orchestrating multiple API calls and transforming the result. Would be a lot of args though since you need the locale and time zone variables. Just a thought, I'm OK with it as is.

Thanks for all the reviews, Nick 🙏 
Not sure why I didn't think of moving this into a service function, that should be the right call.
I will capture this as a follow up ticket for now, as a must-do.

PR #2174, opened 2024-02-16
SW-4874 Initialize completion status of optional steps when editing a draft site

karthikbtf:
- when user edits a draft site, initialize completion status of optional steps based on existing information in the draft site
- define a new type for optional edit steps for cleaner code
- misc fixes: set default zone name generation for default zones, replace url from 'new' to 'plantingSiteId/edit' when draft is created so reload works


nickgraz (approved):
Nice work

PR #2176, opened 2024-02-17
SW-4877 Refactor draft planting site editor error validation code looking ahead 

karthikbtf:
- extract out error validation code into a single function `findErrors` which currently has an internal implementation but will be swapped out for BE API at some point
- add code to generate a draft site from temporary changes, which will be used for error validation
- with this change, there should be one place to update error validation down the road with minimal changes


src/scenes/PlantingSitesRouter/edit/editor/Exclusions.tsx
```diff
@@ -29,20 +31,37 @@ const featureSiteExclusions = (site: DraftPlantingSite): FeatureCollection | und
...
-    featureSiteExclusions(site)
-  );
+  const [exclusionsData, setExclusionsData, undo, redo] = useUndoRedoState<Stack>({
+    exclusions: featureSiteExclusions(site),
+  });
   const getRenderAttributes = useRenderAttributes();
   const { activeLocale } = useLocalization();
+  const snackbar = useSnackbar();
+
+  const exclusions = useMemo<FeatureCollection | undefined>(
```
nickgraz, line 48: I think this `useMemo` can go

nickgraz (approved):
Nice work, you make this look easy!

karthikbtf:
> Nice work, you make this look easy!

Thanks 🙏 for the reviews at this busy time!!

PR #2179, opened 2024-02-21
SW-4888 Feature flag for deliverables MVP

karthikbtf:
- Nick, do we want to reuse this for the Accelerator Console as well or have 2 flags?

### DIFF
--- a/src/features.ts
+++ b/src/features.ts
@@ -1,7 +1,12 @@
 import env from 'src/utils/useEnvironment';
 import { CachedUserService } from 'src/services';
 
-export type FeatureName = 'Show Production View' | 'Projects' | 'User Detailed Sites' | 'Deliverables MVP';
+export type FeatureName =
+  | 'Show Production View'
+  | 'Projects'
+  | 'User Detailed Sites'
+  | 'Deliverables MVP'
+  | 'Accelerator Console';
 
 export type Feature = {
   name: FeatureName;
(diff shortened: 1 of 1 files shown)


src/features.ts
```diff
@@ -1,7 +1,7 @@
 import env from 'src/utils/useEnvironment';
 import { CachedUserService } from 'src/services';
 
-export type FeatureName = 'Show Production View' | 'Projects' | 'User Detailed Sites';
+export type FeatureName = 'Show Production View' | 'Projects' | 'User Detailed Sites' | 'Deliverables MVP';
```
karthikbtf, line 4: I can't seem to list these on separate lines, the formatter brings them back to the same line.

nickgraz:
I thought we'd decided to just do one feature flag for the whole "accelerator in terraware" stuff?

karthikbtf:
> I thought we'd decided to just do one feature flag for the whole "accelerator in terraware" stuff?

Was wondering if we wanted to use this flag as the accelerator flag - since this entire effort goes out as one.
I can separate the flags, might be less confusing.

nickgraz:
I was thinking you can just rename this flag to "Accelerator" or something and we can use it across the board.

karthikbtf:
> I was thinking you can just rename this flag to "Accelerator" or something and we can use it across the board.

Cool, will do.

nickgraz (approved):

PR #2180, opened 2024-02-21
SW-4889 Add routes/placeholder-components for deliverables

karthikbtf:
- also gate Accelerator content with feature flag


nickgraz (approved):
⚡

PR #2192, opened 2024-02-23
SW-4919 Use service function to gate display of deliverables menu

karthikbtf:
- check if org has deliverables and show the menu accordingly


joshbuchea (approved):
Looks clear and straight forward 🎉

PR #2195, opened 2024-02-23
Look up project from the project id in the deliverable mock data

karthikbtf:
- this is looking ahead since fetching deliverable by id will return the deliverable schema, which has a project id

### DIFF
--- a/src/components/DeliverableView/TitleBar.tsx
+++ b/src/components/DeliverableView/TitleBar.tsx
@@ -2,11 +2,13 @@ import { Box, Typography, useTheme } from '@mui/material';
 import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
 import { categoryLabel } from 'src/types/Deliverables';
+import { useProjects } from 'src/hooks/useProjects';
 import BackToLink from 'src/components/common/BackToLink';
 import { EditProps } from './types';
 
 const TitleBar = ({ callToAction, deliverable, isAcceleratorConsole }: EditProps): JSX.Element => {
-  const { category, name, projectName } = deliverable;
+  const { category, name } = deliverable;
+  const { selectedProject } = useProjects(deliverable);
   const theme = useTheme();
 
   return (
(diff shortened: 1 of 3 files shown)


src/components/DeliverableView/TitleBar.tsx
```diff
@@ -2,11 +2,13 @@ import { Box, Typography, useTheme } from '@mui/material';
...
 import strings from 'src/strings';
 import { categoryLabel } from 'src/types/Deliverables';
+import { useProjects } from 'src/hooks/useProjects';
 import BackToLink from 'src/components/common/BackToLink';
 import { EditProps } from './types';
 
 const TitleBar = ({ callToAction, deliverable, isAcceleratorConsole }: EditProps): JSX.Element => {
-  const { category, name, projectName } = deliverable;
+  const { category, name } = deliverable;
+  const { selectedProject } = useProjects(deliverable);
```
joshbuchea, line 11: I took a look at `useProjects` because I was confused why the hook accepts a `deliverable` rather than a `project` or `projectId`. It looks like the hook will pull a `projectId` from whatever record is passed into it. That's slick. Perhaps a little too slick? If `projectId` is the only value used from the record, it seems like it would be clearer to pass in a `projectId` in directly.

Not a request for this PR, just thinking out loud here.

joshbuchea (approved):

src/components/DeliverableView/TitleBar.tsx
nickgraz: The idea is that the record extends `{ projectId?: number }` so that we can determine if a project has already been selected on the given record. This makes it so the consumers (which there are a few at this point) don't have to additionally pull the projects in from state to determine what the `selectedProject` is, if there is one. I agree it may be a bit too slick, but it saves a bit of copy/paste.

joshbuchea: I may be missing something but it seems like the only change necessary aside from updating the hook to accept a `projectId` param rather than a `record` object would be to update the usage to pass in the value of the `projectId` property directly rather than the entire record: `useProjects(deliverable)` --> `useProjects(deliverable?.projectId)`.

nickgraz: You are correct, my previous statement about consumers needing to pull projects from state was incorrect. I like to think there was a good reason why I did it that way in the first place but I can't think of it now 😓 . Feel free to update it if you want, the `useAccessions` hook uses the same pattern if you're feeling ambitious.

PR #2196, opened 2024-02-23
SW-4917 Add call-to-action buttons/flows for the accelerator-console view of deliverable

karthikbtf:
- added update status function in service with redux support
- added call to action buttons to update the various statuses
- reject reason modal
- using a temporary confirmation for non-reject statuses as that confirmatoin design/screen is WIP

Terraware App 2024-02-23 10-06-07

Terraware App 2024-02-23 11-25-21

Terraware App 2024-02-23 11-22-19


src/types/Deliverables.ts
```diff
@@ -4,7 +4,13 @@ export type DeliverableTypeType = 'Document';
...
 export type DeliverableCategoryType = 'All' | 'Legal' | 'Financial' | 'GIS' | 'Forestry' | 'Stakeholder Engagement';
 // export type DeliverableCategories = ['All', 'Legal', 'Financial', 'GIS', 'Forestry', 'Stakeholder Engagement'];
-export type DeliverableStatusType = 'Not Submitted' | 'In Review' | 'Rejected' | 'Approved';
+export type DeliverableStatusType =
+  | 'Not Submitted'
+  | 'In Review'
+  | 'Rejected'
+  | 'Approved'
+  | 'Needs Translation'
+  | 'Not Needed';
```
karthikbtf, line 13: Added these two as they were in the designs/prd. We can update it later with BE types.

karthikbtf:
@nickgraz @joshbuchea take a quick look at this. I won't be in Monday and this would unblock the next tasks for Josh.

src/scenes/AcceleratorRouter/AcceleratorDeliverablesView.tsx
```diff
@@ -1,23 +1,112 @@
...
+          setStatus('Not Submitted');
+          break;
+        }
+        case 'needs_translation': {
+          setStatus('Needs Translation');
+          break;
+        }
+        case 'not_needed': {
+          setStatus('Not Needed');
+          break;
```
nickgraz, line 59: These values are passed into the update request, shouldn't they be the `optionItem.value` instead of the "readable" version? Maybe you've got it this way because there aren't backend types, disregard if that's the case

src/services/DeliverablesService.ts
```diff
@@ -17,38 +19,46 @@ const SEARCH_FIELDS_DELIVERABLES_BASE = ['documentCount', 'id', 'name', 'project
...
+    'Company Name\n' +
+    'Legal Form / Structure – (e.g. corporation, limited liability company, etc.)\n' +
+    'Date of Formation/Incorporation\n' +
+    'Company Number (if applicable)',
+};
+
 const getDeliverable = async (deliverableId: number): Promise<Response & DeliverableData> => {
   // TODO replace with axios
-  return {
+  return Promise.resolve({
```
nickgraz, line 52: Not that it matters since this is mock data, but if you put `async` in the function declaration it will automatically wrap your return in a promise, so you can just `return { ... }` 🚀

nickgraz (approved):
Looks good to me!

src/services/DeliverablesService.ts
karthikbtf: Yes, that was my understanding. I was trying some things to mock a delay (to test the spinner) - will remove this.

karthikbtf:
> Looks good to me!

Thanks @nickgraz , this is going to introduce some conflicts for your deliverables list PR... hopefully simple enough to deal with.

nickgraz:
> > Looks good to me!
> 
> Thanks @nickgraz , this is going to introduce some conflicts for your deliverables list PR... hopefully simple enough to deal with.

No problem I can sort them out

karthikbtf:
Added one missed change, snackbar toast messages to show success/warning/info.

nickgraz (approved):

PR #2203, opened 2024-02-24
SW-4903: Shared detail view status + feedback

joshbuchea:
This PR includes changes to support showing the deliverable status badge and the reason text if the status is ‘Rejected’.

### "Not Submitted" Status (Non-Accelerator)

localhost_3000_deliverables_1_organizationId=1

### "Not Submitted" Status (Accelerator)

localhost_3000_accelerator_deliverables_1_organizationId=1 (2)

### "Rejected" Status (Accelerator)

localhost_3000_accelerator_deliverables_1_organizationId=1 (1)


src/components/DeliverableView/DeliverableStatusBadge.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const { status } = props;
+  const theme = useTheme();
+
+  const getBadgePropsForStatus = useMemo(
+    () => (): BadgeProps => {
+      let badgeProps: BadgeProps;
+
+      switch (status) {
+        case 'Approved':
+          badgeProps = {
```
nickgraz, line 22: You can just return the object from each of these and save the let / assignment.

src/components/DeliverableView/DeliverableStatusBadge.tsx
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+      return badgeProps;
+    },
+    [status, theme]
+  );
+
+  return (
+    <div style={{ float: 'right', marginBottom: '0px', marginLeft: '16px' }}>
+      <Badge {...getBadgePropsForStatus()} />
```
nickgraz, line 85: Can you conditionally show this if `activeLocale` is not undefined? This way it will make sure the `strings.` values are localized.

src/components/DeliverableView/DeliverableStatusBadge.tsx
```diff
@@ -0,0 +1,90 @@
...
+type DeliverableStatusBadgeProps = {
+  status: DeliverableStatusType;
+};
+
+const DeliverableStatusBadge = (props: DeliverableStatusBadgeProps): JSX.Element => {
+  const { status } = props;
+  const theme = useTheme();
+
+  const getBadgePropsForStatus = useMemo(
+    () => (): BadgeProps => {
```
nickgraz, line 17: You can remove the extra function declaration, `useMemo` usually just returns a value. `useCallback` can be used if you need to memoize a function

nickgraz (changes requested):
Looks great! Just need to add the `activeLocale` stuff in and it should be good to go. A couple other suggestions that you can take or leave.

src/components/DeliverableView/DeliverableStatusBadge.tsx
joshbuchea: This was an intentional decision so there would only be one `return` statement in the function, but I updated it since this change reduces the lines of code and is more direct.

src/components/DeliverableView/DeliverableStatusBadge.tsx
joshbuchea: You got it.

src/components/DeliverableView/DeliverableStatusBadge.tsx
joshbuchea: Not sure what I was thinking here, thanks! 😅

src/components/DeliverableView/DeliverableStatusBadge.tsx
```diff
@@ -0,0 +1,83 @@
...
+type DeliverableStatusBadgeProps = {
+  status: DeliverableStatusType;
+};
+
+const DeliverableStatusBadge = (props: DeliverableStatusBadgeProps): JSX.Element => {
+  const { status } = props;
+  const { activeLocale } = useLocalization();
+  const theme = useTheme();
+
+  const getBadgePropsForStatus = useCallback((): BadgeProps => {
```
karthikbtf, line 18: this could be a useMemo instead of a useCallback

src/components/DeliverableView/Description.tsx
```diff
@@ -1,10 +1,18 @@
...
+import DeliverableStatusBadge from 'src/components/DeliverableView/DeliverableStatusBadge';
 
 const Description = (props: ViewProps): JSX.Element => {
+  const { deliverable, isAcceleratorConsole } = props;
+
   return (
     <Box display='flex' flexDirection='column'>
-      <div>Description coming soon {props.isAcceleratorConsole}!</div>
+      <div>
+        {deliverable.status !== 'Rejected' && !isAcceleratorConsole && (
```
karthikbtf, line 11: If you think it makes sense to combine the description and status bar components into one, we can do that. I created the placeholder components just as an initial draft. What you have here is fine too.

src/components/DeliverableView/StatusBar.tsx
```diff
@@ -1,12 +1,32 @@
...
+          borderRadius='8px'
+          marginBottom='16px'
+          padding='16px'
+        >
+          <DeliverableStatusBadge status={deliverable.status} />
+          <strong>{strings.REASON}</strong> {deliverable.reason}
+        </Box>
+      )}
+
+      <Box border={`1px solid ${theme.palette.TwClrBaseGray100}`} borderRadius='8px' marginBottom='16px' padding='16px'>
```
karthikbtf, line 24: this entire box should only be visible to accelerator console (internal comments are not shared with external users/participants).

karthikbtf (commented):
Looks good, a few comments.

src/components/DeliverableView/Description.tsx
joshbuchea: Yeah I didn't love the logic being split across two components here either. I'm up for combining the two, any name ideas for the new combined component?

src/components/DeliverableView/StatusBar.tsx
joshbuchea: Good catch, thanks! That makes sense that they wouldn't be shared externally, will update.

src/components/DeliverableView/Description.tsx
karthikbtf: `Metadata`?

karthikbtf (approved):
Thanks for the changes!

src/components/DeliverableView/Description.tsx
joshbuchea: `DescriptionWithStatus`?

karthikbtf: Sure your suggestion is more descriptive and clear 👍

joshbuchea: I definitely prefer `Metadata`, thanks @karthikbtf 🙌

karthikbtf (approved):

nickgraz (approved):
Nice work

PR #2219, opened 2024-02-27
Rename the AcceleratorCohort* views to Cohort*

nickgraz:
No need to disambiguate since these views will only exist within the AcceleratorRouter. No functional changes here.

### DIFF
--- a/src/scenes/AcceleratorRouter/CohortEditView.tsx
+++ b/src/scenes/AcceleratorRouter/CohortEditView.tsx
@@ -11,7 +11,7 @@ import { useAppDispatch, useAppSelector } from 'src/redux/store';
 import AcceleratorMain from 'src/scenes/AcceleratorRouter/AcceleratorMain';
 import CohortForm from 'src/scenes/AcceleratorRouter/CohortForm';
 
-export default function AcceleratorCohortEditView(): JSX.Element {
+export default function CohortEditView(): JSX.Element {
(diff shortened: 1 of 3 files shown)


karthikbtf (approved):
Nice. Should we do the same for other AcceleratorRouter/Accelerator* components?

nickgraz:
Yea... I was thinking about that. Personally I think we should, I think it is redundant. It is nice having it baked in for disambiguation but if your IDE supports fuzzy file opening then it is a moot point (IE typing in acceleratorcohortnew should still bring you to the correct view regardless if the filename is AcceleratorCohortNewView or CohortNewView)

karthikbtf:
> Yea... I was thinking about that. Personally I think we should, I think it is redundant. It is nice having it baked in for disambiguation but if your IDE supports fuzzy file opening then it is a moot point (IE typing in acceleratorcohortnew should still bring you to the correct view regardless if the filename is AcceleratorCohortNewView or CohortNewView)

It will look cleaner with the rename. The parent directory should be enough to disambiguate. I can do this cleanup (it was my suggestion after all) in a separate PR without increasing scope here.

nickgraz:
### Merge activity

* **Feb 27, 7:18 PM EST**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2219).

PR #2221, opened 2024-02-28
SW-4933 Merge the redundant Page components 

karthikbtf:
- removed new Page component
- moved OG Page from BreadCrumbs into a standalone component
- refactored deliverables view to work with the OG Page


nickgraz (approved):
10/10 thanks for doing this!

PR #2224, opened 2024-02-28
SW-4930: Show page message for participant view when deliverable is rejected

joshbuchea:
This PR includes changes to support showing a page message in the deliverable view when a deliverable is rejected.

**NOTE:** The ticket only required adding the page message for the participant view, but the page message is also shown in the accelerator view (just with an added "Edit Reason" button). I went ahead and added the page message to both versions of the view and updated the page message component to show the added button in the accelerator view.

### Participant View

localhost_3000_deliverables_1_organizationId=1

### Accelerator View

localhost_3000_accelerator_deliverables_1_organizationId=1


src/scenes/DeliverablesRouter/DeliverablePageMessage.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { Box, useTheme } from '@mui/material';
+import { Button, Message } from '@terraware/web-components';
+import { ViewProps } from 'src/components/DeliverableView/types';
+import { useLocalization } from 'src/providers/hooks';
+import strings from 'src/strings';
+
+const DeliverableView = (props: ViewProps): JSX.Element => {
```
karthikbtf, line 8: It might be ok to have 2 separate implementations, one for the participant view and one for the accelerator console view. Or move this under src/components. Two separate implementations might be ok.
I have posed a question on what to do for edit reason. We have a reject reason flow, maybe we can reuse that in the accelerator console view's implementation (change the title of the modal to 'Edit Reason' and button name to 'Save' or something like that). Thoughts?

karthikbtf (commented):
Thanks for this work Josh, looks good, a few suggestions here and there.
Also looks like we need @chudo76 to guide us on what to do with the 'Edit Reason' button and the redundant 'Reason' boxes.

src/scenes/DeliverablesRouter/DeliverablePageMessage.tsx
joshbuchea: I could see the case for either approach, but two implementations sounds reasonable to me. I'm not sure about the reject reason flow, but only because I'm not familiar with it and how much we'd get/save by using it.

karthikbtf: Thanks, Josh. Yes two implementations might be cleaner for now.
Reusing the reject/reason flow but changing the modal title and button text will get you an 'edit reason' feature. It will show a text area to accept new 'reason' (soon to be 'feedback') and updates the deliverable.
That's totally fine if we do this as a separate task, I can take it since I added the original flow. Also, waiting on Chudo to respond in figma on how he sees the edit reason flow.

joshbuchea: Are we still waiting on other feedback related to this ticket?

karthikbtf: @joshbuchea looks like the accelerator console version of the page message is missing now.

joshbuchea: @karthikbtf please clarify exactly which changes you expect to see so I know I'm making the correct changes 🙏

karthikbtf: No worries, my mistake for adding confusion. I was referring to two separate views instead of a shared view. The separate views were being referred to as implementations. We can do this in a separate PR that's totally fine.

karthikbtf: @joshbuchea we could capture the other effort as a separate JIRA.

src/scenes/DeliverablesRouter/PageMessage.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
karthikbtf, line 1: Thoughts on a more descriptive filename? RejectedDeliverableMessage?

karthikbtf (approved):
One suggestion on renaming the PageMessage file. Feel free to merge anytime.

src/scenes/DeliverablesRouter/PageMessage.tsx
joshbuchea: I do agree that `PageMessage` feels a bit too generic and in my mind could be confused to be a more general purpose component. `RejectedDeliverableMessage` seems reasonable, I'll rename it to that then merge the PR.

src/scenes/DeliverablesRouter/DeliverablePageMessage.tsx
joshbuchea: I'll create a separate ticket.

PR #2226, opened 2024-02-28
Do not show the organization dropdown if we are in the accelerator console

nickgraz:
Do not show the organization dropdown if we are in the accelerator console

Do not block app load for organizations bootstrap and remove org query param if we are within the accelerator console

### DIFF
--- a/src/AppBootstrap.tsx
+++ b/src/AppBootstrap.tsx
@@ -9,6 +9,8 @@ import {
   OrganizationProvider,
   UserProvider,
 } from 'src/providers';
+import { useRouteMatch } from 'react-router-dom';
+import { APP_PATHS } from './constants';
 
 const useStyles = makeStyles((theme: Theme) => ({
   spinner: {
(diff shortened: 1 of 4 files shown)


nickgraz:
* **#2226** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2226?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2226?utm_source=stack-comment).

nickgraz:
This kinda goes against our "encapsulate the console away from terraware" idea, but extracting out the bootstrap and top bar seem a bit out of scope for now. If we run into a lot more changes like this we should reconsider, but there are none that I can think of on the horizon.

src/AppBootstrap.tsx
```diff
@@ -38,10 +40,14 @@ function BlockingBootstrap({ children }: BlockingBootstrapProps): JSX.Element {
   const { bootstrapped: userBootstrapped } = useUser();
   const { bootstrapped: organizationBootstrapped } = useOrganization();
   const { bootstrapped: localizationBootstrapped } = useLocalization();
+  const isAcceleratorRoute = useRouteMatch(APP_PATHS.ACCELERATOR);
```
karthikbtf, line 43: Thoughts on exposing this through a hook? might be good to remove the route match semantics dependency in multiple places.

karthikbtf (approved):
Looks good, and totally understandable to have these semantics in the top level. One suggestion.

src/AppBootstrap.tsx
nickgraz: Good call, it seems a bit small for its own hook but this is sprinkled in a lot of places and would be nice to have a location to put more things like this if they come up. Will do that in a follow up PR.

karthikbtf: Sounds good 👍

nickgraz:
### Merge activity

* **Feb 28, 1:14 PM EST**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2226).

PR #2229, opened 2024-02-28
Replace isAcceleratorConsole flags with useAcceleratorConsole hook

karthikbtf:
- also fixed the hook to return a boolean, otherwise it was returning the matched data whose reference changes on render, triggering re-renders
- removed isAcceleratorConsole from props and plumbing

### DIFF
--- a/src/hooks/useAcceleratorConsole.ts
+++ b/src/hooks/useAcceleratorConsole.ts
@@ -4,7 +4,7 @@ import isEnabled from 'src/features';
 
 const useAcceleratorConsole = () => {
   const featureFlagAccelerator = isEnabled('Accelerator');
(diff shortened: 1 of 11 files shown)


src/hooks/useAcceleratorConsole.ts
```diff
@@ -4,7 +4,7 @@ import isEnabled from 'src/features';
 
 const useAcceleratorConsole = () => {
   const featureFlagAccelerator = isEnabled('Accelerator');
-  const isAcceleratorRoute = useRouteMatch(APP_PATHS.ACCELERATOR);
+  const isAcceleratorRoute = !!useRouteMatch(APP_PATHS.ACCELERATOR);
```
nickgraz, line 7: 🙏

nickgraz (approved):
Excellent!

PR #2232, opened 2024-02-28
SW-4935: Update hardcoded styles to use accent tokens

joshbuchea:
This PR includes a quick change to replace hardcoded uses of a color with a token from the theme.


nickgraz (approved):
🚀

PR #2236, opened 2024-02-29
SW-4938: Rename reason to feedback

joshbuchea:
This PR includes changes to rename "reason" to "feedback".

**NOTE:** Current changes include updating all instances of "reason" in the UI to "feedback" and all related code. There are a couple of potential changes I could also see making but wanted to confirm first:

- Rename “Reject Reason” to “Reject Feedback”?
- Change `Deliverable.reason` to `Deliverable.feedback` in the type for `Deliverable` and related mocks? (I’m not sure if the type for `Deliverable` is already defined in the BE API, and if so, if we want to update or leave as-is.)

### DIFF
--- a/src/components/DeliverableView/Metadata.tsx
+++ b/src/components/DeliverableView/Metadata.tsx
@@ -21,7 +21,7 @@ const Metadata = (props: ViewProps): JSX.Element => {
           padding='16px'
         >
           <DeliverableStatusBadge status={deliverable.status} />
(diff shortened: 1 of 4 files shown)


karthikbtf (approved):
Looks good @joshbuchea 
Yes, no need to change the types, it was mostly for the visual content.
Nick is updating the types from the BE, which will capture the `feedback` property.

PR #2235, opened 2024-02-29
Rewire deliverable APIs to the backend 

nickgraz:
- Use latest backend types
- Update these APIs to use the backend instead of mock data:
  - List deliverables
  - Get deliverable
  - Update deliverable

TODO:
[x] Rename 'reason' to 'feedback' in a few other places
[x] Hook up the `update` call
[x] Figure out where the `templateUrl` is supposed to come from 
[x] Search and sort on deliverables on the frontend (in #2242)

PR #2242, opened 2024-02-29
Create searchAndSort function which acts as a local Search API replacement that can use the same interfaces

nickgraz:
I did my best replicating the Search API functionality locally without sacrificing the payloads that are normally sent. This means that we don't need to change the way that the `SearchFiltersWrapper` works, even if we aren't actually using the Search API. 

I don't think this is a bulletproof implementation, and things like `Range` and `ExactOrFuzzy` aren't implemented (although the latter sounds like it would be trivial). Hopefully the test coverage is adequate to cover the most common use cases.

### DIFF
--- a/src/utils/searchAndSort.ts
+++ b/src/utils/searchAndSort.ts
@@ -72,6 +72,33 @@ const searchConditionMet = <T extends Record<string, unknown>>(result: T, condit
   return false;
 };
 
+const sortResults = <T extends Record<string, unknown>>(
+  results: T[],
+  locale: string | null,
+  sortOrder: SearchSortOrder,
+  numberFields?: string[]
+) => {
+  const field = sortOrder.field;
+  // Defaults to ascending if not provided
+  const isDescending = sortOrder.direction === 'Descending';
+  const isNumberField = (numberFields || []).includes(field);
+
+  if (isNumberField) {
+    results = results.sort((a, b) => Number(a[field]) - Number(b[field]));
+  } else {
+    results = results.sort((a, b) =>
+      // TODO this might cause issues if the field is undefined on the result
+      `${a[field]}`.localeCompare(`${b[field]}`, locale || undefined)
+    );
+  }
+
+  if (isDescending) {
+    results.reverse();
+  }
+
+  return results;
+};
+
 /**
  * In-memory search (filter) and sort on a result list using the Search API search and sortOrder interfaces
  * The search currently only supports `Exact` and 'Fuzzy' type searches, 'Range' and 'ExactOrFuzzy' are not supported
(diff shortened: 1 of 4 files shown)


nickgraz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2242?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#2242** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2242?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2235** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2235?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2242?utm_source=stack-comment).

src/utils/searchAndSort.ts
```diff
@@ -0,0 +1,122 @@
...
+  return false;
+};
+
+/**
+ * In-memory search (filter) and sort on a result list using the Search API search and sortOrder interfaces
+ * The search currently only supports `Exact` and 'Fuzzy' type searches, 'Range' and 'ExactOrFuzzy' are not supported
+ * If the result type contains number fields, those must be supplied in the `sortOrderConfig` if you wish to sort on them
+ * @param results         The list of results you want to search and sort
+ * @param search          (optional) The SearchNodePayload which contains filter conditions to apply to the results
+ * @param sortOrderConfig (optional) The sort order configuration contains:
```
nickgraz, line 81: I tried to use the exact same structure as the Search API, but since it only deals with strings, and sometimes (like in the case of deliverables), the entity we want to sort will have numbers. This config allows the consumer to "tell" the function which fields are numbers so they can be sorted appropriately. 

I originally though about just checking to see if the field value was a number, but this would get complicated because the Search API search node payload accepts strings as values, even if they are numbers. So there might be a mismatch between the (hopefully untouched) search node payload structure and the data type on the entities that we are sorting. Another benefit of this approach is that we can additionally tell the function which fields _should_ be numbers, even if they are strings. This helps with sorting `SearchResponseElement`s which always deal in strings, even if the underlying field is supposed to be a number.

karthikbtf: This is cool. One suggestion is to extract out the sort functionality as a separate function and call it here. We can update the planting sites code to use the sort function.. just so we don't have similarish code scattered across. Does similar handling of number semantics on strings, etc.

src/utils/searchAndSort.ts
```diff
@@ -0,0 +1,122 @@
...
+        trigrams.push(word.substring(i, position).toLowerCase());
+        position += 1;
+      }
+    }
+  }
+
+  return trigrams;
+};
+
+const searchConditionMet = <T extends Record<string, unknown>>(result: T, condition: SearchNodePayload): boolean => {
```
karthikbtf, line 43: This is very cool 😎 ... a tad bit of a reservation that search should be done on the BE side maybe. Have a consistent search behavior across pages. I understand for MVP this is our decision,  the endpoints are not search table implementations.

karthikbtf (approved):
Nice 🎉

src/utils/searchAndSort.ts
nickgraz: I agree 100%. That's why I tried to make it as close to the original API as possible, in the hopes that we can just swap it over easily when that day comes.

karthikbtf: Agree on that. This is very cool and should be pretty consistent 🙌

karthikbtf (approved):
This looks great and a cool utility to reuse, thanks for the additional changes 🙏

nickgraz:
### Merge activity

* **Feb 29, 6:45 PM EST**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2242).

PR #2243, opened 2024-02-29
Fix Deliverable page message layout issue

joshbuchea:
This PR includes changes to fix a layout issue that resulted in the Deliverable Rejected page message showing beside the rest of the page content on browsers with higher/wider resolution.

### BEFORE

localhost_3000_accelerator_deliverables_1

### AFTER

localhost_3000_accelerator_deliverables_1 (1)

### DIFF
--- a/src/scenes/AcceleratorRouter/DeliverableView.tsx
+++ b/src/scenes/AcceleratorRouter/DeliverableView.tsx
@@ -42,7 +42,7 @@ const DeliverableView = (props: Props): JSX.Element => {
   return (
     <Page title={<TitleBar {...viewProps} />} rightComponent={props.callToAction} crumbs={crumbs}>
       {props.isBusy && <BusySpinner />}
-      <Box display='flex' flexDirection='column' sx={{ flexGrow: 1 }}>
+      <Box display='flex' flexDirection='column' flexGrow={1}>
         <RejectedDeliverableMessage {...viewProps} />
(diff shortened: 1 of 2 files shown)


src/scenes/AcceleratorRouter/DeliverableView.tsx
```diff
@@ -40,15 +41,15 @@ const DeliverableView = (props: Props): JSX.Element => {
 
   return (
     <Page title={<TitleBar {...viewProps} />} rightComponent={props.callToAction} crumbs={crumbs}>
-      <>
-        {props.isBusy && <BusySpinner />}
+      {props.isBusy && <BusySpinner />}
+      <Box display='flex' flexDirection='column' sx={{ flexGrow: 1 }}>
```
joshbuchea, line 45: - It looked like `BusySpinner` was defined outside of grid components/containers in other Views
- `flexGrow` was necessary here otherwise the page content wasn't stretching full-width (perhaps there's a preferred way to achieve this instead?)

karthikbtf: flexGrow is good 👍 
BusySpinner is absolutely positioned, good to be consistent anyways, thanks for that change.

karthikbtf: you can do `flexGrow={1}` instead of `sx` if you want

karthikbtf (approved):

PR #2246, opened 2024-03-01
SW-4949: Implement Edit Feedback flow

joshbuchea:
This PR includes changes to implement the "edit feedback" flow in the `DeliverableView` within the `AcceleratorRouter`.


karthikbtf:
> @karthikbtf I know in the ticket you mentioned:
> 
> > We can take this one step further and handle the backend update within the Reject Feedback dialog, instead of relying on the client. This way ‘Reject Deliverable' and ‘Edit Feedback’ actions can reuse the same dialog efficiently.
> 
> but it seems like this is pretty efficient as-is. It seems like moving where the update is handled is just relocating code, but I admit I may be missing something.

@joshbuchea I saw the changes.. passing the show dialog function into the message component. This is simpler/cleaner, ignore my suggestions.

src/scenes/AcceleratorRouter/DeliverableView.tsx
```diff
@@ -18,6 +18,7 @@ import RejectedDeliverableMessage from 'src/scenes/AcceleratorRouter/RejectedDel
 
 export type Props = EditProps & {
   isBusy?: boolean;
+  showRejectDialog?: () => void;
```
karthikbtf, line 21: this can be non-optional

src/scenes/AcceleratorRouter/RejectedDeliverableMessage.tsx
```diff
@@ -5,12 +5,16 @@ import { ViewProps } from 'src/components/DeliverableView/types';
 import { useLocalization } from 'src/providers/hooks';
 import strings from 'src/strings';
 
-const RejectedDeliverableMessage = ({ deliverable }: ViewProps): JSX.Element => {
+type Props = ViewProps & {
+  showRejectDialog?: () => void;
```
karthikbtf, line 9: make this required

karthikbtf (approved):
some suggestions but feel free to merge

src/scenes/AcceleratorRouter/RejectedDeliverableMessage.tsx
joshbuchea: Good call 👍

PR #2248, opened 2024-03-01
SW-4954: Show status call to action buttons all the time, grey out button that matches current status

joshbuchea:
This PR includes changes to support always showing the status buttons for deliverables, while disabling the button corresponding to the current status.


karthikbtf (approved):
Nice!

PR #2245, opened 2024-03-01
SW-4928 Deliverable file upload dialog that accepts description, without API integration

karthikbtf:
- dialog pops up after user selects files
- user enters descriptions and hits submit
- currently mocks a delay upon submit
- TODO in follow up: implement service/redux to upload files and dispatch requests here (Jira SW-4956)

Terraware App 2024-02-29 16-30-16

Terraware App 2024-02-29 15-49-37

screencast 2024-02-29 16-27-46

PR #2252, opened 2024-03-01
SW-4945: FE workflow to reset submission status with new document upload

joshbuchea:
This PR includes changes to reset the submission status when a new document is uploaded.

NOTE: The approach I took seemed like the most straight forward, but I'm happy to adjust. Perhaps we'd rather handle the update in a function in the `DeliverableViewWrapper` like `setStatus` and `rejectDeliverable` that we pass into the `DocumentsUploader`?


nickgraz (commented):
I just merged https://github.com/terraware/terraware-web/pull/2245 in. Can you rebase your changes on top of that and "do the thing" when the dialog is submitted? I think doing it when the files are selected might be a bit premature because they still have the ability to cancel the upload after files have been selected.

nickgraz:
Somewhere in here https://github.com/terraware/terraware-web/pull/2245/files#diff-b5eebc38459917be8772eb042b0736956958a496290577aeaed234a561de7f83R45-R56

karthikbtf:
> This PR includes changes to reset the submission status when a new document is uploaded.
> 
> NOTE: The approach I took seemed like the most straight forward, but I'm happy to adjust. Perhaps we'd rather handle the update in a function in the `DeliverableViewWrapper` like `setStatus` and `rejectDeliverable` that we pass into the `DocumentsUploader`?

@joshbuchea sorryI didn't include more details in the ticket, just did that. I have some open [questions](https://www.figma.com/file/OlTouFTKgxUPX8ejhY6mMv?node-id=2045:87020&mode=design#722626317) for @chudo76

nickgraz (approved):
🔥

karthikbtf:
> 🔥

Added more details in the ticket. The participant won't have permissions to update the status, BE will automatically reset status to in-review when a document is uploaded. We need to show a modal informing user that status will be reset, a bunch of open questions in figma.

nickgraz:
I see, we can use the new `isAllowed` mechanism on the user provider to gate this. There isn't a permission for "update status" specifically but it can be added easily.

nickgraz:
Also I suppose the status should only get updated if it isn't "In review" already.

nickgraz:
> > > Also I suppose the status should only get updated if it isn't "In review" already.
> > 
> > 
> > > Also I suppose the status should only get updated if it isn't "In review" already.
> > 
> > 
> > See lines 108/115 in SubmissionService here https://github.com/terraware/terraware-server/pull/1712/files, BE takes care of this already. I made a mistake not clarifying what needs to be done for this ticket. Scope of this effort is to show a modal that status will be reset. Unless status is already not-submitted (implied reset) or in-review (no-op). We don't need to worry about the actual status update or permissions.
> 
> My intention was for one of us to look at the designs and follow up on details for this effort.

Sounds good to me. I think ideally we match the permissions in the front end so we aren't doing things that we expect to fail (or 403) in the back end.

karthikbtf:
> > > > Also I suppose the status should only get updated if it isn't "In review" already.
> > > 
> > > 
> > > > Also I suppose the status should only get updated if it isn't "In review" already.
> > > 
> > > 
> > > See lines 108/115 in SubmissionService here https://github.com/terraware/terraware-server/pull/1712/files, BE takes care of this already. I made a mistake not clarifying what needs to be done for this ticket. Scope of this effort is to show a modal that status will be reset. Unless status is already not-submitted (implied reset) or in-review (no-op). We don't need to worry about the actual status update or permissions.
> > 
> > 
> > My intention was for one of us to look at the designs and follow up on details for this effort.
> 
> Sounds good to me. I think ideally we match the permissions in the front end so we aren't doing things that we expect to fail (or 403) in the back end.

Yes we should be doing that in our codebase and have been for the most part. 
I had created a ticket earlier to gate accelerator flows I think you may have already handled permissions here. 
Participants and all global roles (except read-only) have permissions to upload files.
The scope of this effort is to show a modal assuming user already has permissions to upload a file.

joshbuchea:
@karthikbtf no worries at all, I can rework this PR once the open questions have been answered.

karthikbtf:
@joshbuchea the merge is flagging a bunch of already existing code as new changes. May be cleaner to start from a new branch if you have issues with the merge.

nickgraz:
It doesn't look like the status is being updated, are we assuming this will be done in the backend?

nickgraz:
Scratch that, I found the figma thread that indicates this.

src/components/DeliverableView/DocumentsUploader.tsx
```diff
@@ -21,13 +25,44 @@ const DocumentsUploader = ({ deliverable }: ViewProps): JSX.Element => {
...
+
+    // participant view
+    if (!isAcceleratorRoute && !['Not Submitted', 'In Review', 'Needs Translation'].includes(deliverable.status)) {
+      return true;
+    }
+
+    // accelerator view
+    if (isAcceleratorRoute && !['Not Submitted', 'In Review'].includes(deliverable.status)) {
+      return true;
+    }
```
nickgraz, line 43: I think you can ignore the `isAcceleratorRoute`. According to Figma the condition is always checking for these 3 statuses. Can just `return ['Not Submitted', 'In Review', 'Needs Translation'].includes(deliverable.status);`

joshbuchea: @nickgraz "Needs Translation" should be for the participant view only according to Karthik's comment in the ticket for SW-4945.

nickgraz: That is for display only, the underlying status of the deliverable will not change. In other words, a participant can still interact with a deliverable that has a status of "needs translation", they just won't know it because to them it will _appear_ "in review". Ultimately, this component should be view-agnostic anyways.

karthikbtf: For participants yes we hide the 'needs translation' semantic, it's still technically 'in review'.
For internal users, we show the 'needs translation' semantic, we need to let them know the status will change from 'needs translation' back to 'in review' on a new document upload.
Status changes are handled in the backend.
I can't find the figma thread that discussed this change, the original thread does not mention the change.

If we don't want the isAccelerator dependency here, we could pass the list of statuses to ignore as a prop.

karthikbtf (approved):

src/components/DeliverableView/DocumentsUploader.tsx
nickgraz: But the underlying status will still be "needs translation", even if we don't show it to participants, so the logic is the same, right? In other words, if a participant uploads a document to a deliverable that is "needs translation" (to them it looks like "in review"), the status is still reset and therefore we should still show them a dialog that the status is going to be reset. 

I suppose it will appear as though nothing has changed (on the participant side), since the "reset" status is also "in review". Maybe that is why you are saying it should not be shown? Since it doesn't explicitly call out that it will put it back into "in review" it still seems like useful information for the user that their upload is going to modify the deliverable's status, but if this has already been sorted out in Figma or something then I'm OK with having it do two different things. But I would prefer that the statuses are passed in and the component becomes view agnostic.

karthikbtf: Yes, that is correct - we don't need to say it will be reset because it already appears as 'In Review' to the participant (even if it is 'Translation Needed' in actual data). The BE will reset it from 'Translation Needed' to 'In Review' behind the scenes. There is some resetting going on, we just won't flag it to the participant user.
Passing these statuses into the component sounds good to me 👍

src/components/DeliverableView/DocumentsUploader.tsx
```diff
@@ -3,18 +3,25 @@ import { useCallback, useMemo, useState } from 'react';
...
 
+import StatusChangeConfirmationDialog from 'src/components/DeliverableView/StatusChangeConfirmationDialog';
 import { useLocalization } from 'src/providers';
 import strings from 'src/strings';
 
 import FileUploadDialog from './FileUploadDialog';
 import { ViewProps } from './types';
 
 type DocumentUploaderProps = ViewProps & {
+  deliverableStatusesToIgnore: string[];
```
nickgraz, line 14: Can you make this `DeliverableStatusType[]` so we have type safety?

nickgraz (approved):
One small change, otherwise looks good!

src/components/DeliverableView/DocumentsUploader.tsx
joshbuchea: I actually meant to change that but forgot, you got it 👍

joshbuchea:
more feedback

PR #2253, opened 2024-03-01
SW-3877: Added prettier plugin for sorting imports

tommylau523:
The PR made changes to almost every file as part of reformatting. Please take a look at `.prettierrc`, `.prettierignore`, and `package.json`.

The imports are arranged via the following order:

- react packages
- third-party packages
- third-party packages initialed with \@
- paths initialized with src/
- relative paths

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -4,5 +4,9 @@
   "singleQuote": true,
   "jsxSingleQuote": true,
   "tabWidth": 2,
-  "printWidth": 120
+  "printWidth": 120,
+  "plugins": ["@trivago/prettier-plugin-sort-imports"],
+  "importOrder": ["^react(.*)", "<THIRD_PARTY_MODULES>", "@/(.*)", "^src/(.*)$", "^[./]"],
+  "importOrderSeparation": true,
+  "importOrderSortSpecifiers": true
 }
--- a/.prettierignore
+++ b/.prettierignore
@@ -30,3 +30,6 @@ yarn.lock
 
 # npm license report
 license-report.html
+
+# Test
+**/*.test.ts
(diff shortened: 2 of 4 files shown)


.prettierrc
```diff
@@ -4,5 +4,9 @@
   "singleQuote": true,
   "jsxSingleQuote": true,
   "tabWidth": 2,
-  "printWidth": 120
+  "printWidth": 120,
+  "plugins": ["@trivago/prettier-plugin-sort-imports"],
+  "importOrder": ["^react(.*)", "<THIRD_PARTY_MODULES>", "@/(.*)", "^src/(.*)$", "^[./]"],
```
nickgraz, line 9: This is great! I've been doing this manually for way too long

tommylau523:
I would also like some advice on the preferred conventions on making sweeping formatting changes, and how to split them up into different commits of ease of review. @nickgraz

nickgraz:
I haven't been here long enough to see something like this roll through, so I'm not sure if we have any conventions around these kinds of sweeping changes. I think since this one is only dealing with imports, I think it is alright to have one giant PR.

If you really want to split it up, doing each subdirectory under `src` in its own PR would be sufficient. `src/services`, `src/components`, etc...

tommylau523:
I needed to roll back the import changes to avoid merge conflicts. I plan to merge this first, and create another PR for merging changes to import statements wholesale. 

Future changes and commit can run `yarn format` to avoid merge conflicts

nickgraz:
Unfortunately, we have prettier check in GH actions, so I think you have to install it and do all the formatting in the same PR.

alexTerraformation:
> Unfortunately, we have prettier check in GH actions, so I think you have to install it and do all the formatting in the same PR.

This feels like a reasonable protection bypass candidate

.prettierrc
joshbuchea: +1 this is great, thank you!

joshbuchea (approved):

PR #2257, opened 2024-03-04
SW-4956 Integrate file upload with BE

karthikbtf:
- added types/redux/service
- updated component
- also added roles/guard for deliverable data

### DIFF
--- a/src/services/DeliverablesService.ts
+++ b/src/services/DeliverablesService.ts
@@ -5,9 +5,12 @@ import {
   DeliverableData,
   DeliverablesData,
   ListDeliverablesResponsePayload,
+  UploadDeliverableDocumentRequest,
 } from 'src/types/Deliverables';
 import { SearchNodePayload, SearchSortOrder } from 'src/types/Search';
 import { SearchOrderConfig, searchAndSort } from 'src/utils/searchAndSort';
+
+import { getPromisesResponse } from './utils';
 
 /**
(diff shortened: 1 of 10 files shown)


karthikbtf:
This works not but needs BE PR https://github.com/terraware/terraware-server/pull/1716

karthikbtf:
> This works now but needs BE PR [terraware/terraware-server#1716](https://github.com/terraware/terraware-server/pull/1716)

typo, works "now"

src/services/DeliverablesService.ts
```diff
@@ -84,10 +89,31 @@ const list = async (
...
+    };
+    const entity = { description, file };
+
+    return httpDocumentUpload.post({ urlReplacements, entity, headers });
+  });
+
+  const results = await getPromisesResponse<Response>(promises);
+  return results.every((result) => result !== null && result.requestSucceeded === true);
+};
+
```
joshbuchea, line 111: How do we handle a partial failure? (one or more document uploads completed successfully, while one or more documents uploads failed)

joshbuchea (approved):

src/services/DeliverablesService.ts
karthikbtf: We don't have this pattern today, something we can add later. For now if any one fails, we show an error but we refresh the documents list so user knows which ones succeeded.

PR #2259, opened 2024-03-04
SW-4960 Lazy load routes in terraware web instead of prefetching before we know the user principal context

karthikbtf:
- this code will auto code-split during build time as well
- use spinner fallback while chunks are loading/being-fetched
- try it in vercel


nickgraz (approved):
🔥

PR #2260, opened 2024-03-04
SW-4959 Handle race condition in accessions search when page is loaded with filters in the url

karthikbtf:
- when we first load the page and it already has filters in the url (such as when visiting from the dashboard links), there are 
a bunch of requests in-flight in succession and existing mechanism to handle races is faulty
- applying another mechanism which we use across the board, not the most elegant but it works

### DIFF
--- a/src/components/seeds/database/index.tsx
+++ b/src/components/seeds/database/index.tsx
@@ -36,6 +36,7 @@ import { SearchCriteria, SearchNodePayload, SearchResponseElementWithId, SearchS
 import { useSessionFilters } from 'src/utils/filterHooks/useSessionFilters';
 import { getAllSeedBanks } from 'src/utils/organization';
 import { isAdmin } from 'src/utils/organization';
+import { getRequestId, setRequestId } from 'src/utils/requestsId';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
 import useQuery from 'src/utils/useQuery';
 import useSnackbar, { SNACKBAR_PAGE_CLOSE_KEY } from 'src/utils/useSnackbar';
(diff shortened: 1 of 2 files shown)


src/components/seeds/database/index.tsx
```diff
@@ -423,86 +424,80 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
-        let columnsNamesToSearch: string[];
-        if (searchColumns.includes('active')) {
-          columnsNamesToSearch = [...searchColumns, 'id'];
-        } else {
-          columnsNamesToSearch = [...searchColumns, 'active', 'id'];
-        }
-
-        return columnsNamesToSearch;
-      };
+  const initAccessions = useCallback(() => {
```
karthikbtf, line 427: This looks like a lot of changes but it's only indentation (removed the useCallback argument).
Look for get/setRequestId, those are the changes.

nickgraz (approved):
This looks good as-is, but I wonder if splitting `populateFieldOptions`, `populateSearchResults`, and `populateAvailableFieldOptions` up into their own `useEffect` statements might help with the race condition and reduce complexity.

karthikbtf:
> This looks good as-is, but I wonder if splitting `populateFieldOptions`, `populateSearchResults`, and `populateAvailableFieldOptions` up into their own `useEffect` statements might help with the race condition and reduce complexity.

Every time the filters change, we need to call these 3 populate functions. The race condition is really on the entire group of searches useEffect. On page load it is called once, once more if the page was loaded with filters in the url. The first one to win sets the results.
We could debounce the entire group of searches but wasn't too comfortable with that.
Actually let me refactor this fix to consider all 3 functions winning at the same time.

PR #2261, opened 2024-03-04
SW-4961 Global roles redux and service layer

nickgraz:
- Add new types from global roles API from BE
- Add redux for the global roles users search and user update
- Stub out the request dispatch / select within the accelerator `PeopleView`

### DIFF
--- a/src/redux/features/globalRoles/globalRolesSlice.ts
+++ b/src/redux/features/globalRoles/globalRolesSlice.ts
@@ -1,0 +1,38 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+import { StatusT, buildReducers } from 'src/redux/features/asyncUtils';
+import { requestListGlobalRolesUsers, requestUpdateGlobalRolesUser } from './globalRolesAsyncThunks';
+import { GlobalRolesUsersData } from 'src/types/GlobalRoles';
+
+/**
+ * Users with global roles list
+ */
+const initialStateGlobalRolesUsersList: { [key: string]: StatusT<GlobalRolesUsersData> } = {};
+
+export const globalRolesUsersListSlice = createSlice({
+  name: 'globalRolesUsersListSlice',
+  initialState: initialStateGlobalRolesUsersList,
+  reducers: {},
+  extraReducers: (builder) => {
+    buildReducers(requestListGlobalRolesUsers)(builder);
+  },
+});
+
+export const globalRolesUsersListReducer = globalRolesUsersListSlice.reducer;
+
+/**
+ * Simple OK/response for requests updating a user's global roles, keeps
+ * state of user id that was edited.
+ */
+const initialStateGlobalRolesUserUpdate: { [key: number | string]: StatusT<number> } = {};
+
+export const globalRolesUserUpdateSlice = createSlice({
+  name: 'globalRolesUserUpdateSlice',
+  initialState: initialStateGlobalRolesUserUpdate,
+  reducers: {},
+  extraReducers: (builder) => {
+    buildReducers(requestUpdateGlobalRolesUser)(builder);
+  },
+});
+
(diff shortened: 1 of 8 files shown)


nickgraz:
* **#2261** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/scenes/AcceleratorRouter/PeopleView.tsx
```diff
@@ -1,8 +1,44 @@
...
+    }
+  }, [listRequest, snackbar]);
+
+  useEffect(() => {
+    const request = dispatch(requestListGlobalRolesUsers({ locale: activeLocale }));
+    setRequestId(request.requestId);
+  }, [activeLocale, dispatch]);
+
+  // tslint:disable:no-console
+  console.log('globalRoleUsers', globalRoleUsers);
```
karthikbtf, line 40: remove now/later?

src/redux/features/globalRoles/globalRolesSlice.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Users with global roles list
+ */
+const initialStateGlobalRolesUsersList: { [key: string]: StatusT<GlobalRolesUsersData> } = {};
+
+export const globalRolesUsersListSlice = createSlice({
+  name: 'globalRolesUsersListSlice',
+  initialState: initialStateGlobalRolesUsersList,
+  reducers: {},
+  extraReducers: (builder) => {
+    buildReducers(requestListGlobalRolesUsers)(builder);
```
karthikbtf, line 18: We could probably use a pre-defined key for this, 'list' or something like that. Might need a small tweak in the buildReducers function to accept an optional callback to get the key.

src/services/GlobalRolesService.ts
```diff
@@ -0,0 +1,67 @@
...
+      numberFields: ['id'],
+    };
+  }
+
+  return httpGlobalRolesUsers.get<GlobalRoleUsersListResponsePayload, GlobalRolesUsersData>(
+    {
+      params: {} as Params,
+    },
+    (data) => ({
+      users: searchAndSort(data?.users || [], search, searchOrderConfig),
```
karthikbtf, line 57: 🔥

karthikbtf (approved):
Nice 🙌 , one comment if you want to follow up.

src/scenes/AcceleratorRouter/PeopleView.tsx
nickgraz: Yea this is just here to show it works, will be removed when the table is added in the next PR

src/redux/features/globalRoles/globalRolesSlice.ts
nickgraz: I don't follow, what would that be for? Just so we don't have to track the request ID in the consumer?

karthikbtf: Yes that and keeping one state in the store for list data.. and we can show data that's already in the store before dispatch gets a chance to complete (on repeat visits).
Just something we did before assuming list/search sorta follows a get by id pattern, just that the id here is a predetermined 'list'. 
Sorry about the confusion, this is fine as a current pattern, no need to change.

karthikbtf: We would potentially show incorrect data prefetched from the store, if the search/sort/filters are different in the current view. Scrap this.

nickgraz: Oh I see what you're saying. Yea I think maybe this is where I should be using `createCachedSelector`. I am going to table this work now that the foundation is done since it is going to end up in another PRD. When we come back to it maybe we can implement `re-reselect`, and it seems like we should use it anywhere we are doing locally filtered search results too.

PR #2262, opened 2024-03-05
SW-4953: UI to assign cohorts to modules

joshbuchea:
This PR includes changes to support showing related modules in the `CohortView`.

I'm not sure if there's an existing pattern for displaying a list of values as a form field, so I went the MVP route and just joined the text to create an inline list. I imagine we'd prefer to display this in a list at the very least and I'm happy to update, but I also imagine most of what's in this PR is incorrect/temporary and will be updated so wasn't sure if I should spend the time.

## Example

localhost_3000_accelerator_cohorts_1

### DIFF
--- a/src/scenes/AcceleratorRouter/CohortView.tsx
+++ b/src/scenes/AcceleratorRouter/CohortView.tsx
@@ -71,7 +71,13 @@ const CohortView = () => {
             <TextField label={strings.PHASE} id='phase' type='text' value={cohort?.phase} display={true} />
           </Grid>
           <Grid item xs={12}>
-            <TextField label="Modules" id='phase' type='text' value={cohortModules?.map(module => module.name).join(', ')} display={true} />
+            <TextField
+              label='Modules'
+              id='phase'
+              type='text'
+              value={cohortModules?.map((module) => module.name).join(', ')}
+              display={true}
+            />
           </Grid>
         </Grid>
       </Card>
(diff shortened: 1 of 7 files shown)


src/services/CohortService.ts
```diff
@@ -66,15 +66,94 @@ const deleteCohort = (cohortId: number) =>
     urlReplacements: { '{cohortId}': `${cohortId}` },
   });
 
+/**
+ * List module(s) for a cohort
+ */
+type MockModule = {
```
karthikbtf, line 72: MockCohortModule ?
Add a cohortId and a moduleId in here.
Should we be reusing the MockModule under src/types/Cohort.ts?

src/types/Cohort.ts
```diff
@@ -4,3 +4,14 @@ export type Cohort = components['schemas']['CohortPayload'];
 
 export type CreateCohortRequestPayload = components['schemas']['CreateCohortRequestPayload'];
 export type UpdateCohortRequestPayload = components['schemas']['UpdateCohortRequestPayload'];
+
+export type MockModule = {
```
karthikbtf, line 8: Suggest calling this MockCohortModule and add cohortId and moduleId properties.

src/scenes/AcceleratorRouter/CohortView.tsx
```diff
@@ -61,13 +63,22 @@ const CohortView = () => {
...
           </Grid>
           <Grid item xs={8}>
             <TextField label={strings.PHASE} id='phase' type='text' value={cohort?.phase} display={true} />
           </Grid>
+          <Grid item xs={12}>
+            <TextField
+              label='Modules'
+              id='phase'
+              type='text'
+              value={cohortModules?.map((module) => module.name).join(', ')}
```
karthikbtf, line 78: This should be fine for the MVP.

karthikbtf (approved):
Looks good to me, few comments on the types.

src/services/CohortService.ts
joshbuchea: I like the rename suggestion and oops I didn't mean to leave a redundant type definition 😅

PR #2263, opened 2024-03-05
Revert "SW-4960 Lazy load routes in terraware web instead of prefetching before we know user principal context

karthikbtf:
- Just realized this will lead to issues with our current single build distribution
-  If user has a cached index.js and we release a new build.. and if user triggers a lazy load from within the cached client.. the chunks will not be found (overwritten)
- I will bring up supporting a minimal history of builds in the next roundtable
- This reverts commit f1a1ec1961a5fcdae26d597b5b10c2feffb47fbb.


joshbuchea (approved):

karthikbtf:
Multiple build task already captured earlier https://terraformation.atlassian.net/browse/SW-4427

PR #2268, opened 2024-03-06
SW-4965: Follow up on max # of files limit related UX in file upload dialog

joshbuchea:
This PR includes changes to the UX of deliverable document limit:

- Document limit and document limit reached message only exists for participants
- No document limit for accelerator

## Before document limit reached

localhost_3000_deliverables_1_organizationId=1 (1)

## After document limit reached

localhost_3000_deliverables_1_organizationId=1 (2)


src/components/DeliverableView/DocumentsUploader.tsx
```diff
@@ -1,18 +1,33 @@
-import { useCallback, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { Box } from '@mui/material';
 import { FileChooser } from '@terraware/web-components';
 
 import { useLocalization } from 'src/providers';
+import { MAX_FILES_LIMIT } from 'src/scenes/DeliverablesRouter/DeliverableView';
```
karthikbtf, line 7: Move this into a constants file maybe within this directory or under `src/constants`
Ideally `scenes` depend on `components`. 
`components` are not aware of `scenes`.

src/scenes/DeliverablesRouter/DeliverableView.tsx
```diff
@@ -28,6 +32,8 @@ const DeliverableView = (props: Props): JSX.Element => {
   const { activeLocale } = useLocalization();
   const theme = useTheme();
 
+  const [showDocumentLimitReachedMessage, setShowDocumentLimitReachedMessage] = useState(false);
```
karthikbtf, line 35: We could show the message without this state.
Maybe do the calculation in this component, does not need to be triggered by the DocumentsUploader.
We can hide the DocumentsUploader in this component conditionally.

src/components/DeliverableView/DocumentsUploader.tsx
```diff
@@ -23,21 +38,29 @@ const DocumentsUploader = ({ deliverable }: ViewProps): JSX.Element => {
 
   const onCloseFileUploadDialog = useCallback(() => void setFiles([]), []);
 
+  useEffect(() => {
+    if (!canEdit && documentLimitReached && showDocumentLimitReachedMessage) {
```
karthikbtf, line 42: We don't want to show this message for READ_ONLY internal users.
The canEdit property coming from AcceleratorRouters captures which global roles can upload files.
I added a comment in DeliverableRouter/DeliverableView suggesting we isolate logic to show the message and hide the documents uploader locally - similar to how we hide the DocumentsUploader in AcceleratorRouter if user does not have edit permissions.
We can keep this component clean of changes except maybe the MAX_FILES_LIMIT change.

karthikbtf (commented):
Thanks for this change, I have a few suggestions.

src/components/DeliverableView/DocumentsUploader.tsx
```diff
@@ -23,21 +38,29 @@ const DocumentsUploader = ({ deliverable }: ViewProps): JSX.Element => {
...
-        multipleSelection
-        uploadDescription={strings.UPLOAD_FILES_DESCRIPTION}
-        uploadText={strings.UPLOAD_FILES_TITLE}
-        template={template}
-      />
+      {!documentLimitReached && (
+        <FileChooser
+          acceptFileType='image/*,application/*'
+          chooseFileText={strings.CHOOSE_FILE}
+          maxFiles={canEdit ? undefined : MAX_FILES_LIMIT - deliverable.documents.length}
```
karthikbtf, line 56: We could pass the max files into the DocumentsUploader component as an optional prop and remove the canEdit prop.
Technically both participants and global-roles with edit permissions can upload (edit) files.

src/scenes/DeliverablesRouter/DocumentLimitReachedMessage.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { MAX_FILES_LIMIT } from './DeliverableView';
+
+const DocumentLimitReachedMessage = (): JSX.Element => {
+  const { activeLocale } = useLocalization();
+  const theme = useTheme();
+
+  return (
+    <>
+      {activeLocale && (
```
karthikbtf, line 17: nice 👍

karthikbtf (commented):
Missed a few.

src/scenes/DeliverablesRouter/DeliverableView.tsx
joshbuchea: I was thinking I would have needed to hoist the `files` state from the `DocumentUploader` up a level into the view, but the calculation needs to be performed using `deliverable.documents.length` rather than `files.length` 🤦 Good call, will update and remove this state 👍

joshbuchea:
I added `maxFiles` props to the child components consuming `MAX_FILES_LIMIT` so it no longer needs to be exported and imported, is it ok to leave in the `DeliverablesRouter/DeliverableView` file or would we prefer it's still moved to a constants file?

karthikbtf (commented):
Looking good Josh, thanks for the changes, just one more suggestion.

joshbuchea:
It's cleaner yet, great feedback 🙌

src/components/DeliverableView/DocumentsUploader.tsx
```diff
@@ -31,12 +35,12 @@ const DocumentsUploader = ({ deliverable }: ViewProps): JSX.Element => {
       <FileChooser
         acceptFileType='image/*,application/*'
         chooseFileText={strings.CHOOSE_FILE}
-        setFiles={setFiles}
-        maxFiles={15 - deliverable.documents.length}
+        maxFiles={maxFiles}
         multipleSelection
+        setFiles={setFiles}
+        template={template}
```
karthikbtf, line 41: Thanks for ordering these 🙏

src/scenes/DeliverablesRouter/DeliverableView.tsx
```diff
@@ -13,11 +13,14 @@ import Page from 'src/components/Page';
 import Card from 'src/components/common/Card';
 import { APP_PATHS } from 'src/constants';
 import { useLocalization } from 'src/providers';
+import DocumentLimitReachedMessage from 'src/scenes/DeliverablesRouter/DocumentLimitReachedMessage';
```
karthikbtf, line 16: If you prefer `./` is fine for files within this directory/hierarchy.

src/strings/csv/en.csv
```diff
@@ -318,6 +318,8 @@ DESTINATION_REQUIRED,Destination *,
 DETAILED_SITE,Detailed Site,
 DETAILS,Details,
 DIRECTION_OR_DESCRIPTION,Direction or Description,
+DOCUMENT_LIMIT_REACHED,Document Limit Reached,
+DOCUMENT_LIMIT_REACHED_MESSAGE,You may only upload {0} documents. Talk to your Terraformation representative to have them remove files if you would like to upload more.,
```
karthikbtf, line 322: Oops missed this, we won't be removing files for now. Not sure that is decided yet. TF representative can upload more files on their behalf. Maybe just `Talk to your Terraformation representative if you would like to upload more.`?

karthikbtf (approved):
LGTM, thanks much!
One feedback on a string, feel free to merge after.

src/strings/csv/en.csv
joshbuchea: I also thought this string wasn't very well considered. Good call, will update 👍

PR #2269, opened 2024-03-06
Update types for deliverable document upload API and update the service call

karthikbtf:
- projectId is back in the payload, removed from path


karthikbtf:
The BE changes have already gone in. FE will fail with uploads until this change goes in.

nickgraz (approved):
Looks good, sorry for the delay!

PR #2270, opened 2024-03-06
SW-4872 Enable species filtering in nursery detail batches view

karthikbtf:
- updated code to pass batches results to Search filters component for nursery detail view
- updated code to look at species_scientificName

Terraware App 2024-03-06 10-46-15


nickgraz (approved):
Nice work

PR #2271, opened 2024-03-06
SW4970: Added search values functionality for querying available filter values

tommylau523:


src/scenes/Species/SpeciesListView.tsx
```diff
@@ -293,14 +293,7 @@ export default function SpeciesListView({ reloadData, species }: SpeciesListProp
...
         prefix: 'species',
-        fields: [
-          'id',
-          'growthForm',
-          'seedStorageBehavior',
-          'ecosystemTypes.ecosystemType',
-          'conservationCategory',
-          'rare',
-        ],
+        fields: ['growthForm', 'seedStorageBehavior', 'ecosystemTypes_ecosystemType', 'conservationCategory', 'rare'],
```
karthikbtf, line 296: Just curious, why the flattened field restriction?

src/scenes/Species/SpeciesListView.tsx
```diff
@@ -312,29 +305,18 @@ export default function SpeciesListView({ reloadData, species }: SpeciesListProp
...
-          } else {
-            (innerAcc[newKey] as Record<string, any>).values.push(value);
-          }
-          return innerAcc;
-        }, acc);
-      }, {}) as FieldOptionsMap;
+      const result = (data ?? []) as FieldOptionsMap;
+
+      // remap ecosystemTypes to nested field from flattened field
+      result['ecosystemTypes.ecosystemType'] = result['ecosystemTypes_ecosystemType'];
```
karthikbtf, line 317: this remapping is totally fine

src/services/SearchService.ts
```diff
@@ -67,6 +80,17 @@ async function search<T extends SearchResponseElement>(entity: RawSearchRequestP
   }
 }
 
+async function searchValues<T extends SearchValuesResponseElement>(
+  entity: RawSearchValuesRequestPayload
+): Promise<T | null> {
+  try {
+    const response: SearchValuesResponsePayload = (await httpSearchValues.post({ entity })).data;
```
karthikbtf, line 87: Can data be undefined or null here?

karthikbtf: I believe the HttpService handles errors and returns corresponding state.
You could remove the try/catch and check the status of response.
```
  const response: SearchValueRes..... post({...});
    return response.data?.results as T | null;
  } else {
    return null;
  }
```

karthikbtf (commented):
Looking good, a few suggestions.

src/services/SeedBankService.ts
```diff
@@ -149,13 +146,13 @@ const searchFieldValues = async (
...
+    const params: SearchRequestPayload = {
+      prefix: 'facilities.accessions',
       fields,
-      search: formattedSearch,
+      search: SearchService.convertToSearchNodePayload(searchCriteria ?? {}, organizationId),
+      count: 1000,
     };
-    const apiResponse: Response = await HttpService.root(FIELD_VALUES_ENDPOINT).post({ entity });
-    return apiResponse.data.results;
+    return SearchService.searchValues(params);
```
karthikbtf, line 155: Since we aren't await-ing, we could remove the `async` from this function.

src/scenes/Species/SpeciesListView.tsx
karthikbtf, line 296: Remove the `rare`, looks like it is hardcoded on line 322.

src/scenes/Species/SpeciesListView.tsx
```diff
@@ -312,29 +305,20 @@ export default function SpeciesListView({ reloadData, species }: SpeciesListProp
...
-          const value = isEcosystemTypes ? (d[k] as any[]).map((et) => et.ecosystemType) : d[k];
-          if (Array.isArray(value)) {
-            (innerAcc[newKey] as Record<string, any>).values.push(...value);
-          } else {
-            (innerAcc[newKey] as Record<string, any>).values.push(value);
-          }
-          return innerAcc;
-        }, acc);
-      }, {}) as FieldOptionsMap;
+      const result = (data ?? []) as FieldOptionsMap;
```
karthikbtf, line 314: `data` needs to be an object, it will not be undefined since the searchValues function returns null as a fallback.
This could be 
```
  const result = (data || {}) as FieldOptionsMap;
```

src/scenes/Species/SpeciesListView.tsx
```diff
@@ -312,29 +305,20 @@ export default function SpeciesListView({ reloadData, species }: SpeciesListProp
...
+
+      // remap ecosystemTypes to nested field from flattened field
+      /* tslint:disable:no-string-literal */
+      result['ecosystemTypes.ecosystemType'] = result['ecosystemTypes_ecosystemType'];
```
karthikbtf, line 318: set only if `result['ecosystemTypes_ecosystemType']` is not undefined? Not too sure what happens with undefined values for keys in the filter. (in most cases there should be values for ecosystem types)

karthikbtf (commented):
One more review, looking good, a few comments.
Works great.

src/scenes/Species/SpeciesListView.tsx
```diff
@@ -278,7 +278,6 @@ export default function SpeciesListView({ reloadData, species }: SpeciesListProp
         ? [
             { name: 'growthForm', label: strings.GROWTH_FORM, type: 'multiple_selection' },
             { name: 'conservationCategory', label: strings.CONSERVATION_CATEGORY, type: 'multiple_selection' },
-            { name: 'rare', label: strings.RARE, type: 'multiple_selection' },
```
karthikbtf, line 281: Oops, I should have clarified, we need this table column.
I meant removing `rare` from the list of fields on new line # 295. We don't use the values of `rare` from the response.

karthikbtf (approved):
Nice 👍

PR #2278, opened 2024-03-07
SW-4989 Reorg accelerator router code

karthikbtf:
- moved code into individual subdirs
- extracted out routing for cohorts and deliverables
- replaced AcceleratorMain usage with the 'Page' component which already has built-in header wrappers and snackbar message
- kept the original AcceleratorMain usage in DeliverablesList for a later cleanup
- if there are any styling/spacing issues we can cleanup later


nickgraz (approved):
Nice, thanks for the refactor and getting getting rid of `AcceleratorMain`

karthikbtf:
> Nice, thanks for the refactor and getting getting rid of `AcceleratorMain`

We still have AcceleratorMain and an equivalent used for both deliverables list use-cases. I will clean those up separately.

PR #2280, opened 2024-03-07
Rename 'Voting' flag to 'Console'

karthikbtf:
- just that it makes more sense as that's the milestone name


tommylau523 (approved):
Looks good!

PR #2285, opened 2024-03-11
Sync up with deliverables backend

nickgraz:
- Rework the deliverable fetch since we have to pass in project ID and can't use a simple arg key anymore
- Fix the issue with the org ID being sent from the accelerator console

### DIFF
--- a/src/components/DeliverableView/useFetchDeliverable.ts
+++ b/src/components/DeliverableView/useFetchDeliverable.ts
@@ -1,14 +1,11 @@
-import { useCallback, useEffect, useMemo } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 
 import { APP_PATHS } from 'src/constants';
 import useAcceleratorConsole from 'src/hooks/useAcceleratorConsole';
 import { Statuses } from 'src/redux/features/asyncUtils';
 import { requestGetDeliverable } from 'src/redux/features/deliverables/deliverablesAsyncThunks';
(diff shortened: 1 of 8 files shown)


nickgraz:
* **#2285** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2285?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

joshbuchea (approved):

src/components/DeliverableView/useFetchDeliverable.ts
```diff
@@ -33,16 +30,20 @@ export default function useFetchDeliverable({ deliverableId, projectId }: Props)
   const snackbar = useSnackbar();
   const history = useHistory();
   const dispatch = useAppDispatch();
-  const deliverableResult = useAppSelector(selectDeliverableFetchRequest(deliverableId));
-  const deliverableDataResult = useAppSelector(selectDeliverable(deliverableId));
```
karthikbtf, line 37: Thanks for the change. Curious why this earlier version didn't work. The dispatch should have triggered a new fetch and data update in the store. I'm probably missing something.

nickgraz: It also might be confusing because the original change to the service was done in a previous PR

nickgraz: @karthikbtf here's my crack at that https://github.com/terraware/terraware-web/pull/2289

PR #2286, opened 2024-03-11
SW-4975: FE - Voting mock redux and service layer

joshbuchea:
This PR includes changes to implement a redux store and `VotingService` that will return mock data until the BE is ready.


joshbuchea:
Having just reviewed Tommy's PR for [[WIP] SW-5008 Created scaffolding for Votes API](https://github.com/terraware/terraware-server/pull/1745), I see I need to make some changes. Working on that now.

src/redux/features/votes/votesSelectors.ts
```diff
@@ -0,0 +1,5 @@
+import { RootState } from 'src/redux/rootReducer';
+
+export const selectProjectVotes = (state: RootState) => state.votes.votes;
+
+export const selectProjectVotesRequest = (state: RootState, requestId: string) => state.votesRequests[requestId];
```
karthikbtf, line 5: selectProjectVotesEditRequest?

src/redux/features/votes/votesAsyncThunks.ts
```diff
@@ -0,0 +1,43 @@
...
+      request.payload
+    );
+
+    if (response !== null && response.requestSucceeded) {
+      return response.data;
+    }
+
+    return rejectWithValue(strings.GENERIC_ERROR);
+  }
+);
```
karthikbtf, line 43: do we need a requestListProjectVotes?

karthikbtf (commented):
looking good

src/redux/features/votes/votesAsyncThunks.ts
joshbuchea: Ooh you're right, will add 👍

joshbuchea:
Thanks for the review @karthikbtf 🙌

src/redux/features/votes/votesAsyncThunks.ts
karthikbtf: Thanks Josh! Just fyi I have branched off yours for the page view.

joshbuchea:
I realize I got caught up trying to match the BE types and that likely wasn't the most efficient use of my time. I'll try to wrap this PR up real quick so I can move on to something else.

tommylau523:
> I realize I got caught up trying to match the BE types and that likely wasn't the most efficient use of my time. I'll try to wrap this PR up real quick so I can move on to something else.

I apologize for the service endpoints types changing!

src/redux/features/votes/votesAsyncThunks.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const requestProjectVotesUpdate = createAsyncThunk(
+  'votes/update',
+  async (request: { projectId: number; payload: UpsertProjectVotesRequestPayload }, { rejectWithValue }) => {
+    const response: Response2<UpsertProjectVotesResponsePayload> = await VotesService.setProjectVotes(
+      request.projectId,
+      request.payload
+    );
+
+    if (response !== null && response.requestSucceeded) {
```
karthikbtf, line 21: Response will not be null, you should get a valid response object. The service layer catches errors and set the status in requestSucceeded / error properties.

src/redux/features/votes/votesSlice.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { StatusT, buildReducers } from 'src/redux/features/asyncUtils';
+import { requestProjectVotesDelete, requestProjectVotesUpdate } from 'src/redux/features/votes/votesAsyncThunks';
+import { ProjectVotesPayload, ProjectVotesResponsePayload } from 'src/types/Votes';
+
+// Define a type for the slice state
+type Data = {
+  error?: string;
+  votes?: ProjectVotesPayload;
+};
```
karthikbtf, line 11: Suggest making this a Record of projectId to ProjectVotesPayload. And use the buildReducers approach, with useArgAsKey param set to true.

src/services/VotesService.ts
```diff
@@ -0,0 +1,99 @@
...
+  deleteProjectVotes,
+  getProjectVotes,
+  setProjectVotes,
+};
+
+export default VotesService;
+
+// TODO: remove mock data once BE is ready
+mockGetProjectVotesResponseData = {
+  projectId: 1,
```
karthikbtf, line 76: @tommylau523 can we also get the projectName with this payload?

karthikbtf: As of now, users with global roles cannot fetch the project name using project API.

src/redux/features/votes/votesSlice.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { StatusT, buildReducers } from 'src/redux/features/asyncUtils';
+import { requestProjectVotesDelete, requestProjectVotesUpdate } from 'src/redux/features/votes/votesAsyncThunks';
+import { GetProjectVotesResponse, GetProjectVotesResponsePayload } from 'src/types/Votes';
+
+// Define a type for the slice state
+type Data = {
+  error?: string;
+  votes?: GetProjectVotesResponsePayload;
+};
```
karthikbtf, line 11: Can't find my earlier comment.
Suggest making this a record of project id to votes response.
example:
```
// Define a type for the slice state
type Data = Record<number, StatusT<GetProjectVotesResponsePayload>>;

// Define the initial state
const initialState: Data = {};

export const votesSlice = createSlice({
  name: 'votesSlice',
  initialState,
  reducers: {},
  extraReducers: (builder: ActionReducerMapBuilder<Data>) => {
    buildReducers(requestProjectVotes, true)(builder);
  },
});

```

src/services/VotesService.ts
karthikbtf: Thanks @tommylau523 , saw the PR update.

src/types/Votes.ts
```diff
@@ -0,0 +1,65 @@
...
+export type DeleteProjectVotesResponsePayload = components['schemas']['SimpleSuccessResponsePayload'];
+
+export type DeleteVoteSelection = {
+  projectId: number;
+  phase?: CohortPhase;
+  userId?: number;
+};
+
+export type GetProjectVotesResponsePayload = {
+  projectId: number;
```
karthikbtf, line 23: @joshbuchea can we also add projectName? BE PR was updated.

joshbuchea:
> > I realize I got caught up trying to match the BE types and that likely wasn't the most efficient use of my time. I'll try to wrap this PR up real quick so I can move on to something else.
> 
> I apologize for the service types changing so frequently!

No worries at all @tommylau523 🙂

src/redux/features/votes/votesSlice.ts
joshbuchea: I was wondering about doing that when I set it up. I peaked in a few slices and it didn't seem like we were doing so in any of the other reducers, at least for projectId.

src/redux/features/votes/votesSlice.ts
joshbuchea: I was wondering about keying off of the projectId when I set it up. I peaked in a few slices and it didn't seem like we're currently doing that in any of the other reducers, at least for projectId.

src/types/Votes.ts
joshbuchea: You got it 👍

joshbuchea: @karthikbtf @tommylau523 should we also add `projectName` to the `UpsertProjectVotesResponsePayload`?

src/redux/features/votes/votesSelectors.ts
joshbuchea: So I added an async thunk to get the project votes that also uses the `votesRequests` store, seems like the naming should stay generic in this case?

src/types/Votes.ts
karthikbtf: Not needed, that is keyed off the projectId.
Project name is only for the client display purposes.

src/redux/features/votes/votesAsyncThunks.ts
joshbuchea: Ok thanks for the heads up, hopefully the changes I'm making haven't been too disruptive to your work

src/types/Votes.ts
joshbuchea: Gotcha

src/redux/features/votes/votesAsyncThunks.ts
karthikbtf: The changes were helpful, I needed them :)

src/redux/features/votes/votesSlice.ts
joshbuchea: I don't mind making the change, it just seems like it may be inconsistent with the rest of the app?

karthikbtf: Can you point me to the inconsistency?
Here we are fetching votes by projectId, we need to key the result off the projectId.
Otherwise when we switch to another project, we will end up show stale incorrect data that's in the store, probably relevant for another project.

src/redux/features/votes/votesAsyncThunks.ts
joshbuchea: I added a `requestProjectVotesGet` async thunk, does we need a separate `requestListProjectVotes` thunk?

src/redux/features/votes/votesSlice.ts
karthikbtf: I think I understand what you may have been comparing to.
For org level data, we don't scope the results to any key since we either work one org at a time (or globally such as in the accelerator console).
When operating within an org, storing results by requested id is preferred, [example](https://github.com/terraware/terraware-web/blob/main/src/redux/features/draftPlantingSite/draftPlantingSiteSlice.ts#L43)
There is no global Vote results, these are per project id, hence my request.

src/redux/features/votes/votesAsyncThunks.ts
joshbuchea: I realize the list thunk sorts data by locale, but I think we still needed `requestProjectVotesGet` to return the entire response.

karthikbtf: No we don't need a list project votes. We won't be showing votes across projects.
The only way to show votes is by project. When the user visits a project's scores, there's a link to view the project's votes.

karthikbtf: My original comment/question should have just said 'requestGetProjectVotes' to avoid confusion. Was mostly referring to listing votes within a project, but list also sounds like listing all projects votes. What you have right now is good 👍

src/redux/features/votes/votesSlice.ts
joshbuchea: I was trying to highlight that I couldn't find any instances of `<projectId` in the code base. As far as I can tell we aren't keying off project id anywhere else in the app right now and the change seemed inconsistent. I think the reasoning for it makes sense, just trying to understand why we wanted to make the change here and nowhere else.

src/redux/features/votes/votesSlice.ts
```diff
@@ -0,0 +1,51 @@
...
+type VotingRequestsState = Record<string, StatusT<ProjectVotesResponsesUnion>>;
+
+const initialVotingRequestsState: VotingRequestsState = {};
+
+export const votesRequestsSlice = createSlice({
+  name: 'votesRequestsSlice',
+  initialState: initialVotingRequestsState,
+  reducers: {},
+  extraReducers: (builder: ActionReducerMapBuilder<VotingRequestsState>) => {
+    buildReducers(requestProjectVotesGet)(builder);
```
karthikbtf, line 45: I'm assuming when we edit or delete votes, we only want to know if it succeeded (boolean true/false).
That is the pattern so far, the client refreshes the view after that.
Votes Update/Delete should have a result store of requestId to boolean status, ideally.
The get project votes should be a separate slice on it's own, keyed off the projectId.

src/redux/features/votes/votesSlice.ts
joshbuchea: Will make the change though.

src/redux/features/votes/votesSelectors.ts
karthikbtf: We should have separate slices for the edit requests and the get requests. Added a comment in the reducers section.

Example:
```
export const selectProjectVotes = (state: RootState, projectId: number) => state.votes[projectId];

export const selectProjectVotesEditRequest = (state: RootState, requestId: string) => state.votesRequests[requestId];
```

and the reducers something like this

```
// Define a type for the slice state
type Data = Record<number, StatusT<GetProjectVotesResponsePayload>>;

// Define the initial state
const initialState: Data = {};

export const votesSlice = createSlice({
  name: 'votesSlice',
  initialState,
  reducers: {},
  extraReducers: (builder: ActionReducerMapBuilder<Data>) => {
    buildReducers(requestProjectVotesGet, true)(builder);
  },
});

export const votesReducer = votesSlice.reducer;

type VotingRequestsState = Record<string, StatusT<boolean>>;

const initialVotingRequestsState: VotingRequestsState = {};

export const votesRequestsSlice = createSlice({
  name: 'votesRequestsSlice',
  initialState: initialVotingRequestsState,
  reducers: {},
  extraReducers: (builder: ActionReducerMapBuilder<VotingRequestsState>) => {
    buildReducers(requestProjectVotesUpdate)(builder);
    buildReducers(requestProjectVotesDelete)(builder);
  },
});

```

Have the async thunks return boolean values for the updates/deletes.

src/redux/features/votes/votesSlice.ts
joshbuchea: It seems like the pattern in a lot of the web app is to reset the data when fetching with new params rather than storing the data in redux keyed off of the params.

karthikbtf (commented):
@joshbuchea I think our comments criss-crossed and added to some confusion. Sorry about that.

Just two open requests about the get and update/delete slices and selectors.

src/redux/features/votes/votesSlice.ts
karthikbtf: @joshbuchea we don't reset data, except when user switches orgs.
We have a few patterns, all of them have data stored by id, either directly or indirectly.
In some cases this is built into the response from the BE, example: we list all projects and store the entire list of projects in the redux store. When we want a project by id, we look up this list.
We have cases where we don't list all the data, instead we fetch by id (such as in the example earlier).. in these cases, we key off the id in the store.
We don't store data without knowing what it belongs to.

joshbuchea:
> @joshbuchea I think our comments criss-crossed and added to some confusion. Sorry about that.
> 
> Just two open requests about the get and update/delete slices and selectors.

No worries at all, thank you for your time and effort reviewing my PR 🙌 Will work to get those two items wrapped up ASAP.

src/redux/features/votes/votesSlice.ts
joshbuchea: Ok that makes sense that we'd want to store data differently for those two types of requests, thank you for explaining. Will update.

karthikbtf:
> > @joshbuchea I think our comments criss-crossed and added to some confusion. Sorry about that.
> > Just two open requests about the get and update/delete slices and selectors.
> 
> No worries at all, thank you for your time and effort reviewing my PR 🙌 Will work to get those two items wrapped up ASAP.

Thanks @joshbuchea . Do you think it would be helpful to document the redux workflows to avoid confusion?
The redux work has been a learning experience and we are slowly evolving there in terms of best-practices. There might be some old code that doesn't look like new code, example: using async thunks more often now

src/redux/features/votes/votesSlice.ts
joshbuchea: @karthikbtf so actually I just realized the delete response is boolean only but the upsert response returns some data. In that case how should we handle? Split all three? 🤔

karthikbtf: My thoughts but will leave it to you. For now we can keep the delete and update as booleans, the client will just reload all votes. It's a pattern we use everywhere. At some point if we need to get more efficient, we can split them down the road.

joshbuchea:
> > > @joshbuchea I think our comments criss-crossed and added to some confusion. Sorry about that.
> > > Just two open requests about the get and update/delete slices and selectors.
> > 
> > 
> > No worries at all, thank you for your time and effort reviewing my PR 🙌 Will work to get those two items wrapped up ASAP.
> 
> Thanks @joshbuchea . Do you think it would be helpful to document the redux workflows to avoid confusion? The redux work has been a learning experience and we are slowly evolving there in terms of best-practices. There might be some old code that doesn't look like new code, example: using async thunks more often now

Sure, some documentation for the redux workflows would likely be pretty helpful for me. I do find myself jumping around a bit trying to make best guesses as to which patterns to use when implementing new redux/service layers. It probably doesn't have to be very extensive, could be as simple as describing our current preferred patterns.

src/types/Votes.ts
```diff
@@ -0,0 +1,66 @@
...
+export type UpsertVoteSelection = {
+  projectId: number;
+  phase: CohortPhase;
+  userId: number;
+  voteOption?: VoteOption;
+};
+
+type VoteOption = 'yes' | 'no' | 'conditional';
+
+type VoteSelection = {
```
karthikbtf, line 56: can we export these types?

src/types/Votes.ts
```diff
@@ -0,0 +1,66 @@
+import { components } from 'src/api/types/generated-schema';
+
+type CohortPhase = string;
+
+type PhaseVotes = {
```
karthikbtf, line 5: export these types?

karthikbtf (approved):
Looks good to me @joshbuchea 🙌

joshbuchea:
Sorry for being so thick, I was definitely confused about a few things and your clarifications were a huge help 😅 Thank you @karthikbtf. I made the redux store updates and fixed the typescript error, but please let me know if I missed anything and I'll get it updated right away.

src/types/Votes.ts
joshbuchea: You got it

karthikbtf (approved):
Looks good to me @joshbuchea nice work 🎉

PR #2291, opened 2024-03-12
SW-4974 Implement mock data, redux, service layer, and hooks for listing and updating scores for a project

nickgraz:
This gets us to a point where we can start working on the frontend, but there will likely be a fair amount of changes when the backend is completed.


nickgraz:
* **#2291** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523:
https://github.com/terraware/terraware-server/pull/1753

This may help!

src/types/Score.ts
```diff
@@ -0,0 +1,35 @@
+import strings from 'src/strings';
+
+// This will all be updated with real types when the BE is done
+
+export type Score = {
```
tommylau523, line 5: In the backend, we had a more nested approach for organizing phase:

```
[
  {
    phase: "phase 0",
    scores: [
      {
        category: "Carbon",
        score: null | number
      }, ...
    ]
  },
  {
    phase: "phase 1",
    ...
  },
]
  
```

It may be easier to organize permissions going forward since scores should be frozen in previous phases but editable currently for example.

nickgraz: Excellent, thanks for the info. I am going to wait until the BE is done before making too many modifications here as it is just for the purpose of moving the frontend at the moment.

nickgraz: @tommylau523 I updated it a little bit to match the possible backend a bit closer.

tommylau523 (approved):
Looks good to me! I think some service calls may look a bit different depending on the API schema but can be fixed later!

PR #2293, opened 2024-03-12
SW-5027 Allow navigating back to deliverables list from detail deliverable view

karthikbtf:
- added a route check to flag this case


karthikbtf:
@nickgraz @joshbuchea a very simple bug fix for deliverables

joshbuchea (approved):

PR #2303, opened 2024-03-13
SW-5037 In documents table: fix projectName column, create link to the document, fix createdTime column

nickgraz:


src/components/DeliverableView/DocumentCellRenderer.tsx
```diff
@@ -3,20 +3,25 @@ import React from 'react';
...
         column={column}
-        // TODO this will be a "BE enum to string converter" once we get the types and enums
         value={
-          <Link to={{ pathname: value as string }} target='_blank'>
-            {row.documentType === 'google' ? 'Google Drive' : ''}
+          <Link
+            to={ENDPOINT_DELIVERABLE_DOCUMENT.replace('{deliverableId}', row.deliverableId).replace(
+              '{documentId}',
+              row.id
+            )}
```
nickgraz, line 21: This is definitely not prettier

src/components/DeliverableView/DocumentsList.tsx
```diff
@@ -28,13 +28,23 @@ const DocumentsList = (props: DocumentsListProps): JSX.Element => {
       direction: order === 'asc' ? 'Ascending' : 'Descending',
     });
 
+  const rows = useMemo(
+    () =>
+      props.deliverable.documents.map((document) => ({
+        ...document,
+        projectName: props.deliverable.projectName,
+        deliverableId: props.deliverable.id,
+      })),
```
nickgraz, line 37: I went this route as opposed to plumbing this redundant info into the BE response, let me know if you'd rather I go that route.

src/components/DeliverableView/DocumentCellRenderer.tsx
```diff
@@ -3,20 +3,25 @@ import React from 'react';
...
+            target='_blank'
+          >
+            {row.documentStore}
```
nickgraz, line 24: I am assuming we don't want to localize these values since they are brand names.

karthikbtf: We do the same thing for `Terraware`.
If we keep them in strings, allows us to support languages in other scripts. Super theoretical, but totally fine for now. Let's keep it as-is,  we can do a clean sweep later on if there's a need.

src/components/DeliverableView/DocumentsList.tsx
karthikbtf: This is totally fine. 
Ideally we create reducers to produce these views but overkill here.

src/components/DeliverableView/DocumentCellRenderer.tsx
```diff
@@ -3,20 +3,25 @@ import React from 'react';
...
     return (
       <CellRenderer
         index={index}
         column={column}
-        // TODO this will be a "BE enum to string converter" once we get the types and enums
         value={
-          <Link to={{ pathname: value as string }} target='_blank'>
-            {row.documentType === 'google' ? 'Google Drive' : ''}
+          <Link
+            to={ENDPOINT_DELIVERABLE_DOCUMENT.replace('{deliverableId}', row.deliverableId).replace(
```
karthikbtf, line 18: 👍

karthikbtf (approved):
Looks great! :shipit:

PR #2308, opened 2024-03-13
SW-5041 Integrate with BE voting types

karthikbtf:
- updated types
- removed delete api (no use-case)
- data is still mocked for now


tommylau523:
Hey I just noticed that `projectId` is redundant in the payload, since it is also a path variable. Sorry about that. 

I pushed a patch on the backend here:
https://github.com/terraware/terraware-server/pull/1758

tommylau523 (approved):

PR #2310, opened 2024-03-13
SW-4529 Show org creation success banner after page change

alexTerraformation:
Previously if you were on a different page prior to creating an org the banner would not show up because it the redirect would happen after the banner was drawn.


karthikbtf (approved):
Good catch.

PR #2312, opened 2024-03-13
SW-4961 Start on people list view using TableWithSearchFilters

nickgraz:


src/components/Page/index.tsx
```diff
@@ -13,6 +13,7 @@ export type PageProps = {
   crumbs?: Crumb[];
   hierarchicalCrumbs?: boolean;
   isLoading?: boolean;
+  pageHeaderRef?: RefObject<HTMLDivElement>;
```
nickgraz, line 16: Pass this in optionally so we can use it in the parent component.

src/scenes/AcceleratorRouter/People/ListView.tsx
```diff
@@ -0,0 +1,131 @@
...
+    (locale: string | null, search: SearchNodePayload, searchSortOrder: SearchSortOrder) => {
+      // TODO implement search and sort order into redux and service
+      const request = dispatch(requestListGlobalRolesUsers({ locale }));
+      setRequestId(request.requestId);
+    },
+    [dispatch]
+  );
+
+  // tslint:disable:no-console
+  console.log('globalRoleUsers', globalRoleUsers);
```
karthikbtf, line 96: remove... later?

src/components/Page/index.tsx
```diff
@@ -76,7 +78,7 @@ export default function Page({
       <Grid item xs={12}>
         <PageSnackbar />
       </Grid>
-      <Grid container ref={contentRef} sx={contentStyle}>
+      <Grid container ref={pageHeaderRef ?? contentRef} sx={contentStyle}>
```
karthikbtf, line 81: I think I'm missing when/why we would want to override the ref.

karthikbtf (approved):
Very cool, one question.

src/components/Page/index.tsx
nickgraz: The table container uses the page header ref. I'm not exactly sure what it is for, but since the table component is adjacent to this one, it has to be managed in the parent. I can double check tomorrow to see if it is needed.

karthikbtf: The page header ref is mostly to manage the static header when scrolling. Page component should auto handle it I think. Sure, if you figure it out and it's not needed, feel free to remove it. Looks great otherwise.

karthikbtf (approved):
Nice!

nickgraz:
### Merge activity

* **Mar 14, 11:07 AM EDT**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2312).

PR #2313, opened 2024-03-14
SW-5024-prep Refactor Votes pages for reusability

karthikbtf:
- Voting read-only and edit-views have similar data and scaffolding
- Extracted that out into a VotingWrapper
- Using a VotingProvider react context to provide the votes data as both views leverage url params
- removed the useFetchVotes hook
- TODO: implement voting edit details (Follow up)

Terraware App 2024-03-13 17-35-05

### DIFF
--- a/src/scenes/AcceleratorRouter/Voting/VotingProvider.tsx
+++ b/src/scenes/AcceleratorRouter/Voting/VotingProvider.tsx
@@ -1,32 +1,34 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
-import { useHistory } from 'react-router-dom';
+import { useCallback, useEffect, useState } from 'react';
+import { useHistory, useParams } from 'react-router-dom';
 
 import { APP_PATHS } from 'src/constants';
-import { Statuses } from 'src/redux/features/asyncUtils';
 import { requestProjectVotesGet } from 'src/redux/features/votes/votesAsyncThunks';
 import { selectProjectVotes } from 'src/redux/features/votes/votesSelectors';
 import { useAppDispatch, useAppSelector } from 'src/redux/store';
-import { PhaseVotes } from 'src/types/Votes';
+import { Phase, PhaseVotes } from 'src/types/Votes';
+import useQuery from 'src/utils/useQuery';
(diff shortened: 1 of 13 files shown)


src/scenes/AcceleratorRouter/Voting/VotingProvider.tsx
```diff
@@ -1,33 +1,35 @@
...
+const VotingProvider = ({ children }: Props): JSX.Element => {
   const history = useHistory();
+  const query = useQuery();
   const dispatch = useAppDispatch();
   const snackbar = useSnackbar();
 
-  const [phaseVotes, setPhaseVotes] = useState<PhaseVotes>();
+  const pathParams = useParams<{ projectId: string }>();
+  const projectId = Number(pathParams.projectId);
+  const phase: Phase = (query.get('phase') as Phase) || 'Phase 1 - Feasibility Study'; // default to phase 1?
```
nickgraz, line 26: That seems correct

src/scenes/AcceleratorRouter/Voting/index.tsx
```diff
@@ -2,17 +2,22 @@ import { Redirect, Route, Switch } from 'react-router-dom';
...
 import VotingEdit from './VotingEdit';
+import VotingProvider from './VotingProvider';
+import VotingView from './VotingView';
 
 const VotingRouter = () => {
   return (
     <Switch>
       <Route exact path={APP_PATHS.ACCELERATOR_VOTING}>
-        <VotingView />
+        <VotingProvider>
```
nickgraz, line 13: Not a big deal, but I think you can wrap the whole switch in the provider. Not 100% sure though.

nickgraz (approved):
Nice work!

src/scenes/AcceleratorRouter/Voting/index.tsx
karthikbtf: That would be ideal, let me give that a try. If it works I'll get that in. Thanks for the tip.

karthikbtf: That worked. Thanks!

PR #2314, opened 2024-03-14
SW-5013: Scoring entry component

joshbuchea:
This PR includes a new `ScoreEntry` component for the accelerator scorecards.

## TODO

- [x] Remove hacky demo code
- [ ] Break out sub-components from `ScoreEntry` into discrete files?
- [ ] Move many styles from `sx` prop to style props on the components?
- [ ] Remove local state from `SourceControl` component?

## Example

https://github.com/terraware/terraware-web/assets/1474361/720ec198-2ae3-4a16-8bb9-28548607d510


src/scenes/AcceleratorRouter/Overview/index.tsx
```diff
@@ -1,15 +1,95 @@
...
 import Page from 'src/components/Page';
 import Button from 'src/components/common/button/Button';
 import { APP_PATHS } from 'src/constants';
 import strings from 'src/strings';
+import { Scorecard } from 'src/types/Score';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
 
+import ScoreEntry from '../Scoring/ScoreEntry';
+
+const mockQualitativeInfo =
```
joshbuchea, line 15: This is a hacky demo to test the component, will remove all changes in this scene before merging.

src/scenes/AcceleratorRouter/Scoring/ScoreEntry.tsx
```diff
@@ -0,0 +1,148 @@
...
+};
+
+export default ScoreEntry;
+
+type ScoreControlProps = {
+  disabled?: boolean;
+  onChange: (value: ScoreValue) => void;
+  score: Score;
+};
+
```
joshbuchea, line 56: There are two sub-components in this file: `SourceControl` and `SourceControlButton`. I can break these out into discrete files if preferred.

src/scenes/AcceleratorRouter/Scoring/ScoreEntry.tsx
```diff
@@ -0,0 +1,148 @@
...
+  return (
+    <Box
+      sx={{
+        backgroundColor: theme.palette.TwClrBaseGray050,
+        borderRadius: '8px',
+        display: 'inline-flex',
+        marginBottom: theme.spacing(2),
+        padding: '12px',
+      }}
```
joshbuchea, line 72: There's also a lot of usage of the `sx` prop to set styles, I can move many of these to style props on the component if that's preferred.

src/scenes/AcceleratorRouter/Scoring/ScoreEntry.tsx
```diff
@@ -0,0 +1,148 @@
...
+  onChange: (value: ScoreValue) => void;
+  score: Score;
+};
+
+const ScoreControl = ({ disabled, onChange, score }: ScoreControlProps) => {
+  const theme = useTheme();
+  const castScoreValue = Number(score.value) as ScoreValue;
+  const [scoreValue, setScoreValue] = useState<ScoreValue>(castScoreValue);
+  const scoreValueLabel = getScoreValue(scoreValue);
+  const scoreColors = getScoreColors(scoreValue, theme);
```
joshbuchea, line 62: I added local state in the `ScoreControl` component that's initialized with the value from the `Score` record in the scorecard, but is updated independently after initialization. I think this might be nice so the component feels snappy and responsive like other form controls, but we can remove if we'd prefer to not have this.

karthikbtf (approved):

src/scenes/AcceleratorRouter/Scoring/ScoreEntry.tsx
karthikbtf: Line 62 will change on every render, we should use `useMemo` here

src/scenes/AcceleratorRouter/Scoring/ScoreEntry.tsx
joshbuchea: Ok that's kind of what I was thinking, but wasn't sure if we had a preference. Thanks!

src/scenes/AcceleratorRouter/Scoring/ScoreEntry.tsx
joshbuchea: Good call, will update.

src/scenes/AcceleratorRouter/Scoring/ScoreEntry.tsx
```diff
@@ -0,0 +1,148 @@
...
+  disabled?: boolean;
+  onChange: (value: ScoreValue) => void;
+  score: Score;
+};
+
+const ScoreControl = ({ disabled, onChange, score }: ScoreControlProps) => {
+  const theme = useTheme();
+  const castScoreValue = Number(score.value) as ScoreValue;
+  const [scoreValue, setScoreValue] = useState<ScoreValue>(castScoreValue);
+  const scoreLabel = useMemo(() => getScoreValue(scoreValue), [scoreValue]);
```
karthikbtf, line 61: probably don't need this one since it returns a string and won't change

karthikbtf (approved):
LGTM, one last comment.

joshbuchea:
Thanks for the review, I appreciate it 🙌

PR #2316, opened 2024-03-14
SW-4916 Require accession quantity to do a viability test

alexTerraformation:
Since performing a viability test does a withdrawal, we need to have a positive quantity or estimated quantity in order to add a test.


karthikbtf (approved):
👍

PR #2334, opened 2024-03-15
Implement trigram similarity to reduce search results

nickgraz:
Implement trigram similarity to reduce search results

Make sure single letter searches work as expected

### DIFF
--- a/src/utils/searchAndSort.test.ts
+++ b/src/utils/searchAndSort.test.ts
@@ -5,6 +5,7 @@ import {
   modifySearchNode,
   searchAndSort,
   splitTrigrams,
+  trigramWordSimilarity,
 } from 'src/utils/searchAndSort';
 
 type MockResult = {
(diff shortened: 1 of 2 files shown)


nickgraz:
* **#2334** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2334?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/utils/searchAndSort.ts
```diff
@@ -61,9 +63,21 @@ const searchConditionMet = <T extends Record<string, unknown>>(result: T, condit
       return searchValues.some((value) => resultValue.includes(value));
     } else if (condition.type === 'Fuzzy') {
       return searchValues.some((value) => {
-        // Split the search value into trigrams and see if the result field contains any trigram
+        // Trigrams don't work with single letter searches
+        if (value.length === 1) {
```
tommylau523, line 68: Seems like we may want to include up to 3 length for this case. For reference the console has 25 deliverables under Sampson Road. If I search `sa`, or `sam`. I get 8 of the 25 on the former and no results on the latter. works great for 4 or more!

src/utils/searchAndSort.test.ts
```diff
@@ -20,11 +21,108 @@ type MockResult = {
...
+    /**
+     * SELECT word_similarity('andro', 'project andromeda');
+     * word_similarity
+     * -----------------
+     *       0.8333333
+     */
+    expect(trigramWordSimilarity('andro', 'project andromeda')).toEqual(0.8333333333333334);
+
+    /**
+     * I was unable to make this work exactly like postgres. I spent way too long trying to get to
```
nickgraz, line 85: Read this note

src/utils/searchAndSort.ts
nickgraz: This is fixed in the latest commit

tommylau523 (approved):
Works well! Well done!

src/utils/searchAndSort.ts
```diff
@@ -37,7 +37,24 @@ export const splitTrigrams = (value: string): string[] => {
...
+  return new Set(trigrams);
+};
+
+export const trigramWordSimilarity = (a: string, b: string) => {
+  // Split the search value and result value into trigrams and compare their similarity
+  const aTrigrams = new Set(splitTrigrams(a));
+  const bTrigrams = new Set(splitTrigrams(b));
+
+  const matches: string[] = [];
+  bTrigrams.forEach((trigram) => {
```
tommylau523, line 50: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection

This may work the same? Not sure though

nickgraz: I would love to use that, but unfortunately it isn't available in many browsers (no available at all in FF) - https://caniuse.com/?search=set.intersection.

nickgraz: And I can't seem to find a babel polyfill for it.

tommylau523: Sets are really that rarely used in Javascript I guess!

nickgraz: Yea, it is quite sad. There is an [open proposal](https://tc39.es/proposal-set-methods/) for it so hopefully it will make it into the next release 🤞

PR #2336, opened 2024-03-15
SW-5044: Console: Need a way to filter Deliverables list by Project

joshbuchea:


src/components/DeliverablesTable/index.tsx
```diff
@@ -41,34 +44,65 @@ const DeliverablesTable = ({
...
-            {
-              field: 'category',
-              // Same note as above
-              options: DeliverableCategories,
-              label: strings.CATEGORY,
-            },
-          ]
-        : [],
-    [activeLocale]
+  const getProjectName = useCallback(
```
nickgraz, line 54: Any chance you can throw this in the hook?

src/components/DeliverablesTable/index.tsx
```diff
@@ -41,34 +44,65 @@ const DeliverablesTable = ({
...
+        options: (projects || [])?.map((project: Project) => `${project.id}`),
+        searchNodeCreator: (values: (number | string | null)[]) => ({
+          field: 'projectId',
+          operation: 'field',
+          type: 'Exact',
+          values: values.map((value: number | string | null): string | null => (value === null ? value : `${value}`)),
+        }),
+        label: strings.PROJECTS,
+        renderOption: (id: string | number) => getProjectName(Number(id)),
+        notPresentFilterShown: true,
```
nickgraz, line 91: You can get rid of this and the next line, "no project" shouldn't be an option here.

src/components/DeliverablesTable/index.tsx
```diff
@@ -41,34 +44,65 @@ const DeliverablesTable = ({
...
+        notPresentFilterLabel: strings.NO_PROJECT,
+        pillValueRenderer: (values: unknown[]): string | undefined => {
+          if (values.length === 1 && values[0] === null) {
```
nickgraz, line 94: I think you can get rid of these 3 lines since there can be "no project"

src/components/DeliverablesTable/index.tsx
joshbuchea: Sure 👍

src/components/DeliverablesTable/index.tsx
```diff
@@ -41,33 +44,50 @@ const DeliverablesTable = ({
...
+        field: 'category',
+        // Same note as above
+        options: DeliverableCategories,
+        label: strings.CATEGORY,
+      },
+    ];
+
+    // show the project filter only in the accelerator route
+    // the participant view already has a projects filter above the table
+    if (isAcceleratorRoute) {
```
nickgraz, line 74: These components should be "view unaware". Can you pass this filter in as another prop?

PR #2349, opened 2024-03-19
SW-5085 Empty state for participants list view

karthikbtf:
- added empty state with button to create a participant
- added routes and empty components
- added TODOs for later
- did not add nav in this PR as it mostly involve the overview tab

Terraware App 2024-03-19 09-47-17


src/scenes/AcceleratorRouter/Participants/ParticipantsList.tsx
```diff
@@ -0,0 +1,78 @@
...
+  useEffect(() => {
+    // TODO: fetch participants
+    setParticipants([]);
+  }, []);
+
+  return (
+    <Card style={{ display: 'flex', flexDirection: 'column' }}>
+      <Box sx={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between' }}>
+        <Typography color={theme.palette.TwClrTxt} fontSize='20px' fontWeight={600} lineHeight='28px'>
+          {strings.PARTICIPANTS}
```
nickgraz, line 28: Need to check activeLocale here and in the `EmptyState` component

nickgraz (approved):
Looks good, a couple `activeLocale` checks needed but otherwise good to go

src/scenes/AcceleratorRouter/Participants/ParticipantsList.tsx
karthikbtf: ofc I had to forget this again 🤦

nickgraz: I wonder if we can just do this at the app level, it feels wrong to have to constantly sprinkle it everywhere. If there's no active locale, put up some sort of message. Obviously we'd still need it anywhere we are doing memoization but would at least make us not have to put it in places where the component would rerender if the locale changes anyways.

karthikbtf: This check is a side-effect (and maybe not needed in many places).
We need strings to change when the activeLocale changes, that's what we really need.

We have a bunch of useMemo or useCallbacks that return objects depending on string values (example. dropdown options, column names).
We wanted these to be re-evaluated when the activeLocale changed (user changes the language). 

Adding activeLocale as a dependency in useMemo/useCallback solved that problem..  but... introduced a compile error (unused activeLocale dependency) .. 
Hence the sprinkled check in a bunch of places  `if (!activeLocale) return []` to force a usage.

The other option is to move strings into a hook.
```
import useStrings from 'src/strings/hooks';

....
const strings = useStrings();
... {strings.LABEL}
```

This could totally solve our problem and we can remove activeLocale dependency from :all-the-things

nickgraz: Right, that's what I was talking about, where things are memoized it is needed. But with components that re-render, if the string is just within the JSX only, does it really need to be there if we just force the whole app to reload when `activeLocale` changes. I like the idea of a hook, just something to dry up the checks. Maybe when we get on the other side of this console work and are looking for something new it would be a good thing to check out.

karthikbtf: Got it, yes I agree these checks don't make sense for components.
I tested switching languages without the activeLocale check and the empty state strings did change. I can remove this check in a separate PR.
Yes, we can revisit this when we get more cycles for sure.

PR #2352, opened 2024-03-19
Update usage of TextTruncated to leverage wrapped component

karthikbtf:
- pre-determined string values with certain overrides


nickgraz (approved):
🙌

PR #2355, opened 2024-03-19
SW-5071 Remove drying reminder checkbox

alexTerraformation:
Any invalid date or blank date will clear the drying reminder.


src/scenes/AccessionsRouter/edit/EndDryingReminderModal.tsx
```diff
@@ -56,45 +53,28 @@ export default function EndDryingReminderModal(props: EndDryingReminderModalProp
...
-    const dateMs = new Date(dateValue).getTime();
-    const today = new Date(getTodaysDateFormatted()).getTime();
-
-    if (isNaN(dateMs)) {
+    // Invalid dates or blank entries will remove the reminder date
+    if (!newValue) {
       setDateError(strings.INVALID_DATE);
-      return false;
-    } else if (dateMs < today) {
+      return true;
```
karthikbtf, line 62: Remove the date error on line 61?

src/scenes/AccessionsRouter/edit/EndDryingReminderModal.tsx
```diff
@@ -56,45 +53,28 @@ export default function EndDryingReminderModal(props: EndDryingReminderModalProp
...
+    } else if (!isInTheFuture(newValue)) {
```
karthikbtf, line 63: Using today's date shows an error of 'No past dates'. Maybe something like 'Must be in the future'?

karthikbtf (approved):
Looks good to me, tested good.
A few comments/suggestions.

src/scenes/AccessionsRouter/edit/EndDryingReminderModal.tsx
alexTerraformation: I originally left it as a visual indicator that the date isn't formatted correctly, like if someone was trying to type in a date.  Do you think it would be better without?

karthikbtf: Got it, we need to show an error in that case.
Some thoughts, if it makes sense.
We could allow the user to clear the date and save, without showing an error. Essentially removing the drying reminder.
If the date is non-empty and invalid, we could show the error and prevent save.

alexTerraformation: That was originally what I wanted to do, but the MUI datepicker version that we use doesn't provide easy access to the raw input data.  If/when we update MUI to 6.x there are a lot of improvements to datepicker that we could utilize.
I think in the meantime this isn't a bad compromise considering it's a corner case.

karthikbtf (approved):

PR #2357, opened 2024-03-19
Redux to create participant and list accelerator orgs 

karthikbtf:
- added service and redux
- made assumptions with data model, to be synced with BE later


nickgraz (approved):
Looks good, can't believe I missed "create" 🤦‍♂️

PR #2363, opened 2024-03-20
SW-5086 Add participant workflow with mock data

karthikbtf:
- added 'Add Participant' action menu in participants list (for non-empty state)
- added hooks to fetch cohorts and accelerator orgs
- added ParticipantForm (leveraged CohortForm as a template)
- implemented the sections for org/projects selection
- plumbed save back to ParticipantsNew component
- ParticipantForm can be reused for edit Participant
- TODO: sync with BE types when ready (follow up)
- TODO: add permission checks in the UI (follow up)

Terraware App 2024-03-20 11-29-58

### DIFF
--- a/src/scenes/AcceleratorRouter/Participants/ParticipantForm.tsx
+++ b/src/scenes/AcceleratorRouter/Participants/ParticipantForm.tsx
@@ -1,0 +1,204 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+
+import { Box, Grid, useTheme } from '@mui/material';
+import { Dropdown, Textfield } from '@terraware/web-components';
+
+import AddLink from 'src/components/common/AddLink';
+import Card from 'src/components/common/Card';
+import PageForm from 'src/components/common/PageForm';
+import { useAcceleratorOrgs } from 'src/hooks/useAcceleratorOrgs';
+import { useCohorts } from 'src/hooks/useCohorts';
+import strings from 'src/strings';
+import { AcceleratorOrg } from 'src/types/Accelerator';
+import { Cohort } from 'src/types/Cohort';
+import { ParticipantCreateRequest, ParticipantUpdateRequest } from 'src/types/Participant';
+import useDeviceInfo from 'src/utils/useDeviceInfo';
+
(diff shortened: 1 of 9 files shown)


src/scenes/AcceleratorRouter/Participants/ParticipantForm.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  const onOrgSelect = useCallback(
+    (sectionId: number, orgId: number) => {
+      const org = acceleratorOrgs?.find((o) => o.id === orgId);
+      if (!org) {
+        return;
+      }
+      setOrgProjectsSections((prev) => {
+        const updated = [...prev];
+        updated[sectionId - 1].org = org;
```
nickgraz, line 92: I get why this works, but is a bit tricky. I guess the alternative is either passing the index through (not great) or `[...prev].map(_return original if ID is different, otherwise update_)`. Anyways, not a big deal, but perhaps a comment as to why the id is being used as an index might be helpful.

nickgraz (approved):
Looks good, one small comment but no changes needed.

src/scenes/AcceleratorRouter/Participants/ParticipantForm.tsx
karthikbtf: I can use the index approach and remove 'id' semantics since it might be easier to read.
I didn't get the `[...prev].map(...` suggestion, can you clarify? If that's better I'd be up for using it.

nickgraz:
```
const updated = [...prev].map((section) => section.id !== sectionId 
  ? section 
  : {
    ...section,
    org,
    selectedProjectIds: []
  }
```

karthikbtf: Ah ok, instead of an index access, thanks. I will use this and add a comment on the 'id'. Realized using an index will remove the ability to use the 'useCallback' efficiently.

karthikbtf: Thanks for the suggestion, definitely cleaner and hides the index into array semantics.

PR #2366, opened 2024-03-20
Update timestamp in Phase Score header

nickgraz:
- Update Score Phase header to use ISO date string

### DIFF
--- a/src/scenes/AcceleratorRouter/Scoring/PhaseScores/LastUpdated.tsx
+++ b/src/scenes/AcceleratorRouter/Scoring/PhaseScores/LastUpdated.tsx
@@ -1,10 +1,10 @@
 import React, { useMemo } from 'react';
 
 import { Typography } from '@mui/material';
+import { getDateDisplayValue } from '@terraware/web-components/utils';
 
 import strings from 'src/strings';
 import { PhaseScores, Score } from 'src/types/Score';
-import { getISODate } from 'src/utils/dateFormatter';
 
 interface LastUpdatedProps {
   phaseScores: PhaseScores;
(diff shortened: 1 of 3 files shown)


nickgraz:
* **#2366** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

karthikbtf:
@nickgraz should we be using existing date formats throughout the app? Maybe worth asking design.
We currently use '-' instead of '/' as the separator.

karthikbtf (approved):
Looks good but maybe we should use the same date format as in other parts of the app.
https://github.com/terraware/web-components/blob/main/src/utils/index.ts#L7

nickgraz:
@karthikbtf I asked [in Figma](https://www.figma.com/file/Ec5QnRsnzqEKeifzwPKR3k?node-id=1667:34404&mode=design#742770074) about this, I think for now we will use the numerical representation for the locale, which seems correct. I will ask about the rest of the app, I agree, if we are using ISO as standard everywhere else we should probably use that here also to avoid confusion.

karthikbtf:
> @karthikbtf I asked [in Figma](https://www.figma.com/file/Ec5QnRsnzqEKeifzwPKR3k?node-id=1667:34404&mode=design#742770074) about this, I think for now we will use the numerical representation for the locale, which seems correct. I will ask about the rest of the app, I agree, if we are using ISO as standard everywhere else we should probably use that here also to avoid confusion.

Sounds good @nickgraz 👍 let's go with this for now.

src/scenes/AcceleratorRouter/Scoring/PhaseScores/LastUpdated.tsx
```diff
@@ -2,18 +2,15 @@ import React, { useMemo } from 'react';
 
 import { Typography } from '@mui/material';
 
-import { useLocalization } from 'src/providers';
 import strings from 'src/strings';
 import { PhaseScores, Score } from 'src/types/Score';
-import { getLongDate } from 'src/utils/dateFormatter';
+import { getISODate } from 'src/utils/dateFormatter';
```
karthikbtf, line 7: Use `import { getDateDisplayValue } from '@terraware/web-components/utils';`?
Sorta related, plan is to move away from moment (sunset) and into luxon. web-components has both implementations just need to make the switch at some point.

karthikbtf (commented):
Sorry about the back and forth on formats.

karthikbtf (approved):
🙏🏻

PR #2371, opened 2024-03-21
SW-5080 Edit participant view with mock data

karthikbtf:
- implemented ParticipantEdit using ParticipantNew as a template
- reused ParticipantForm
- updated ParticipantForm to pre-populate orgs/projects based on initialized data
- misc refactor

Terraware App 2024-03-20 17-21-54

Terraware App 2024-03-20 17-22-32


nickgraz (approved):
⚡

src/scenes/AcceleratorRouter/Participants/useNavigateToParticipants.tsx
```diff
@@ -0,0 +1,17 @@
+import { useCallback } from 'react';
+import { useHistory } from 'react-router-dom';
+
+import { APP_PATHS } from 'src/constants';
+
+export type Response = () => void;
+
+export default function useNavigateToParticipants(): Response {
```
nickgraz, line 8: This is nice, maybe moving forward we should wrap all the relevant routes for an area in a hook like this so we can easily sprinkle them around

karthikbtf: That's a nice idea, we could have a library of navigation hooks.. same for table cell renderers. I will create a tech investment ticket.

karthikbtf: Moved the useNavigateTo.. under `src/hooks/navigation` so we can add to that directory.

nickgraz (approved):

src/scenes/AcceleratorRouter/Participants/useNavigateToParticipants.tsx
karthikbtf: Created https://terraformation.atlassian.net/browse/SW-5104

PR #2373, opened 2024-03-21

joshbuchea:
This PR includes changes to update the projects filter options when the participant filter is changed.

### Demo

https://github.com/terraware/terraware-web/assets/1474361/e0cf1851-14a2-482b-be02-60912597a425

### Outstanding issue: Featured filters are not reset when participants dropdown changes

https://github.com/terraware/terraware-web/assets/1474361/35eafe3e-c9a5-4d54-a6cc-6d7172269329


joshbuchea:
This seemed to work to update the filter options when the participant filter is changed, but I think we need to reset the featured filters when the participant filter is changed and I was having trouble working out how to reset the featured filters.

src/components/DeliverablesTable/index.tsx
```diff
@@ -40,17 +43,28 @@ const DeliverablesTable = ({
...
   const dispatch = useAppDispatch();
   const { activeLocale } = useLocalization();
   const { availableProjects: projects, getProjectName } = useProjects();
+  const { selectedParticipant } = useParticipants(participantId);
 
   const [deliverables, setDeliverables] = useState<ListDeliverablesElement[]>([]);
   const [deliverablesSearchRequestId, setDeliverablesSearchRequestId] = useState('');
   const deliverablesSearchRequest = useAppSelector(selectDeliverablesSearchRequest(deliverablesSearchRequestId));
 
+  const getFilterProjectName = useCallback(
```
nickgraz, line 59: Can you throw this into the `useParticipants` hook?

joshbuchea: You got it 👍

joshbuchea: @graz just moved that over, it depended on `selectedParticipant` and `getProjectName` so I opted to pass them both in. Not sure if that's what you were thinking?

nickgraz: I sent you down the wrong path, since we are essentially combining two hooks, I think it was cleaner the way you originally had it. Sorry for the run around!

joshbuchea: No worries 😄

nickgraz (approved):
👍

joshbuchea:
@graz could you take a look at the second example video I added in the description when you have a moment? It seems like the featured filters need to be reset whenever the participant filter is changed, but I'm having trouble working out exactly how to do that.

src/components/TableWithSearchFilters/index.tsx
```diff
@@ -111,6 +111,21 @@ const TableWithSearchFilters = (props: TableWithSearchFiltersProps) => {
...
+        const filter = {
+          ...newFilters[key],
+          values: [],
+        };
+        newFilters[key] = filter;
+      });
+      return newFilters;
+    });
+  }, [extraTableFilters]);
+
```
joshbuchea, line 128: This works and I think we should be resetting the filters any time the extra filters change anyways, but let me know if this is problematic for any reason.

joshbuchea:
Checking why the deployment failed now 🔎

src/components/TableWithSearchFilters/index.tsx
nickgraz: Looks good to me! Nice work!

nickgraz (approved):

PR #2389, opened 2024-03-21
SW-5110 Update accelerator org service and components to use new types

karthikbtf:
- updated with BE PR branch 'accel-org'
- org role property in org payload is now optional, updated code to handle that
- updated accelerator org service/redux and participant form to use updated properties
- this will need to go in after BE PR is merged

screenshot with real data

Terraware App 2024-03-21 16-11-57


karthikbtf:
BE PR was merged, opening this for review.

nickgraz (approved):
Looks good

PR #2392, opened 2024-03-22
SW-4990: Add new columns as needed

joshbuchea:
This PR includes changes to add new column(s) to the Cohorts list view.

### DIFF
--- a/src/scenes/AcceleratorRouter/Cohorts/CohortCellRenderer.tsx
+++ b/src/scenes/AcceleratorRouter/Cohorts/CohortCellRenderer.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import Link from 'src/components/common/Link';
+import TextTruncated from 'src/components/common/TextTruncated';
 import CellRenderer, { TableRowType } from 'src/components/common/table/TableCellRenderer';
 import { RendererProps } from 'src/components/common/table/types';
 import { APP_PATHS } from 'src/constants';
@@ -17,6 +18,10 @@ export default function CohortCellRenderer(props: RendererProps<TableRowType>): 
         row={row}
       />
     );
+  }
+
+  if (column.key === 'participantIds') {
+    return <CellRenderer {...props} value={<TextTruncated stringList={(value as string[]) || []} />} />;
   }
 
   return <CellRenderer {...props} />;
--- a/src/scenes/AcceleratorRouter/Cohorts/CohortListView.tsx
+++ b/src/scenes/AcceleratorRouter/Cohorts/CohortListView.tsx
@@ -31,6 +31,11 @@ const columns = (activeLocale: string | null): TableColumnType[] =>
           name: strings.PHASE,
           type: 'string',
         },
+        {
+          key: 'participantIds',
+          name: strings.PARTICIPANTS,
+          type: 'string',
+        },
       ]
     : [];
 


src/scenes/AcceleratorRouter/Cohorts/CohortCellRenderer.tsx
```diff
@@ -19,5 +20,9 @@ export default function CohortCellRenderer(props: RendererProps<TableRowType>):
     );
   }
 
+  if (column.key === 'participantIds') {
+    return <CellRenderer {...props} value={<TextTruncated stringList={(value as string[]) || []} />} />;
+  }
+
```
joshbuchea, line 26: I'm sure we want to show participant names here rather than participant ids, just did this as a placeholder start for now.

joshbuchea: Turns out in the designs we're showing a count rather than a list of names:

Screenshot 2024-03-26 at 11 24 06 AM

src/scenes/AcceleratorRouter/Cohorts/CohortListView.tsx
```diff
@@ -25,6 +25,11 @@ const columns = (activeLocale: string | null): TableColumnType[] =>
           name: strings.PHASE,
           type: 'string',
         },
+        {
```
karthikbtf, line 28: The `columns` passed in from this component is unused. We could remove `columns` from here.

karthikbtf (approved):
Looks good. One comment about unused `columns`.

src/scenes/AcceleratorRouter/Cohorts/CohortsTable.tsx
```diff
@@ -43,6 +43,11 @@ const columns = (activeLocale: string | null): TableColumnType[] =>
           name: strings.PHASE,
           type: 'string',
         },
+        {
+          key: 'participantIds',
```
karthikbtf, line 47: Looks like the BE search will return a `numParticipants` count. You can use that, might simplify the rendering too.

src/scenes/AcceleratorRouter/Cohorts/CohortListView.tsx
joshbuchea: Oh it's pretty obvious too 😅 Thanks for catching 🙌

src/scenes/AcceleratorRouter/Cohorts/CohortsTable.tsx
joshbuchea: Didn't realize that, thanks for mentioning. Will update.

joshbuchea: I just checked the `cohorts` data with the `depth` param set to `'Participant'` for `requestCohorts` and the `numParticipants` property doesn't seem to be available. Did a quick search and wasn't able to find it in the API schema either.

karthikbtf: It is in an upcoming PR. It will be part of the search response, see `SearchResponsePayload` in the generated-types (it is very generic).
We can make the change later for sure.

karthikbtf (approved):
Nice 👍

PR #2410, opened 2024-03-26
SW-5117 Redux and service functions to search/download participant projects list

karthikbtf:
- created service functions and redux
- assumed data model for now, did not merge with existing data model


src/services/ParticipantProjectService.ts
```diff
@@ -60,6 +66,78 @@ const get = async (participantProjectId: number): Promise<Response2<ParticipantP
...
 
+const downloadList = async (search?: SearchNodePayload, sortOrder?: SearchSortOrder): Promise<string | null> => {
+  return (
+    'Project Name,Participant,Cohort,Phase,Country,Region,Restorable Land,Land Use Model Type\r' +
+    'Andromeda,Cartwheel,Cohort 1,Phase 0 - Due Diligence,Ghana,Sub-Saharan Africa,500,Native Forest'
+  );
+};
+
+const list = async (
+  // TODO: remove locale if we are using BE search API
```
tommylau523, line 77: For this to use BE search API we need to add CohortsSearchTable and ParticipantsSearchTable on the BE.

src/types/ParticipantProject.ts
```diff
@@ -30,3 +31,16 @@ export type ParticipantProject = {
   votingDecision: VoteOption;
   whatNeedsToBeTrue: string;
 };
+
+export type ParticipantProjectSearchResult = {
+  cohortId: number;
+  cohortName: string;
+  country: string;
+  id: number;
+  landUseModelType: string;
```
tommylau523, line 40: LandUseModel type is multi-select so string[]

tommylau523 (approved):

src/types/ParticipantProject.ts
karthikbtf: Thanks for the info, yes we can change this later as needed.
Can a participant project have more than one value for its land use type? If so, we need to change how it is rendered on the table.

tommylau523: Yes it is a multi-select in the requirements: 

-Native Forest
-Monoculture
-Sustainable Timber
-Other Timber
-Mangroves
-Agroforestry
-Silvopasture
-Other Land-Use Model

Figma designs still need to be updated to reflect this I think.

karthikbtf: Got it, thanks @tommylau523 ... I can update the table later once the BE types are available.

PR #2413, opened 2024-03-26
SW-5116 Participant Projects Table with results from Search API

karthikbtf:
- added table with mock data
- had to fix FeaturedFilters.tsx which was setting all search nodes to use the last set field

Terraware App 2024-03-26 12-52-25

Terraware App 2024-03-26 12-54-32

Terraware App 2024-03-26 12-54-10

Terraware App 2024-03-26 11-21-16

Terraware App 2024-03-26 12-52-47

Terraware App 2024-03-27 10-44-53

### DIFF
--- a/src/scenes/AcceleratorRouter/ParticipantProjects/ListView.tsx
+++ b/src/scenes/AcceleratorRouter/ParticipantProjects/ListView.tsx
@@ -1,0 +1,180 @@
+import { useCallback, useEffect, useMemo, useState } from 'react';
+
+import { TableColumnType } from '@terraware/web-components';
+
+import TableWithSearchFilters from 'src/components/TableWithSearchFilters';
+import { FilterConfig } from 'src/components/common/SearchFiltersWrapperV2';
+import { useLocalization } from 'src/providers';
+import { requestListParticipantProjects } from 'src/redux/features/participantProjects/participantProjectsAsyncThunks';
+import { selectParticipantProjectsListRequest } from 'src/redux/features/participantProjects/participantProjectsSelectors';
+import { useAppDispatch, useAppSelector } from 'src/redux/store';
+import strings from 'src/strings';
+import { getPhaseNumber } from 'src/types/Cohort';
+import { ParticipantProjectSearchResult } from 'src/types/ParticipantProject';
+import { SearchNodePayload, SearchSortOrder } from 'src/types/Search';
+import useSnackbar from 'src/utils/useSnackbar';
+
+import CellRenderer from './CellRenderer';
+
+const columns = (activeLocale: string | null): TableColumnType[] =>
+  activeLocale
+    ? [
+        {
+          key: 'name',
+          name: strings.PROJECT,
+          type: 'string',
+        },
+        {
+          key: 'participantName',
+          name: strings.PARTICIPANT,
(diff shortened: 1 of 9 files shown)


src/types/Cohort.ts
```diff
@@ -12,3 +12,11 @@ export const CohortPhases: CohortPhaseType[] = [
   'Phase 2 - Plan and Scale',
   'Phase 3 - Implement and Monitor',
 ];
+
+/**
+ * Get string value of phase number.
+ */
+export const getPhaseNumber = (phase: CohortPhaseType): string => {
```
karthikbtf, line 19: The table renders phases as numbers instead of the full phase name. I have a question on figma suggesting we show the full phase (like we do on the cohorts list).

karthikbtf:
Updated to use BE search response.

joshbuchea (commented):
This looks great, thanks for alphabetizing object props 🙌 💯

One comment/question. @nickgraz mentioned to me in a recent PR review that `useStyles` was deprecated in the MUI library and that aside from a few exceptions, we should be setting styles using the `sx` or `style` props. Was trying to find the comment to confirm I got the details correct, but couldn't find it with a quick check.

karthikbtf:
> This looks great, thanks for alphabetizing object props 🙌 💯
> 
> One comment/question. @nickgraz mentioned to me in a recent PR review that `useStyles` was deprecated in the MUI library and that aside from a few exceptions, we should be setting styles using the `sx` or `style` props. Was trying to find the comment to confirm I got the details correct, but couldn't find it with a quick check.

That is correct, makeStyles is deprecated, we use `sx` or `style` props on MUI components. We use makeStyles as a last resort. 
We also have some existing web-components that accept `className` properties, these are not MUI components but wrappers around MUI, we haven't had time to rewrite our web-components to accept `sx` props as a pass through. That's probably what you see in this PR (the cell renderer code, which also happens to be copy/pasted, sigh, from other parts of our codebase).
I have one more small change to this PR coming up soon.

joshbuchea:
That makes sense, thank you for the context.

karthikbtf:
@joshbuchea can you take a look at this again when you get a chance?

src/scenes/AcceleratorRouter/ParticipantProjects/ListView.tsx
```diff
@@ -0,0 +1,154 @@
...
+        dispatchSearchRequest={dispatchSearchRequest}
+        featuredFilters={featuredFilters}
+        fuzzySearchColumns={fuzzySearchColumns}
+        id='accelerator-participan-projects-table'
+        Renderer={CellRenderer}
+        rows={projects}
+        title={strings.PROJECTS}
+      />
+    </>
+  );
```
joshbuchea, line 153: Is the fragment wrapper necessary here?

joshbuchea (approved):
Nice!

src/scenes/AcceleratorRouter/ParticipantProjects/ListView.tsx
karthikbtf: Good catch, not needed, there were other elements in there but I moved them into the Table component. Will remove it.

PR #2414, opened 2024-03-26
SW-5115 Export Participant Projects from the list view

karthikbtf:
- added export hooks
- refactored code to skip creating an export wrapper

Terraware App 2024-03-26 14-27-58


joshbuchea (approved):
Looks good

PR #2416, opened 2024-03-26
SW-4981: Move cohorts from nav into tabbed view

joshbuchea:
This PR moves the cohorts list into the tabbed views within the overview view.

### Before

staging terraware io_accelerator_cohorts (1)

### After

localhost_accelerator_overview_tab=cohorts


src/scenes/AcceleratorRouter/Overview/index.tsx
```diff
@@ -6,15 +6,14 @@ import { makeStyles } from '@mui/styles';
...
-import Button from 'src/components/common/button/Button';
-import { APP_PATHS } from 'src/constants';
 import { useLocalization, useUser } from 'src/providers';
 import ParticipantsList from 'src/scenes/AcceleratorRouter/Participants/ParticipantsList';
 import strings from 'src/strings';
-import useDeviceInfo from 'src/utils/useDeviceInfo';
 import useQuery from 'src/utils/useQuery';
 import useStateLocation, { getLocation } from 'src/utils/useStateLocation';
 
+import CohortsListView from '../Cohorts/CohortsListView';
```
karthikbtf, line 15: Use absolute path for directories not within this file's hierarchy?

src/scenes/AcceleratorRouter/Cohorts/CohortsListView.tsx
```diff
@@ -96,42 +118,46 @@ const CohortsTable = ({ filterModifiers, extraTableFilters }: CohortTableProps)
...
       Renderer={CohortCellRenderer}
+      rightComponent={actionMenus}
       rows={cohorts || []}
+      title={strings.COHORTS}
     />
   );
 };
 
-export default CohortsTable;
+export default CohortsListView;
```
karthikbtf, line 128: Move to end of file? I don't know if there's a preference but having one place for non-implicit exports (either beginning or ending of file). We tend to use the end of file.

src/scenes/AcceleratorRouter/Cohorts/CohortsListView.tsx
```diff
@@ -96,42 +118,46 @@ const CohortsTable = ({ filterModifiers, extraTableFilters }: CohortTableProps)
       fuzzySearchColumns={fuzzySearchColumns}
       id='cohortsTable'
       Renderer={CohortCellRenderer}
+      rightComponent={actionMenus}
       rows={cohorts || []}
+      title={strings.COHORTS}
```
karthikbtf, line 123: Nice, thanks. You could optionally use the `busy= prop` if your selector has a status field (if cohorts request uses an async thunks). Otherwise no worries.

src/scenes/AcceleratorRouter/Cohorts/CohortsListView.tsx
```diff
@@ -46,12 +46,15 @@ const columns = (activeLocale: string | null): TableColumnType[] =>
...
-const CohortsTable = ({ filterModifiers, extraTableFilters }: CohortTableProps) => {
+const CohortsListView = ({ filterModifiers, extraTableFilters }: CohortsListViewProps) => {
   const dispatch = useAppDispatch();
   const { activeLocale } = useLocalization();
   const history = useHistory();
+  const { isAllowed } = useUser();
+  const { isMobile } = useDeviceInfo();
 
   const cohorts = useAppSelector(selectCohorts);
+  const isEmptyState = !cohorts?.length;
```
karthikbtf, line 57: We probably don't want to flash the empty state while the search is happening.
Also don't want to show the empty state if search or filters ends up returning empty results.
Suggestion:
`const isEmptyState = cohorts?.length === 0 && noSearchAndNoFilters;`
There should be a way to find out if there are search params. The Participants list view does something like this, if you'd like a reference.

karthikbtf (commented):
Nice. A few comments.

src/scenes/AcceleratorRouter/Overview/index.tsx
joshbuchea: Oops missed this one, thanks!

src/scenes/AcceleratorRouter/Cohorts/CohortsListView.tsx
joshbuchea: I copied this pattern from another view somewhere, but tend to agree that's it's good to have one preferred place. Will move.

src/scenes/AcceleratorRouter/Cohorts/CohortsListView.tsx
joshbuchea: Good call 👍

karthikbtf (approved):
Thanks for the changes @joshbuchea looks great!

joshbuchea:
@karthikbtf thanks for your help cleaning it up and pushing it through 🙌

PR #2421, opened 2024-03-27

joshbuchea:
This PR includes changes to the projects filter dropdown options in the Deliverables list.

I'm trying to get the correct data set for the projects filter dropdown, but I'm not 100% confident that what I have is correct. I'm using the new BE API that's available to get a list of accelerator organizations and their projects, so I believe the data I'm providing as options to the projects filter dropdown is a list of accelerator projects, but there are several projects that are not associated with any deliverables, including "Alex's Super Secret Project".

Also, in a previous PR for this ticket I made changes that when the participant filter dropdown is applied, the projects filter dropdown options are updated to only show projects associated with that participant. This feature was previously requested from Leanne in Slack, so I'm not entirely sure whether it should be left in-place or if the changes for this should be reverted.

So it'd be wonderful if someone could help me confirm the project options are correct and make sense 🙏  If it isn't correct I'm happy to make any changes, as I'm aiming to get this one done right rather than done quickly.

## Examples

Screenshot 2024-03-27 at 12 33 13 PM

Screenshot 2024-03-27 at 12 34 28 PM


src/providers/OrganizationProvider.tsx
```diff
@@ -32,6 +32,7 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
   const [orgPreferenceForId, setOrgPreferenceForId] = useState<number>(defaultSelectedOrg.id);
   const [orgAPIRequestStatus, setOrgAPIRequestStatus] = useState<APIRequestStatus>(APIRequestStatus.AWAITING);
   const [organizations, setOrganizations] = useState<Organization[]>();
+  const [acceleratorOrganizations, setAcceleratorOrganizations] = useState<AcceleratorOrganization[]>();
```
joshbuchea, line 35: It seemed like this provider might be the best place for this, but I'm happy to move it if we'd prefer to.

src/providers/OrganizationProvider.tsx
```diff
@@ -66,6 +67,17 @@ export default function OrganizationProvider({ children }: OrganizationProviderP
...
+  const reloadAcceleratorOrganizations = useCallback(async () => {
+    const populateAcceleratorOrganizations = async () => {
+      const response = await OrganizationService.getAcceleratorOrganizations();
+      if (!response.error) {
+        setAcceleratorOrganizations(response.organizations);
+      }
+    };
+
+    await populateAcceleratorOrganizations();
+  }, []);
```
joshbuchea, line 79: I didn't add the ability to tracking the request status or pass a projectId and select an accelerator project because I wasn't sure if we needed it, but I can add it if we want either.

src/services/OrganizationService.ts
```diff
@@ -54,10 +69,36 @@ type UpdateOrganizationRequestPayload =
...
+    if (response.statusCode === 401) {
+      response.error = 'NotAuthenticated';
+    } else {
+      response.error = 'GenericError';
+    }
+  }
+
+  return response;
+};
+
```
joshbuchea, line 101: Same comment regarding location, it seemed like the best place but I can move it no problem.

karthikbtf: @joshbuchea the accelerator orgs code already exists, this was needed for participant creation as well.
See PR https://github.com/terraware/terraware-web/pull/2389/files for types/service/redux

src/components/DeliverablesTable/index.tsx
```diff
@@ -50,10 +49,15 @@ const DeliverablesTable = ({
   const dispatch = useAppDispatch();
   const { activeLocale } = useLocalization();
   const { isAllowed } = useUser();
-  const { selectedOrganization } = useOrganization();
-  const { availableProjects: projects, getProjectName } = useProjects();
+  const { acceleratorOrganizations, selectedOrganization } = useOrganization();
   const { selectedParticipant } = useParticipants(participantId);
 
+  const acceleratorProjects = useMemo(() => {
+    return acceleratorOrganizations.reduce((acc: AcceleratorProject[], org: AcceleratorOrganization) => {
```
karthikbtf, line 56: this can be achieved with a flatMap  `acceleratorOrganizations.flatMap(org => org.projects)`, will return a flat list of all the project ids

karthikbtf (commented):
@joshbuchea the accelerator org code was added by me yesterday. The redux/services with mock data was done sometime last week.
If we can reuse that it will be great. We can avoid passing accelerator orgs the same way we do regular orgs, which is mostly for non-accelerator-console context.

karthikbtf:
> @joshbuchea the accelerator org code was added by me yesterday. The redux/services with mock data was done sometime last week. If we can reuse that it will be great. We can avoid passing accelerator orgs the same way we do regular orgs, which is mostly for non-accelerator-console context.

There is a `useAcceleratorOrgs` hook that should be easy to use.

joshbuchea:
Comments are great, thanks for the direction! Will make the requested changes 👍

src/components/DeliverablesTable/index.tsx
```diff
@@ -51,9 +51,13 @@ const DeliverablesTable = ({
   const { activeLocale } = useLocalization();
   const { isAllowed } = useUser();
   const { selectedOrganization } = useOrganization();
-  const { availableProjects: projects, getProjectName } = useProjects();
+  const { acceleratorOrgs } = useAcceleratorOrgs(true);
   const { selectedParticipant } = useParticipants(participantId);
 
+  const acceleratorProjects = useMemo(() => {
```
karthikbtf, line 57: nit: since this is a single line
```
const acceleratorProjects = useMemo(() =>  acceleratorOrgs?.flatMap((org) => org.projects), [acceleratorOrgs]);
```

src/components/DeliverablesTable/index.tsx
```diff
@@ -64,9 +68,9 @@ const DeliverablesTable = ({
     (projectId: number | string) => {
       return participantId
         ? selectedParticipant?.projects?.find((p) => p.id === Number(projectId))?.name || ''
-        : getProjectName(Number(projectId));
+        : acceleratorProjects?.find((p) => p.id === Number(projectId))?.name || '';
```
karthikbtf, line 71: nit: wrap the ternary with an ` || ''`
```
  return (participantId ? selectedPart... : acceleratorPro...) || '';
```

karthikbtf (approved):
Thanks for the changes 🙏 wow this is a much simpler change.
A few comments but looks good to merge. Regarding solution, if you want to wait for Nick's thoughts that's totally fine.

joshbuchea:
Nits are excellent suggestions, updated.

joshbuchea:
The set of projects is compiled from `acceleratorOrgs` from the `useAcceleratorOrgs` hook. @nickgraz does this seem like the correct set of data for the projects filter on this view?

nickgraz:
> The set of projects is compiled from `acceleratorOrgs` from the `useAcceleratorOrgs` hook. @nickgraz does this seem like the correct set of data for the projects filter on this view?

I believe that is correct. I had to make a similar call on my end earlier, it seems like they would only want to filter on projects that _could_ have projects, IE "accelerator projects", so I think you made the right call.

joshbuchea:
Ok awesome, thanks for the quick review/response! 🙌 💯

nickgraz (approved):

PR #2445, opened 2024-04-02

joshbuchea:
Fast follow-up to limit projects filter options to projects represented in the table data.


nickgraz (approved):
🚀

PR #2447, opened 2024-04-02
SW-5142 Parsed search terms to use PhraseMatch when double-quoted

tommylau523:
Every instance we have a searchbox, where user can specify an input. We will try to `removeDoubleQuotes` and use PhraseMatching. Most file changed here are just wholesale changes to implement this behavior.

### DIFF
--- a/src/components/ProjectNewView/flow/SelectBatches.tsx
+++ b/src/components/ProjectNewView/flow/SelectBatches.tsx
@@ -22,7 +22,7 @@ import strings from 'src/strings';
 import { CreateProjectRequest } from 'src/types/Project';
 import { FieldNodePayload, SearchNodePayload, SearchSortOrder } from 'src/types/Search';
 import { getAllNurseries } from 'src/utils/organization';
-import { removeDoubleQuotes } from 'src/utils/search';
+import { parseSearchTerm } from 'src/utils/search';
 
 import { EntitySpecificFilterConfig } from './ProjectEntityFilter';
 
@@ -131,13 +131,13 @@ export default function SelectBatches(props: SelectBatchesProps): JSX.Element | 
   );
 
   const getSearchFields = useCallback((debouncedSearchTerm: string): FieldNodePayload[] => {
-    const phraseMatchQuery = removeDoubleQuotes(debouncedSearchTerm);
+    const { type, values } = parseSearchTerm(debouncedSearchTerm);
     return [
       {
         operation: 'field',
         field: 'batchNumber',
-        type: phraseMatchQuery ? 'PhraseMatch' : 'Fuzzy',
-        values: [phraseMatchQuery || debouncedSearchTerm],
+        type,
+        values,
       },
     ];
   }, []);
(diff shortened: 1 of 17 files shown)


tommylau523:
* **#2447** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2447?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/services/SearchService.ts
```diff
@@ -73,7 +78,8 @@ function convertToSearchNodePayload(
 
 async function search<T extends SearchResponseElement>(entity: RawSearchRequestPayload): Promise<T[] | null> {
   try {
-    const response: SearchResponsePayload = (await httpSearch.post({ entity })).data;
+    const newEntity = entity.search ? {...entity, search: convertToPhraseMatch(entity.search)} : entity
```
tommylau523, line 81: We convert the search before calling the API, but the correct approach is likely calling it upstream, where the query term is parsed.

tommylau523: This gives us fine-grained control on which search boxes need to support PhraseMatch. For example, weights do not support Phrase Search and we don't want to convert that.

src/services/SearchService.ts
```diff
@@ -104,6 +111,27 @@ async function searchCsv(entity: RawSearchRequestPayload): Promise<any> {
   }
 }
 
+function convertToPhraseMatch(searchNode: SearchNodePayload): SearchNodePayload {
```
tommylau523, line 114: But I think it is worth reviewing this function. Likely need to export and add tests as well.

src/components/ProjectNewView/flow/SelectBatches.tsx
```diff
@@ -129,17 +130,17 @@ export default function SelectBatches(props: SelectBatchesProps): JSX.Element |
...
+  const getSearchFields = useCallback((debouncedSearchTerm: string): FieldNodePayload[] => {
+    const phraseMatchQuery = removeDoubleQuotes(debouncedSearchTerm);
+    return [
       {
         operation: 'field',
         field: 'batchNumber',
-        type: 'Fuzzy',
-        values: [debouncedSearchTerm],
+        type: phraseMatchQuery ? 'PhraseMatch' : 'Fuzzy',
+        values: [phraseMatchQuery || debouncedSearchTerm],
```
karthikbtf, line 140: Suggestion: create a util that takes the search term and type(s) as input and returns a `{ type, values }`
Util can do the remove double quotes, and return the type and values.
We could in other places import the util and spread the result of using it.
Number of changes will be about the same but logic will be centralized.

karthikbtf (commented):
Nice change, one suggestion. Apologies for the suggestion being just as wide of a change but curious to hear your thoughts.

tommylau523:
> Nice change, one suggestion. Apologies for the suggestion being just as wide of a change but curious to hear your thoughts.
> 
> Suggestion: create a util that takes the search term and type(s) as input and returns a `{ type, values }`
> Util can do the remove double quotes, and return the type and values.
> We could in other places import the util and spread the result of using it.
> Number of changes will be about the same but logic will be centralized.
> 
> Or maybe have a util that takes in the entire param payload + search term and returns the adjusted payload.

I think it is a great idea. I first started with parsing the entire param payload and calling it under SearchService but I think there will be edge cases we miss here. I opted for the first suggestion with `{type, values}`.

karthikbtf (approved):
Nice, thanks for the changes!

PR #2449, opened 2024-04-03
SW-5145 Fix participants edit crash

karthikbtf:
- If the list of unassociated accelerator org/projects are [{org: ABC, projects: [DEF]}]
- and
- If the participant had an org/project association [{org: ABC, projects: [XYZ]}, {org: ...}]
- we merged the ABC org's projects list to contain both DEF and XYZ (for the multiselect)
- The projects array from redux is immutable, hence the crash (and also not wise to edit redux store data)
- Updated code to make a copy instead.

Original error:
```
 Cannot add property 1, object is not extensible
    at Array.push (<anonymous>)
    at ParticipantForm.tsx:65:29
```


nickgraz (approved):
Nice fix, these ones are always tricky

karthikbtf:
> Nice fix, these ones are always tricky

Thanks, yes tricky one. I caused/fixed a similar bug a few months back. Need to avoid editing redux store data.

PR #2451, opened 2024-04-03
SW-5144 Fix initial value for accession remaining quantity changed description text

karthikbtf:
- description should not be initialized to any existing value, should be empty by default
- intention is to capture 'quantity changed' description when quantity changes
- was erroneously initialized to the 'notes' values (plant description), result of copy/pasta, leading to some ux issues
- edit/save and history values still work as expected


constanzauanini (approved):
Good catch

PR #2452, opened 2024-04-03
SW-5016 Updates to Accessions field names

karthikbtf:
- to be consistent across mobile/web apps
- changes as per Jira ticket https://terraformation.atlassian.net/browse/SW-5016
- there was one change to increase grid width to allow larger label


joshbuchea:
Love to see UX improvements like these! 💚

PR #2456, opened 2024-04-03
SW-5159 Fix participant project view redirecting back to participant project list even when there are votes

karthikbtf:
- there was a race condition where the project id is -1 and then a valid value, the voting provider fetches votes for project -1 and fails on subsequent views
- On first visit vote request for -1 and actually project both have network latency and we had enough time to transition from project id -1 to valid project id
- On subsequent visits, the result for project id -1 votes is already in redux store and is processed right away before the transition to real project id
- added a guard in the project provider to wait until we have a valid project
- added another redirect guard in case someone actually visits a project id of -1 in the url (so we aren't spinning forever)
- tested ok for first and subsequent views (assumption here is there really are votes in the db)
- Note: I had initially set about adding -1 guards in voting provider and scoring provider.. decided it was better to handle at the root

### DIFF
--- a/src/providers/Project/ProjectProvider.tsx
+++ b/src/providers/Project/ProjectProvider.tsx
@@ -1,6 +1,9 @@
 import { useCallback, useEffect, useState } from 'react';
 import { useParams } from 'react-router-dom';
 
+import { BusySpinner } from '@terraware/web-components';
+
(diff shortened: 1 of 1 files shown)


joshbuchea (approved):
Looks good

PR #2457, opened 2024-04-03
SW-5158 Show participant name in participant project details header

karthikbtf:
- added participant to the react context
- fetched participant in the provider and set the context value
- updated header to use participant name

### DIFF
--- a/src/hooks/useParticipant.ts
+++ b/src/hooks/useParticipant.ts
@@ -21,8 +21,10 @@ export const useParticipant = (participantId: number) => {
   const result = useAppSelector(selectParticipantGetRequest(requestId));
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


src/hooks/useParticipant.ts
```diff
@@ -21,8 +21,10 @@ export const useParticipant = (participantId: number) => {
   const result = useAppSelector(selectParticipantGetRequest(requestId));
 
   useEffect(() => {
-    const request = dispatch(requestGetParticipant(participantId));
-    setRequestId(request.requestId);
+    if (participantId !== -1) {
+      const request = dispatch(requestGetParticipant(participantId));
+      setRequestId(request.requestId);
+    }
```
joshbuchea, line 27: Nice! 💯

joshbuchea (approved):

PR #2463, opened 2024-04-03
SW-5152 / SW-5169 Redirect to score edit if there is no scores/qualitative data

karthikbtf:
- added a `hasData` property on the context which checks if there is data
- in read-only view, redirect to edit if there is no data
- in edit-view, redirect to project details on cancel if there is no data (to avoid ending up back in edit view)
- since we need to deterministically identify if there is no data, added blocking checks in the update code to wait until the refresh list completes
- this is a tad ugly but needed, otherwise redirecting back to read-only view after save will only end up back in edit view (hasData hasn't had a chance to update)
- also fixed bug 5169 where a new score entry only keeps the latest edit (was using originalScore as copy instead of updated score), fixed that

### DIFF
--- a/src/scenes/AcceleratorRouter/Scoring/EditView.tsx
+++ b/src/scenes/AcceleratorRouter/Scoring/EditView.tsx
@@ -9,6 +9,7 @@ import Page from 'src/components/Page';
 import Card from 'src/components/common/Card';
 import PageForm from 'src/components/common/PageForm';
 import { APP_PATHS } from 'src/constants';
+import useNavigateTo from 'src/hooks/useNavigateTo';
 import strings from 'src/strings';
 import { Score, ScoreCategory, ScoreValue } from 'src/types/Score';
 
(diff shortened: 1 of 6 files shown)


src/scenes/AcceleratorRouter/Scoring/useScoresUpdate.ts
```diff
@@ -37,15 +39,18 @@ export default function useScoresUpdate(projectId: number): Response {
...
-      dispatch(requestListScores(projectId));
+      const dispatched = dispatch(requestListScores(projectId));
+      setListRequestId(dispatched.requestId);
     }
   }, [dispatch, projectId, result, snackbar]);
 
   return useMemo<Response>(
     () => ({
-      status: result?.status,
+      // Update is considered successful if both update and refresh list are successful.
```
tommylau523, line 49: Nice!

src/scenes/AcceleratorRouter/Scoring/EditView.tsx
```diff
@@ -41,16 +43,20 @@ const ScorecardEditView = () => {
   }, [history, projectId]);
 
   const onCancel = useCallback(() => {
-    goToScorecardView();
-  }, [goToScorecardView]);
+    if (hasData === false) {
```
tommylau523, line 46: should this be `!hasData` to handle `hasData === undefined`?

tommylau523 (approved):

src/scenes/AcceleratorRouter/Scoring/EditView.tsx
karthikbtf: If it is undefined, just means that the calculation hasn't happened yet. In this edit use-case, it won't be undefined.
Mostly helps in the view use-case.

PR #2465, opened 2024-04-03
SW-5160 Allow viewing participant project details even if there are no voters

karthikbtf:
- removed the redirect to projects list if there are no voters, so user can still see the project details
- added a 'no voters' message box in votes view for the project if there are no voters


tommylau523 (approved):
Looks good. Should we add something like "Contact an accelerator admin to add voters." to the error message?

karthikbtf:
> Looks good. Should we add something like "Contact an accelerator admin to add voters." to the error message?

That's a good suggestion, yes I can add that.

PR #2467, opened 2024-04-04
SW-5125: FE - Full Module View

joshbuchea:
This PR adds the "Full Module View", as well as a placeholder page for Module Events (1:1 Session, Live Session).

See [designs](https://www.figma.com/file/xo8UmzHZo4sTjGjU0fElUG/Participant-Experience?type=design&node-id=1561-4980&mode=design&t=LFdIfNpENC81eycS-0).

## Review Notes

There are still a few details outstanding:

- [ ] Update page title date format?
- [ ] Update module event date/time format?
- [ ] Wire up links & join session buttons

## Screenshots

localhost_modules_project_36_content_1_organizationId=28


src/constants.ts
```diff
@@ -43,6 +43,7 @@ export enum APP_PATHS {
   MONITORING = '/monitoring',
   MODULES_FOR_PROJECT = '/modules/project/:projectId',
   MODULES_FOR_PROJECT_CONTENT = '/modules/project/:projectId/content/:moduleId',
+  MODULES_FOR_PROJECT_EVENT = '/modules/project/:projectId/content/:moduleId/event/:eventId',
```
joshbuchea, line 46: Made a guess here, will adjust if we'd prefer a different path.

src/scenes/ModulesRouter/ModuleEvent.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const ModuleEvent = () => {
+  const { activeLocale } = useLocalization();
+  const theme = useTheme();
+  const pathParams = useParams<{ eventId: string; moduleId: string; projectId: string }>();
+  const eventId = Number(pathParams.eventId);
+  const moduleId = Number(pathParams.moduleId);
+  const projectId = Number(pathParams.projectId);
+  const module = useAppSelector(selectModule(moduleId));
+  const event = module?.events.find((e) => e.id === eventId);
```
joshbuchea, line 23: Perhaps this should be memoized?

joshbuchea, line 23:
```
  const event = module?.events.find((e) => e.id === eventId);
```

Perhaps this should be memoized?

src/scenes/ModulesRouter/ModuleView.tsx
```diff
@@ -0,0 +1,169 @@
...
+  label,
+  leftBorder = true,
+  onClickButton,
+  value,
+}: {
+  label: string;
+  leftBorder?: boolean;
+  onClickButton?: () => void;
+  value: string;
+}) => {
```
joshbuchea, line 31: Perhaps this should be extracted out into it's own file?

src/utils/dateFormatter.ts
```diff
@@ -14,6 +14,16 @@ export const getLongDate = (date: string, locale: string | undefined | null): st
...
 
+export const getLongDateTime = (date: string, locale: string | undefined | null): string =>
+  new Intl.DateTimeFormat(locale || 'en-US', {
+    month: 'long',
+    day: 'numeric',
+    year: 'numeric',
+    hour: 'numeric',
+    hour12: true,
+    timeZoneName: 'short',
+  }).format(new Date(date));
```
joshbuchea, line 25: The result from this function doesn't match the format in the designs 100%, but it's close. Still need to adjust the result from "March 16, 2024 at 3 PM HST" --> "March 16, 2024, 3pm HST".

src/scenes/ModulesRouter/ModuleViewTitle.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type ModulePageTitleProps = {
+  module: Module | undefined;
+};
+
+const ModuleViewTitle = ({ module }: ModulePageTitleProps) => {
+  return (
+    <Box alignItems='center' display='flex' flexDirection='row' flexWrap='wrap' marginY='24px' width='100%'>
+      <Typography fontSize={'24px'} lineHeight={'32px'} fontWeight={600} paddingRight='24px' whiteSpace='nowrap'>
+        {module ? 'AND_Climate Action Network-Andromeda' : ''}
```
joshbuchea, line 15: Where should the name in the page title come from, the project? Participant? Something else? (`'AND_Climate Action Network-Andromeda'`)

src/scenes/ModulesRouter/ModuleView.tsx
```diff
@@ -0,0 +1,169 @@
...
+                <>
+                  <ModuleContentSection>
+                    <Typography fontSize={'20px'} lineHeight={'28px'} fontWeight={600}>
+                      {strings.THIS_MODULE_CONTAINS}
+                    </Typography>
+                  </ModuleContentSection>
+
+                  {module.contents.map((content) => (
+                    <ModuleContentSection key={content.id}>
+                      <Link fontSize='16px' to='#'>
```
joshbuchea, line 133: Will set a value in the `to` prop for `Links` once I have a chance to create the route and placeholder page(s). What route(s) do we think would make sense these content items?

src/scenes/ModulesRouter/ModuleEvent.tsx
karthikbtf: Might be a good idea to memoize, theoretically the array can be large in size to search.

src/scenes/ModulesRouter/ModuleEvent.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const crumbs: Crumb[] = useMemo(
+    () => [
+      {
+        name: activeLocale ? 'Module' : '',
+        to: APP_PATHS.MODULES_FOR_PROJECT_CONTENT.replace(':projectId', `${projectId}`)
+          .replace(':moduleId', `${moduleId}`)
+          .replace(':eventId', `${eventId}`),
```
karthikbtf, line 31: remove the eventId

src/scenes/ModulesRouter/ModuleView.tsx
```diff
@@ -0,0 +1,169 @@
...
+            </Grid>
+
+            {module.events.length && (
+              <Grid item>
+                {module.events.map((event) => (
+                  <ModuleFieldDisplay
+                    key={event.id}
+                    label={event.name}
+                    onClickButton={() => goToModuleEvent(projectId, event.id, module.id)}
+                    value={event.eventTime ? getLongDateTime(event.eventTime.replace('PST', ''), activeLocale) : ''}
```
karthikbtf, line 157: curious why the `replace`

src/scenes/ModulesRouter/ModuleView.tsx
karthikbtf: Sounds like a good idea.

src/utils/dateFormatter.ts
karthikbtf: Here's an example of how we show similar dates in the app today (example Notifications dropdown) `March 14, 2024 at 4:38:49 PM`
Might be good to use a similar pattern, maybe suggest in the designs.

karthikbtf (commented):
Some comments. I am not up to speed on the content/designs and unable to answer some of your questions. Will let Nick take a look as well.

src/scenes/ModulesRouter/ModuleView.tsx
joshbuchea: Oh great question, this was a hack because the datetime format in the mock data needs to be adjusted because it doesn't appear to be valid as-is. (JS date is unable to parse it and passing to one of the datetime formatter functions was throwing an error.) I'll fix the datetime format in the mock data is it's parseable by JS date and we can adjust the format if it isn't quite right.

src/constants.ts
tommylau523: Likely we can just use `/modules/project/:projectId/event:eventId`

Because we probably won't use `moduleId` to query for the full event details. But I do like the this path showing the hierarchy of things.

src/scenes/ModulesRouter/ModuleFieldDisplay.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+import { Box, Grid, Typography, useTheme } from '@mui/material';
+import { Button } from '@terraware/web-components';
+
+import strings from 'src/strings';
+
+const ModuleFieldDisplay = ({
```
tommylau523, line 8: I think we can name this to be ModuleEventCard or something like that, and make the props take in a dateTime specifically.

Or if you are imagining this will be used in other places, we can move it to components/

Seems like this pattern is only used for events according to Figma.

src/constants.ts
```diff
@@ -43,6 +43,7 @@ export enum APP_PATHS {
   MONITORING = '/monitoring',
   MODULES_FOR_PROJECT = '/modules/project/:projectId',
```
tommylau523, line 44: Not related to this PR, perhaps we can rehash this a bit?

`/project/{projectId}/modules` for the all modules view
`/project/{projectId}/modules/{modulesId}` for the full modules view

Then, in the contents link, we can use 
`/project/{projectId}/modules/{modulesId}/contents/{contentId}` for any content details or additional resources. 

Perhaps also
`/project/{projectId}/modules/{modulesId}/events/{eventId}` as well for events. If we go this route.

src/scenes/ModulesRouter/ModuleView.tsx
tommylau523: We do need to link to deliverables or contents depending on the payload:

for deliverables: `/deliverables/:deliverableId/submissions/:projectId`

for contents, I left a comment on the routes above, hopefully it makes sense there.

src/constants.ts
tommylau523: If we do stick with `/modules` prefix, (since the rest of our routes do organize like that), another option is to add 

`/modules/project/{projectId}/content/{modulesId}/detail/{contentId}`

to show additional resources or details of module contents.

src/scenes/ModulesRouter/ModuleViewTitle.tsx
tommylau523: This is the project name! I will make sure to let design and product know to standardize these in the future in our FIgma.

tommylau523 (commented):
Thanks Josh! I know the specs are spotty so thanks for filling in the gaps! I left some comments to hopefully give you a bit of details.

joshbuchea:
Thanks for the review @tommylau523 🙌 Great comments and suggestions, will make changes based on feedback and will re-request a review soon.

joshbuchea:
@tommylau523 do you know if we're planning to add an endpoint to fetch module events or just populate events content data within the modules?

joshbuchea:
Just realized you said "Because we probably won't use `moduleId` to query for the full event details." in your previous comment. I'll assume we'll make a separate fetch to get module event details and will make the changes required to support that.

joshbuchea:
I think I still need to add redux for module events. Should we include that as part of the `ModuleService` or would we prefer to create a new service? If the latter, what should we name it? `ModuleEventService`? Something else?

Also, what URL params do we expect to be required when fetching module event data?

@karthikbtf @tommylau523 would love to hear your thoughts when you have a free moment 🙏

tommylau523:
The `ModuleService` should be able to supply event information, details as well as links that are specific to a project.

karthikbtf:
> I think I still need to add redux for module events. Should we include that as part of the `ModuleService` or would we prefer to create a new service? If the latter, what should we name it? `ModuleEventService`? Something else?
> 
> Also, what URL params do we expect to be required when fetching module event data?
> 
> @karthikbtf @tommylau523 would love to hear your thoughts when you have a free moment 🙏

@joshbuchea my thought would be to leverage ModuleService for events, unless the distinction at a service level really matters.
Not up to speed on Participant Experience enough to comment on url params for module event data, will defer to @tommylau523

karthikbtf:
> > I think I still need to add redux for module events. Should we include that as part of the `ModuleService` or would we prefer to create a new service? If the latter, what should we name it? `ModuleEventService`? Something else?
> > Also, what URL params do we expect to be required when fetching module event data?
> > @karthikbtf @tommylau523 would love to hear your thoughts when you have a free moment 🙏
> 
> @joshbuchea my thought would be to leverage ModuleService for events, unless the distinction at a service level really matters. Not up to speed on Participant Experience enough to comment on url params for module event data, will defer to @tommylau523

Also, if you prefer to do this in stages, you can get an initial version in and iterate on that until gaps are filled.

src/constants.ts
karthikbtf: In general, especially for BE API routes, resource based hierarchy does make sense.
For FE routes, my suggestion would be to make them page-view/use-case driven hierarchy. It affects isolation of route-handling code, potential lazy loading, other semantics such as highlighting left nav.
Of course we can make any combination work if there's a strong preference one way or another.

karthikbtf (approved):
@joshbuchea this looks good for a first check-in, if you want to follow up separately on pending items. One concern still is the introduction of a new date display format, if design wasn't aware we already have a pattern.. I'd recommend following up on that as well.

PR #2468, opened 2024-04-04
SW-4795 Apply "font-variant-numeric: tabular-nums" to all table columns in Terraware, and right-align all number table columns

constanzauanini:
"font-variant-numeric: tabular-nums" is applied in web-components to all table cells.

Right-aligments are applied according [Jira ticket](https://terraformation.atlassian.net/browse/SW-4795)


src/components/BatchWithdrawFlow/flow/SelectBatchesWithdrawnQuantity.tsx
```diff
@@ -72,18 +72,21 @@ const defaultTableColumns = (): TableColumnType[] => [
     key: 'germinatingQuantityWithdrawn',
     name: strings.GERMINATING_QUANTITY,
     type: 'string',
+    alignment: 'right',
```
karthikbtf, line 75: Looks like we are doing this for all number type columns.
Should we fix the `type` in all the places as `number`, under common/table if the column type is a number we can update the columns to set alignment to right? Just a thought. It will work for all future columns as well as long as we set the type to number correctly.

karthikbtf (approved):
lgtm, one comment

src/components/BatchWithdrawFlow/flow/SelectBatchesWithdrawnQuantity.tsx
constanzauanini: I thought about it, but wasn't sure if setting type number would change anything at tables, but I was checking the code and it looks like it doesn't, so I will try this approach, sounds more correct!

karthikbtf: To be honest, I don't know if the other approach has advantages. Either way we need to ensure setting a column property correctly in the future (type as number or alignment as right). Maybe we should just stick with your current approach. The other suggestion is not justified.

constanzauanini: In case we have other columns of type number, we would want them to be right aligned, so I think it's correct to do it like that

src/components/common/table/index.tsx
```diff
@@ -56,6 +56,15 @@ interface TableProps<T> extends Omit<Props<T>, keyof LocalizationProps> {
 }
 
 export function BaseTable<T extends TableRowType>(props: TableProps<T>): JSX.Element {
+  const addAlignment = () => {
```
karthikbtf, line 59: leverage `useMemo` here?

karthikbtf (approved):

PR #2474, opened 2024-04-04
SW-5163 Fix top/left padding on page with module timeline

karthikbtf:
- The TfMain construction padding was reduced, to be consistent with other crumb layout views
- A cleaner option would be to support the timeline equivalent layout within the Page component as it assumes that TfMain is the root level component of the page
- I can take a look at that later

Terraware App 2024-04-04 09-11-23


nickgraz (approved):
✨

PR #2483, opened 2024-04-04
SW-5153 Show tooltip/message about accelerator orgs in create participant workflow

karthikbtf:
- org dropdown tooltip letting users know what it represents
- warning message when there are no orgs to list

Screen Shot 2024-04-04 at 2 31 03 PM

Terraware App 2024-04-04 14-41-51


nickgraz (approved):
Excellent

PR #2486, opened 2024-04-05
SW-4795 Bump web-components version to correctly show right-aligned table headers

constanzauanini:
Before:
Screenshot 2024-04-05 at 12 52 41

After:
Screenshot 2024-04-05 at 12 50 46


joshbuchea (approved):
🚢

PR #2488, opened 2024-04-05
 SW-5186 Batches at Nursery table view: Withdrawn column is not populated

constanzauanini:
Withdrawn column was not populated because we were not requiring the value in the search request


alexTerraformation (approved):
Thank you!

PR #2495, opened 2024-04-09
SW-5191: Sorting Project tables in Console Overview crashes

joshbuchea:
This PR includes changes to fix the crash that was occurring when sorting by column in the Console Overview Projects table.

I removed some mapping that looked like it was in-place to replace long/awkward keys and possibly reduce the differences between `ParticipantProject` and `ParticipantProjectSearchResult`, but that mapping seemed to be what was causing the crash as the updated field names weren't recognized by the BE when performing the sort function.

I'm not sure if there was another way to fix this crash without removing the mapping, but I'm happy to make changes if there's another way we'd prefer to approach this issue.

### Before

https://github.com/terraware/terraware-web/assets/1474361/f64b3a72-bbd9-4ff3-a28c-402e1eb9d1c5

### After

https://github.com/terraware/terraware-web/assets/1474361/10047fc1-b383-4f34-aea4-40ffa49677e0

### DIFF
--- a/src/services/ParticipantProjectService.ts
+++ b/src/services/ParticipantProjectService.ts
@@ -123,7 +123,7 @@ const get = async (participantProjectId: number): Promise<Response2<ParticipantP
   });
 
 const downloadList = async (search?: SearchNodePayload, sortOrder?: SearchSortOrder): Promise<string | null> =>
-  await SearchService.searchCsv(getSearchParams(search, sortOrder, true));
+  await SearchService.searchCsv(getSearchParams(search, sortOrder));
 
 const list = async (
   search?: SearchNodePayload,
--- a/src/services/ParticipantProjectService.ts
+++ b/src/services/ParticipantProjectService.ts
@@ -39,11 +39,7 @@ const COHORT_ID_EXISTS_PREDICATE: SearchNodePayload = {
   },
 };
 
-const getSearchParams = (
-  search?: SearchNodePayload,
-  sortOrder?: SearchSortOrder,
-  isCsv?: boolean
-): SearchRequestPayload => {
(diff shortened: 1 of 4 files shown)


src/services/ParticipantProjectService.ts
```diff
@@ -156,17 +156,21 @@ const list = async (
     type LandUse = { landUseModelType: string };
 
     return {
-      cohortName: participant_cohort_name,
-      country: country_name,
+      acceleratorDetails_confirmedReforestableLand,
+      'acceleratorDetails_confirmedReforestableLand(raw)': Number(
```
karthikbtf, line 160: Looks like we no longer need this `(raw)` property. Remove it from the service and other places.
Also remove it from line 59 in this file.
This is all my doing, apologies.

karthikbtf (approved):
Thanks for the fix 👍

src/services/ParticipantProjectService.ts
```diff
@@ -56,7 +56,6 @@ const getSearchParams = (
       'country_name',
       'country_region',
       'acceleratorDetails_confirmedReforestableLand',
-      ...(isCsv ? [] : ['acceleratorDetails_confirmedReforestableLand(raw)']),
```
karthikbtf, line 59: remove the isCsv option from function signature as well?

karthikbtf (approved):
Thanks 🙏 one suggestion

src/services/ParticipantProjectService.ts
joshbuchea: Nice catch, you got it

karthikbtf (approved):
👍

PR #2496, opened 2024-04-09
SW-5192 Add participant flow UX tweaks when there are no additional accelerator orgs to select from

karthikbtf:
- we already have messaging when there are zero accelerator orgs to select from
- in the scenario where there are orgs but all of them have been selected, show the 'add org' link disabled with a tool explaining why
- previously in this scenario the add org link was hidden when there were no more orgs
- waiting on Chudo's approval of this approach

Screen Shot 2024-04-09 at 9 53 00 AM


chudo76 (approved):
LGTM!

PR #2497, opened 2024-04-09
SW-5176 Reconcile 'accession collecting site' labels across create/edit/view flows

karthikbtf:
- we were using three different labels, reconciled them to use same label (as per latest requirements)
- we do have 2 unused labels at this point

Terraware App 2024-04-09 10-16-12

Terraware App 2024-04-09 10-15-56

Terraware App 2024-04-09 10-14-33


joshbuchea (approved):
Nice! 🚢

PR #2500, opened 2024-04-09
SW-5125: FE - Full Module View (Follow-ups)

joshbuchea:
This PR has some follow-ups including:

- Add module events to `ModuleService` & Redux
- Join Session buttons opens new browser window with event
- Add session call details box
- Update due date styles
- Connect module event links

## Tasks

- [ ] Update routes?
- [ ] Update `ModuleEvent` to include `module.name` & `module.dateRange` and remove module fetch?

### DIFF
--- a/src/scenes/ModulesRouter/ModuleEventView.tsx
+++ b/src/scenes/ModulesRouter/ModuleEventView.tsx
@@ -8,10 +8,9 @@ import { Crumb } from 'src/components/BreadCrumbs';
 import Link from 'src/components/common/Link';
 import PageWithModuleTimeline from 'src/components/common/PageWithModuleTimeline';
 import { APP_PATHS } from 'src/constants';
-import { useLocalization } from 'src/providers';
+import { useLocalization, useProject } from 'src/providers';
 import { requestGetModule, requestGetModuleEvent } from 'src/redux/features/modules/modulesAsyncThunks';
 import { selectModule, selectModuleEvent } from 'src/redux/features/modules/modulesSelectors';
-import { selectProject } from 'src/redux/features/projects/projectsSelectors';
 import { useAppDispatch, useAppSelector } from 'src/redux/store';
 import strings from 'src/strings';
 import { getLongDateTime } from 'src/utils/dateFormatter';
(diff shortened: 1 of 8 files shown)


joshbuchea:
@tommylau523 @karthikbtf @nickgraz now that Nick is back, I think it would be good to get his opinion as well re: route changes. I'm happy to adjust the routes to whatever we'd prefer them to be. I don't have a preference for one convention vs another, so I'm happy to defer to the group.

@tommy I apologize for asking you to repeat yourself, but would you be willing share your proposed route changes again here in this PR? 🙏

tommylau523:
Not a problem at all! 

currently we are following something like this:

`modules/project/:projectId`
`modules/project/:projectId/content/:moduleId`
`modules/project/:projectId/content/:moduleId/event/:eventId`
`modules/project/:projectId/content/:moduleId/resource/:contentId`

Which I think is totally fine as is. 

I had a suggestion that can reorganize routes to follow projects first:

`project/:projectId/module/:moduleId/event or content/:id`

It fits the resource hierarchy a bit better in my opinion.

joshbuchea:
Awesome thanks @tommylau523!

joshbuchea:
Also, right now the module needs to be fetched on the module events page because the module name & date range is used to render the page title. But if those properties were available in the module event data we could remove the module fetch. Any preference here?

src/scenes/ModulesRouter/ModuleEventView.tsx
```diff
@@ -111,58 +69,85 @@ const ModuleEventView = () => {
...
+
+                {event?.additionalLinks?.length && (
+                  <>
+                    {event?.additionalLinks?.map((link, index) => (
+                      <Box key={index} marginBottom={theme.spacing(2)}>
+                        <Link
+                          fontSize='16px'
+                          onClick={() => {
+                            if (link.url) {
+                              window.open(link.url, '_blank', 'noopener noreferrer');
```
karthikbtf, line 138: Alternatively use the `to=` and `target=` properties of `Link`? Not sure if the `rel=` is auto added, maybe we could update Link to do that.

karthikbtf (approved):
Looks good to me, one suggestion.

src/scenes/ModulesRouter/ModuleEventView.tsx
joshbuchea: So these URLs are external and it seemed like `Link` was specific for internal app routes, but I'll try it out and will update if `Link` handles external URLs.

joshbuchea:
@nickgraz would love to get your thoughts when you have a free moment re: @tommylau523's comment about potentially updating the routes.

nickgraz:
@joshbuchea I think those route suggestions are great. The ones I put in there were a complete shot in the dark, @tommylau523 has a better idea of how it will really be structured.

joshbuchea:
Sounds good @nickgraz and @tommylau523, let's update the routes.

PR #2506, opened 2024-04-11
SW-5129 Added module tab to nav bar

tommylau523:
Added side nav for modules. Will need a better way of finding whether projects in an organization have modules at some point.


tommylau523:
* **#2506** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2506?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -109,6 +114,31 @@ export default function NavBar({
     }
   }, [selectedOrganization]);
 
+  useEffect(() => {
+    const checkHasModules = async () => {
+      const projectsResult = await ProjectsService.listProjects(selectedOrganization.id);
```
nickgraz, line 119: Can you use redux for this?

nickgraz: I think the only addition to redux would be a new selector that takes a list of project IDs and returns the list of modules for those projects.

```
export const selectModuleLists = (projectIds: number[]) => (state: RootState) => projectIds.map((projectId) => state.moduleList[projectId]?.data);
```

Then you would just hook up the dispatch and selector for projects, `requestProjects` and `selectProjects`, and you'd have to dispatch the requests for the individual module lists too. Saying all that, it seems like perhaps this would all be better as its own function in one of the services that handles everything you have below. It just seems a bit heavy handed to have all of this within a component, let alone a component called "nav bar". Ideally this stuff is isolated elsewhere and the usage within this component is as simple as dispatch/select.

tommylau523: Done! Could you review to see if my approach makes sense?

tommylau523: I am wondering if I am using the feature flag correctly as well.

nickgraz: Looks good, thanks for changing that up! A couple small suggestions.

src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -256,9 +291,10 @@ export default function NavBar({
           )}
         </SubNavbar>
       </NavItem>
-      {!hasDeliverables && reportsMenu && (
+      {!hasDeliverables && ((isEnabled('Participant Experience') && modulesMenu) || reportsMenu) && (
         <>
-          <NavSection />
+          <NavSection title={strings.ACCELERATOR.toUpperCase()} />
+          {isEnabled('Participant Experience') ? modulesMenu : null}
```
nickgraz, line 297: Can you set the flag check to a variable? `{featureFlagParticipantExperience && modulesMenu}`

src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -274,6 +310,7 @@ export default function NavBar({
             }}
             id='deliverables'
           />
+          {isEnabled('Participant Experience') ? modulesMenu : null}
```
nickgraz, line 313: `{featureFlagParticipantExperience && modulesMenu}`

nickgraz (approved):
A couple small suggestions that can probably be ignored if you want, nice work!

src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -256,9 +291,10 @@ export default function NavBar({
           )}
         </SubNavbar>
       </NavItem>
-      {!hasDeliverables && reportsMenu && (
+      {!hasDeliverables && ((isEnabled('Participant Experience') && modulesMenu) || reportsMenu) && (
```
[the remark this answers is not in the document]
tommylau523: I was a bit confused as well. Seems like reports may not be part of accelerator? Let me check with design and also simplify this section.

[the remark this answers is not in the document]
tommylau523: I refactored the deliverables item to the similar menu format. Could you take a look?

src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -256,27 +314,16 @@ export default function NavBar({
...
       </NavItem>
-      {!hasDeliverables && reportsMenu && (
-        <>
-          <NavSection />
-          {reportsMenu}
-        </>
-      )}
-      {hasDeliverables && (
+
+      {(deliverablesMenu || modulesMenu || reportsMenu) && (
```
nickgraz, line 318: Looks great! I _think_ the reports menu should show up though even if they don't have deliverables or modules. Might want to verify with product, otherwise it looks great

nickgraz (approved):

src/scenes/OrgRouter/NavBar.tsx
tommylau523: I think it will show up if `reportsMenu` is not null! And that is decided by `reports.length` and `canSubmitReports`. The behavior should be unchanged.

nickgraz: Oh I see! `acceleratorSectionTitle` is just an empty string without modules or deliverables, thanks!

PR #2508, opened 2024-04-12
SW-5206: Update module routes

joshbuchea:
This PR includes updates to the module route keys & paths.

Note: Plural naming was used to follow existing route conventions.


nickgraz (approved):
Looks good!

joshbuchea:
@nickgraz so it turns out these path changes break the module pages, I believe because the route is now being picked up and handled by the `ProjectsRouter` rather than the `ModulesRouter`. Is it ok to move the three project module related routes into the `ProjectsRouter` or is there an alternative approach we're prefer?

joshbuchea:
Ah but then we need to move the views as well 🤔

I think we may be able to just move the `ModulesRouter` above the `ProjectsRouter` and it would be picked up, but then we rely on route order within the `OrgRouter`. Perhaps that's ok?

joshbuchea:
Went ahead and made the latter change since it's the simplest.

PR #2525, opened 2024-04-17
SW-5225 Fix date validation for 'readyByDate' in nursery transfer withdrawals

karthikbtf:
- there are 2 dates in this flow, 'withdrawnDate' and 'readyByDate'
- common date validation for both these dates was preventing future dates
- the future date check only applies to 'withdrawnDate', whereas 'readyByDate' can be in the future and most likely will
- updated validation code to support a 'allowFutureDates' option, set option to true for 'readyByDate'


nickgraz (approved):
Looks good

PR #2532, opened 2024-04-22
Fix modules selector to be memoized

karthikbtf:
- use cached selector


constanzauanini (approved):
Thanks for caching this correctly! I tried different ways using useMemo and didn't work

nickgraz (approved):
🙌

PR #2537, opened 2024-04-23
SW-5258 Accelerator Overview/Projects page horizontal scroll is broken

constanzauanini:


src/scenes/AcceleratorRouter/Overview/index.tsx
```diff
@@ -70,11 +72,12 @@ const OverviewView = () => {
   });
 
   return (
-    <Page title={strings.OVERVIEW}>
+    <AcceleratorMain>
+      <PageHeader title={strings.OVERVIEW} />
```
karthikbtf, line 76: Any chance we can get this to work with `Page` so we aren't depending more on other components? Is there some style we can pick up?

karthikbtf (approved):
Works great but one question/suggestion.

karthikbtf (approved):
Thanks!

PR #2538, opened 2024-04-23
SW-5257 Fix my-account/edit redirect for no-org view

karthikbtf:
- the router was doing an exact match, updated that
- also fixed styles since there is no nav
- also reverted some styles from yesterday since I had a cached build that didn't show proper syles

Terraware App 2024-04-23 09-15-58

Terraware App 2024-04-23 09-15-40


constanzauanini (approved):
Thanks for fixing this. I missed this use case

PR #2541, opened 2024-04-23
SW-5109 Refactor a few of the confirmation dialogs to use the 'Confirm' component

karthikbtf:
- this is a small batch of refactor, there's a lot more that needs to be migrated


joshbuchea (approved):
Nice start, thanks for doing this @karthikbtf

PR #2544, opened 2024-04-24
SW-5239: Withdrawals: List View is titled "Withdrawal History" but breadcrumb says "Withdrawal Log"

joshbuchea:
This PR includes a quick fix to update "Withdrawal Log" to "Withdrawal History".

### After

localhost_nursery_withdrawals_tab=withdrawal_history organizationId=28

### DIFF
--- a/src/scenes/NurseryRouter/NurseryReassignmentView.tsx
+++ b/src/scenes/NurseryRouter/NurseryReassignmentView.tsx
@@ -233,7 +233,7 @@ export default function NurseryReassignmentView(): JSX.Element {
             id='back'
             to={`${APP_PATHS.NURSERY_WITHDRAWALS}?tab=withdrawal_history`}
             className={classes.backToWithdrawals}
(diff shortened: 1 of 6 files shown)


src/scenes/NurseryRouter/NurseryReassignmentView.tsx
```diff
@@ -233,7 +233,7 @@ export default function NurseryReassignmentView(): JSX.Element {
             id='back'
             to={`${APP_PATHS.NURSERY_WITHDRAWALS}?tab=withdrawal_history`}
             className={classes.backToWithdrawals}
-            name={strings.WITHDRAWAL_LOG}
```
constanzauanini, line 236: Should we remove the string from the string file too?

joshbuchea: I haven't been removing unused strings because they get added back automatically unless they're also removed from Phrase. Just requested access to Phrase though so will remove once I get access.

PR #2555, opened 2024-04-26
SW-5272: FE - Set up mock data for species deliverable

joshbuchea:
This PR includes:

- “Participant Project Species” service and redux added to FE with mock data
- Add 2 mocked “species list” deliverables to the “get deliverable(s)” API calls

**NOTE:** I'm not sure if I got the fields quite right, but I was able to fetch and select a participant project species by dispatching the async thunk and using the selector and I'm able to see the two added species list deliverables in the deliverables list view.

## Screenshots

localhost_deliverables_organizationId=28


src/redux/features/participantProjectSpecies/participantProjectSpeciesAsyncThunks.ts
```diff
@@ -0,0 +1,76 @@
...
+      return response.data.projectSpecies;
+    }
+
+    return rejectWithValue(strings.GENERIC_ERROR);
+  }
+);
+
+export const requestListParticipantProjectSpecies = createAsyncThunk(
+  'participantProjectSpecies/list',
+  async (projectId: number, { rejectWithValue }) => {
```
constanzauanini, line 26: We need a requestListParticipantProjectSpecies for all projects (for the species list view), not just for a specific project, maybe this projectId could be optional?

constanzauanini: Oh, never mind. I just realize I can use the getDeliverables and filter the result by type

constanzauanini: @nickgraz I know this wasn’t in the Jira ticket, buy maybe we missed it?

joshbuchea: That wasn't in the AC in the ticket (the ticket actually had great AC this time, thanks @nickgraz), so that's why it's not included in the PR. But I can definitely add that if we need it, no problem.

Great suggestion @constanzauanini! I'll give that a try.

src/services/ParticipantProjectSpeciesService.ts
```diff
@@ -62,9 +62,17 @@ const get = async (projectId: number, projectSpeciesId: number): Promise<Project
...
-const list = async (projectId: number): Promise<AllSpeciesResponse> => {
+const list = async (projectId?: number): Promise<AllSpeciesResponse> => {
   return new Promise((resolve) => {
-    resolve({ data: { projectSpecies: mockProjectSpecies?.[projectId] || [] }, requestSucceeded: true });
+    if (projectId === undefined) {
+      const allProjectsSpecies: ProjectSpecies[] = [];
+      Object.values(mockProjectSpecies).forEach((speciesList) => {
+        allProjectsSpecies.push(...speciesList);
+      });
+      resolve({ data: { projectSpecies: allProjectsSpecies }, requestSucceeded: true });
```
constanzauanini, line 72: Thanks for doing this! Could we also make the projectId optional in the thunk?

src/redux/features/participantProjectSpecies/participantProjectSpeciesAsyncThunks.ts
nickgraz: This request is used to show all of the participant project species that have already been associated. They will be shown within the "species deliverable" view. A project ID must be required here because it is the basis of the relationship between the participant project and the organization species. 

We already have selectors elsewhere for "show all species for an organization", which will be used in the "add species to my participant project" flow.

src/services/ParticipantProjectSpeciesService.ts
joshbuchea: Oh yeah oops! Sorry will push the change up ASAP.

src/services/ParticipantProjectSpeciesService.ts
```diff
@@ -0,0 +1,118 @@
+import { components } from 'src/api/types/generated-schema';
+
+import { Response } from './HttpService';
+
+export type ProjectSpecies = {
```
nickgraz, line 5: Can you rename this to `ParticipantProjectSpecies`?

src/services/ParticipantProjectSpeciesService.ts
```diff
@@ -0,0 +1,118 @@
...
+  deliverableId: number;
```
nickgraz, line 6: No deliverable ID will be needed on this entity. Basically, the species deliverables just exist, and implicitly "point to" the entire list of participant project species that exist at that moment. There is no direct link from a deliverable to a participant project species.

src/services/ParticipantProjectSpeciesService.ts
```diff
@@ -0,0 +1,118 @@
...
+  id: number;
+  projectId: number;
+  speciesId: number;
```
nickgraz, line 9: You can add `rationale`, `status`, `feedback` to this type.

src/services/ParticipantProjectSpeciesService.ts
nickgraz: Project ID needs to be required. Participant Project Species exist in addition to Organization Species and are accessed through the project ID

src/redux/features/participantProjectSpecies/participantProjectSpeciesAsyncThunks.ts
```diff
@@ -0,0 +1,76 @@
...
+  AllSpeciesResponse,
+  DeleteProjectSpeciesResponse,
+  ProjectSpecies,
+  ProjectSpeciesResponse,
+} from 'src/services/ParticipantProjectSpeciesService';
+import strings from 'src/strings';
+
+export const requestGetParticipantProjectSpecies = createAsyncThunk(
+  'participantProjectSpecies/get-one',
+  async ({ projectId, projectSpeciesId }: { projectId: number; projectSpeciesId: number }, { rejectWithValue }) => {
```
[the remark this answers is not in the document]
joshbuchea: Ok that makes sense, will remove  it

src/services/ParticipantProjectSpeciesService.ts
joshbuchea: You got it

src/services/ParticipantProjectSpeciesService.ts
joshbuchea: Ok interesting, but makes sense. Will remove.

src/services/ParticipantProjectSpeciesService.ts
joshbuchea: You got it. Are `rationale` and `feedback` both `string`s and is `status` an enum?

joshbuchea:
Thanks for the review @nickgraz 🙌

src/redux/features/participantProjectSpecies/participantProjectSpeciesAsyncThunks.ts
joshbuchea: Ok that makes sense, thanks for the context. Will revert the optional `projectId` param changes 👍

constanzauanini: Thanks @nickgraz for context, so then in the "search all species for an organization" we need to add a field to get the projects that species is part of? Could we mock that too?

src/services/ParticipantProjectSpeciesService.ts
nickgraz: You can do all strings for now, enum will come from BE when it is done

src/redux/features/participantProjectSpecies/participantProjectSpeciesAsyncThunks.ts
nickgraz: I believe that feature (projects within species) is only available to participants, so we will just combine the two lists. If they are a participant, get the org species and participant project species. If they aren't, only fetch the org species and don't mention "project" anywhere. I will double check this assumption

nickgraz (approved):
Nice work!

src/redux/features/participantProjectSpecies/participantProjectSpeciesAsyncThunks.ts
constanzauanini: I thought same thing, that's why I think we need a "get all participant project species" selector

nickgraz: I believe that is what this selector does `selectParticipantProjectSpeciesListRequest`

constanzauanini: Correct, but what would be the corresponding thunk to dispatch? We need a projectId to use this one

nickgraz: Oh, I see what you're saying. Since there is only a single argument being passed (projectId) we can replace the request ID with it by adding `true` to the `buildReducers` call in the slice that uses this thunk. Good catch!

nickgraz: https://github.com/terraware/terraware-web/pull/2567

constanzauanini: Thanks for the PR with the change, but we still need to get all project species, without a projectId. I made a comment in the other PR, I think the selectParticipantProjectSpeciesListRequest needs a requestId, but not a projectId, and the selectParticipantProjectSpeciesRequest needs a projectId. But do we need to change the thunk to allow not having a projectId?

nickgraz: Why do we need to "get all project species" without a project ID? Project species only exist in the context of a participant project, and the only access patterns for getting all project species is in the context of a specific participant project. 

Also, the "participant project species" model will have `id, project_id, species_id, status, feedback, rationale`. So we actually don't need `project_id` to "get single participant project species" since the `id` of the row is sufficient to retrieve all the data for that single participant project species.

nickgraz: I need to remove `projectId` from `requestGetParticipantProjectSpecies`, it should only need the participant project species ID.

nickgraz: Oh, I see. Yea I think we will be able to add that in the search API request. The species stable uses the Search API to show the rows, and we will add `participant_project_species` as a Search API sublist, so all we will need to do is add a search field like `participantProjectSpecies.projects.name`.

constanzauanini: Thanks! Sorry for all the confusion!

nickgraz: No problem! 🤜 🤛

PR #2559, opened 2024-04-26
SW-5296 Pre-populate seedling withdrawal quantities with 0

alexTerraformation:
For purposes other than outplant we allow withdrawing from all three statuses of seedling batches.  Usually users only want to withdraw from a single status, or two statuses.  This causes a high number of errors on submission when users have not entered zero for all of the statuses that they don't want to withdraw.  

This change pre-fills all fields with 0 so that any non-edited fields will not cause errors upon submission.


joshbuchea (approved):
Looks good!

PR #2577, opened 2024-05-01
SW-5279: Create SpeciesDeliverableTable

joshbuchea:
This PR adds the `SpeciesDeliverableTable` component, as well as SW-5282 (Dialog flow for "bulk action - remove species from project").


src/components/SpeciesDeliverableTable/RemoveSpeciesDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+          label={strings.REMOVE}
+          type='destructive'
+          onClick={removeSelectedSpeciesFromParticipantProject}
+          size='medium'
+          key='button-2'
+        />,
+      ]}
+      skrim={true}
+    >
+      <Typography sx={{ paddingTop: 3 }}>{strings.ARE_YOU_SURE}</Typography>
```
constanzauanini, line 76: If this is the only content of the modal you could use the "message" prop of DialogBox

constanzauanini (approved):
Just one comment but looking good!

src/components/SpeciesDeliverableTable/RemoveSpeciesDialog.tsx
joshbuchea: Great suggestion! 🙌 Will update

PR #2578, opened 2024-05-01
Move header and content to new components

constanzauanini:
Looks like to use the context correctly we need to wrap all the components that depends on a context change


tommylau523 (approved):

nickgraz (approved):
Nice!

PR #2589, opened 2024-05-06
SW-5287 Console Page "right buttons"

constanzauanini:
Button were already created for console document deliverables. I renamed DeliverableView to DocumentDeliverableView in Console side and added a SpeciesDeliverableView to add future changes for table there.

The only change in buttons for species is the approve modal copy

Screenshot 2024-05-06 at 1 33 09 PM

This also includes SW-5288 (comments )


joshbuchea (approved):
Looks great!

PR #2592, opened 2024-05-07
SW-5289 Species table row modifications

constanzauanini:
Screenshot 2024-05-07 at 12 06 34 PM

Table changes:
Scientific name click goes to “participant project species view” (this view is not done yet)
Added Reject and approve button in each row.
Removed add species button.
Added “reject species” dialog when clicking reject button in row


src/components/SpeciesDeliverableTable/index.tsx
```diff
@@ -78,20 +80,20 @@ const SpeciesDeliverableTable = (): JSX.Element => {
...
+            {!isAcceleratorRoute && (
+              <Button
+                icon='plus'
+                id='add-species-to-project'
+                label='Add Species to Project'
+                onClick={() => setOpenedAddSpeciesModal(true)}
+                priority='secondary'
+                size='medium'
+              />
+            )}
```
joshbuchea, line 92: Is it safe to assume all accelerator users have permission to use this feature or are there any permission checks that should be made here?

src/services/ParticipantProjectSpeciesService.ts
```diff
@@ -107,7 +107,7 @@ const update = async (
     const index = speciesList.findIndex((ps) => ps.id === participantProjectSpeciesId);
 
     if (index !== -1) {
-      speciesList[index] = participantProjectSpecies;
+      // speciesList[index] = participantProjectSpecies;
```
joshbuchea, line 110: Was this for testing? Does this change need to be included?

joshbuchea (approved):

src/components/SpeciesDeliverableTable/index.tsx
constanzauanini: Hmm I'm not sure if all users can add species. We can add condition later if needed

src/services/ParticipantProjectSpeciesService.ts
constanzauanini: It was for testing, anyways mocks are removed now I think

PR #2596, opened 2024-05-08
SW-5290  Console - Participant Project Species view

constanzauanini:
Screenshot 2024-05-08 at 4 42 29 PM

TODO: Additional Species Data is not added to the view yet


src/components/SpeciesDeliverableTable/TableCellRenderer.tsx
```diff
@@ -41,7 +45,16 @@ export default function SpeciesDeliverableCellRenderer(props: RendererProps<Tabl
 
   const createLinkToAcceleratorSpecies = (iValue: React.ReactNode | unknown[]) => {
     return (
-      <Link to={`${APP_PATHS.ACCELERATOR_SPECIES.replace(':speciesId', row.speciesId.toString())}`}>
+      <Link
+        onClick={() => {
+          setCurrentParticipantProjectSpecies(row.id);
+          navigate(
```
nickgraz, line 51: Can you add this to https://github.com/terraware/terraware-web/blob/main/src/hooks/useNavigateTo.ts#L7?

src/scenes/AcceleratorRouter/Deliverables/index.tsx
```diff
@@ -15,7 +18,33 @@ const DeliverablesRouter = () => {
         element={
           <ProjectProvider>
             <ParticipantProvider>
-              <DeliverableViewWrapper />
+              <ParticipantProjectSpeciesProvider>
```
nickgraz, line 21: Can you move this lower in the tree? Ideally this provider only wraps species related views. I think the "[currentDeliverable](https://github.com/terraware/terraware-web/pull/2596/files#diff-08120ee089f7fa1ecb53e3f3942ddf0fbccd748b043a9f92ee4ae98c341e90e0R22)" logic can be moved to the `ParticipantProvider`. This way you can do `const { currentDeliverable } = useParticipantData()` within the `ParticipantProjectSpeciesProvider`.

In order to do this, you will need to move the "get current deliverable from params" [code](https://github.com/terraware/terraware-web/pull/2596/files#diff-d1ed142f473884830fdf70b706960a57f9f9969f94e42deb200a8a081fcf8389R37) within the `DeliverableViewWrapper` into the `ParticipantProvider`. 

That provider currently has information about active module, phase, etc... so I think it makes sense to also add active deliverable there too.

Another option, that might be better but slightly more work, is to create a new `DeliverableProvider` that manages the current deliverable.

In the router:
```
<ProjectProvider>
  <ParticipantProvider>
    <DeliverableProvider>
      <DeliverableViewWrapper />
    </DeliverableProvider>
  </ParticipantProvider>
</ProjectProvider>
```

In the wrapper:
```
        return (
          <ParticipantProjectSpeciesProvider>
            <SpeciesDeliverableView deliverable={deliverable} />
          </ParticipantProjectSpeciesProvider>
        );
```

src/scenes/AcceleratorRouter/Species/RejectedBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+const useStyles = makeStyles(() => ({
+  icon: {
+    marginLeft: '-1px',
+    marginTop: '-1px',
+  },
+}));
+
+interface RejectedBoxProps {
+  participantProjectSpecies: ParticipantProjectSpecies;
+  onSubmit: (feedbaxck: string) => void;
```
nickgraz, line 22:
```suggestion
  onSubmit: (feedback: string) => void;
```

src/providers/ParticipantProject/ParticipantProjectSpeciesProvider.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const dispatch = useAppDispatch();
+
+  const { projectId } = useProjectData();
+
+  const [currentParticipantProjectSpecies, setCurrentParticipantProjectSpecies] = useState<ParticipantProjectSpecies>();
+  const [currentDeliverable, setCurrentDeliverable] = useState<Deliverable>();
+  const participantProjectSpeciesApi = useAppSelector(selectParticipantProjectSpeciesListRequest(projectId));
+
+  const reload = useCallback(() => {
+    void dispatch(requestListParticipantProjectSpecies(projectId));
```
nickgraz, line 26: Can you pull the current species by using the path param instead of the requesting the entire list? You can use `requestGetParticipantProjectSpecies(participantProjectSpeciesId)`

src/scenes/AcceleratorRouter/Deliverables/index.tsx
constanzauanini: Sure, I will create DeliverableProvider, sounds better

constanzauanini: I tried a lot of different ways to have the ParticipantProjectSpeciesProvider only before the <SpeciesDeliverableView> but then when I tried to get the ParticipantProjectSpecies in the others view, it was always undefined. Maybe I'm doing something wrong. 

I added new routers to make it easier to follow

constanzauanini: In the new views I created for View species (src/scenes/AcceleratorRouter/Species/index.tsx) and edit species (src/scenes/AcceleratorRouter/Species/SpeciesEditView.tsx)

nickgraz: Can you take a look at this branch and see if it works for you? https://github.com/terraware/terraware-web/pull/2623/files

nickgraz: Oh nevermind, I think I see the issue... will push up another commit here in a bit

constanzauanini: No, it doesn't work. It still doesn't get the participantPojectSpecies correctly, because it is set in the Table (that it is inside the SpeciesDeliverableView, inside the delierableViewWrapper). So we need the provider there too

nickgraz:
I merged in my branch that was based on this, I'll fix up the merge conflicts and get this merged in!

nickgraz (approved):

PR #2603, opened 2024-05-09
SW-5356 Added modules dispatch into ParticipantProvider, used projects with modules only for Participant Experience

tommylau523:


tommylau523:
* **#2603** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2603?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/providers/Participant/ParticipantProvider.tsx
```diff
@@ -20,13 +25,27 @@ const ParticipantProvider = ({ children }: Props) => {
...
   const projects = useAppSelector(selectProjects);
+  const allModules = useAppSelector((state) =>
+    selectAllModuleList(
+      state,
+      (projects || []).map((project) => project.id)
+    )
+  );
+
+  useEffect(() => {
+    if (projects && projects.length > 0) {
```
tommylau523, line 44: Some deeper issues with our selectors here. We need to refresh when our dispatch is completed here so that we can accurately determine which projects have modules, and show the screens accordingly. 

May need to rework our selectors a bit for this.

nickgraz: I think we should use the async thunks for this, that way we can know when the request is completed.

nickgraz: If you remove `, true` from [this](https://github.com/terraware/terraware-web/blob/main/src/redux/features/modules/modulesSlice.ts#L32), you can get the request ID and use that to determine if the request has been satisfied or not.

constanzauanini (approved):
Looks good. 

Not a blocker, but just noticed that when changing from an org with no modules to one with modules, it takes some time to update the home page, maybe we could show a spinner while fetching?

tommylau523:
We have this ticket tracking it:
https://terraformation.atlassian.net/browse/SW-5331

PR #2606, opened 2024-05-10
SW-5331 Used search to imporve find module projects performances

tommylau523:
This fixes a slow down in the home screen, which is very visible especailly when switch organizations.

### DIFF
--- a/src/providers/Participant/ParticipantProvider.tsx
+++ b/src/providers/Participant/ParticipantProvider.tsx
@@ -1,8 +1,8 @@
 import React, { useCallback, useEffect, useState } from 'react';
 
 import { useLocalization, useOrganization } from 'src/providers/hooks';
-import { requestListAllModules, requestListModules } from 'src/redux/features/modules/modulesAsyncThunks';
-import { selectAllProjectModuleList, selectProjectModuleList } from 'src/redux/features/modules/modulesSelectors';
+import { requestListModuleProjects, requestListModules } from 'src/redux/features/modules/modulesAsyncThunks';
+import { selectModuleProjects, selectProjectModuleList } from 'src/redux/features/modules/modulesSelectors';
 import { requestGetParticipant } from 'src/redux/features/participants/participantsAsyncThunks';
 import { selectParticipant } from 'src/redux/features/participants/participantsSelectors';
 import { selectProjects } from 'src/redux/features/projects/projectsSelectors';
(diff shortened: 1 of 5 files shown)


tommylau523:
* **#2606** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2606?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/providers/Participant/ParticipantProvider.tsx
```diff
@@ -90,21 +89,19 @@ const ParticipantProvider = ({ children }: Props) => {
...
-    if (allProjectModuleList && allProjectModuleList.status === 'success' && allProjectModuleList.data) {
-      const allProjectModules = allProjectModuleList.data;
-      const nextModuleProjects = (participantProjects || []).filter(
-        (project) =>
-          allProjectModules.findIndex(({ id, modules }) => id === project.id && modules && modules.length > 0) !== -1
-      );
+    if (moduleProjectsList && moduleProjectsList.status === 'success' && moduleProjectsList.data) {
+      const nextModuleProjects = moduleProjectsList.data
+        .map((id) => participantProjects.find((project) => project.id === id))
+        .filter((project): project is Project => !!project);
```
constanzauanini, line 95: Is this filter necessary?  Can a project from participant projects not be a project?

joshbuchea: Good question, I don't believe we're filtering like this elsewhere for projects so it doesn't seem like it should be necessary here.

tommylau523: The type will end up being Project | undefined because of the behavior of the ‘find’ method. We do use filter with type predicates in some areas.

tommylau523: This ends up with a TS2322 error otherwise: https://terraformationteam.slack.com/archives/C01C69KA23Y/p1712783339082759

src/providers/Participant/ParticipantProvider.tsx
```diff
@@ -90,21 +89,19 @@ const ParticipantProvider = ({ children }: Props) => {
...
 
   useEffect(() => {
-    if (allProjectModuleList && allProjectModuleList.status === 'success' && allProjectModuleList.data) {
-      const allProjectModules = allProjectModuleList.data;
-      const nextModuleProjects = (participantProjects || []).filter(
-        (project) =>
-          allProjectModules.findIndex(({ id, modules }) => id === project.id && modules && modules.length > 0) !== -1
-      );
+    if (moduleProjectsList && moduleProjectsList.status === 'success' && moduleProjectsList.data) {
+      const nextModuleProjects = moduleProjectsList.data
```
nickgraz, line 93: I wonder if you can satisfy all the data with the search query alone. This is fine how you have it, but it seems like we might be able to get everything in one shot.

nickgraz (approved):
One optional thing to consider as an improvement, but not required to merge this in.

src/providers/Participant/ParticipantProvider.tsx
tommylau523: Should be able to! This is likely a backlog ticket for improving our API usages in general. Maybe we can start backlog of what API calls can be consolidated.

tommylau523:
### Merge activity

* **May 13, 6:44 PM EDT**: @tommylau523 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/2606).

PR #2610, opened 2024-05-10
SW-5339: Deliverables: Status - change "Rejected" to "Not Accepted"

joshbuchea:
This PR includes changes to update the deliverable status label from "Rejected" to "Not Accepted" for participant users.

## Screenshots

localhost_accelerator_deliverables (1)

localhost_accelerator_deliverables (2)


tommylau523 (approved):
Looks good to me! Do you think we need a new state in the backend tables? That change should come in a different PR anyways. One case I can think of is how search APIs end up working.

joshbuchea:
@tommylau523 great question — I think the intention may have been to limit the scope of the work to a simple FE UI update, but I'm not certain. So I asked Jennifer in a comment on the Jira ticket: https://terraformation.atlassian.net/browse/SW-5339

joshbuchea:
@tommylau523 for the case you mentioned with the search API, are additional FE updates needed to handle that case?

tommylau523:
I think let's keep it a FE only change for now. When we get to updating the BE tables, we have to make sure the FE types and requests get updated accordingly.

PR #2634, opened 2024-05-17
Hide projects table in accelerator species edit

constanzauanini:


nickgraz:
I think we want to keep the table here for all routes, if the org has a participant. It should be visible to participants as well.

https://www.figma.com/design/OlTouFTKgxUPX8ejhY6mMv/Participant-and-Accelerator-Workflows?node-id=4555-268130&t=iFfVb9uvnh8Z9rLK-0

constanzauanini:
> I think we want to keep the table here for all routes, if the org has a participant. It should be visible to participants as well.
> 
> https://www.figma.com/design/OlTouFTKgxUPX8ejhY6mMv/Participant-and-Accelerator-Workflows?node-id=4555-268130&t=iFfVb9uvnh8Z9rLK-0

When showing the edit species view from the console side, you get to that page having a "Selected project" (Project Andromeda in the figma file). So I think it doesn't make sense to show the projects there. This is the view I'm refering to: https://www.figma.com/file/OlTouFTKgxUPX8ejhY6mMv/Participant-and-Accelerator-Workflows?type=design&node-id=5028-138715&mode=design&t=4pvMD8xYTfwrGIWv-0 

But I'm not sure. Maybe we can leave it for now

nickgraz:
Oh! I see. My bad, looks great!

nickgraz (approved):

PR #2635, opened 2024-05-17
SW-5385 Add list icon on Project table in Species View

constanzauanini:
Screenshot 2024-05-17 at 12 45 25 PM

### DIFF
--- a/src/services/ParticipantProjectSpeciesService.ts
+++ b/src/services/ParticipantProjectSpeciesService.ts
@@ -1,7 +1,9 @@
+import { DateTime } from 'luxon';
+
 import { components, paths } from 'src/api/types/generated-schema';
+import { DeliverableTypeType, getDeliverableTypeLabel } from 'src/types/Deliverables';
 import { SearchNodePayload, SearchRequestPayload, SearchSortOrder } from 'src/types/Search';
 
-import DeliverablesService from './DeliverablesService';
 import HttpService, { Response2 } from './HttpService';
 import SearchService from './SearchService';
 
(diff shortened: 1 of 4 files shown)


src/services/ParticipantProjectSpeciesService.ts
```diff
@@ -190,14 +192,37 @@ const getProjectsForSpecies = async (speciesId: number, organizationId: number):
   const apiSearchResults = (await SearchService.search(searchParams)) as SpeciesProjectsSearchResult[];
   // it will always be one result since we are searching by species id
   const firstApiResult = apiSearchResults[0];
```
nickgraz, line 194: I just noticed that this has the possibility of returning `undefined`, in which case this would fail during runtime. `SearchService.search` returns `SearchResponseElement[] | undefined`.

src/services/ParticipantProjectSpeciesService.ts
```diff
@@ -190,14 +192,37 @@ const getProjectsForSpecies = async (speciesId: number, organizationId: number):
...
-  const projects = firstApiResult?.participantProjectSpecies.map((pps) => {
-    return {
-      submissionStatus: pps.submissionStatus,
-      projectName: pps.project.name,
-      projectId: pps.project.id,
-      id: pps.id,
-    } as SpeciesProjectsResult;
-  });
+  const projects = await Promise.all(
+    firstApiResult?.participantProjectSpecies.map(async (pps) => {
```
nickgraz, line 196: I think this can all be done in a single request to the search API. I will put up a PR

nickgraz: @constanzauanini let me know what you think https://github.com/terraware/terraware-web/pull/2638

nickgraz (approved):

PR #2637, opened 2024-05-17
SW-2243 Web-App build should flag unused string constants from the strings files

constanzauanini:
- Added a script that prints unused strings. It will not remove them, just log them
- It will run when generating strings.
- Removed all unused strings up to this point.


nickgraz (approved):
This is awesome! Do we need to remove these in Phrase later or are they automatically removed?

joshbuchea:
Awesome, love the clean up of all of the unused strings! 💯

constanzauanini:
> This is awesome! Do we need to remove these in Phrase later or are they automatically removed?

We need to remove them in phrase. I don't know how to do that, do I need a user?

PR #2640, opened 2024-05-18
SW-5023: Accessions: Ability to add and edit photo files in web app

joshbuchea:
This PR adds the ability to add photos to accessions.

NOTE: This PR currently includes changes to support adding photos when creating accessions, will follow-up with changes to support removing photos after consulting with product and/or design.

## Example

https://github.com/terraware/terraware-web/assets/1474361/18001cb5-b1c7-4436-b385-302378104cd1


src/components/common/SelectPhotos.tsx
```diff
@@ -13,10 +13,10 @@ export default function SelectPhotos(props: PhotoChooserProps): JSX.Element {
   return (
     <PhotoChooser
       {...props}
-      uploadText={strings.UPLOAD_PHOTOS}
```
constanzauanini, line 16: We are using this selector for withdrawals and reports. Are we sure we want to change the upload text in all the views?

constanzauanini (approved):

joshbuchea:
@constanzauanini I re-requested a review from you since I made quite a few changes adding the ability to add/remove photos when editing accessions, thank you in advance 🙌

PR #2650, opened 2024-05-21
SW-5407: Use updated Tab components for Accession Details and Reports

joshbuchea:
This PR includes changes to use the `Tabs` component anywhere that tabs are being rendered with other components.

### DIFF
--- a/src/components/Reports/ReportsView.tsx
+++ b/src/components/Reports/ReportsView.tsx
@@ -1,16 +1,14 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
-import { TabContext, TabList, TabPanel } from '@mui/lab';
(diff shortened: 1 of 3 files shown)


constanzauanini (approved):
Looks good!

nickgraz (approved):
This looks great. One thing you might consider looking at is the `useStickyTabs` [hook](https://github.com/terraware/terraware-web/blob/main/src/utils/useStickyTabs.ts). It consolidates a fair bit of the tab logic that ends up in views and has the benefit of making the tabbed-view of "remembering" what tab they were on. Here's an example of it being used - https://github.com/terraware/terraware-web/blob/057360a73995bcf4fc0fd52844ea82e500bbaa38/src/scenes/AcceleratorRouter/Overview/index.tsx#L38-L70

joshbuchea:
Great suggestion @nickgraz! Will update to use the `useStickyTabs` hook.

PR #2658, opened 2024-05-22

terraware-phrase:


nickgraz:
@constanzauanini does this look correct? There seem to be a lot of deletions. Also `SPECIES_LIST` is still in the source code so I'm a bit confused why this PR would be removing that.

constanzauanini:
> @constanzauanini does this look correct? There seem to be a lot of deletions. Also `SPECIES_LIST` is still in the source code so I'm a bit confused why this PR would be removing that.

Removed strings seems fine. They are all the unused strings. But don't know why SPECIES_LIST is removed. Let me check that

constanzauanini:
@alexTerraformation I think we missed removing this two strings from Phrase: RECEIVED_ON and RECEIVING_NURSERY_REQUIRED (that's why the are added again in this PR), and maybe we removed SPECIES_LIST in phrase accidentally ?

sgrimm:
Closing this PR and will have Phrase make a new one to clean up merge conflicts

PR #2664, opened 2024-05-23
SW-5418 Add native/non-native to edit species modal and other changes

constanzauanini:
This requires BE changes

### DIFF
--- a/src/api/types/generated-schema.ts
+++ b/src/api/types/generated-schema.ts
@@ -1709,6 +1709,7 @@ export interface components {
       rationale?: string;
       /** Format: int64 */
       speciesId: number;
+      speciesNativeCategory?: "Native" | "Non-native";
     };
     CreateParticipantRequestPayload: {
       /**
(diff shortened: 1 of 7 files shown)


src/api/types/generated-schema.ts
```diff
@@ -103,10 +103,6 @@ export interface paths {
     /** Gets all species associated to a participant project. */
     get: operations["getSpeciesForProject"];
   };
-  "/api/v1/accelerator/projects/{projectId}/species/snapshots/{deliverableId}": {
```
constanzauanini, line 106: @nickgraz this endpoint is deleted in the backend and I'm getting an error in the build because it is used in the FE. Was the removal intentional? Are we missing removing something in the FE?

nickgraz: It hasn't been merged in yet, sorry. It should be merged in today

nickgraz (approved):
Looks good! Sorry you had to wait on the BE stuff

PR #2665, opened 2024-05-23
SW-5023: Accessions: Ability to add and edit photo files in web app

joshbuchea:
This PR adds the ability to add/remove accession photos.


tommylau523:
Added DELETE endpoint here:
https://github.com/terraware/terraware-server/pull/2012

PR #2674, opened 2024-05-29
SW-5417: Remove makeStyles everywhere in Terraware web (#1)

joshbuchea:

NOTE: Will break changes for this ticket out into multiple PRs, with this being the first.


nickgraz:
Nice work, this is awesome and much needed

nickgraz (approved):

constanzauanini:
Thanks for doing this @joshbuchea !

PR #2676, opened 2024-05-31
SW-5428: Saving while adding Project to species fails with 404

joshbuchea:
This PR includes a change to disable the "Add Species to Project" button in the `SpeciesDeliverableTable`, if the deliverable is not "current".


joshbuchea:
I think these changes work to indirectly determine if the species deliverable's module is active, by checking to see if the species deliverable id exists within `currentDeliverables` (it looks like `currentDeliverables` contains deliverables for the active module). But please let me know if these changes won't work like I think they will, thanks!

tommylau523 (approved):

PR #2678, opened 2024-05-31
SW-5470 Don't overflow container if text is too long

alexTerraformation:
Grid entry wrapper requires a height.  Use a regular Grid item instead with a minHeight to ensure we can grow vertically with additional text


tommylau523 (approved):

nickgraz (approved):
Nice

PR #2686, opened 2024-06-03
SW-5417: Remove makeStyles everywhere in Terraware web (#4)

joshbuchea:

NOTE: Will break changes for this ticket out into multiple PRs, with this being the **fourth**.

### DIFF
--- a/src/scenes/NoOrgRouter/index.tsx
+++ b/src/scenes/NoOrgRouter/index.tsx
@@ -1,8 +1,6 @@
 import React from 'react';
 import { Navigate, Route, Routes } from 'react-router-dom';
 
-import { makeStyles } from '@mui/styles';
-
 import NoOrgLandingPage from 'src/components/emptyStatePages/NoOrgLandingPage';
 import { APP_PATHS } from 'src/constants';
 import { useUser } from 'src/providers';
@@ -10,25 +8,13 @@ import MyAccountRouter from 'src/scenes/MyAccountRouter';
 import OptInFeaturesView from 'src/scenes/OptInFeatures';
 import useEnvironment from 'src/utils/useEnvironment';
 
-const useStyles = makeStyles(() => ({
-  container: {
-    '& .MuiAppBar-root': {
-      width: '100%',
-    },
-  },
-}));
-
 const NoOrgRouter = () => {
   const { isProduction } = useEnvironment();
   const { reloadUserPreferences: reloadPreferences } = useUser();
-  const classes = useStyles();
 
   return (
     <Routes>
-      <Route
-        path={`${APP_PATHS.MY_ACCOUNT}/*`}
-        element={<MyAccountRouter className={classes.container} hasNav={false} />}
-      />
+      <Route path={`${APP_PATHS.MY_ACCOUNT}/*`} element={<MyAccountRouter hasNav={false} />} />
       <Route path={APP_PATHS.WELCOME} element={<NoOrgLandingPage />} />
       {!isProduction && <Route path={APP_PATHS.OPT_IN} element={<OptInFeaturesView refresh={reloadPreferences} />} />}
       <Route path='*' element={<Navigate to={APP_PATHS.WELCOME} />} />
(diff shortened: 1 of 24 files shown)


src/scenes/NoOrgRouter/index.tsx
```diff
@@ -1,34 +1,20 @@
...
   const { reloadUserPreferences: reloadPreferences } = useUser();
-  const classes = useStyles();
 
   return (
     <Routes>
-      <Route
-        path={`${APP_PATHS.MY_ACCOUNT}/*`}
-        element={<MyAccountRouter className={classes.container} hasNav={false} />}
-      />
+      <Route path={`${APP_PATHS.MY_ACCOUNT}/*`} element={<MyAccountRouter hasNav={false} />} />
```
constanzauanini, line 17: Don't we need the container class here?

src/scenes/PlantsDashboardRouter/components/LiveDeadPlantsPerSpeciesCard.tsx
```diff
@@ -88,8 +80,12 @@ export default function LiveDeadPlantsPerSpeciesCard({
             label=''
             options={allSpecies}
             selectedValue={selectedSpecies}
-            className={classes.maxDropdownWidth}
             fullWidth={true}
+            selectStyles={{
+              inputContainer: {
+                maxWidth: '228px',
```
constanzauanini, line 86: is this working correctly? Because inputContainer style is added to the div that contains the input og the dropdown, but the className is added to the div that wraps the whole component (outside the label)

constanzauanini (commented):
Looks good, just a couple of questions

src/scenes/NoOrgRouter/index.tsx
joshbuchea: Good catch! 🙌 💯

So for this one I wasn't sure how I was going to apply the styles to the `Route` component since it doesn't support `sx` props and we have no control over this component as it comes from `react-router-dom`...

When I looked into updating the `MyAccountRouter` component, I discovered the `className` prop is passed down through several layers of components (`MyAccountRouter` > `MyAccountView` > `MyAccountContent`), but is ultimately never used.

src/scenes/PlantsDashboardRouter/components/LiveDeadPlantsPerSpeciesCard.tsx
joshbuchea: Thank you flagging this 🙌 I was hoping it would and I believe I checked the component looked right in Storybook, but  I'll take another look and make sure this is working and fix if not.

joshbuchea:
Thank you for the review @constanzauanini! 🙌

src/scenes/PlantsDashboardRouter/components/LiveDeadPlantsPerSpeciesCard.tsx
joshbuchea: Just realized I actually checked a similar change in another component, but not this component. Will still take another look and fix if needed.

src/scenes/NoOrgRouter/index.tsx
nickgraz: Everything looks good in the views under that router, perhaps it was not actually being used.

src/scenes/PlantsDashboardRouter/components/LiveDeadPlantsPerSpeciesCard.tsx
nickgraz: I think you may want to add another "select style" prop like in [this PR](https://github.com/terraware/web-components/commit/415e6ba5b9900296881c214f1ad50bd6fce9fe9a#diff-845080ba173a86b66b8d07537880ffd2a1e0136b6f2a6505f5a1a843be016ae8R11-R17) for `selectContainer` and then apply it to [this](https://github.com/terraware/web-components/commit/415e6ba5b9900296881c214f1ad50bd6fce9fe9a#diff-845080ba173a86b66b8d07537880ffd2a1e0136b6f2a6505f5a1a843be016ae8R235) top level div.

joshbuchea: Ok I just tested this change locally and it appears to limit the width of the dropdown 🎉 Thanks again for flagging Constanza, I like confirming changes work as expected 🙌

joshbuchea:
Thank you for your review also @nickgraz, I appreciate it! 🙌

PR #2703, opened 2024-06-05
SW-5417: Remove makeStyles everywhere in Terraware web (#6)

joshbuchea:

NOTE: Will break changes for this ticket out into multiple PRs, with this being the **sixth** (and FINAL PR 🤞). Sorry this one is a bit larger than the others 😅


src/components/Map/MapRenderUtils.tsx
```diff
@@ -1,88 +1,9 @@
...
-    '& .mapboxgl-popup-close-button': {
-      display: 'none',
-    },
-  },
-});
-
-/**
- * Species / plants renderer
- */
-export function useSpeciesPlantsRenderer(subzonesWithPlants: any): MapPopupRenderer {
```
constanzauanini, line 50: Is this not used?

constanzauanini (approved):
This was a big one! Lgtm!

src/components/Map/MapRenderUtils.tsx
joshbuchea: Surprisingly no, it was only referenced in one other file where it was just being imported and re-exported

joshbuchea:
Looks like there's a conflict now, will fix

PR #2707, opened 2024-06-07
SW-5502: FE - Create utility to check users consent status

joshbuchea:
This PR includes changes to support getting/setting the current user's cookie consent preferences.

**NOTE:** Cookie consent preference is returned as properties of the user object, so all that was required to get the user's cookie consent preference was to update the `getUser` method to include the new properties. We also need a way to set the user's cookie consent preference, so changes are included to support that.

### DIFF
--- a/src/api/types/generated-schema.ts
+++ b/src/api/types/generated-schema.ts
@@ -770,10 +770,6 @@ export interface paths {
     /** Creates a new planting site. */
     post: operations["createPlantingSite"];
   };
-  "/api/v1/tracking/sites/validate": {
-    /** Validates the definition of a new planting site. */
-    post: operations["validatePlantingSite"];
-  };
   "/api/v1/tracking/sites/{id}": {
     /**
      * Gets information about a specific planting site.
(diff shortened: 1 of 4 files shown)


src/api/types/generated-schema.ts
```diff
@@ -8602,6 +8652,22 @@ export interface operations {
       };
     };
   };
+  /** Validates the definition of a new planting site. */
+  validatePlantingSite: {
```
nickgraz, line 8656: Is this supposed to be in here?

src/services/UserService.ts
```diff
@@ -55,21 +55,25 @@ const cachedTimeZone: CachedTimeZone = {
...
-          globalRoles: data.user.globalRoles,
-          id: data.user.id,
-          lastName: data.user.lastName,
-          locale: data.user.locale,
-          timeZone: data.user.timeZone,
-        }
-      : undefined,
-  }));
+  const response: UserResponse = await httpCurrentUser.get<UserServerResponse, UserData>({}, (data) => {
+    return {
```
nickgraz, line 59: Is this necessary?

nickgraz (approved):
Looks pretty good, just a couple small questions but nothing blocking.

src/api/types/generated-schema.ts
joshbuchea: I re-generated the types and just pulled in all of the changes, but I'll remove the unrelated changes to keep the PR clean.

src/services/UserService.ts
joshbuchea: Yeah this change is necessary to include the new cookie consent properties in the user object

PR #2713, opened 2024-06-07
SW-5437 Add accession GPS coordinates to table view

sgrimm:
Allow users to add a "Latitude, Longitude" column to the accessions table to show
the list of GPS coordinates from the seed collection. This column will appear in
exported CSV files, with newline-separated values in the case of multiple
coordinates for an accession.

### DIFF
--- a/src/components/seeds/database/TableCellRenderer.tsx
+++ b/src/components/seeds/database/TableCellRenderer.tsx
@@ -84,5 +84,10 @@ export default function SearchCellRenderer(props: RendererProps<SearchResponseEl
     return <CellRenderer index={index} column={column} value={`${value}%`} row={row} />;
   }
 
+  if (column.key === 'geolocations.coordinates') {
+    const coordinatesValue = ((row.geolocations || []) as any[]).map((gl) => gl.coordinates).join('; ');
+    return <CellRenderer index={index} column={column} value={coordinatesValue} row={row} />;
+  }
+
   return <CellRenderer {...props} />;
 }
--- a/src/components/seeds/database/TableCellRenderer.tsx
+++ b/src/components/seeds/database/TableCellRenderer.tsx
@@ -4,9 +4,11 @@ import { FiberManualRecord } from '@mui/icons-material';
 import { Box, TableCell, Typography, useTheme } from '@mui/material';
 
 import Link from 'src/components/common/Link';
+import TextTruncated from 'src/components/common/TextTruncated';
 import CellRenderer from 'src/components/common/table/TableCellRenderer';
 import { RendererProps } from 'src/components/common/table/types';
 import { APP_PATHS } from 'src/constants';
+import strings from 'src/strings';
 import { SearchResponseElement } from 'src/types/Search';
 
 export default function SearchCellRenderer(props: RendererProps<SearchResponseElement>): JSX.Element {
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#2713** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2713?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/components/seeds/database/index.tsx
```diff
@@ -368,7 +368,26 @@ export default function Database(props: DatabaseProps): JSX.Element {
...
+                coordinatesList = gl.coordinates;
+              } else {
+                coordinatesList += `\r${gl.coordinates}`;
+              }
+            });
+
+            return {
+              ...result,
+              'geolocations.coordinates': coordinatesList,
+            };
```
sgrimm, line 387: This is modeled after the way we handle the list of sub-locations in the nursery inventory table (InventoryListByBatch.tsx).

nickgraz: You can handle the table view display by modifying this table's "table cell renderer" [here](https://github.com/terraware/terraware-web/blob/main/src/components/seeds/database/TableCellRenderer.tsx).

Something like this should get you close:
```
if (column.key === 'geolocations.coordinates') {
    const coordinatesValue = row.geolocations?.map((gl) => gl.coordinates).join('\r');
    return <CellRenderer index={index} column={column} value={coordinatesValue} row={row} />;
}
```

nickgraz: I suspect the `'\r'` should get replaced with a `', '`, the `\r` won't show up as intended in the UI. Alternatively you could use `<br>` if you are truly trying to break up the text within the table cell.

sgrimm: Thanks! Didn't know there was a renderer like that. Should we change the nursery inventory table to use that approach too?

I changed the separator to a semicolon since there are commas in the values. Tried `<br>` but it got escaped, and anyway I don't think we want to make the rows variable-height.

nickgraz: Yes, we should change that. I'll make a tech debt ticket, thanks.

src/components/seeds/database/TableCellRenderer.tsx
```diff
@@ -84,5 +84,10 @@ export default function SearchCellRenderer(props: RendererProps<SearchResponseEl
     return <CellRenderer index={index} column={column} value={`${value}%`} row={row} />;
   }
 
+  if (column.key === 'geolocations.coordinates') {
+    const coordinatesValue = ((row.geolocations || []) as any[]).map((gl) => gl.coordinates).join('; ');
+    return <CellRenderer index={index} column={column} value={coordinatesValue} row={row} />;
```
nickgraz, line 89: This is great. Sorry I did not mention this earlier, but if there is a chance that this text might be too long for the table cell, you can use the `TextTruncated` component.

[Here's](https://github.com/terraware/terraware-web/blob/dd8702d485590e41233c7cbef0d5a9bf945ba7ea/src/scenes/Species/TableCellRenderer.tsx#L119) an example.

nickgraz (approved):
Looks great, using the `TextTruncated` component isn't strictly necessary but we might consider if it seems like the cell values are too wide.

src/components/seeds/database/TableCellRenderer.tsx
sgrimm: Ah, TextTruncated is perfect for this. Thanks.

PR #2719, opened 2024-06-11
SW-5503: FE - Update My Account page to include cookie consent status & SW-5501: FE - Add cookie consent banner to home page

joshbuchea:
This PR includes changes to:

- Display and/or update the current user's cookie consent status in the My Account view
- Show new cookie consent banner to user's have not yet consented to cookies using the new banner

### Banner (desktop)

localhost_3000_myaccount_organizationId=41

### My Account

localhost_3000_myaccount_organizationId=41 (2)


alexTerraformation:
Tried this on the vercel deployment.  Seems good!  One small change to make is that we shouldn't show the banner anymore if/when the user declines cookies.  The behavior should be the same whether user accepts or declines cookies.

joshbuchea:
@alexTerraformation oops! That definitely was the intention, but I see the logic needs to be adjusted. Thanks for trying this out 🙌

src/components/CookieConsentBanner.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  useEffect(() => {
+    const cookiesConsentedTimeIsGreaterThanOneYear = user?.cookiesConsentedTime
+      ? Date.now() - new Date(user.cookiesConsentedTime).valueOf() > ONE_YEAR_IN_MILLISECONDS
+      : false;
+
+    if (user && (typeof user?.cookiesConsented !== 'boolean' || cookiesConsentedTimeIsGreaterThanOneYear)) {
+      setVisible(true);
+    }
+  }, [user]);
```
nickgraz, line 41: It seems like you can return out early if there's no `user.cookiesConsentedTime`

Something like this might be easier to read:
```
useEffect(() => {
    if (!(user && user.cookiesConsentedTime)) {
      return;
    }

    const cookiesConsentedTimeIsGreaterThanOneYear =
      Date.now() - new Date(user.cookiesConsentedTime).valueOf() > ONE_YEAR_IN_MILLISECONDS;

    if (cookiesConsentedTimeIsGreaterThanOneYear) {
      setVisible(true);
    }
  }, [user]);
```

src/scenes/MyAccountRouter/MyAccountView.tsx
```diff
@@ -205,6 +208,7 @@ const MyAccountContent = ({
     setPreferredWeightSystemSelected((userPreferences?.preferredWeightSystem as string) || 'metric');
     setLocaleSelected(selectedLocale);
     setSelectedRows([]);
+    onChange('cookiesConsented', user.cookiesConsented);
```
nickgraz, line 211: Shouldn't the value of this already be set? It seems like the change on 555 should be sufficient.

src/services/UserService.ts
```diff
@@ -116,8 +116,9 @@ const updateUser = async (user: User, options: UpdateOptions = {}): Promise<Resp
   }
 
   const response: Response = await httpCurrentUser.put({ entity });
-  if (response.requestSucceeded && typeof user.cookiesConsented === 'boolean') {
-    await PreferencesService.updateUserCookieConsentPreferences({ cookiesConsented: user.cookiesConsented });
+  if (response.requestSucceeded && typeof user.cookiesConsented === 'string') {
```
nickgraz, line 119: When would this be set to a string? It looks like you're dealing with bools in the places where you can interact with it

src/scenes/MyAccountRouter/MyAccountView.tsx
joshbuchea: This resets the form value when cancel is pressed.

src/services/UserService.ts
joshbuchea: I discovered the value is cast to a string by the `RadioGroup` component and this wasn't working as expected.

src/components/CookieConsentBanner.tsx
joshbuchea: This won't work as-is because we also need to display the cookie consent banner when `user.cookiesConsented` is `null` (before the user has made a selection). I'll see what I can do to improve how this reads using your suggestion.

joshbuchea: We also need to ask the user again after one year has passed.

src/scenes/MyAccountRouter/MyAccountView.tsx
nickgraz: But aren't you navigating away from the form? Is the form state still there if you navigate away and come back to the form?

src/components/CookieConsentBanner.tsx
nickgraz: > This won't work as-is because we also need to display the cookie consent banner when `user.cookiesConsented` is `null` (before the user has made a selection). I'll see what I can do to improve how this reads using your suggestion.

I see, it sounds like that functionality should be another use effect.

nickgraz: > if (cookiesConsentedTimeIsGreaterThanOneYear) {
>       setVisible(true);
>     }

Is that not handled by this part?
```
    if (cookiesConsentedTimeIsGreaterThanOneYear) {
      setVisible(true);
    }
```

src/services/UserService.ts
nickgraz: Oh I see, that's a bummer. Is there any way you can rectify that within the component that has the radio group? The `UserService` shouldn't be doing strange things because of the behavior of something that is implementing it. Ideally the strange things, like casting the radio group value to a boolean, are done where the effect occurs in the first place.

src/components/CookieConsentBanner.tsx
joshbuchea: @nickgraz I'm not sure I see why or agree that it needs to be another `useEffect`. It seems perfectly reasonable to handle these two cases in one `useEffect` in my mind since both cases are related to the user's cookie consent.

src/scenes/MyAccountRouter/MyAccountView.tsx
joshbuchea: I'm not sure exactly, but this change was needed otherwise the value stayed modified after pressing cancel in the form.

src/services/UserService.ts
joshbuchea: The `RadioGroup` component is part of the MUI library and this comment is included for the `value` prop in the `RadioGroupProps` type:

> Value of the selected radio button. The DOM API casts this to a string.

joshbuchea: Good point about handling the casting in a more appropriate location, I reverted the change to the `UserService` and handled the issue in the `onChange` handler of the `RadioGroup` component.

src/components/CookieConsentBanner.tsx
joshbuchea: I modified the `useEffect` based on your suggestion to improve readability a bit, but this `useEffect` is working now and the code doesn't seem particularly complex or clever so I think this can stay as-is.

nickgraz: Looks good to me. I generally try to split up business logic into separate effects, because then you can have more targeted effects, which means less re-execution due to changed dependencies. "Show the banner if the user hasn't told us yes or no" and "show the banner if the consent is too old" are separate requirements that are looking at different pieces of data. 

Either way, this is good the way you have it now, I am probably just over thinking it!

nickgraz (approved):
Looks good, thanks for the modifications!

joshbuchea:
@nickgraz you bet, thanks for the review! 🙌

PR #2724, opened 2024-06-11
SW-5032 Remove background illustration

sgrimm:
Remove the background SVG image from the app. This eliminates the
need for some component parameters that only controlled whether or
not the background was used.


sgrimm:
* **#2724** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2724?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/components/common/TfMain.tsx
```diff
@@ -23,28 +22,8 @@ export default function TfMain({ backgroundImageVisible, children, style }: Prop
...
-            : 'url(/assets/home-bg-right-layer-z4.svg) no-repeat 100% 100%/auto 285px, ' +
-              'url(/assets/home-bg-left-layer-z4.svg) no-repeat 0 100%/auto 295px, ' +
-              'url(/assets/home-bg-water-z2.svg) repeat-x 0 100%/auto 180px, ' +
-              'url(/assets/home-bg-left-z4.svg) no-repeat 0 100%/auto 295px, ' +
-              'url(/assets/home-bg-right-z3.svg) no-repeat 100% 100%/auto 400px',
-          backgroundAttachment: 'fixed !important',
-          ...style,
-        },
         flexDirection: 'column',
+        ...style,
```
sgrimm, line 26: This seems to produce the correct visual appearance, but I'm not sure if it's the right change. Previously, we were using a background of `initial` on the breakpoints, and the `...style` was tied to a breakpoint too.

nickgraz: It looks like `...style` was added recently [here](https://github.com/terraware/terraware-web/commit/d72b3a587acecf2b21a70ac902b14c391765c1ea#diff-bf8032494da0bcb13c3a130a194b54182198d3e607977ddef02d91818692a2a1R44). I think your change is correct. The mobile / not mobile determination should be determined by any parent components that are passing down a `style` object. 

I would just give it a double check on mobile to make sure it looks correct. I am assuming it will look correct because the `style` object passed in is determined by the device already [here](https://github.com/terraware/terraware-web/commit/d72b3a587acecf2b21a70ac902b14c391765c1ea#diff-6273b33daf307fa182ac466012236fc4eca750ed3209ae405b0ee9c10c38e9dfR29).

The `Page` component is the only one passing styles down into `TfMain`, and it looks like styles are only passed into `Page` in that one location I linked above.

The `MyAccountContent` component allows for `style` to be passed in, and down into `TfMain`, but it doesn't look like anything is actually passing styles into `MyAccountContent`.

nickgraz (approved):
Looks good, thanks for the deletions!

PR #2731, opened 2024-06-12
SW-5528, SW-5529 Line breaks in profile fields

sgrimm:
Display explicit line breaks in participant project profile textarea
fields so multi-paragraph values aren't folded into a single long
paragraph.

Wrap long words (URLs, etc.) to prevent them from overflowing into
other parts of the UI.

### DIFF
--- a/src/components/ProjectField/TextAreaDisplay.tsx
+++ b/src/components/ProjectField/TextAreaDisplay.tsx
@@ -6,6 +6,8 @@ import { ProjectFieldProps } from '.';
 
 const ProjectFieldTextAreaDisplay = ({ label, value }: ProjectFieldProps) => {
   const theme = useTheme();
+  const valueLines = value ? `${value}`.split('\n') : null;
+  const valueElements = valueLines?.flatMap((line) => [<br key='break' />, line])?.slice(1);
 
   return (
     <Grid
@@ -21,8 +23,13 @@ const ProjectFieldTextAreaDisplay = ({ label, value }: ProjectFieldProps) => {
         <Typography fontSize={'16px'} lineHeight={'24px'} fontWeight={600} marginBottom={theme.spacing(1)}>
           {label}
         </Typography>
-        <Typography fontSize={'16px'} lineHeight={'24px'} fontWeight={400}>
-          {value}
+        <Typography
+          fontSize={'16px'}
+          lineHeight={'24px'}
+          fontWeight={400}
+          sx={{ wordBreak: 'break-word' }}
+        >
+          {valueElements}
         </Typography>
       </Box>
     </Grid>
--- a/src/components/ProjectField/TextAreaDisplay.tsx
+++ b/src/components/ProjectField/TextAreaDisplay.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 
 import { Box, Grid, Typography, useTheme } from '@mui/material';
-import TextField from '@terraware/web-components/components/Textfield/Textfield';
 
 import { ProjectFieldProps } from '.';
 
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#2731** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2731?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/components/ProjectField/TextAreaDisplay.tsx
```diff
@@ -6,6 +6,8 @@ import { ProjectFieldProps } from '.';
 
 const ProjectFieldTextAreaDisplay = ({ label, value }: ProjectFieldProps) => {
   const theme = useTheme();
+  const valueLines = value ? `${value}`.split('\n') : null;
+  const valueElements = valueLines?.flatMap((line) => [<br key='break' />, line])?.slice(1);
```
sgrimm, line 10: This was the cleanest way I could think of to turn newlines into React `<br/>` objects to preserve line breaks in the text, but maybe there's a better approach.

sgrimm: I looked around quickly but didn't see a built-in component to do line-break-preserving text values, but it wouldn't surprise me if there is one and I just missed it.

nickgraz: You could try using the `web-components:TextField` component, there is a `display` prop to show the value instead of an input. 

[Here's an example](https://github.com/terraware/terraware-web/blob/50eb53c2a01c619f5b6acbad5120c772715ed5ad/src/scenes/ObservationsRouter/plot/index.tsx#L122-L129).

The component also has a `preserveNewlines` prop that may be needed.

nickgraz: You can try copy pasting the value you have into the Storybook [here](https://web-components-1nd6ccvl0-terraformation.vercel.app/?path=/story/textfield-new--text-area&args=helperText:;display:!true) and see if it appears as you expect.

src/components/ProjectField/TextAreaDisplay.tsx
```diff
@@ -21,9 +22,19 @@ const ProjectFieldTextAreaDisplay = ({ label, value }: ProjectFieldProps) => {
...
           {label}
         </Typography>
-        <Typography fontSize={'16px'} lineHeight={'24px'} fontWeight={400}>
-          {value}
-        </Typography>
+        {typeof value === 'string' && (
+          <Typography fontSize={'16px'} lineHeight={'24px'} fontWeight={400}>
+            <TextField
+              id=''
+              label=''
```
sgrimm, line 29: There's no actual `<textarea>` being produced here so it didn't seem like the ID or label would be meaningful. If using empty strings is a massive no-no, I can figure out values to put there.

src/components/ProjectField/TextAreaDisplay.tsx
```diff
@@ -21,9 +22,19 @@ const ProjectFieldTextAreaDisplay = ({ label, value }: ProjectFieldProps) => {
         <Typography fontSize={'16px'} lineHeight={'24px'} fontWeight={600} marginBottom={theme.spacing(1)}>
           {label}
         </Typography>
-        <Typography fontSize={'16px'} lineHeight={'24px'} fontWeight={400}>
-          {value}
-        </Typography>
+        {typeof value === 'string' && (
+          <Typography fontSize={'16px'} lineHeight={'24px'} fontWeight={400}>
```
tommylau523, line 26: I think instead of using a TextField, you can just add `whiteSpace={'pre-wrap'}` here.

sgrimm: Oh, nice! Didn't know that existed. It does the trick.

nickgraz (approved):

PR #2734, opened 2024-06-13
Removed support feature flag

tommylau523:


tommylau523:
* **#2734** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2734?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/scenes/HelpSupportRouter/provider/index.tsx
```diff
@@ -29,11 +27,11 @@ const SupportProvider = ({ children }: Props) => {
   });
 
   useEffect(() => {
-    if (featureEnabled) {
+    if (!supportRequestTypes || supportRequestTypes.length < 3) {
```
nickgraz, line 30: Where does this magic number come from?

tommylau523: It is because we have 3 support types total. Let me actually refactor some stuff now to make that clear.

tommylau523: done!

nickgraz: I'm a bit confused, if the `supportRequestTypes` are just strings, which we define as a constant, why not use that? Why are they being pulled from the server?

tommylau523: The server has an endpoint that replies which support types are configured. 

So the client is checking with the server that each of the three types are valid before letting users submit the forms. This is to prevent a error that causes users to not be able to submit a support request.

The types can become invalid if Jira is unavailable for example.

nickgraz: I see. In your example, Jira being unavailable and the server returning 2 out of the 3 support types, wouldn't this keep firing? Because there are only 2 coming back, it would dispatch the list request again, which would update the `supportRequestTypes`, and the cycle would continue, no? Because the `useState` is receiving and array, even if the two arrays _look_ the same, they will still technically be different, which would make this `useEffect` fire again.

I am not sure if this is possible, but something to consider. Perhaps you need to only dispatch this request if `supportRequestTypes.length === 0` and initialize the state with an empty array here - `useState<SupportRequestType[]>([]);`

tommylau523: With that consideration, I will change this to only check for `=== undefined`. The server can return that no support types is valid, which will be an empty list. So we don't want to keep firing for that. 

Thanks for catching this Nick! You are absolutely right that this will cause an infinite firing.

nickgraz: Understood, thanks!

nickgraz (approved):

PR #2743, opened 2024-06-17
SW-5534: Error when deleting Species from active Species Deliverable (works from Species Details page)

joshbuchea:
This PR includes a fix for an issue that was causing an error when attempting to remove a species from a species deliverable.


nickgraz (approved):
🙌

PR #2748, opened 2024-06-18
SW-5533: Can't add Species to Project (from Species page) even when Deliverable is active

joshbuchea:
This PR includes changes to resolve an issue that was preventing the ability to add a project to a species from the species detail view.

### DIFF
--- a/src/redux/features/modules/modulesAsyncThunks.ts
+++ b/src/redux/features/modules/modulesAsyncThunks.ts
@@ -99,12 +99,13 @@ export const requestListModuleDeliverables = createAsyncThunk(
     return deliverableSearchResults.map(
       (result) =>
         ({
+          category: result.category,
           moduleId: result.module_id,
           projectId: result.project_id,
-          ...result,
           dueDate: DateTime.fromISO(result.dueDate),
-          module_id: undefined,
-          project_id: undefined,
+          id: result.id,
+          name: result.name,
+          status: result.status,
           type: result['type(raw)'],
         }) as ModuleDeliverable
     );
--- a/src/redux/features/modules/modulesAsyncThunks.ts
+++ b/src/redux/features/modules/modulesAsyncThunks.ts
@@ -67,7 +67,7 @@ export const requestListModuleDeliverables = createAsyncThunk(
   async (request: { moduleId: number; projectId: number }) => {
     const searchParams: SearchRequestPayload = {
       prefix: 'projects.projectDeliverables',
-      fields: ['id', 'module_id', 'project_id', 'name', 'category', 'dueDate', 'status', 'type'],
+      fields: ['id', 'module_id', 'project_id', 'name', 'category', 'dueDate', 'status', 'type', 'type(raw)'],
       search: {
         operation: 'and',
         children: [
--- a/src/redux/features/modules/modulesAsyncThunks.ts
+++ b/src/redux/features/modules/modulesAsyncThunks.ts
@@ -105,6 +105,7 @@ export const requestListModuleDeliverables = createAsyncThunk(
           dueDate: DateTime.fromISO(result.dueDate),
           module_id: undefined,
           project_id: undefined,
+          type: result['type(raw)'],
         }) as ModuleDeliverable
     );
   }
--- a/src/types/Module.ts
+++ b/src/types/Module.ts
@@ -78,6 +78,7 @@ export type ModuleDeliverable = {
   dueDate: DateTime;
   status: DeliverableStatusType;
   type: DeliverableTypeType;
+  'type(raw)'?: string;
 };
 
 export type ModuleContentType = keyof Pick<Module, 'additionalResources' | 'preparationMaterials'>;
--- a/src/types/Module.ts
+++ b/src/types/Module.ts
@@ -66,7 +66,8 @@ export type ModuleDeliverableSearchResult = {
   category: DeliverableCategoryType;
   dueDate: string;
   status: DeliverableStatusType;
-  type: DeliverableTypeType;
+  type: string;
+  'type(raw)'?: DeliverableTypeType;
 };
 
 export type ModuleDeliverable = {
@@ -78,7 +79,6 @@ export type ModuleDeliverable = {
   dueDate: DateTime;
   status: DeliverableStatusType;
   type: DeliverableTypeType;
-  'type(raw)'?: string;
 };
 
 export type ModuleContentType = keyof Pick<Module, 'additionalResources' | 'preparationMaterials'>;
--- a/src/types/ProjectToDo.ts
+++ b/src/types/ProjectToDo.ts
@@ -57,7 +57,8 @@ export type DeliverableSearchResultType = {
   name: string;
   position: number;
   status: DeliverableStatusType | null;
-  type: DeliverableTypeType;
+  type: string;
+  'type(raw)': DeliverableTypeType;
 };
 
 export class DeliverableToDoItem implements ToDoItem {
@@ -82,7 +83,7 @@ export class DeliverableToDoItem implements ToDoItem {
     this.position = searchResult.position;
     this.projectId = searchResult.project_id;
     this.status = searchResult.status ?? 'Not Submitted';
-    this.type = searchResult.type;
+    this.type = searchResult['type(raw)'];
   }
 
   isCompleted = (): boolean => this.status == 'Approved' || this.status == 'Not Needed';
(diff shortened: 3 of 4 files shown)


src/types/Module.ts
```diff
@@ -78,6 +78,7 @@ export type ModuleDeliverable = {
   dueDate: DateTime;
   status: DeliverableStatusType;
   type: DeliverableTypeType;
```
tommylau523, line 80: type should be `string` here and type(raw) should be `DeliverableTypeType`. But for clarity maybe we can rename `type` to `typeName` and `type(raw)` to `type`.

src/redux/features/modules/modulesAsyncThunks.ts
```diff
@@ -67,7 +67,7 @@ export const requestListModuleDeliverables = createAsyncThunk(
   async (request: { moduleId: number; projectId: number }) => {
     const searchParams: SearchRequestPayload = {
       prefix: 'projects.projectDeliverables',
-      fields: ['id', 'module_id', 'project_id', 'name', 'category', 'dueDate', 'status', 'type'],
+      fields: ['id', 'module_id', 'project_id', 'name', 'category', 'dueDate', 'status', 'type', 'type(raw)'],
       search: {
         operation: 'and',
         children: [
```
tommylau523, line 73: <i class='graphite__hidden'>[Re: line 104]<br/><br/></i>

If you do go with the renaming of variables to match types: 

add above the line:
```
typeName: result.type,
type: result['type(raw)'],
```

below this line: (to remove the remapped fields
```
'type(raw)': undefined
```
<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2748?utm_source=unchanged-line-comment">Graphite</a>.</span>

tommylau523 (approved):

src/types/Module.ts
```diff
@@ -78,6 +78,7 @@ export type ModuleDeliverable = {
   dueDate: DateTime;
```
tommylau523, line 78: <i class='graphite__hidden'>[Re: line 69]<br/><br/></i>

Sorry this was a lot of boilerplate. 

```
type: string
 'type(raw)': DeliverableTypeType
```<!--__GRAPHITE_HTML_TAG_START__--><span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2748?utm_source=unchanged-line-comment">Graphite</a>.</span>

<!--__GRAPHITE_HTML_TAG_END__-->

Comment (tommylau523) [reply src/types/Module.ts:78]: Add if you don't mind, could you also update
src/types/ProjectToDo.ts line 85?

```
this.type = searchResult['type(raw)'];
```

Should prevent the same typing error from happening there as well.

File: src/types/Module.ts (hunk 1 above)
Comment (nickgraz) [reply src/types/Module.ts:80]: I usually try to differentiate between the "search results" type and the general type. I think `ModuleDeliverable` should only have the hydrated and verified `type: DeliverableTypeType`, whereas the `ModuleDeliverableSearchResult` can have `type(raw): string` because that is specific to the search API.

File: src/redux/features/modules/modulesAsyncThunks.ts (hunk 2 above)
Comment (joshbuchea) [reply src/redux/features/modules/modulesAsyncThunks.ts:73]: I revert changes to `ModuleDeliverable` type and modified `ModuleDeliverableSearchResult` type instead, then updated the `.map` to explicitly map each field. I think these changes incorporate most of your feedback and cleans things up a bit, but please let me know if this looks ok to you.

File: src/types/Module.ts (hunk 3 above)
Comment (joshbuchea) [reply src/types/Module.ts]: @tommylau523 you bet, will update 👍

File: src/types/Module.ts (hunk 1 above)
Comment (joshbuchea) [reply src/types/Module.ts]: @nickgraz I agree and that's exactly the change I just made

File: src/types/ProjectToDo.ts [hunk 4]
```diff
@@ -57,7 +57,8 @@ export type DeliverableSearchResultType = {
   name: string;
   position: number;
   status: DeliverableStatusType | null;
-  type: DeliverableTypeType;
+  type: string;
+  'type(raw)': DeliverableTypeType;
```
nickgraz, line 61: Are these reversed?

joshbuchea: I think that's correct because `type` should be a `string` since it contains a localized value of `type(raw)`, but perhaps I'm thinking about it incorrectly or mixed things up a bit here...

nickgraz: If this is the direct search result, everything is a string. Things can only be cast to `DeliverableTypeType`, `number`, or anything that isn't a `string` explicitly through some sort of hydration. It looks like it is used in `projectToDoAsyncThunk` and `ProjectToDo`, but it is cast to the response from `SearchService.search`. So technically these types are incorrect.

nickgraz: In other words, since we are casting the value, and the `SearchResponseElement` type returned from `SearchService.search` allows the cast, we are losing type safety because we are kinda tricking the compiler. This could lead to runtime errors if we attempt to do number things to non-numbers. Also since we aren't using a typeguard, the types that are specific values (like `DeliverableStatusType`, `DeliverableCategoryType`, `DeliverableTypeType`) can actually hold any string here, so a switch statement somewhere else might incorrectly think all cases have been exhausted.

tommylau523: Sorry this was probably my mistake to begin with. I misused casting in order to get search results to fit with our API enum types.

nickgraz: Ideally a service method is responsible for executing the API call and hydrating the response into the expected / usable type. This way you can use "normal" types around the app and contain the Search API response type

joshbuchea: Ok thanks for the explanation @nickgraz 🙌  Based on that what changes would you recommend here?

nickgraz: I think [all of this](https://github.com/terraware/terraware-web/blob/main/src/redux/features/projectToDo/projectToDoAsyncThunk.ts#L15-L57) should go into a service file somewhere (TodoService?) and we should hydrate the result into the expected type similar to [this](https://github.com/terraware/terraware-web/blob/main/src/services/ParticipantsService.ts#L78).

tommylau523: I can get started on refactoring this logic into ToDoService, and you can rebase this PR on top of it. Does that sound okay with you Josh?

joshbuchea: Ok got it, thanks @nickgraz 🙌  And that sounds great, thanks @tommylau523 🙌

joshbuchea: @tommylau523 if you refactor this logic into a ToDoService and wire it up, I'm not sure my PR will be needed any longer. Please feel free to take over this ticket.

tommylau523: See here!
https://github.com/terraware/terraware-web/pull/2751

I was able to add species on an active deliverable. Would appreciate some help with testing as well.

nickgraz: @tommylau523 do you think this PR is needed anymore after your change was merged in?

tommylau523: I think this is not needed!

joshbuchea: Ok so @tommylau523's PR did resolve the issue I was trying to fix in this PR with the types, but it looks like there's also a `404` happening on Staging when I try to add a species to a project from the edit species view.

Your PR cleaned things up a ton, thanks for doing that work @tommylau523 🙌 💯

I'm guessing there's just a path that needs to be updated, but I'll work out the issue with the `404` and open a new PR.

joshbuchea: Ok so I don't think it's a path issue after all, I'm getting this error message when I try to add the species _Actinidia deliciosa_ to the project New project 2001 in the Graziano Tree Farm organization on Staging:

> "Species Deliverable for project 38 not found"

joshbuchea: Perhaps there are projects available in the dropdown that shouldn't be available... 🤔 Will poke around a bit more and work out the issue.

joshbuchea: Ok yeah so I tried again with the “Sampson Road” project and it worked as expected. I think we likely need to filter the list of projects available in the dropdown to a reduced subset.

PR #2750, opened 2024-06-19
SW-5601 Remove GA4 and enable Mixpanel

alexTerraformation:
Disable cookies and event tracking by default.  Enable once we have user cookie settings


src/App.tsx
```diff
@@ -15,21 +17,59 @@ import { useLocalization, useUser } from 'src/providers';
...
 import { getRgbaFromHex } from 'src/utils/color';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
+import useEnvironment from 'src/utils/useEnvironment';
 
 const AcceleratorRouter = React.lazy(() => import('src/scenes/AcceleratorRouter'));
 const TerrawareRouter = React.lazy(() => import('src/scenes/TerrawareRouter'));
 
+// Mixpanel setup
+// Set this to true to enable Mixpanel tracking for the Terraware Dev project
+const enableMixpanelDev = true;
```
alexTerraformation, line 27: Will remove this before merging

src/App.tsx
```diff
@@ -15,21 +17,59 @@ import { useLocalization, useUser } from 'src/providers';
...
-  const { isAllowed } = useUser();
+  const { user, isAllowed } = useUser();
   const { isAcceleratorRoute } = useAcceleratorConsole();
   const theme = useTheme();
+  const mixpanel = useMixpanel();
 
   const [showNavBar, setShowNavBar] = useState(true);
 
+  useEffect(() => {
+    console.log(user?.cookiesConsented);
```
constanzauanini, line 55: Remove logs?

src/components/TopBar/TopBarContent.tsx
```diff
@@ -55,14 +57,21 @@ export default function TopBarContent(props: TopBarProps): JSX.Element | null {
...
 
   const onHandleLogout = () => {
+    mixpanel?.reset();
     window.location.href = `/sso/logout`;
   };
 
+  const handleTopBarClick = () => {
+    mixpanel?.track('TopBarHome', { my_custom_prop: 'foo' });
+    navigate(APP_PATHS.HOME);
+    console.log('Here!');
```
constanzauanini, line 67: Remove

src/components/TopBar/TopBarContent.tsx
```diff
@@ -55,14 +57,21 @@ export default function TopBarContent(props: TopBarProps): JSX.Element | null {
   };
 
   const onHandleLogout = () => {
+    mixpanel?.reset();
     window.location.href = `/sso/logout`;
   };
 
+  const handleTopBarClick = () => {
+    mixpanel?.track('TopBarHome', { my_custom_prop: 'foo' });
```
nickgraz, line 65: 🙌

src/providers/LocalizationProvider.tsx
```diff
@@ -49,19 +49,6 @@ export default function LocalizationProvider({
     fetchTimeZones();
   }, [selectedLocale]);
 
-  useEffect(() => {
```
nickgraz, line 52: Why was this removed?

alexTerraformation: This was used to update the locale of the cookie-script pop-up.  Since we don't have that pop-up anymore we don't need to update it's locale.

src/App.tsx
```diff
@@ -15,21 +17,57 @@ import { useLocalization, useUser } from 'src/providers';
...
+import useEnvironment from 'src/utils/useEnvironment';
 
 const AcceleratorRouter = React.lazy(() => import('src/scenes/AcceleratorRouter'));
 const TerrawareRouter = React.lazy(() => import('src/scenes/TerrawareRouter'));
 
+// Mixpanel setup
+// Set this to true to enable Mixpanel tracking for the Terraware Dev project
+const enableMixpanelDev = true;
+const { isProduction, isStaging, isDev } = useEnvironment();
+const MIXPANEL_TOKEN = isProduction
```
nickgraz, line 29: Instead of determining which env var to use here, and exposing all of them, we should provide the correct token for the environment. Then you can just use `process.env.MIXPANEL_TOKEN` instead of determining at runtime.

In the `build` step of the GH workflow you can add something like `REACT_APP_MIXPANEL_TOKEN: ${{ secrets.REACT_APP_MIXPANEL_TOKEN }}`. 

You will also need to add a line in the script that sets the "tier based env var" to the environment before build [here](https://github.com/terraware/terraware-web/blob/main/.github/scripts/set-environment.sh#L28), it should look something like this `REACT_APP_MIXPANEL_TOKEN=${TIER}_REACT_APP_MIXPANEL_TOKEN`. 

The last step is to add the secrets to the repo for `PROD_REACT_APP_MIXPANEL_TOKEN` and `STAGING_REACT_APP_MIXPANEL_TOKEN`.

alexTerraformation: These tokens are not secrets which is why I didn't mind doing it this way (https://developer.mixpanel.com/reference/project-token).  I can still add it to the env vars if you think that would be an improvement but I won't be creating any secrets for these.

nickgraz: Env vars are fine. I only mentioned the secrets way because that's currently how other non-secret env vars are passed into the React app. See `REACT_APP_MAPBOX_TOKEN` and `REACT_APP_GOOGLE_KEY`. I'm not sure if there would be more fandanglin' if you are using the proper GH repo env vars.

Either way, I do think it should be passed into the environment instead of determined at runtime.

.github/workflows/workflow.yml
```diff
@@ -106,6 +106,7 @@ jobs:
         env:
           REACT_APP_TERRAWARE_API: ''
           REACT_APP_TERRAWARE_FE_BUILD_VERSION: ${{ env.APP_VERSION }}
+          REACT_APP_MIXPANEL_TOKEN: ${{ env.MIXPANEL_TOKEN }}
```
nickgraz, line 109:
```suggestion
          REACT_APP_MIXPANEL_TOKEN: ${{ secrets[env. MIXPANEL_TOKEN_NAME] }}
```

.github/scripts/set-environment.sh
```diff
@@ -36,5 +36,10 @@ COMMIT_SHA=$commit_sha
 DOCKER_TAGS=$docker_tags
 APP_VERSION=$APP_VERSION" >> $GITHUB_ENV
 
+if [[ $IS_CD ]]; then
+  echo "MIXPANEL_TOKEN=$MIXPANEL_{TIER}_TOKEN" >> $GITHUB_ENV
+else
+  echo "MIXPANEL_TOKEN=$MIXPANEL_DEV_TOKEN" >> $GITHUB_ENV
```
nickgraz, line 42:
```suggestion
  echo "MIXPANEL_TOKEN_NAME=DEV_MIXPANEL_TOKEN" >> $GITHUB_ENV
```

.github/scripts/set-environment.sh
```diff
@@ -36,5 +36,10 @@ COMMIT_SHA=$commit_sha
 DOCKER_TAGS=$docker_tags
 APP_VERSION=$APP_VERSION" >> $GITHUB_ENV
 
+if [[ $IS_CD ]]; then
+  echo "MIXPANEL_TOKEN=$MIXPANEL_{TIER}_TOKEN" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
alexTerraformation: Unless we're really picky about this I prefer the tier to not be the first thing in the name because it's more difficult to scan the list for what you're looking for.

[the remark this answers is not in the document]
nickgraz: I agree, I generally like the variable part to be a suffix.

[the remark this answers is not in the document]
nickgraz: But what you have here is fine too

.github/scripts/set-environment.sh
```diff
@@ -36,5 +36,10 @@ COMMIT_SHA=$commit_sha
 DOCKER_TAGS=$docker_tags
 APP_VERSION=$APP_VERSION" >> $GITHUB_ENV
 
+if [[ $IS_CD ]]; then
+  echo "MIXPANEL_TOKEN=MIXPANEL_${TIER}_TOKEN" >> $GITHUB_ENV
```
nickgraz, line 40: A slight nit pick, technically this is the name of the secret, not the token.

nickgraz (approved):
One small nit pick (no need to change if you don't want to), otherwise looks good.

joshbuchea:
🎉

PR #2753, opened 2024-06-19
SW-5539: Native/Non-Native is marked required but adding a species without a value is allowed

joshbuchea:
This PR includes changes to ensure the native field is required in the add species modal, as well as updates to improve the error UX for other form fields.


constanzauanini (approved):

nickgraz (approved):
⚡

PR #2757, opened 2024-06-19
SW-5533: Can't add Species to Project (from Species page) even when Deliverable is active

joshbuchea:
This PR includes changes to resolve an issue that was preventing adding a species to a project from the species detail view.

Context: The previously available projects included projects without a species deliverable, but that is a requirement from the API when adding a species to a project.

### DIFF
--- a/src/scenes/Species/SpeciesProjectsTable.tsx
+++ b/src/scenes/Species/SpeciesProjectsTable.tsx
@@ -108,7 +108,7 @@ export default function SpeciesProjectsTable({
       );
     });
     setSelectableProjects(pendingProjects || []);
-  }, [allProjects, filteredResults, currentDeliverables]);
+  }, [allProjects, currentDeliverables, filteredResults]);
 
   useEffect(() => {
     let updatedResults = searchResults ?? [];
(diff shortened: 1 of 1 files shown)


nickgraz (approved):
Nice catch

joshbuchea:
I'm a bit confused. Earlier today, I received a `404` error when trying to add the species _Actinidia deliciosa_ (Kiwi) to **New project 2001** in the **Graziano Tree Farm** organization on Staging:

> Species Deliverable for project 38 not found

To address this, I created this PR to limit the projects in the dropdown to those with a current species deliverable, leaving only "Sampson Road" as an option. However, Leanne commented that both "Sampson Road" and "Jennifer's QA" have active species deliverables and should be available.

`currentDeliverables` only includes two deliverables for "Sampson Road." I'm unsure if there's another dataset to filter the options to both "Sampson Road" and "Jennifer's QA."

Adding to the confusion, I reverted the PR changes locally, making four projects available in the dropdown, and was able to add the species to all four successfully. It seems the API results have changed during the day, possibly due to some data/state update. This one has been fun! 😅

PR #2761, opened 2024-06-20
SW-5541: Filtering for AC Deliverables table not working properly

joshbuchea:
This PR includes a fix for the projects filter not working in the deliverables table for the accelerator console.


nickgraz (approved):
Love the deletions

PR #2768, opened 2024-06-20
Reload user after cookie consent selection

joshbuchea:
A quick PR to reload `user` when the cookie consent selection is made.

### DIFF
--- a/src/redux/features/user/usersAsyncThunks.ts
+++ b/src/redux/features/user/usersAsyncThunks.ts
@@ -34,15 +34,11 @@ export const requestSearchUserByEmail = createAsyncThunk(
 
 export const requestUserCookieConsentUpdate = createAsyncThunk(
   'users/cookie-consent/update',
-  async (
-    { payload, reloadUser }: { payload: UpdateUserCookieConsentRequestPayload; reloadUser: () => void },
-    { rejectWithValue }
-  ) => {
+  async (payload: UpdateUserCookieConsentRequestPayload, { rejectWithValue }) => {
     const response: Response2<UpdateUserCookieConsentResponsePayload> =
       await PreferencesService.updateUserCookieConsentPreferences(payload);
 
     if (response.requestSucceeded) {
-      reloadUser();
       return response;
     }
 
--- a/src/redux/features/user/usersAsyncThunks.ts
+++ b/src/redux/features/user/usersAsyncThunks.ts
@@ -34,11 +34,15 @@ export const requestSearchUserByEmail = createAsyncThunk(
 
 export const requestUserCookieConsentUpdate = createAsyncThunk(
   'users/cookie-consent/update',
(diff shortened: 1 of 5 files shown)


src/redux/features/user/usersAsyncThunks.ts
```diff
@@ -34,11 +34,15 @@ export const requestSearchUserByEmail = createAsyncThunk(
 
 export const requestUserCookieConsentUpdate = createAsyncThunk(
   'users/cookie-consent/update',
-  async (payload: UpdateUserCookieConsentRequestPayload, { rejectWithValue }) => {
+  async (
+    { payload, reloadUser }: { payload: UpdateUserCookieConsentRequestPayload; reloadUser: () => void },
+    { rejectWithValue }
+  ) => {
```
joshbuchea, line 40: This works as-is, but I'm happy to `reloadUser` elsewhere if we'd prefer that in another location.

nickgraz: Yea it seems a bit squirrelly to have the `CookieConsentBanner` component be the glue between redux and a provider. 

It is a bit more work, but it seems like the `dispatch(requestUserCookieConsentUpdate(...` should be inside the `UserProvider`. This way it can look at the response and refresh itself when the response comes back.

joshbuchea: Yeah I agree it wasn't the cleanest. I reverted the changes in the thunk and opted to call `reloadUser` after awaiting the thunk dispatch, which is in-line with how we're reloading the user elsewhere in the app.

nickgraz (approved):
I am still not a fan of the way this is being done. A component shouldn't be the glue between redux and providers, it should really only be consuming state (whether from redux or provider) and dispatching actions. I don't want to hold this up for the release though if it works as expected.

src/redux/features/user/usersAsyncThunks.ts
joshbuchea: Moving the function into the `UserProvider` now...

nickgraz: Sorry, I didn't mean to hold this up! It can be followed up on later

joshbuchea:
@nickgraz I don't disagree with your reasoning and I went ahead and made the change, but why do we have `reloadUser` available via the `useUser` hook? Because it seems like that's all it's being used for.

nickgraz:
> @nickgraz I don't disagree with your reasoning and I went ahead and made the change, but why do we have `reloadUser` available via the `useUser` hook? Because it seems like that's all it's being used for.

I think that function is in there because of the exact problem I am talking about. Instead of using redux as the glue, components became the glue.

joshbuchea:
I don't understand the point you're trying to make, components are calling `reloadUser`, which is what I was previously doing within the cookie consent banner after the first change.

It seems like given your reasoning, reloading the user should be handled via providers or services. Obviously I must be missing something.

PR #2769, opened 2024-06-21
SW-5538: Table column sorting for Species table for Species List Deliverable doesn't seem to work

joshbuchea:
This PR includes changes to resolve an issue that was causing table sorting to not working in the species deliverable table.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@mui/styled-engine-sc": "^5.14.12",
     "@mui/styles": "^5.15.15",
     "@reduxjs/toolkit": "^1.9.3",
-    "@terraware/web-components": "^3.0.27",
+    "@terraware/web-components": "^3.0.24",
     "@testing-library/jest-dom": "^6.0.0",
     "@testing-library/react": "^14.3.0",
     "@testing-library/user-event": "^14.4.3",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@mui/styled-engine-sc": "^5.14.12",
     "@mui/styles": "^5.15.15",
     "@reduxjs/toolkit": "^1.9.3",
-    "@terraware/web-components": "^3.0.24",
+    "@terraware/web-components": "^3.0.27",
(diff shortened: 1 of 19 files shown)


nickgraz:
Is this all needed only because of the fact that the column name is a nested property? Or are there other reasons?

joshbuchea:
@nickgraz changes were made because of the nested property, no other reasons

nickgraz:
> @nickgraz changes were made because of the nested property, no other reasons

This can be solved by transforming the data into a flat structure.

joshbuchea:
It's also solved in this PR, I'm not sure it'd be worth spending more time to change this now.

nickgraz:
> It's also solved in this PR, I'm not sure it'd be worth spending more time to change this now.

If you are introducing more static functions then you should add unit tests for them. In my opinion this is being solved "the hard way" and is introducing more complexity when it is unnecessary. At a minimum, if you want to do it this way, there should be unit tests and you should try to avoid using the `any` type.

joshbuchea:
I assumed the data was structured with nested properties for a reason. I don't agree that it's adding much complexity and think it's fine to do it this way.

nickgraz:
> I assumed the data was structured with nested properties for a reason. I don't agree that it's adding much complexity and think it's fine to do it this way.

We have divergent opinions on complexity. No problem. If you must do it this way, can you at least write some unit tests for the static utility functions you've created and remove the `any` type?

nickgraz:
[Here](https://github.com/terraware/terraware-web/pull/2785) is the way I was thinking it would be solved. Either way, I am OK with your approach if some tests are added and the `any` type is removed.

joshbuchea:
I don't have a strong preference for how the fix is implemented, so I'm happy to close this PR in favor of yours.

joshbuchea:
FWIW, I just disagreed with the notion that a change was necessary and didn't want to invest more time on this issue.

joshbuchea:
Closing in favor of #2785

PR #2772, opened 2024-06-21
SW-5545 Document deliverable approve button should be disabled immediately after approving

constanzauanini:
The deliverable was refreshed correctly, but actions where inside useMemo without the correct dependency array


nickgraz (approved):
🙌

PR #2795, opened 2024-06-27
SW-5559 Accelerator admin deliverable categories

sgrimm:
Add an editable list of deliverable categories for each person to the accelerator
console's "People" views.

Deliverable categories are read and updated separately from global roles in the
API, so this required adding support for the new API endpoints and calling them
from the scene code. Since both reading and writing now involve multiple API
requests, and we need the data to stay in sync between the edit and read-only
views, the API interactions are moved to `PersonProvider`, which now exposes
some callback functions to its child components.


sgrimm:
* **#2795** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2795?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/scenes/AcceleratorRouter/People/index.tsx
```diff
@@ -11,7 +11,14 @@ import SingleView from './SingleView';
 const PeopleRouter = () => {
   return (
     <Routes>
-      <Route path={'new'} element={<NewView />} />
+      <Route
+        path={'new'}
+        element={
+          <PersonProvider>
+            <NewView />
+          </PersonProvider>
```
sgrimm, line 19: Not sure if having the same provider in two different places in a tree of routes is considered bad practice. I tried moving it to the top level but then it isn't able to get the user ID from the path on the edit and single views.

src/scenes/AcceleratorRouter/People/PersonProvider.tsx
```diff
@@ -18,37 +29,112 @@ const PersonProvider = ({ children }: Props) => {
...
+
+  const reloadUser = useCallback(() => {
     if (userId !== -1) {
       void dispatch(requestGetUser(userId));
     }
-  }, [userId, dispatch]);
+  }, [dispatch, userId]);
+
+  const reload = useCallback(() => {
+    console.log(userId);
```
nickgraz, line 60: Make sure to remove this before merging in

src/scenes/AcceleratorRouter/People/PersonProvider.tsx
```diff
@@ -18,37 +29,112 @@ const PersonProvider = ({ children }: Props) => {
...
+    if (
+      updateCategoriesRequest?.status === 'success' &&
+      updateGlobalRolesRequest?.status === 'success' &&
+      onUpdateSuccess
+    ) {
+      onUpdateSuccess();
+    }
+  }, [onUpdateSuccess, updateCategoriesRequest, updateGlobalRolesRequest]);
+
+  useEffect(() => reload(), [reload]);
```
nickgraz, line 137:
```suggestion
  useEffect(reload, [reload]);
```

src/scenes/AcceleratorRouter/People/PersonProvider.tsx
sgrimm: 🤦  Thanks.

src/scenes/AcceleratorRouter/People/index.tsx
nickgraz: It isn't ideal, because the provider will be instantiated twice, which negates the benefit of a shared provider.

I think in this case you don't need the provider to be wrapped around the `NewView`. It seems the only functionality being passed into that view is the ability to set the user ID and the `update` function. Since the `update` functionality doesn't really provide any meaningful state that should be shared, I think it should just be thrown into a custom hook if you want to reduce the duplicate code. See [this hook](https://github.com/terraware/terraware-web/blob/main/src/components/DeliverableView/useUpdateDeliverable.ts).

The idea is that the state related to the request flow is encapsulated in the hook, which makes it easy to interact with the APIs without putting a bunch of boilerplate in separate views/components. Since the hook will be instantiated in multiple places, they are also not "shared", but this is OK because there isn't anything that needs to be shared.

src/scenes/AcceleratorRouter/People/PersonProvider.tsx
```diff
@@ -18,37 +29,111 @@ const PersonProvider = ({ children }: Props) => {
...
 
-  const [user, setUser] = useState<User>();
-  const userRequest = useAppSelector(selectUserRequest(userId));
+  const [updateGlobalRolesRequestId, setUpdateGlobalRolesRequestId] = useState('');
+  const updateGlobalRolesRequest = useAppSelector(selectGlobalRolesUserUpdateRequest(updateGlobalRolesRequestId));
+  const [updateCategoriesRequestId, setUpdateCategoriesRequestId] = useState('');
+  const updateCategoriesRequest = useAppSelector(
+    selectUserDeliverableCategoriesUpdateRequest(updateCategoriesRequestId)
+  );
+  const [onUpdateSuccess, setOnUpdateSuccess] = useState<() => void>();
```
nickgraz, line 44: I can't say that this is an anti-pattern, but it is somewhat abnormal. We generally don't set a function to state to be called later, instead we will watch a value and do things when the value changes to what we want. 

In combination with my other comment about the `PersonProvider`, I think what you should do here is create a `useUpdatePerson` hook that returns `{ update: (payload) -> void, success: boolean }`. Within the hook, if the update request fails, the hook can be responsible for sending the toast to the snackbar, and if the request is successful, it can update the `success` value within the hook's state so that downstream consumers know they can do specific things when the request is successful. 

In the case of its usage here, since you are navigating to a new view, any usages of the custom `useUpdatePerson` hook will be a new instantiation, so you don't have to worry about the `success` value "sticking" across views.

src/scenes/AcceleratorRouter/People/index.tsx
sgrimm: > It seems the only functionality being passed into that view is the ability to set the user ID and the update function.

Setting the user ID triggers the provider to fetch that user's details which are then passed back down to `NewView` to show in the form. How would I make that work with a hook?

Alternately, I could maybe grab the user ID from the URL in `EditView` and `SingleView` instead of in the provider, and then call the provider's `setUserId` to trigger loading the user's data. Then the whole tree of routes can be wrapped in a single provider rather than needing two of them. Does that seem like a sane approach?

sgrimm: I am going to give that a try just to see what it looks like (but happy to throw it out if it's not the right thing to do).

src/scenes/AcceleratorRouter/People/PersonForm.tsx
```diff
@@ -6,31 +6,47 @@ import { MultiSelect, Textfield } from '@terraware/web-components';
...
 type PersonFormProps = {
-  busy?: boolean;
   emailEnabled?: boolean;
   emailError?: string;
-  user?: User;
+  personData?: PersonData;
   onCancel: () => void;
-  onChange?: (person: User) => void;
-  onSave: (person: User) => void;
+  onChange?: (person: PersonData) => void;
```
nickgraz, line 22: I think this part is a bit confusing. The "record" used in a form generally shouldn't contain functions in it, it should be a somewhat standard key value pair. I see that you are trying to pass around both the `user` and the `deliverableCategories` as separate objects for convenience, but in my opinion it makes the code a bit more tightly coupled. 

I think the proper way to go about this is to have a `UserWithDeliverableCategories` type that the form is concerned with, and let the provider or hook that is responsible for updating pull the data apart and make the separate API calls. This way the form can be unaware of how the data is ultimately stored.

nickgraz (commented):
I know I am asking a lot here, and it may or may not make sense the way I am explaining myself. I think what you have here is mostly really great, there are just a few things I would do slightly differently. If you want, I would be happy to make a draft PR with the changes I am talking about.

src/scenes/AcceleratorRouter/People/index.tsx
nickgraz: I see. I think that makes sense the way you have it then. I may be over complicating this, since this is already a bit complicated with the fact that we have a form that is eventually pre-populated with some remote data.

It seems that you do in fact need a way to have the consumer set the user ID within the provider in the event that it isn't available in the URL.

src/scenes/AcceleratorRouter/People/PersonProvider.tsx
sgrimm: I've made this change locally, but I'm not sure how best to get the provider to refresh its data after an edit succeeds. The problem is that I'm calling `useUpdatePerson()` from `EditView` and thus its success status isn't visible to the provider and can't trigger a refresh in the provider.

If I were to call `useUpdatePerson()` from the provider, I assume your last paragraph would then no longer apply since the hook would no longer be invoked from a new instantiation.

I can think of a couple ways to make it work but they feel like ugly hacks to me. I assume there's a clean way to do this.

Will update the PR with what I've done so you can see if it's just that I messed something up.

sgrimm:
With the update I just pushed, you see stale data if you edit a user and hit the Save button, because nothing triggers the provider to reload the user data. I suspect I misunderstood part of the suggestion.

src/scenes/AcceleratorRouter/People/PersonProvider.tsx
nickgraz: I think in this case I would dispatch the fetch when the update resolves in the thunk, right above [this line](https://github.com/terraware/terraware-web/pull/2795/files#diff-316ee36fa58bf952f4168c6cdcea40825d639a60cff9cbae8a15f24cbc408494R27).

nickgraz: Similar to [this](https://github.com/terraware/terraware-web/blob/main/src/redux/features/participants/participantsAsyncThunks.ts#L78)

src/scenes/AcceleratorRouter/People/PersonForm.tsx
```diff
@@ -6,20 +6,22 @@ import { MultiSelect, Textfield } from '@terraware/web-components';
...
 type PersonFormProps = {
   busy?: boolean;
   emailEnabled?: boolean;
   emailError?: string;
-  user?: User;
+  user?: UserWithDeliverableCategories;
   onCancel: () => void;
-  onChange?: (person: User) => void;
-  onSave: (person: User) => void;
+  onChange?: (person: UserWithDeliverableCategories) => void;
```
nickgraz, line 23: These changes look great, thank you

sgrimm:
Thanks for that pointer. The thing I was missing was that I needed to use the user ID, not a unique request ID, as the key in the Redux store. It's all working now and ready for another review.

nickgraz:
> Thanks for that pointer. The thing I was missing was that I needed to use the user ID, not a unique request ID, as the key in the Redux store. It's all working now and ready for another review.

Ah I see. In case you are wondering how that works, if the thunk that is dispatched accepts a single, primitive value, then you are able to use the `useArgAsKey` feature of the `buildReducers` function [here](https://github.com/terraware/terraware-web/blob/main/src/redux/features/asyncUtils.ts#L10). It is a nice thing Karthik added for simple requests using IDs for example.

src/redux/features/userDeliverableCategories/userDeliverableCategoriesSlice.ts
```diff
@@ -11,14 +11,14 @@ import {
...
-const initialStateUserDeliverableCategoriesGet: { [requestId: string]: StatusT<UserDeliverableCategoriesData> } = {};
+const initialStateUserDeliverableCategoriesGet: { [userId: number]: StatusT<UserDeliverableCategoriesData> } = {};
 
 export const userDeliverableCategoriesGetSlice = createSlice({
   name: 'userDeliverableCategoriesGetSlice',
   initialState: initialStateUserDeliverableCategoriesGet,
   reducers: {},
   extraReducers: (builder) => {
-    buildReducers(requestGetUserDeliverableCategories)(builder);
+    buildReducers(requestGetUserDeliverableCategories, true)(builder);
```
nickgraz, line 21: Ignore my previous comment, looks like you already know about the feature!

src/scenes/AcceleratorRouter/People/PersonProvider.tsx
```diff
@@ -18,37 +19,42 @@ const PersonProvider = ({ children }: Props) => {
...
-  }, [userRequest, snackbar]);
+  }, [getCategoriesRequest, getUserRequest, userId, snackbar]);
 
-  useEffect(() => {
-    setPersonData({
-      user,
-      userId,
-    });
-  }, [userId, user]);
+  // useEffect(reload, [reload]);
```
nickgraz, line 57: 🔪

nickgraz:
This looks great @sgrimm! Thanks for being amenable to my feedback.

sgrimm:
I removed that commented-out line; any other changes needed to make this ready to go?

nickgraz:
> I removed that commented-out line; any other changes needed to make this ready to go?

Sorry, I did a regular comment instead of an approval comment. This is good to go!

nickgraz (approved):

PR #2812, opened 2024-07-04
Save value when editing

constanzauanini:


src/redux/features/documentProducer/variables/variablesThunks.ts
```diff
@@ -28,3 +28,22 @@ export const requestListDeliverableVariables = createAsyncThunk(
...
+    {
+      projectId,
+      variableId,
+      ...rest
+    }: UpdateVariableWorkflowDetailsPayload & { projectId: number; variableId: number },
+    { rejectWithValue }
+  ) => {
+    const response = await VariableService.updateVariableWorkflowDetails(variableId, projectId, rest);
+    if (response.requestSucceeded) {
+      return Boolean(true);
```
nickgraz, line 44:
```suggestion
      return true;
```

src/redux/features/documentProducer/variables/variablesThunks.ts
```diff
@@ -28,3 +28,22 @@ export const requestListDeliverableVariables = createAsyncThunk(
...
 );
+
+export const requestUpdateVariableWorkflowDetails = createAsyncThunk(
+  'updateVariableWorkflowDetails',
+  async (
+    {
+      projectId,
+      variableId,
+      ...rest
+    }: UpdateVariableWorkflowDetailsPayload & { projectId: number; variableId: number },
```
nickgraz, line 39: Nice!

src/scenes/AcceleratorRouter/Deliverables/QuestionsDeliverableView.tsx
```diff
@@ -32,141 +49,312 @@ export type Props = EditProps & {
...
+    setStatus(currentStatus, undefined, internalComment);
+  };
+
+  const onValuesChanged = (variableId: number, values: VariableValueValue[]) => {
+    setValues(values);
+  };
+
+  const onSave = async () => {
+    const operations: components['schemas']['AppendValueOperationPayload'][] = [];
+    values.forEach((val) => {
```
nickgraz, line 116: I wonder if you also need to implement "update". This was in the code I copied over from the document producer variable update.

https://github.com/terraware/terraware-web/pull/2808/files#diff-4de89dc1f5fbcbba072393e8890f56c09bc4cbc145f052cf8e25400efc68ae6bR143

I will check to see what the difference is in the back end

constanzauanini: I was thinking about that. Maybe if the variable can have multiple values I'm just adding a new one instead of replacing the old one. But for the two I was testing with, looks like they can have only one value (project size and

nickgraz: The different operation descriptions are in here: https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/documentproducer/api/ValueOperationPayloads.kt#L58

nickgraz: It seems like what you have will work for simple variable types like single line text, single select dropdown, number, date, etc... but it might not work when dealing with "list" type variables like multi line text, multi select dropdown, etc...

nickgraz: So I think we should probably re-use the existing mechanism, since it was already tested on the document producer side and probably works how we need for all the variable types. The rest of this PR looks good, so I am OK with you merging this in as-is and I can create a custom hook or something that will encapsulate that functionality and make sure it is the same across participant and accelerator sides. The "workflow details" stuff doesn't need to change at all.

nickgraz (approved):

src/scenes/AcceleratorRouter/Deliverables/QuestionsDeliverableView.tsx
constanzauanini: Ok, sounds good! Will fix conflicts and merge

PR #2824, opened 2024-07-08
SW-5397 Add deliverable type column and filter

constanzauanini:


joshbuchea (approved):
I left a comment on the Jira ticket. It seems the "Type" column was removed as part of SW-5548. I’m not sure why we removed the Type column only to add it back now 🤔

Changes look good though 👍

nickgraz (approved):

PR #2825, opened 2024-07-08
Migrate to Docker Compose v2

joshbuchea:
This PR migrates the repo to Docker Compose v2 for compatibility with the latest version of Docker Desktop `4.32.0`:

- Replace `docker-compose` with `docker compose`

**Background:**

After upgrading to the latest version of Docker Desktop `4.32.0` (`157355`) today, when I attempted to run `docker-compose up -d` I ran into the following error: `zsh: command not found: docker-compose`.

Dropping the dash from `docker-compose` resolves the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,10 @@ For the moment, we're not asking for code contributions from the community. (Che
 
 You may see references to some private repositories in the documentation. We're working toward opening more of our code, but not everything is ready yet.
 
+## Requirements
+
+- Docker Desktop version `4.32.0` or later
+
 ## How to Run the App in Development Mode
 
 1. Configure the `.env` file using the sample file `.env.sample` located at the root directory of this repo.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -99,7 +99,7 @@ yarn docker:stop    # don't forget to stop the server when you're done
 If you want to examine or change the database that's used for your local testing, you can run the PostgreSQL interactive command-line client. This will only work after you've run `scripts/resetdb.sh` or `yarn docker:start`.
 
 ```shell
-docker-compose exec postgres psql -U postgres terraware
+docker compose exec postgres psql -U postgres terraware
```
nickgraz, line 102: Can you update the readme to mention what version of `docker` the user needs to be running?

nickgraz (approved):
This looks good, if you can add the recommended `docker` version to the README, this is good to go!

README.md
joshbuchea: So it looks like support for Docker Compose v2 was added to Docker Desktop back in version `3.4`, so `3.4` or later is all that's required. I'm not sure if we want to list that as the required version or if we want to use a later version? The latest version is `4.32.0`.

joshbuchea: Went ahead and added a requirements section with Docker Desktop `4.32.0` or later. Happy to change the version or place a note elsewhere if we'd prefer to not add a new section.

PR #2830, opened 2024-07-10
SW-5618: Edit "In Place" for long Questionnaires

joshbuchea:
This PR includes changes to support auto-scrolling to the same questions position when the edit button is pressed in the Questions Deliverable View.


constanzauanini (approved):
Lgtm! Looks like code is not formatted

PR #2850, opened 2024-07-13
SW-5650: FE - Add editing support for Table field

joshbuchea:
Add support for editing `Table` fields in Questions Deliverables.


constanzauanini (approved):
Lgtm. We can remove the "image" support in the PR title, since this is only for table

nickgraz (approved):
Nice work!

PR #2854, opened 2024-07-15
SW-5650: FE - Add editing support for Image field

joshbuchea:
This PR includes changes to support editing image fields for Questions Deliverables.

### DIFF
--- a/src/components/DocumentProducer/EditImagesModal/PhotoSelector.tsx
+++ b/src/components/DocumentProducer/EditImagesModal/PhotoSelector.tsx
@@ -60,6 +60,7 @@ export default function PhotoChooser(props: PhotoChooserProps): JSX.Element {
   const divRef = useRef<HTMLDivElement>(null);
   const theme = useTheme();
   const [editing, setEditing] = useState<boolean>(false);
+  const [filesDataChanged, setFilesDataChanged] = useState<boolean>(false);
 
   useEffect(() => {
     setEditing(!!selectedFile);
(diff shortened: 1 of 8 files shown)


src/components/DocumentProducer/EditImagesModal/PhotoSelector.tsx
```diff
@@ -142,14 +145,20 @@ export default function PhotoChooser(props: PhotoChooserProps): JSX.Element {
...
 
   useEffect(() => {
+    if (!filesDataChanged) {
+      return;
+    }
+
     onPhotosChanged(filesData);
-  }, [filesData, onPhotosChanged]);
+    setFilesDataChanged(false);
+  }, [filesData, onPhotosChanged, filesDataChanged]);
```
joshbuchea, line 154: Changes were made in this file to prevent `onPhotosChanged` from being called until the field is modified and also prevent it from being called multiple times. Those changes seemed safe to make here, but I can clone this file and create a deliverable variation if preferred.

src/hooks/useProjectVariablesUpdate/index.ts
```diff
@@ -71,7 +103,65 @@ export const useProjectVariablesUpdate = (
...
+            file: newImage.file,
+            caption: newImage.caption,
+            citation: newImage.citation,
+            projectId,
+          })
+        );
+
+        // set request id with last image request
+        if (pendingValues.length - 1 === index) {
+          setUploadRequestId(upRequest.requestId);
```
joshbuchea, line 162: This is only tracking the progress of one image upload for now. Do we have a method for tracking the progress of multiple dispatched requests?

constanzauanini: No. What we have in other parts of the app, is a thunk that receives all the request, and the response is only one, after all are completed: https://github.com/terraware/terraware-web/blob/main/src/redux/features/participantProjectSpecies/participantProjectSpeciesAsyncThunks.ts#L49

joshbuchea: Ok I'll create a new thunk using that as a reference, thanks @constanzauanini! 🙌

constanzauanini (approved):
Lgtm!

PR #2874, opened 2024-07-18
SW-5668 Update prescreen overview to use buttons instead of links for deliverables

constanzauanini:


src/components/ModuleDetailsCard/index.tsx
```diff
@@ -216,17 +217,36 @@ const ModuleDetailsCard = ({
...
-                        showSimplifiedStatus={showSimplifiedStatus}
-                        status={deliverable.status || 'Not Submitted'}
-                      />
-                    </Grid>
-                    <Grid item flexGrow={0} xs={gridSize()}>
-                      <Link fontSize='16px' onClick={deliverable.onClick} style={{ textAlign: 'left' }}>
-                        {deliverable.name}
-                      </Link>
-                    </Grid>
+                    {module.name === 'Pre-screen' ? (
```
tommylau523, line 220: Can we use a prop for this? I think DeliverableDetails can probably have a useButton prop.

constanzauanini: Sure!

tommylau523 (approved):

PR #2882, opened 2024-07-19
SW-5638: FE - Add document outline

joshbuchea:
This PR adds a new document outline panel in the doc producer tool.

### DIFF
--- a/src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
+++ b/src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 
 import {
   Box,
@@ -13,6 +13,7 @@ import {
 } from '@mui/material';
 
 import Icon from 'src/components/common/icon/Icon';
+import { useIsVisible } from 'src/hooks/useIsVisible';
 import { requestListVariablesValues } from 'src/redux/features/documentProducer/values/valuesThunks';
 import { selectVariablesWithValues } from 'src/redux/features/documentProducer/variables/variablesSelector';
 import { requestListVariables } from 'src/redux/features/documentProducer/variables/variablesThunks';
@@ -106,6 +107,8 @@ const SectionItem = ({
           sx={{
             height: '4px',
             backgroundColor: isSelected ? theme.palette.TwClrBgBrand : 'transparent',
+            borderBottomRightRadius: '2px',
+            borderTopRightRadius: '2px',
             marginRight: '8px',
             width: '16px',
           }}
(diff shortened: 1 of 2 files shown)


src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+  return (
+    rect.top >= 0 &&
+    rect.left >= 0 &&
+    rect.bottom <= (window.innerHeight || document.documentElement.clientHeight) &&
+    rect.right <= (window.innerWidth || document.documentElement.clientWidth)
+  );
+};
+
+const StyledTooltip = styled(({ className, ...props }: TooltipProps) => (
```
constanzauanini, line 36: I think the tooltip component from web-components already has this colors

src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+    if (!isHTMLElementInViewport(element)) {
+      // scroll to section heading in the document, if not already in view
+      element.scrollIntoView({ behavior: 'smooth' });
+    }
+  }, [section.id, selectedSectionId, setSelectedSectionId]);
+
+  return (
+    <Box>
+      <ButtonBase
```
constanzauanini, line 94: Would the web-components <Link> component work here? It has an onClick prop

src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
joshbuchea: I thought these changes felt familiar, will update. Thanks! 🙌

src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
joshbuchea: I think my initial thoughts were that it wouldn't work for some reason, but perhaps it will. Good suggestion, will give it a try.

src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
joshbuchea: I tried using `Tooltip` from `web-components`, but it broke the layout for some reason. Will spend a few minutes trying to fix the layout issues, but will revert if I'm unable to work it out relatively quick.

## Before

localhost_3000_accelerator_documents_1_tab=document (1)

## After

localhost_3000_accelerator_documents_1_tab=document

src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
joshbuchea: It seems that our `Button` component in `web-components` doesn't accept `children`, so it won't work for this case.

src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentOutlinePanel.tsx
constanzauanini: Oh, that's strange. Yeah, don't worry to much if it takes time. I thought it was going to be an easy change

nickgraz (approved):
Looks good, thanks for trying those other avenues

PR #2886, opened 2024-07-22
SW-5689: Auto scrolls down when going to edit view from Participant View

joshbuchea:
This PR includes changes to resolve an issue that caused the Questions Deliverable Edit View to always auto-scroll, regardless of how much of the view has been viewed/scrolled.

Changes include:

- Only auto-scroll when the document is scrolled beyond the viewport height
- Remove default option values for `element.scrollIntoView` method (they were previously included unnecessarily)


nickgraz (approved):
Nice!

PR #2891, opened 2024-07-23
SW-5691: Can't type in single line text fields (console and participant view)

joshbuchea:
This PR includes changes to allow editing text fields in Questions Deliverables on both the console and participant sides.

### DIFF
--- a/src/scenes/AcceleratorRouter/Deliverables/QuestionsDeliverableView.tsx
+++ b/src/scenes/AcceleratorRouter/Deliverables/QuestionsDeliverableView.tsx
@@ -91,6 +91,11 @@ const QuestionBox = ({
 
   const [modalFeedback, setModalFeedback] = useState(firstVariableValue?.feedback || '');
 
+  const pendingValues: VariableValueValue[] | undefined = useMemo(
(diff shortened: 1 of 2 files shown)


src/scenes/AcceleratorRouter/Deliverables/QuestionsDeliverableView.tsx
```diff
@@ -322,7 +327,7 @@ const QuestionBox = ({
               <Grid item xs={12}></Grid>
               <Grid item xs={12}>
                 <DeliverableVariableDetailsInput
-                  values={variable.values}
+                  values={pendingValues || variable.values}
```
joshbuchea, line 330: @nickgraz this change resolves the issue of not being able to type into text fields when editing a Questionnaire Deliverable, but I'm not sure if there's a better approach or anything else to consider here. I tested editing with other field types and it seems like everything works as expected.

nickgraz: This makes sense to me. Seems like a strange bug but it makes sense to pass it in like this since this view is the container.

nickgraz (approved):

PR #2900, opened 2024-07-25
SW-5659: FE - Application success screen

joshbuchea:
This PR includes changes to add the application status submit success screen.

## Screenshots

localhost_3000_applications_1_review_organizationId=141


src/scenes/ApplicationRouter/portal/Review/index.tsx
```diff
@@ -9,6 +13,50 @@ import strings from 'src/strings';
...
+  );
+};
+
+const ApplicationStatusSubmitted = () => (
+  <ApplicationStatus
+    buttonLabel={strings.EXIT_APPLICATION}
+    feedback='<p>Your Application has been submitted to the Accelerator team. We will review and score your answers.</p><p>Check back to see updates? Look for an email? [need copy]</p>'
+    isFailure={false}
+    onClickButton={() => {
+      // TODO: what does "Exit Application" do?
```
tommylau523, line 54: Exit application just brings us back to home page. I have been using goToHome from useNavigateTo

src/scenes/ApplicationRouter/portal/Review/index.tsx
```diff
@@ -32,6 +80,7 @@ const ReviewView = () => {
 
   return (
     <ApplicationPage crumbs={crumbs}>
+      {selectedApplication.status === 'Submitted' && <ApplicationStatusSubmitted />}
```
tommylau523, line 83: We can change this to 'In Review' instead. Submitted is an internal one.

The list of statuses are here:
https://terraformation.atlassian.net/wiki/spaces/PM/pages/1295319074/PRD+Application+-+Pre-screen#Data-%2F-Calculations

tommylau523: Adding on to this, I think we will need slightly different text/screens for Accepted, Waitlist, Not Accepted as well. But we can follow up with more PRs instead.

tommylau523 (approved):
Looks great! I think we need to rework the status to be a switch case at some point to make sure we don't miss any statuses. But this can happen later.

joshbuchea:
@tommylau523 thanks for the feedback! 🙌 I agree that reworking the code to utilize a switch case is a good idea. Will get the requested changes implemented.

PR #2897, opened 2024-07-25
Refactored deliverables to component cards

tommylau523:
Refactored out deliverable views so that application portal can use cards and inject different buttons and navbar accordingly.

This is in preparation for implementing application deliverable screen down the line.

Some design notes:
- Crumbs, navigation are injected on the router, in this case `DeliverablePage`. This is so when Application portal implements deliverable views, it can look different.

PR #2902, opened 2024-07-25
SW-5666 Added submit button to prescreen, make API calls into hooks

tommylau523:
Added submit button to prescreen, make API calls into hooks.

Used submit and restart hooks in the prescreen pages.


tommylau523:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2902?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#2902** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2902?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2898** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2898?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2897** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2897?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2895** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/2895?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/scenes/ApplicationRouter/portal/ApplicationPage.tsx
```diff
@@ -14,16 +16,14 @@ type Props = {
 };
 
 const ApplicationPage = ({ children, crumbs, hierarchicalCrumbs, rightComponent }: Props) => {
-  const { allApplications, selectedApplication, setSelectedApplication, reload } = useApplicationData();
```
nickgraz, line 17: 🙏

src/scenes/ApplicationRouter/portal/Sections/SectionView.tsx
```diff
@@ -54,13 +57,56 @@ const SectionView = ({ section, sectionDeliverables }: SectionViewProp) => {
...
+  const handleRestart = useCallback(async () => {
+    if (selectedApplication) {
+      await restart();
+      await reload();
+      goToApplication(selectedApplication.id);
+    }
+  }, [selectedApplication, reload, restart, goToApplication]);
+
+  const handleSubmit = useCallback(async () => {
+    console.log(submit);
```
nickgraz, line 74: 🔪

src/scenes/ApplicationRouter/provider/index.tsx
```diff
@@ -50,20 +55,63 @@ const ApplicationProvider = ({ children }: Props) => {
     [allApplications]
   );
 
-  const _reload = useCallback(() => {
+  const _reload = useCallback(async () => {
+    if (selectedApplication) {
+      const result = await ApplicationService.getApplication(selectedApplication.id);
```
nickgraz, line 60: Should this go into redux and our normal request flow?

src/scenes/ApplicationRouter/NewApplicationModal.tsx
```diff
@@ -114,12 +118,12 @@ const NewApplicationModal = ({ open, onClose, onSave }: NewApplicationModalProps
     }
 
     if (!!projectId) {
-      const result = await ApplicationService.createApplication(projectId);
-      if (result.requestSucceeded && result.data) {
-        onSave(result.data?.id);
+      const newApplicationId = await create(projectId);
```
nickgraz, line 121: I think this should go through our normal async thunk request flow.

tommylau523: I am thinking that for these one-off calls, we can handle the result directly instead of going through a selector. Especially since we are not re-using or re-querying the results. Same for `reload`.

nickgraz: Makes sense. It is quite a bit of boilerplate

nickgraz (approved):

src/scenes/ApplicationRouter/NewApplicationModal.tsx
tommylau523: Yeah to add to that, I think it gets a bit tricky with selectors because for use cases like this, we want to make sure we are not looking at "stale" results as well. So I think it will be a lot of extra logic to sort that out. Just my 2 cents!!

src/scenes/ApplicationRouter/provider/index.tsx
tommylau523: This is done this way because a navigate will occur before reload is completed without this synchronized step.

I am hoping to actually not have to fetch one off application at all if possible. `requestListApplications` will end up refetching the data. I think at some point I will need to add some indication of `isLoading` in the provider. But one for the later.

src/scenes/ApplicationRouter/NewApplicationModal.tsx
nickgraz: I think the idea of "stale" results is handled by the request selector taking a request ID. So if it changes, the stale results will go away immediately. But I could be misunderstanding what you mean by stale here. Also, things like "create an application then save data to it" sounds like logic that should exist in the service layer as opposed to a modal.

Another edge case here is that the modal can be closed between the two requests, which I think would break the onSave flow. Probably not something we have to be concerned with here, but generally the ability to unmount a component would be disabled if there are multiple requests that execute in succession.

Either way, what you have here is fine, no changes needed.

src/scenes/ApplicationRouter/provider/index.tsx
nickgraz: I see, if you use redux for this, you can just select the application out of state using the ID. This would allow you to select a single application out of the "list applications" redux state.

src/scenes/ApplicationRouter/NewApplicationModal.tsx
tommylau523: Good call! I think I will implement a new service call for "creating a project and then an application with it".

src/scenes/ApplicationRouter/provider/index.tsx
tommylau523: Good call here! I can likely keep track of selectedApplicationId, and set selectedApplication to null for loading state.

tommylau523: Tracking this task as a future upgrade!

PR #2906, opened 2024-07-26
SW-5715: FE - Participant Project Dashboard Updates

joshbuchea:
This PR includes changes to the Participant project dashboard.

## Tasks

- [x] Replace hardcoded carbon data values once new properties are available from API

## Screenshots

localhost_3000_accelerator_projects_34 (1)

localhost_3000_accelerator_projects_34 (3)

localhost_3000_accelerator_projects_34 (4)


src/scenes/AcceleratorRouter/ParticipantProjects/EditView/index.tsx
```diff
@@ -40,8 +42,10 @@ const EditView = () => {
   const { phaseVotes } = useVotingData();
   const { goToParticipantProject } = useNavigateTo();
   const { isAllowed } = useUser();
+  const { allApplications } = useApplicationData();
 
   const isAllowedEdit = isAllowed('UPDATE_PARTICIPANT_PROJECT');
+  const isAllowedViewScoreAndVoting = isAllowed('VIEW_PARTICIPANT_PROJECT_SCORING_VOTING');
```
joshbuchea, line 48: This permission is used to gate the fields on the view side, so it seems like we should also be gating on the edit side.

There doesn't currently seem to be an edit equivalent for this permission (`EDIT_PARTICIPANT_PROJECT_SCORING_VOTING`) defined in the ACL, so I used the view permission for now.

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/index.tsx
```diff
@@ -58,6 +62,11 @@ const EditView = () => {
 
   const [confirmProjectNameModalOpen, setConfirmProjectNameModalOpen] = useState(false);
 
+  const projectApplication = useMemo(
+    () => allApplications?.find((app) => app.projectId === projectId),
+    [allApplications, projectId]
+  );
+
```
joshbuchea, line 69: I'm not sure if there's a more convenient method available to get an application for a given project?

src/scenes/AcceleratorRouter/ParticipantProjects/SingleView.tsx
```diff
@@ -212,6 +217,24 @@ const SingleView = () => {
...
+                label={strings.MIN_MAX_CARBON_ACCUMULATION}
+                value={
+                  participantProject?.minCarbonAccumulation && participantProject?.maxCarbonAccumulation
+                    ? `${participantProject.minCarbonAccumulation}-${participantProject.maxCarbonAccumulation}`
+                    : false
+                }
+              />
+              <ProjectFieldDisplay label={strings.CARBON_CAPACITY_TC02_HA} value='770' />
+              <ProjectFieldDisplay label={strings.ANNUAL_CARBON_T} value='82,566' />
+              <ProjectFieldDisplay label={strings.TOTAL_CARBON_T} value='3,302,640' />
```
joshbuchea, line 236: I wasn't able to find new carbon properties for these fields in the API schema with a quick search, but these hardcoded values need to be replaced once the properties are available.

src/scenes/ApplicationRouter/portal/PrescreenResult/index.tsx
```diff
@@ -1,9 +1,10 @@
 import React, { useCallback, useMemo } from 'react';
 
-import { Box, Card, Typography, useTheme } from '@mui/material';
+import { Box, Typography, useTheme } from '@mui/material';
 import { Button } from '@terraware/web-components';
 
 import { Crumb } from 'src/components/BreadCrumbs';
+import Card from 'src/components/common/Card';
```
joshbuchea, line 7: @tommylau523 I think this was causing that layout issue you noticed where it looked like the top of the card was being clipped.

joshbuchea:
@tommylau523 I couldn't find an existing mapping of application status values to colors and the newly added carbon properties to participant projects, so those still need to be added. It's also definitely possible they do exist somewhere and I just overlooked them.

src/scenes/AcceleratorRouter/ParticipantProjects/SingleView.tsx
tommylau523: Good catch. I think they need to be added to existing Accelerator project details as well. Let me look into new columns in the BE.

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/index.tsx
nickgraz: Can you add one?

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/index.tsx
```diff
@@ -263,6 +270,40 @@ const EditView = () => {
...
+              onChange={onChangeParticipantProject}
+              valueMax={participantProjectRecord?.maxCarbonAccumulation}
+              valueMin={participantProjectRecord?.minCarbonAccumulation}
+            />
+            <ProjectFieldTextfield
+              id={'carbonCapacity'}
+              label={strings.CARBON_CAPACITY_TC02_HA}
+              onChange={onChangeParticipantProject}
+              type={'number'}
+              value='770'
```
nickgraz, line 290: Should these be empty instead?

src/scenes/AcceleratorRouter/ParticipantProjects/SingleView.tsx
tommylau523: https://github.com/terraware/terraware-server/pull/2212

tommylau523: Ready now!

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/index.tsx
joshbuchea: Values are available from the API now, will update

tommylau523:
I posted this on Slack as well but thought I would mention here. We want to disable being able to change country, since the country is determined by project boundary.

tommylau523:
> I posted this on Slack as well but thought I would mention here. We want to disable being able to change country, since the country is determined by project boundary.

Sorry we can actually leave that in.

PR #2929, opened 2024-07-31
SW-5598 PartEx Project Dropdowns in Home and Modules Screens: do not show carat when there is only one Project

constanzauanini:


src/scenes/Home/ParticipantHomeView/Header.tsx
```diff
@@ -43,12 +43,16 @@ const Header = () => {
...
-            />
+            {options?.length > 1 ? (
+              <Dropdown
+                onChange={setCurrentParticipantProject}
+                options={options}
+                selectStyles={selectStyles}
+                selectedValue={currentParticipantProject?.id}
+              />
+            ) : (
+              <Typography sx={selectStyles.input}>{options[0].label}</Typography>
```
nickgraz, line 54: If there is only 1 option, should it also be auto selected?

constanzauanini: Yes, but it's already selected

nickgraz: Oh I see, thanks

nickgraz (approved):

PR #2937, opened 2024-08-01
SW-5738 Sort number of participants on the FE

constanzauanini:
Move sorting of participants number to the FE

### DIFF
--- a/src/scenes/AcceleratorRouter/Cohorts/CohortsListView.tsx
+++ b/src/scenes/AcceleratorRouter/Cohorts/CohortsListView.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, useCallback, useMemo, useState } from 'react';
+import React, { ReactNode, useCallback, useEffect, useMemo, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
 import { Box, Card, Typography, useTheme } from '@mui/material';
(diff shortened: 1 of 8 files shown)


src/scenes/AcceleratorRouter/Cohorts/CohortsListView.tsx
```diff
@@ -60,8 +62,19 @@ const CohortsListView = ({ filterModifiers, extraTableFilters }: CohortsListView
...
 
+  useEffect(() => {
+    if (cohorts?.length) {
+      const newCohorts: CohortWithParticipantNum[] = cohorts.map((c) => ({
+        ...c,
+        numOfParticipants: c.participantIds?.length || 0,
+      }));
+      setCohortsWithParticipantNum(newCohorts);
+    }
+  }, [cohorts]);
```
nickgraz, line 76: Instead of creating more state, you can determine this within the cell renderer. Inside the `CohortCellRenderer`, do something like this:

```
if (column.key === 'numOfParticipants') {
    return <CellRenderer {...props} value={((row as Cohort)?.participantIds || []).length} />;
  }
```

constanzauanini: Yes, this was done like this before, but the sorting uses the original value, not the rendered one, so we need the rows to have the quantity number value

nickgraz: Oh I see. In that case maybe it should be hydrated into the data at the service level?

nickgraz (approved):

PR #2967, opened 2024-08-06
SW-5579 Link to a Project's Deliverables from Project Profile page

constanzauanini:
- Modified the TableWithSearchFilters component to allow having a filter applied from the creation of the table.
- Added the projectName filter to the DeliverablesTable, when coming from the link in the Participant Project page

### DIFF
--- a/src/components/DeliverablesTable/index.tsx
+++ b/src/components/DeliverablesTable/index.tsx
@@ -1,8 +1,9 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { useNavigate } from 'react-router-dom';
 
 import { TableColumnType } from '@terraware/web-components';
 
-import { FilterConfig } from 'src/components/common/SearchFiltersWrapperV2';
+import { FilterConfig, FilterConfigWithValues } from 'src/components/common/SearchFiltersWrapperV2';
 import { useParticipants } from 'src/hooks/useParticipants';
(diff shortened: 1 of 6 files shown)


nickgraz (commented):
Have you looked at `useSessionFilters`? It might be better to implement that into the `TableWithSearchFilters` since we are likely going to want to include that functionality when we "combine all tables". If that is out of scope for this ticket, that's fine, but I'd be interested to see if you think that hook might work here. 

When it is implemented, you can send someone to the view with something like `&filter_projectId=3` and it will automatically apply the filter and save the filter value to session storage.

constanzauanini:
> Have you looked at `useSessionFilters`? It might be better to implement that into the `TableWithSearchFilters` since we are likely going to want to include that functionality when we "combine all tables". If that is out of scope for this ticket, that's fine, but I'd be interested to see if you think that hook might work here.
> 
> When it is implemented, you can send someone to the view with something like `&filter_projectId=3` and it will automatically apply the filter and save the filter value to session storage.

Yes, I modified `TableWithSearchFilters` to include that functionality, so we can reuse it later in all tables, but I didn't know about the useSessionFilters. I will check that, thanks

nickgraz (approved):

constanzauanini:
@nickgraz I will create a subtask in this ticket https://terraformation.atlassian.net/browse/SW-5750 to take care of the useSessionFilters in the table component, since it was not a small change to be done it in this same PR

PR #2973, opened 2024-08-07
SW-5712: "Invasive species to remove" table field appears to have "Photo" column w/o input option

joshbuchea:


src/components/DocumentProducer/DeliverableEditTable/index.tsx
```diff
@@ -97,10 +110,17 @@ const DeliverableEditableTableEdit = ({ variable, onChange }: DeliverableEditabl
...
+                      <DeliverableEditableCell
                         id={`${cell.rowId}-${cell.colId}`}
                         column={columns[colNum]}
                         onChange={(value) => setCellValue(rowNum, colNum, value as string | number)}
+                        projectId={projectId}
+                        setDeletedImages={setDeletedImages}
+                        setImages={setImages}
+                        setNewImages={(variableId, values) => {
+                          setNewImages(variableId, values);
+                          // setCellValue(rowNum, colNum, values as PhotoWithAttributesAndUrl[]);
```
joshbuchea, line 122: Do we need to also update the cell value?

src/components/DocumentProducer/EditableTableModal/helpers.ts
```diff
@@ -48,6 +53,16 @@ export const newValueFromEntry = (value: string | number, type: CellVariableType
...
+    case 'Image':
+      const newPhoto = (value as PhotoWithAttributesAndUrl[])[0];
+
+      return {
+        id: -1,
+        listPosition: 0,
+        type: 'Image',
+        caption: newPhoto.caption ?? '',
+        citation: newPhoto.citation ?? '',
+      };
```
joshbuchea, line 65: Right now the call to `setCellValue` is commented out, but this value will be returned as a new cell value for a new image. I'm not sure if this data is correct or if the step to update the cell value is needed.

joshbuchea:
As discussed in the [Jira ticket](https://terraformation.atlassian.net/browse/SW-5712), we've reclassified this task as a story due to additional specification and design needs.

Will keep this branch around in case this work is helpful when we re-visit it.

PR #2976, opened 2024-08-08
SW-5759: When Min/Max Carbon Accumulation is not filled out, put N/A to match other fields

joshbuchea:
Use `undefined` rather `false` as fallback value so `ProjectField` will display N/A when there's no value.

### Before

localhost_3000_accelerator_projects_34 (2)

### After

localhost_3000_accelerator_projects_34


constanzauanini (approved):
Nice!

nickgraz (approved):

PR #2986, opened 2024-08-09
SW-5797: Missing table columns for application console

joshbuchea:
This PR includes changes to add missing table columns & filters to the applications table on the console side.

### Before

staging terraware io_accelerator_applications

### After

localhost_3000_accelerator_applications


src/scenes/AcceleratorRouter/Applications/ApplicationsList.tsx
```diff
@@ -60,10 +125,15 @@ const ApplicationList = () => {
         result.data
           .filter((application) => application.status !== 'Not Submitted')
           .map((application) => ({
+            // TODO: get country code from API search results
+            countryCode: '',
```
joshbuchea, line 129: @nickgraz and/or @sgrimm, we need to display a country value in the table of applications on the console side, but currently there's no data for country being returned in `ApplicationPayload`. Can we add/include either the country code or name to the payload? If not, what other data can be used to determine country?

src/scenes/AcceleratorRouter/Applications/ApplicationsList.tsx
```diff
@@ -44,11 +72,48 @@ const defaultSearchOrder: SearchSortOrder = {
...
+      const populateCountries = async () => {
+        const response = await LocationService.getCountries();
+        if (response) {
+          setCountries(response);
+        }
+      };
+
+      populateCountries();
+    }
+  }, [activeLocale]);
```
nickgraz, line 116: This poor snippet of code is now in 7 locations

src/scenes/AcceleratorRouter/Applications/ApplicationsList.tsx
nickgraz: It looks like the "project accelerator details" are [filled in](https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/accelerator/ApplicationService.kt#L54) when an application passes the pre-screen, which includes the country code, determined by the project boundary.

nickgraz: @sgrimm do you know if the country is collected during the application process? Or is it only available after the pre-screen is passed?

nickgraz: It looks like [this](https://github.com/terraware/terraware-web/blob/8d690c8601acb67644193b30be55540b61c0c184/src/services/ApplicationService.ts#L44) is the underlying API call that the view is making to populate the table. It is a custom endpoint (as opposed to the Search API) so we will need to add the data in there on the backend [here](https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/accelerator/api/ApplicationsController.kt#L145-L157)

nickgraz: Scratch that, the country code is stored directly on the `projects` row. So I think you can add `APPLICATIONS.projects.COUNTRY_CODE` to [this select](https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/accelerator/db/ApplicationStore.kt#L603-L609) to get what you're looking for, since that fetch is what runs for the "list applications" API. It isn't guaranteed to have a value though

src/scenes/AcceleratorRouter/Applications/ApplicationsList.tsx
joshbuchea: Poor snippet indeed 😢

joshbuchea:
Circling back to this PR, @tommylau523 do you have bandwidth available to add `countryCode` to the `ApplicationPayload`?

tommylau523:
> Circling back to this PR, @tommylau523 do you have bandwidth available to add `countryCode` to the `ApplicationPayload`?

Missed this! I will add it today!

joshbuchea:
@tommylau523 can you also add `type` to `ApplicationPayload`? I added it client-side to `ApplicationRow` but then searching/sorting doesn't work correctly in the table because it happens prior to the client-side data changes.

tommylau523 (approved):

tommylau523:
@joshbuchea For the type, we can do some FE parsing for that. Essentially, they are by status and I think only "Not Submitted" and "Failed Pre-screen" will be in the Pre-screen types. 

I think it is worth checking with product folks to see if we should remove it. Since applications in "Not Submitted" are not even supposed to show up on this table. It doesn't really make sense to me to have a column and filter that just checks if a status is equal to one option.

jenniferyimTF:
@joshbuchea @tommylau523 Based on the above information, if it was up to me, I would say, remove the Type column and filter from the Applications table.

joshbuchea:
@jenniferyimTF posted my comment before I noticed yours, I'll go ahead and remove the Type column & filter. Thanks for following up here in GitHub 🙌

PR #3009, opened 2024-08-13
SW-5672 Implement confirmation dialog for species being added to projects with active or past deliverables

nickgraz:
- Allow the deliverable search to take an array of project IDs
- Encapsulate the species deliverable search behavior into a hook
- Add `AddToProjectConfirmModal` that appears if a user attempts to add a species to a project that has an active or past deliverable


nickgraz:
* **#3009** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3009?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/scenes/Species/SpeciesProjectsTable.tsx
```diff
@@ -181,9 +180,11 @@ export default function SpeciesProjectsTable({
           onSubmit={onRemoveHandler}
         />
       )}
-      {openedAddToProjectModal && selectableProjects && (
```
constanzauanini, line 184: Is it ok to remove this condition? _ selectableProjects_ ?

constanzauanini (approved):
Looks good! Just one question

src/scenes/Species/SpeciesProjectsTable.tsx
nickgraz: I believe so, because `selectableProjects` is initialized as an empty array which is truthy. So this condition would never be false anyways.

PR #3016, opened 2024-08-15
SW-5849: Console: Internal Comments should display regardless of variable's status

joshbuchea:
This PR includes changes to fix an issue that caused feedback and/or internal comments values to get wiped erroneously while "updating variable workflow details" (status, feedback, or internal comments).

I also included several changes to improve the UX of the Internal Comments field:

- Internal comments dialog should not close prematurely when a user deletes all of the text in the textarea
- Internal comments should be reset to the original value when the user presses Cancel in the dialog
- Internal comments should not update the value under the dialog until the user presses Save in the dialog

### DIFF
--- a/src/components/Variables/VariableInternalComment.tsx
+++ b/src/components/Variables/VariableInternalComment.tsx
@@ -34,6 +34,7 @@ function VariableInternalComment({ variable, update, editing, sx }: VariableInte
 
   const toggleDialog = useCallback(() => {
     setIsDialogOpen((prev) => !prev);
+    setInternalComment(variableValue?.internalComment || '');
   }, []);
 
   const handleUpdate = () => {
@@ -42,7 +43,7 @@ function VariableInternalComment({ variable, update, editing, sx }: VariableInte
   };
 
   return (
-    (editing || internalComment) && (
+    (editing || internalComment || isDialogOpen) && (
       <Box sx={[...(Array.isArray(sx) ? sx : [sx])]}>
         <Box
           display='flex'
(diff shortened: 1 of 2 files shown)


src/components/Variables/VariableInternalComment.tsx
```diff
@@ -34,6 +34,7 @@ function VariableInternalComment({ variable, update, editing, sx }: VariableInte
 
   const toggleDialog = useCallback(() => {
     setIsDialogOpen((prev) => !prev);
+    setInternalComment(variableValue?.internalComment || '');
```
joshbuchea, line 37: This resets internal comments to the original value every time the dialog is closed

src/components/Variables/VariableInternalComment.tsx
```diff
@@ -42,7 +43,7 @@ function VariableInternalComment({ variable, update, editing, sx }: VariableInte
   };
 
   return (
-    (editing || internalComment) && (
+    (editing || internalComment || isDialogOpen) && (
```
joshbuchea, line 46: Without `isDialogOpen` the Dialog closes prematurely (no longer rendered) whenever the text is set to an empty string

src/components/Variables/VariableInternalComment.tsx
```diff
@@ -59,7 +60,7 @@ function VariableInternalComment({ variable, update, editing, sx }: VariableInte
             onChange={(value) => setInternalComment(value as string)}
             preserveNewlines
             type='textarea'
-            value={internalComment || strings.NO_COMMENTS_ADDED}
+            value={variableValue?.internalComment || strings.NO_COMMENTS_ADDED}
```
joshbuchea, line 63: The field value shouldn't change as the user is typing in the dialog, it shouldn't change until the user presses Save

constanzauanini (approved):
Thanks for this fixes!

PR #3018, opened 2024-08-15
SW-5814: An org with no projects that are not currently associated with an application autoselects "Select Existing Project" option

joshbuchea:
This PR includes changes to resolve an issue where the "Select Existing Project" option is auto-selected when there are no available projects without an application.

### DIFF
--- a/src/scenes/ApplicationRouter/NewApplicationModal.tsx
+++ b/src/scenes/ApplicationRouter/NewApplicationModal.tsx
@@ -1,7 +1,7 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 
 import { FormControlLabel, Grid, Radio, RadioGroup, useTheme } from '@mui/material';
-import { BusySpinner, Dropdown } from '@terraware/web-components';
+import { BusySpinner, Dropdown, DropdownItem } from '@terraware/web-components';
 
 import DialogBox from 'src/components/common/DialogBox/DialogBox';
 import TextField from 'src/components/common/Textfield/Textfield';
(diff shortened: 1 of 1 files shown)


src/scenes/ApplicationRouter/NewApplicationModal.tsx
```diff
@@ -58,9 +58,9 @@ const NewApplicationModal = ({ open, onClose }: NewApplicationModalProps): JSX.E
   );
   const createApplicationResult = useAppSelector(selectApplicationCreate(createApplicationRequestId));
 
-  const [newApplication, , onChange] = useForm<NewApplication>({
-    projectType: availableProjects && availableProjects.length > 0 ? 'Existing' : 'New',
-    projectId: availableProjects && availableProjects.length > 0 ? availableProjects[0].id : undefined,
+  const [newApplication, setNewApplication, onChange] = useForm<NewApplication>({
+    projectType: 'New',
+    projectId: undefined,
```
joshbuchea, line 63: We should be checking `projectOptions` rather than `availableProjects` here

src/scenes/ApplicationRouter/NewApplicationModal.tsx
```diff
@@ -71,6 +71,15 @@ const NewApplicationModal = ({ open, onClose }: NewApplicationModalProps): JSX.E
...
 
+  useEffect(() => {
+    if (newApplication.projectType === 'New' && projectOptions.length) {
+      setNewApplication({
+        projectType: 'Existing',
+        projectId: undefined,
+      });
+    }
+  }, [newApplication, projectOptions, setNewApplication]);
+
```
joshbuchea, line 82: Since we can't determine if there are available projects without an application until `projectOptions` is computed, I set the default to `New` and added a `useEffect` to select `Existing` if any such projects are found.

I tried setting `projectId` to `projectOptions[0].value` to align with existing logic, but while `newApplication` updated, the Dropdown didn't reflect this change. Testing in Staging confirmed the Dropdown doesn't show the auto-selected `projectId`. Instead of spending more time on this, I set `projectId` to `undefined`. If auto-selecting in the Dropdown is necessary, I can revisit and modify that behavior.

nickgraz: I think the issue is that the computation of the project options and the initialization of the form are happening in two different places. Also, the `useMemo` which is computing `projectOptions` is missing `allApplications` from its dependency array, so if `availableProjects` is done updating before the final value for `allApplications` is received, it will be incorrect. 

Generally these things should be done in a single `useEffect` statement, so you aren't juggling state values between multiple memoized functions and hoping the order of resolution is correct. I will push up a PR into this branch with the fix.

joshbuchea: Sounds good, thanks @nickgraz 🙌

nickgraz (approved):

joshbuchea:
Thanks for the assist @nickgraz 🙌

PR #3020, opened 2024-08-15
SW-5596 Deliverables List in PartEx: Address issues with Project dropdown behavior

constanzauanini:
Make project dropdown a selector instead of a filter.
For single-Project Participants, the Project is automatically selected.
Change the “No Project” value text to “All”.

### DIFF
--- a/src/components/ProjectsDropdown/index.tsx
+++ b/src/components/ProjectsDropdown/index.tsx
@@ -6,22 +6,24 @@ import { useLocalization } from 'src/providers';
 import strings from 'src/strings';
 import { Project } from 'src/types/Project';
 
-type ProjectsDropdownProps<T extends { projectId?: number } | undefined> = {
+type ProjectsDropdownProps<T extends { projectId?: number | string } | undefined> = {
   allowUnselect?: boolean;
   availableProjects: Project[] | undefined;
   label?: string | undefined;
   record: T;
   required?: boolean;
   setRecord: (setFn: (previousValue: T) => T) => void;
+  unselectLabel?: string;
 };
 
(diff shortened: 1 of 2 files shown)


nickgraz (commented):
It seems the best way to do this is to allow the consumer to pass in the "no project" filter label. The functionality seems identical (aside from the auto selection if there's only one project).

constanzauanini:
> It seems the best way to do this is to allow the consumer to pass in the "no project" filter label. The functionality seems identical (aside from the auto selection if there's only one project).

I'm also keeping the selected project to the other pages (modules, home, etc) with this change. Do you think that is not needed? I used the same selector as in home and modules page

nickgraz:
> > It seems the best way to do this is to allow the consumer to pass in the "no project" filter label. The functionality seems identical (aside from the auto selection if there's only one project).
> 
> I'm also keeping the selected project to the other pages (modules, home, etc) with this change. Do you think that is not needed? I used the same selector as in home and modules page

Oh OK I see. That makes sense. I guess I am still unsure why the entire dropdown is being rebuilt.

alexTerraformation:
Not sure if it's related to any changes for this task, but I found today that none of the deliverables filters are working at all at the moment.  https://terraformation.atlassian.net/browse/SW-5865

constanzauanini:
> Not sure if it's related to any changes for this task, but I found today that none of the deliverables filters are working at all at the moment. https://terraformation.atlassian.net/browse/SW-5865

They work on this branch, probably some change from staging. Will review

nickgraz (approved):

PR #3048, opened 2024-08-20
SW-5754 Allow organizations to be marked as accelerator orgs in console

constanzauanini:


src/scenes/AcceleratorRouter/Organizations/AddAcceleratorOrganizationModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+export interface AddAcceleratorOrganizationModalProps {
+  onClose: (reload?: boolean) => void;
+  onSubmit: (orgId?: string) => void;
+  acceleratorOrgsIds: number[];
+}
+
+export default function AddAcceleratorOrganizationModal(props: AddAcceleratorOrganizationModalProps): JSX.Element {
+  const { onClose, onSubmit, acceleratorOrgsIds } = props;
+  const theme = useTheme();
+  const { organizations } = useOrganization();
```
constanzauanini, line 20: @nickgraz I think this is not getting all the existing orgs. Do you know why?

nickgraz: The organization provider looks like it is only used for _user_ orgs, not "all" orgs. [Here's](https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt#L57-L60) the API that it hits. Perhaps you are looking to use [this API](https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/accelerator/api/AcceleratorOrganizationsController.kt#L25-L30).

nickgraz: Hmm actually it looks like that backend API only deals with orgs that are already tagged, so it won't work for what you're doing. I think we need a new API or to optionally make the pre-existing one more permissive.

nickgraz (approved):
Looks good to me!

PR #3067, opened 2024-08-22
SW-5475 Add module stepper back in project profile view

constanzauanini:


joshbuchea (approved):
I had to look a little closer to understand why this restores the module stepper, but this looks good! 🚢

constanzauanini:
> I had to look a little closer to understand why this restores the module stepper, but this looks good! 🚢

Because in the console side, there is no "current participant" to get the project id from. The project profile page has to get the id from the url

PR #3079, opened 2024-08-23
SW-5826: Add support for multi-select

joshbuchea:
This PR adds support for multi-select fields.


src/components/DocumentProducer/DeliverableVariableDetailsInput/index.tsx
```diff
@@ -307,14 +309,32 @@ const DeliverableVariableDetailsInput = ({
...
         <Dropdown
-          onChange={(newValue: any) => onChangeValueHandler(newValue, 'value')}
+          fullWidth
           label=''
+          onChange={(newValue: any) => onChangeValueHandler(newValue, 'value')}
           options={getOptions()}
-          selectedValue={value}
-          fullWidth={true}
+          selectedValue={(value as number[])?.[0]}
+          sx={[formElementStyles, { paddingBottom: theme.spacing(1) }]}
```
joshbuchea, line 319: Most of these lines are just re-ordering props, but I also removed the redundant `={true}` from `fullWidth` and changed the value of `selectedValue` from `value` to `(value as number[])?.[0]`.

constanzauanini (approved):
Working correctly in Vercel!

PR #3086, opened 2024-08-26
SW-5806 Add more details about the Accelerator in the Apply card

constanzauanini:
Screenshot 2024-08-26 at 10 35 09 AM


nickgraz (approved):

joshbuchea:
I know previously when Karthik was still here, we pushed back against inline links within copy and we made the decision to move in that direction for all future copy. I didn't realize how simple it actually is to include a link inline, this looks great! 🙌

PR #3064, opened 2024-08-22
SW-5825: Can't type accents in Questionnaire Deliverable inputs

joshbuchea:
This PR includes changes to how state is managed for text values in Questionnaire Deliverables. These changes resolve two issues:

- SW-5825: Can't type accents in Questionnaire Deliverable inputs
- SW-5853: Typing mid-word will automatically moved the cursor to the end of the text after one letter is typed

PR #3088, opened 2024-08-26
SW-5902: Full application questionnaire deliverables inputs missing

joshbuchea:
This PR fixes a regression from my [previous PR for SW-5825](https://github.com/terraware/terraware-web/pull/3064), where I missed adding an empty fallback value for text variables without pre-existing values. This update corrects that oversight.


alexTerraformation (approved):
Thanks!

PR #3092, opened 2024-08-27
SW-5754 improvement - Use autocomplete for org selector

constanzauanini:
Depends on [this PR ](https://github.com/terraware/web-components/pull/633) from web-components


nickgraz (approved):
This looks good, it looks like the `package.json` need to be updated though.

PR #3097, opened 2024-08-27
SW-5826: Add support for multi-select

joshbuchea:
This PR includes changes to add support for multi-select fields in Doc Producer, bringing it inline with the recent changes in Deliverables.

This PR includes the same set of functional changes as [the PR that added multi-select support for deliverables](https://github.com/terraware/terraware-web/pull/3079).

## Examples

localhost_3000_accelerator_documents_10_tab=variables

https://github.com/user-attachments/assets/66ad534b-3515-4114-a509-dbe0fd191105

### DIFF
--- a/src/components/DocumentProducer/DeliverableVariableDetailsInput/index.tsx
+++ b/src/components/DocumentProducer/DeliverableVariableDetailsInput/index.tsx
@@ -312,7 +312,7 @@ const DeliverableVariableDetailsInput = ({
       {variable.type === 'Select' && !variable.isMultiple && (
         <Dropdown
           fullWidth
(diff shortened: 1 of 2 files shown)


constanzauanini:
Looks like you are adding/removing values from the "description" field. Shouldn't be the "values" field?

joshbuchea:
@constanzauanini ooh good catch! I forgot to add a label to the multi-select field, will fix.

joshbuchea:
The **Value** label was missing from the value input field, which made it appear to be associated with the **Description** label.

constanzauanini (approved):

PR #3104, opened 2024-08-29
SW-5912: FE display multi-line text variable inputs as a large text area (Follow-up #1)

joshbuchea:
This PR includes a follow-up change to extend the width of multi-line text fields to 100%.

### Before

localhost_3000_accelerator_overview_tab=projects

### After

localhost_3000_accelerator_overview_tab=projects (1)


joshbuchea:
@constanzauanini nice improvements in this PR, great work!

I wanted to mentioned that Léann previously opened a [bug re: auto-scrolling a tiny bit to questions that were already in the viewport](https://terraformation.atlassian.net/browse/SW-5689) (I believe because it resulted in an odd UX). Hopefully these changes don't introduce a regression by restoring that undesired behavior, if we didn't account for that here we'll likely need to address it.

PR #3114, opened 2024-08-30
SW-5919 Document outline should navigate through document on click

constanzauanini:
- Previously, we where checking if the section header was in the viewport to scroll into it or not. Since we always want to center that section, this validation was removed.

- Also, some sections didn't have id, so it was impossible to scroll into them. Added ids to those sections

- Finally, the scroll was not correct because the header of the page was interfering with the position. Added a gap to scroll into the element, taking the header space into account

### DIFF
--- a/src/components/DocumentProducer/EditableSection/Container.tsx
+++ b/src/components/DocumentProducer/EditableSection/Container.tsx
@@ -25,6 +25,7 @@ import Display from './Display';
 import Edit from './Edit';
 
 type EditableSectionProps = {
+  id?: string;
   docId: number;
   projectId: number;
   section: SectionVariableWithValues;
@@ -33,6 +34,7 @@ type EditableSectionProps = {
 };
 
 export default function EditableSectionContainer({
+  id,
   docId,
   projectId,
   section,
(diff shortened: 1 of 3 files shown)


tommylau523 (approved):

joshbuchea:
@constanzauanini nice improvements in this PR, great work!

I wanted to mentioned that Léann previously opened a [bug re: auto-scrolling a tiny bit to questions that were already in the viewport](https://terraformation.atlassian.net/browse/SW-5689) (I believe because it resulted in an odd UX). Hopefully these changes don't introduce a regression by restoring that undesired behavior, if we didn't account for that here we'll likely need to address it.

constanzauanini:
> @constanzauanini nice improvements in this PR, great work!
> 
> I wanted to mentioned that Léann previously opened a [bug re: auto-scrolling a tiny bit to questions that were already in the viewport](https://terraformation.atlassian.net/browse/SW-5689) (I believe because it resulted in an odd UX). Hopefully these changes don't introduce a regression by restoring that undesired behavior, if we didn't account for that here we'll likely need to address it.

Thanks for pointing that out @joshbuchea, will review it!

PR #3122, opened 2024-09-03
SW-5697 Add required question support and completion criteria for deliverables

constanzauanini:
Screenshot 2024-09-03 at 4 32 56 PM


src/components/DeliverableView/QuestionsDeliverableEditForm.tsx
```diff
@@ -181,12 +188,41 @@ const QuestionsDeliverableEditForm = (props: QuestionsDeliverableEditViewProps):
...
     }
 
+    if (!validateForm()) {
+      setValidateFields(true);
+      return;
+    }
+
+    // If Questionnaire Deliverable is part of the Application/Pre-screen mark deliverable as “Completed”
+    if (isPrescreen) {
+      updateDeliverable({ ...deliverable, status: 'Completed' });
```
tommylau523, line 223: This will fail if used by an external user.
Our BE always will run this step so I don't think we need this.

constanzauanini: But in the [Jira ticket](https://terraformation.atlassian.net/browse/SW-5697) Nick and @sgrimm said we needed to do this in the FE. Did I misunderstood?

tommylau523: I missed those updates previously. In this case, we need to use the `completeDeliverable` hook instead.

tommylau523: Which I just realized we don't have yet... So we need to add the Redux for it as well. 

It should be identical to `DeliverableService.submit()`, but just with `/{deliverableId}/submissions/{projectId}/complete` instead. 

https://github.com/terraware/terraware-web/blob/main/src/services/DeliverablesService.ts#L164

https://github.com/terraware/terraware-web/blob/main/src/redux/features/deliverables/deliverablesAsyncThunks.ts#L82

https://github.com/terraware/terraware-web/blob/main/src/redux/features/deliverables/deliverablesSlice.ts#L65

https://github.com/terraware/terraware-web/blob/main/src/components/DeliverableView/useSubmitDeliverable.ts

tommylau523: Sorry this turns out to be a lot of changes required.

constanzauanini: No worries, and thanks for the feedback. Will update it!

tommylau523 (approved):

PR #3123, opened 2024-09-03
SW-5851 Save modified or deleted table cells

sgrimm:
Adding rows to table variable values and editing existing table cells
wasn't working for two reasons:

* There was an optimization that skipped saving the value if it
  looked like the user hadn't changed anything. But that check was
  only looking at changes to simple variables like text and number
  fields, and was ignoring changes to other variable types such as
  tables and images. The code that assembles the server request
  already has logic to skip the request if it would be a no-op,
  so the optimization wasn't needed anyway; it's gone now.

* The code that assembled the server request was comparing the
  initial cell values with themselves and thinking that edited
  cells hadn't changed. Now it compares against the values from
  the edit form.

Also fixes SW-5933.

### DIFF
--- a/src/components/AcceleratorDeliverableView/QuestionsDeliverableCard.tsx
+++ b/src/components/AcceleratorDeliverableView/QuestionsDeliverableCard.tsx
@@ -160,10 +160,6 @@ const QuestionBox = ({
   const onSave = () => {
     setEditingId(undefined);
 
-    if (pendingVariableValues.size === 0) {
-      return;
-    }
-
     update();
   };
 
--- a/src/components/DeliverableView/QuestionsDeliverableEditForm.tsx
+++ b/src/components/DeliverableView/QuestionsDeliverableEditForm.tsx
@@ -181,15 +181,6 @@ const QuestionsDeliverableEditForm = (props: QuestionsDeliverableEditViewProps):
     }
   }, [exit, updateSuccess, uploadSuccess]);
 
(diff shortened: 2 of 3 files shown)


sgrimm:
* **#3123** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3123?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

constanzauanini:
I did similar changes in this [PR](https://github.com/terraware/terraware-web/pull/3076/) but I was not getting the table updated in some cases, let me test with your changes

constanzauanini:
The issue I was seeing before is still happening. For this deliverable https://tree-location-web-git-table-cells-terraformation.vercel.app/deliverables/123/submissions/32?organizationId=141 add values for all the columns in the table, but it just saves the first 2 columns

sgrimm:
Thanks -- will look at that. Sorry I missed your other PR!

sgrimm:
@constanzauanini The bug with the columns not getting saved is actually a symptom of the problems with updated variable definitions not causing values to be migrated properly. The good news is the values _are_ actually getting saved, but they're getting saved to older versions of the column variables that aren't included in the response when you list the values for the deliverable.

The fix for that is the next thing I'll be focusing on; it'll mostly be on the BE with some FE impact that's unrelated to this PR.

tommylau523 (approved):

PR #3136, opened 2024-09-05
SW-5931: Existing Project not recognized when Starting New Application

joshbuchea:
This PR includes a fix for an issue that caused new projects to not be recognized/available as an option when starting a new application.

### Before

localhost_home_organizationId=181

### After

localhost_home_organizationId=181 (1)


tommylau523 (approved):

src/scenes/ApplicationRouter/NewApplicationModal.tsx
```diff
@@ -65,7 +65,7 @@ const NewApplicationModal = ({ open, onClose }: NewApplicationModalProps): JSX.E
 
   // Init the options and auto select the first project option if there is a project without an application
   useEffect(() => {
-    if (!allApplications || allApplications.length === 0) {
```
tommylau523, line 68: I think we need an additional "initiated" flag to make sure this doesn't keep firing.

PR #3148, opened 2024-09-06
SW-5768 Add modal to change document status

constanzauanini:
Screenshot 2024-09-06 at 4 09 16 PM


joshbuchea (dismissed):
The changes look clean, but the example image doesn't quite match the image or linked designs in the ticket. Also, it seems like the top field needs a label?

constanzauanini:
> The changes look clean, but the example image doesn't quite match the image or linked designs in the ticket. Also, it seems like the top field needs a label?

Yes, the design didn't take the status into account, so these requested changes are described in the comments in the Jira ticket.

joshbuchea:
Ok I thought I may have been missing some information, thanks for the context @constanzauanini.

joshbuchea:
Changes look good, but it still seems like there should be a label for the Internal Comments field?

constanzauanini:
> Changes look good, but it still seems like there should be a label for the Internal Comments field?

Ok, I will add it. I wasn't sure if we needed it

joshbuchea (approved):
Thanks for adding the label 🙌

PR #3150, opened 2024-09-06
SW-5978: Saving an unchanged questionnaire deliverable should return the user to the view display mode

joshbuchea:
This PR includes:

- A fix for an issue that was preventing saving questionnaire deliverables (modified or not)
- A fix to return to display mode when pressing save with no pending changes in questionnaire deliverables


src/components/DeliverableView/QuestionsDeliverableEditForm.tsx
```diff
@@ -219,7 +219,6 @@ const QuestionsDeliverableEditForm = (props: QuestionsDeliverableEditViewProps):
       }
     } else {
       setValidateFields(true);
-      return;
```
joshbuchea, line 222: This was preventing saving any non-application questionnaire deliverable

tommylau523 (approved):

PR #3151, opened 2024-09-07
SW-5976: Optional question preventing questionnaire saving

joshbuchea:
This PR includes a fix for a `400` error that occurs when a user enters text and then deletes it before pressing save, for a text variable question with no previous value within a questionnaire deliverable.

### DIFF
--- a/src/hooks/useProjectVariablesUpdate/util.ts
+++ b/src/hooks/useProjectVariablesUpdate/util.ts
@@ -184,7 +184,7 @@ export const makeVariableValueOperations = ({
           value: nV,
           existingValueId: pendingValues[index].id,
         });
-      } else {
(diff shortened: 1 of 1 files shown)


src/hooks/useProjectVariablesUpdate/util.ts
```diff
@@ -184,7 +184,7 @@ export const makeVariableValueOperations = ({
           value: nV,
           existingValueId: pendingValues[index].id,
         });
-      } else {
+      } else if (nV.textValue) {
```
joshbuchea, line 187: If `nV.textValue` isn't set here, there's no operation to perform because no text is being added or updated.

constanzauanini (approved):

PR #3156, opened 2024-09-09
Not allow submitting questionnaire with missing required fields

constanzauanini:


tommylau523 (approved):

src/hooks/useProjectVariablesUpdate/index.ts
```diff
@@ -49,6 +50,7 @@ export const useProjectVariablesUpdate = (
   const [pendingNewImages, setPendingNewImages] = useState<Map<number, PhotoWithAttributes[]>>(new Map());
   const [pendingVariableValues, setPendingVariableValues] = useState<Map<number, VariableValueValue[]>>(new Map());
   const [removedVariableValues, setRemovedVariableValues] = useState<Map<number, VariableValueValue>>(new Map());
+  const [noChanges, setNoChanges] = useState<boolean>(true);
```
joshbuchea, line 53: @constanzauanini I just added a `noOp` flag in this file [here](https://github.com/terraware/terraware-web/blob/main/src/hooks/useProjectVariablesUpdate/index.ts#L60) that serves the same purpose as the `noChanges` flag added in this PR. We should consolidate and only use one.

PR #3163, opened 2024-09-10
SW-5989 Questionnaire deliverable not marked as complete when all required questions are filled out

constanzauanini:
Some required questions were marked as incomplete because of the condition: getRawValue(variable) === undefined || getRawValue(variable) === ''

Only use that condition for text, number and select variables. Other varaibles types return undefined for getRawValue()


tommylau523 (commented):
I actually folded in some of the changes into this PR:
https://github.com/terraware/terraware-web/pull/3165

Sorry for doing that. But I think there are a few related issues causing deliverables to not mark as completed or not check the correct values.

PR #3164, opened 2024-09-10
SW-5990 Some questionnaire deliverables need to be saved 2x to be marked as complete

constanzauanini:


src/utils/documentProducer/variables.ts
```diff
@@ -81,16 +82,22 @@ export const getRawValue = (variable: VariableWithValues): number | number[] | s
...
   const allRequiredVariables = variablesWithValues.filter(
     (v) => v.isRequired && variableDependencyMet(v, variablesWithValues)
   );
-
   const missingRequiredFields = allRequiredVariables.some((variable) => {
     let hasEmptyValue = false;
     const values = variable.values;
     if (!values || values.length === 0 || getRawValue(variable) === undefined || getRawValue(variable) === '') {
       hasEmptyValue = true;
+      if (pendingTableValues?.get(variable.id)?.length) {
```
tommylau523, line 97: This may be a lot to add, but we may need to do the following:
1) Add isRequired to the `VariableTableCell` type, this can be passed down from `TableVariableWithValues` in `getInitialCellValues`.
2) Check if an entire row is empty. If so, we can treat it like the row does not exist
3) For a row, check if a column is required. If it is, check if the value is not empty
4) Then we need to check a required table has at least one row (which we have here!)

tommylau523: If just one cell is needed, we do need to check if at least one cell is non-empty. Basically still step 2 and 4. A user right now can type something, hit backspace and submit an empty string.

tommylau523 (approved):
Looks like this is working correctly. Ok with submitting for now till we figure out what a required table means!

src/hooks/useProjectVariablesUpdate/index.ts
```diff
@@ -96,7 +96,7 @@ export const useProjectVariablesUpdate = (
   }, [pendingVariableValues, variablesWithValues]);
 
   const missingFields = useMemo(
-    () => missingRequiredFields(stagedVariableWithValues),
+    () => missingRequiredFields(stagedVariableWithValues, pendingCellValues),
     [missingRequiredFields, stagedVariableWithValues]
```
tommylau523, line 100: need to add pendingCellValues to deps

constanzauanini:
> Looks like this is working correctly. Ok with submitting for now till we figure out what a required table means!

Just added a small change to check for empty values

tommylau523:
Thanks Constanza! One comment about adding a deps to useCallback

tommylau523 (changes requested):
I think we need to update the cell checking

src/hooks/useProjectVariablesUpdate/index.ts
constanzauanini: Thanks!

src/utils/documentProducer/variables.ts
```diff
@@ -112,15 +113,24 @@ export const isValueEmpty = (
...
+export const missingRequiredFields = (
+  variablesWithValues: VariableWithValues[],
+  pendingTableValues?: Map<number, VariableTableCell[][]>
+) => {
   const allRequiredVariables = variablesWithValues.filter(
     (v) => v.isRequired && variableDependencyMet(v, variablesWithValues)
   );
-
   const missingRequiredFields = allRequiredVariables.some((variable) => {
     const values = variable.values;
```
tommylau523, line 124: I think this may work and we can move it to the top of the function. 

```
if (variable.type === 'Table') {
  const tableCells = !pendingTableValues?.get(variable.id)

  // Return true if there is no row
  if (!tableCells || tableCells.length === 0) {
    return true;
  }
  
  // Return true if every cell is empty
  return tableCells.every((row) => row.every(cell) => {
    if (!cell.values || cell.values.length ===0) {
        return true;
    } 
    return cellValue(cell.values[0]).toString() === ''
  }))
}
```

tommylau523 (approved):

src/utils/documentProducer/variables.ts
tommylau523: Ahh I understand.  Quite tricky because I think we need to catch the case of not adding a row to an empty table, as well as not modifying an existing table.

tommylau523: Perhaps we can add a `stagedCellValues` similar to `stagedVariableValues`

constanzauanini: Yes! That was my first idea but was getting trouble with typescript

constanzauanini: But if there is no row, then values of the table will be empty, so it will be marked as required in the first check

tommylau523:
I did test in Vercel, and submitting nothing in a table will still complete the Deliverable.

constanzauanini:
> I did test in Vercel, and submitting nothing in a table will still complete the Deliverable.

Oh, thanks for testing @tommylau523. I was trying to see what was the values of "pendingValues" when saving an empty table. 

Do you want to create a PR with the code you did to add the pendingValues to the stages variables?

tommylau523:
> > I did test in Vercel, and submitting nothing in a table will still complete the Deliverable.
> 
> Oh, thanks for testing @tommylau523. I was trying to see what was the values of "pendingValues" when saving an empty table. 
> 
> Do you want to create a PR with the code you did to add the pendingValues to the stages variables?

I am trying a few more things for now! I haven't really tested it out yet. Sorry for the back and forth, I was also just thinking of all the edge cases. I will open a PR if it ends up working!

constanzauanini:
> > > I did test in Vercel, and submitting nothing in a table will still complete the Deliverable.
> > 
> > 
> > Oh, thanks for testing @tommylau523. I was trying to see what was the values of "pendingValues" when saving an empty table.
> > Do you want to create a PR with the code you did to add the pendingValues to the stages variables?
> 
> I am trying a few more things for now! I haven't really tested it out yet. Sorry for the back and forth, I was also just thinking of all the edge cases. I will open a PR if it ends up working!

Ok, just checked that pendingValues has the table headers always, that's why it was failing.

tommylau523:
https://github.com/terraware/terraware-web/pull/3166 

This seems to work for me!

PR #3180, opened 2024-09-12
SW-5944 Vertical Stepper: Show the module that’s ending the soonest iVertical Stepper: Show the module that’s ending the soonest in the semantic/warning color

constanzauanini:
Screenshot 2024-09-12 at 3 16 25 PM

### DIFF
--- a/src/components/common/PageWithModuleTimeline/ModuleTimeline.tsx
+++ b/src/components/common/PageWithModuleTimeline/ModuleTimeline.tsx
@@ -12,11 +12,11 @@ import { CohortPhaseType } from 'src/types/Cohort';
 import { Module } from 'src/types/Module';
 
 type AltStepIconProps = {
-  isActive: boolean;
   index: number;
+  bgColor?: string;
 };
 
-const AltStepIcon = ({ isActive, index }: AltStepIconProps) => {
+const AltStepIcon = ({ index, bgColor }: AltStepIconProps) => {
   const theme = useTheme();
   const stepNumber = index + 1;
 
@@ -24,7 +24,7 @@ const AltStepIcon = ({ isActive, index }: AltStepIconProps) => {
     <Box
       sx={{
         alignItems: 'center',
-        bgcolor: isActive ? theme.palette.TwClrBaseGreen500 : theme.palette.TwClrBaseGray300,
+        bgcolor: bgColor,
         borderRadius: '50%',
         color: theme.palette.TwClrBaseWhite,
         display: 'flex',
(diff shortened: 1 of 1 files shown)


tommylau523:
Icon color too:
https://terraformation.atlassian.net/browse/SW-5560

I think there is a bit of customization available here:
https://mui.com/material-ui/api/step-icon/

tommylau523:
Sorry I saw that we can just add a prop to `AltStepIcon`! Didn't know we built that ourselves already!

constanzauanini:
> Sorry I saw that we can just add a prop to `AltStepIcon`! Didn't know we built that ourselves already!

Updated!

src/components/common/PageWithModuleTimeline/ModuleTimeline.tsx
```diff
@@ -46,6 +46,21 @@ export type ModuleTimelineProps = {
...
+  const now = new Date();
+  const futureModules = modules?.filter((module) => new Date(module.endDate) > now);
+
+  const soonestEndingModuleId = futureModules?.reduce((soonest, current) => {
+    const soonestEndDate = new Date(soonest.endDate);
+    const currentEndDate = new Date(current.endDate);
+    return currentEndDate < soonestEndDate ? current : soonest;
+  }, modules[0])?.id;
+
+  const warningLabelStyles = {
```
tommylau523, line 59: Sorry my comments were confusing. Do we need these styles or does the `AltStepIcon` already handle it? 

I think the MuiStepLabel ones are only used if we use the default icon form the `StepLabel`

tommylau523 (approved):

src/components/common/PageWithModuleTimeline/ModuleTimeline.tsx
constanzauanini: We need those styles for the text label color, and the AltStepIcon styles are for the icon

tommylau523: Looks good then! Thanks!!

src/components/common/PageWithModuleTimeline/ModuleTimeline.tsx
```diff
@@ -55,8 +70,23 @@ const ModuleTimeline = ({ cohortPhase, modules }: ModuleTimelineProps) => {
...
+                        : module.isActive
+                          ? theme.palette.TwClrBaseGreen500
+                          : theme.palette.TwClrBaseGray300
+                    }
+                  />
+                }
+                sx={
+                  soonestEndingModuleId === module.id
+                    ? warningLabelStyles
+                    : { fontWeight: 600, '.MuiStepLabel-label.Mui-disabled': { fontWeight: 600 } }
```
tommylau523, line 88: Would it makes sense to inject `fontColor` here directly? instead of making a style class?

tommylau523: Ok to submit either way! Mostly asking because I am not too good with CSS

constanzauanini: Without the classes the color is override by mui active / disabled colors

PR #3184, opened 2024-09-13
SW-5915 Home: Add Mixpanel Events for clicks on links in Application card

constanzauanini:


tommylau523 (approved):

src/mixpanelEvents.ts
```diff
@@ -4,4 +4,6 @@ export enum MIXPANEL_EVENTS {
   TOP_BAR_HOME = 'Top Bar Home',
   PART_EX_LEFT_NAV_DELIVERABLES = 'Participant Deliverables Nav Click',
   PART_EX_LEFT_NAV_MODULES = 'Participant Modules Nav Click',
+  HOME_ACCELERATOR_APPLY_BUTTON = 'Apply to Accelerator Click',
+  HOME_ACCELERATOR_TF_LINK = 'Accelerator TF Link Click',
```
joshbuchea, line 8: Will Mixpanel be used to track event types other than click? If so, I think it may be helpful to append the event type in the event name: `_CLICK` in this case.

joshbuchea (approved):

src/mixpanelEvents.ts
constanzauanini: I'm not sure, I followed the other events name that didn't have "click" in the name

joshbuchea: I see that the event type wasn't included in previously added events, following existing conventions makes sense.

PR #3234, opened 2024-09-23
SW-6052 Dependencies do not work across different questionnaire deliverables

constanzauanini:


src/utils/documentProducer/variables.ts
```diff
@@ -158,3 +158,25 @@ export const missingRequiredFields = (
     return false;
   });
 };
+
+export const getDependingVariablesStableIdsFromOtherDeliverable = (variablesWithValues: VariableWithValues[]) => {
+  const dependingStableVariablesId = variablesWithValues.reduce((acc: number[], v: VariableWithValues) => {
```
tommylau523, line 163: This can be a bit simpler if we use Sets

```
const existingStableIds = new Set<string>()
const dependentStableIds = new Set<string>()
variablesWithValues.forEach((variable) => {
  existingStableIds.add(variable.stableId)
  if (variable.dependencyVariableStableId) {
    dependentStableIds.add(variable.dependencyVariableStableId)
  }
});

// cannot use Set.difference here because it is only available is Node 22
existingStableIds.forEach((stableId) => dependentStableIds.remove(stableId));

return Array.from(dependentStableIds) 
```

src/components/DeliverableView/QuestionsDeliverableCard.tsx
```diff
@@ -87,13 +100,33 @@ const QuestionsDeliverableCard = (props: EditProps): JSX.Element | null => {
     [variablesWithValues]
   );
 
+  useEffect(() => {
+    const ids = getDependingVariablesStableIdsFromOtherDeliverable(variablesWithValues);
+    setSpecificVariables(ids);
```
tommylau523, line 105: we can dispatch here directly instead of going through a state

src/components/AcceleratorDeliverableView/QuestionsDeliverableCard.tsx
```diff
@@ -457,6 +477,15 @@ const QuestionsDeliverableCard = (props: EditProps): JSX.Element => {
     }
   }, [variablesWithValues]);
 
+  const specificVariablesWithValues = useAppSelector((state) =>
+    selectSpecificVariablesWithValues(state, specificVariables, deliverable.projectId)
+  );
+
+  useEffect(() => {
+    const ids = getDependingVariablesStableIdsFromOtherDeliverable(variablesWithValues);
+    setSpecificVariables(ids);
```
tommylau523, line 486: we can dispatch directly here instead of going through a setState

src/components/AcceleratorDeliverableView/QuestionsDeliverableCard.tsx
```diff
@@ -457,6 +477,15 @@ const QuestionsDeliverableCard = (props: EditProps): JSX.Element => {
     }
   }, [variablesWithValues]);
 
+  const specificVariablesWithValues = useAppSelector((state) =>
```
tommylau523, line 480: can we rename this to be dependentVariablesWithValues?

src/components/DeliverableView/QuestionsDeliverableCard.tsx
```diff
@@ -87,13 +100,33 @@ const QuestionsDeliverableCard = (props: EditProps): JSX.Element | null => {
...
+  }, [variablesWithValues]);
+
+  const specificVariablesWithValues = useAppSelector((state) =>
```
tommylau523, line 108: can we rename this to be dependentVariablesWithValues?

tommylau523 (commented):
The redux and service all look good! Some simplification and some renaming stuff

src/components/AcceleratorDeliverableView/QuestionsDeliverableCard.tsx
```diff
@@ -424,6 +432,7 @@ const QuestionsDeliverableCard = (props: EditProps): JSX.Element => {
 
   const [editingId, setEditingId] = useState<number | undefined>();
   const [updatePendingId, setUpdatePendingId] = useState<number | undefined>();
+  const [specificVariables, setSpecificVariables] = useState<number[]>([]);
```
tommylau523, line 435: can we rename this state to be dependentVariableStableIds?

src/components/DeliverableView/QuestionsDeliverableCard.tsx
```diff
@@ -77,6 +89,7 @@ const QuestionsDeliverableCard = (props: EditProps): JSX.Element | null => {
   const dispatch = useAppDispatch();
   const theme = useTheme();
   const { deliverable, hideStatusBadge } = props;
+  const [specificVariables, setSpecificVariables] = useState<number[]>([]);
```
tommylau523, line 92: can we rename this state to be dependentVariableStableIds?

src/components/DeliverableView/QuestionsDeliverableCard.tsx
```diff
@@ -111,7 +144,12 @@ const QuestionsDeliverableCard = (props: EditProps): JSX.Element | null => {
           }}
         >
           {filteredVariablesWithValues.map((variableWithValues: VariableWithValues, index: number) =>
-            variableDependencyMet(variableWithValues, filteredVariablesWithValues) ? (
+            variableDependencyMet(
+              variableWithValues,
+              specificVariablesWithValues
```
tommylau523, line 149: This is really minor, but is it possible to move this to another `useMemo` under `filteredVariablesWithValues`, mostly just to be consistent with where the other variablesWithValues are defined? Maybe can be named `dependentVariablesWithValues`

src/utils/documentProducer/variables.ts
constanzauanini: Looks much better! I did a complex thing! Thanks

tommylau523 (approved):

src/utils/documentProducer/variables.ts
```diff
@@ -158,3 +158,18 @@ export const missingRequiredFields = (
     return false;
   });
 };
+
+export const getDependingVariablesStableIdsFromOtherDeliverable = (variablesWithValues: VariableWithValues[]) => {
+  const existingStableIds = new Set<number>();
```
tommylau523, line 163: nit: in theory our stableIDs are in a string column. In practice they are all numbers in the spreadsheet. This is likely fine for now but we could in theory put strings in the spreadsheet.

constanzauanini: Oh, maybe I should update that then

PR #3237, opened 2024-09-25
SW-6050 Fix page cut off due to responsiveness

constanzauanini:


alexTerraformation (approved):
Wow, such a simple change!

PR #3238, opened 2024-09-25
SW-6014 TF Experts should not be able to edit Participants in Console

constanzauanini:


tommylau523 (approved):

src/scenes/AcceleratorRouter/Participants/ParticipantsView.tsx
```diff
@@ -106,16 +100,6 @@ export default function ParticipantsView(): JSX.Element {
 
     return (
       <Box display='flex' justifyContent='right'>
-        {canUpdateParticipants && (
```
tommylau523, line 109: Are we removing the edit button entirely?

tommylau523: Should AcceleratorAdmin+ be able to edit participants?

constanzauanini: Oh, you are right! We need to allow AcceleratorAdmin+ to edit. Will fix, thanks

PR #3247, opened 2024-09-27
SW-5291 Module Management in Console: Associate Modules with Cohorts

constanzauanini:


src/redux/features/modules/modulesAsyncThunks.ts
```diff
@@ -86,3 +88,112 @@ export const requestListModuleDeliverables = createAsyncThunk(
...
+    { rejectWithValue }
+  ) => {
+    const { modulesId, cohortId } = request;
+    let allSucceeded = true;
+
+    const promises = modulesId.map((moduleId) => CohortModuleService.deleteOne(moduleId, cohortId));
+
+    const results = await Promise.all(promises);
+
+    results.forEach((res) => {
```
tommylau523, line 128: can be if (results.every((result) => result.requestSuceeded))

src/redux/features/modules/modulesAsyncThunks.ts
```diff
@@ -86,3 +88,112 @@ export const requestListModuleDeliverables = createAsyncThunk(
...
+      cohortId: number;
+      requests: (UpdateCohortModuleRequest & {
+        moduleId: number;
+      })[];
+    },
+    { rejectWithValue }
+  ) => {
+    let allSucceeded = true;
+
+    const promises = requests.map((request) => {
```
tommylau523, line 181: can be Promise.all(requests.map((request) => CohortModuleService.update(...))

src/redux/features/modules/modulesAsyncThunks.ts
```diff
@@ -86,3 +88,112 @@ export const requestListModuleDeliverables = createAsyncThunk(
...
+      const { moduleId, ...entity } = request;
+      return CohortModuleService.update({ moduleId: moduleId, cohortId, entity });
+    });
+
+    const results = await Promise.all(promises);
+
+    results.forEach((res) => {
```
tommylau523, line 188: same for using `results.every(...)` as above

src/scenes/AcceleratorRouter/Cohorts/CohortEditView.tsx
```diff
@@ -72,19 +79,14 @@ export default function CohortEditView(): JSX.Element {
   }
 
   return (
-    <PageWithModuleTimeline
```
tommylau523, line 75: We can check with designs but we likely do want to keep module stepper. I think in older mocks they have them.

src/services/CohortModuleService.ts
```diff
@@ -0,0 +1,37 @@
+import { UpdateCohortModuleRequest } from 'src/types/Module';
+
+import HttpService, { Response } from './HttpService';
+
+const COHORT_MODULE_ENDOINT = '/api/v1/accelerator/cohorts/{cohortId}/modules/{moduleId}';
+
+/**
+ * Delete cohort module
+ */
+const deleteOne = async (moduleId: number, cohortId: number): Promise<Response> =>
```
tommylau523, line 10: nit: can just be delete

src/scenes/AcceleratorRouter/Modules/ModulesTable.tsx
```diff
@@ -0,0 +1,260 @@
...
+                label={strings.ADD_MODULE}
+                onClick={() => setAddModuleModalOpened(true)}
+                priority='secondary'
+                size='medium'
+              />
+            )}
+          </Box>
+        </Grid>
+        <Grid item xs={12}>
+          {!modules && !editing ? (
```
tommylau523, line 228: this is not showing up for me. I think we need to check `modules.length === 0`

src/scenes/AcceleratorRouter/Modules/ModulesTable.tsx
```diff
@@ -0,0 +1,260 @@
...
+            <Typography>{strings.EDIT_COHORT_TO_ADD_MODULES}</Typography>
+          ) : (
+            <Table
+              id='modules-table'
+              columns={editing ? columns : viewColumns}
+              rows={allModules}
+              orderBy='name'
```
tommylau523, line 235: We probably want `startDate`

src/hooks/useListModules.ts
```diff
@@ -21,6 +30,32 @@ const useListModules = () => {
     [dispatch, setRequestId]
   );
 
+  const listDeliverables = useCallback(() => {
```
tommylau523, line 33: I think we can make this into a `useEffect` and not export it, or at least add a `useEffect` block to call this so that we don't have to call the `useEffect` to update the list of modules in the component

src/scenes/AcceleratorRouter/Modules/AddModuleModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const { onClose, onSave, moduleToEdit, existingModules } = props;
+
+  const dispatch = useAppDispatch();
+  const theme = useTheme();
+
+  const { modules, listModules } = useListModules();
+  const [record, setRecord, onChange] = useForm<CohortModule>(moduleToEdit || {});
+
+  useEffect(() => {
+    void listModules({});
```
tommylau523, line 33: can be simplified according to the comment in `useListModules`

src/hooks/useListModules.ts
```diff
@@ -21,6 +30,32 @@ const useListModules = () => {
...
+            _deliverablesByModuleId[moduleId].every((existingDeliverable) => existingDeliverable.id !== _deliverable.id)
+          ) {
+            [..._deliverablesByModuleId[moduleId], _deliverable];
+          }
+        } else {
+          _deliverablesByModuleId[moduleId] = [_deliverable];
+        }
+      });
+
+      setDeliverablesByModuleId(_deliverablesByModuleId);
```
tommylau523, line 55: This only shows one deliverable which I think is due to some overwriting.

But also, we have a `fetchDeliverables` (https://github.com/terraware/terraware-server/blob/e4d2afd302085be741681a801c6c9804c9210c91/src/main/kotlin/com/terraformation/backend/accelerator/db/DeliverableStore.kt#L59), it is probably worth exposing this via an API so you have a list of deliverables by modules! I can get started on this API today :)

tommylau523 (changes requested):
Sorry I left a decent number of comments! I did find an API feature that would make the deliverables stuff easier so I will get started on that now!

src/hooks/useListModules.ts
constanzauanini: Yes, I think having it in the BE will be useful. I added in the FE because I need it! Let me know if you are adding it in the BE, or if not I will review why is this returning one deliverable, I think it worked, but I may have changed something and broke it

src/scenes/AcceleratorRouter/Cohorts/CohortEditView.tsx
constanzauanini: @chudo76 do we want to keep module stepper on Cohort View and Edit View?

src/services/CohortModuleService.ts
constanzauanini: I get an error: 'delete' is not allowed as a variable declaration name.

src/hooks/useListModules.ts
constanzauanini: Also in the modules section, there is another modules table where we show the deliverables quantity and the cohort quantity. Maybe we can also add that in the BE?

tommylau523: I think we only have search for project modules and deliverables. Probably easiest to expose an API since we already have to store method. I will submit a PR soon!

src/services/CohortModuleService.ts
tommylau523: Ahh that makes sense

src/hooks/useListModules.ts
constanzauanini: Thanks!

tommylau523: https://github.com/terraware/terraware-server/pull/2470

tommylau523: I couldn't find an easy way to fit it into our existing APIs. you may need to write a cached selector around it and dispatch per module. Sort of a pain :(

src/scenes/AcceleratorRouter/Modules/ModuleDeliverablesModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+          key='button-1'
+        />,
+      ]}
+      scrolled={true}
+    >
+      <Table
+        id='module-deliverables-table'
+        columns={columns}
+        rows={deliverables?.map((del) => ({ ...del, cohortId: cohortId })) || []}
+        orderBy='name'
```
tommylau523, line 59: I think position should be the ordering

src/hooks/useListModules.ts
```diff
@@ -21,6 +30,36 @@ const useListModules = () => {
     [dispatch, setRequestId]
   );
 
+  const listDeliverables = useCallback(() => {
+    const dispatched = dispatch(requestListDeliverables({ locale: activeLocale }));
+    setDeliverablesRequestId(dispatched.requestId);
+  }, [dispatch, setDeliverablesRequestId]);
+
+  useEffect(() => {
+    listModules({});
```
tommylau523, line 39: nit: don't need the empty object in there!

src/scenes/AcceleratorRouter/Cohorts/CohortEditView.tsx
chudo76: Michael probably designed a lot of these screens. I see that we have the stepper in Edit for others (Participant, Project) so we should have it for Cohort, although it feels a bit weird for all of them, and I don't exactly know how useful it is.

tommylau523: on my second review, this seems redundant given the module table on the page. happy to defer to Michael to make the final call!!

chudo76: I also just went back to look at the designs ([Cohort Edit](https://www.figma.com/design/Ec5QnRsnzqEKeifzwPKR3k/Accelerator-Console?node-id=1654-24923&t=oxxdxOmy1AZzHSAN-4), [Project Edit](https://www.figma.com/design/Ec5QnRsnzqEKeifzwPKR3k/Accelerator-Console?node-id=1725-46570&t=oxxdxOmy1AZzHSAN-4)) and it looks like Michael didn't have the stepper for Edit mode. If it's not too much trouble, can we remove the stepper for the Edit modes for all of them?

chudo76: Oh I think I focused on the question about the _Edit_ page and missed the question about the _View_ page for Cohort Details.

I tend to agree and with your second review @tommylau523 , and for Cohort Details view, we don't need the Stepper since we will have the Modules table now. (cc: @constanzauanini)

constanzauanini: Thank you both!

src/hooks/useListModules.ts
constanzauanini: I think it is more complex to do a request for each module, than just getting all deliverables and group them by module. So I fixed what I previously did (it was a small change). Let me know what you think

src/hooks/useListModules.ts
```diff
@@ -1,26 +1,65 @@
...
+    const dispatched = dispatch(requestListDeliverables({ locale: activeLocale }));
+    setDeliverablesRequestId(dispatched.requestId);
+  }, [dispatch, setDeliverablesRequestId]);
+
+  useEffect(() => {
+    listModules();
+  }, [dispatch]);
+
+  useEffect(() => {
+    if (listModulesResponse && allDeliverablesResponse.status === 'success') {
```
tommylau523, line 43: `listModulesRepsponse` should be `allDeliverablesResponse`

src/redux/features/modules/modulesAsyncThunks.ts
```diff
@@ -86,3 +88,97 @@ export const requestListModuleDeliverables = createAsyncThunk(
...
+    if (response && response.requestSucceeded) {
+      return true;
+    }
+
+    return rejectWithValue(strings.GENERIC_ERROR);
+  }
+);
+
+export const requestUpdateManyCohortModule = createAsyncThunk(
+  'cohort-modules/update',
```
tommylau523, line 159: I think we usually use camel case for the names and this should be update many

src/hooks/useListModules.ts
tommylau523: Sounds good! I think this is fine as it is!

If you do want to use the new endpoint, you can do something similar to updateMany at the redux level with a Promise.all. But this looks like it is working well already!

tommylau523 (approved):
Looks good! I found a few more little things.

constanzauanini:
> Looks good! I found a few more little things.

Thanks for all the feedback Tommy! Merging it now

PR #3267, opened 2024-10-08
SW-5791 Added Events tab

constanzauanini:
Depends on https://github.com/terraware/web-components/pull/640


src/scenes/AcceleratorRouter/Modules/ModuleView.tsx
```diff
@@ -44,7 +45,7 @@ export default function ModuleView(): JSX.Element {
   const { activeLocale } = useLocalization();
   const contentRef = useRef(null);
   const { moduleId } = useParams<{ moduleId: string }>();
-  const { getModule, module, deliverables } = useGetModule();
```
tommylau523, line 47: Not sure if it is related but I am seeing duplicated modules in the list view

tommylau523: same for deliverables tab

src/scenes/AcceleratorRouter/Modules/Events.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const navigate = useNavigate();
+
+  const liveSessions = events?.filter((ev) => ev.type === 'Live Session');
+  const oneOnOneSessions = events?.filter((ev) => ev.type === 'One-on-One Session');
+  const recordedSessions = events?.filter((ev) => ev.type === 'Recorded Session');
+  const workshops = events?.filter((ev) => ev.type === 'Workshop');
+
+  return (
+    <Card flushMobile style={{ display: 'flex', flexDirection: 'column', flexGrow: 1 }}>
+      <Box
```
tommylau523, line 73: I forgot how to make this happen but we likely want the table to be scrollable instead of the whole card.

src/scenes/AcceleratorRouter/Modules/AddEventModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                  borderBottom: `1px solid ${theme.palette.TwClrBrdrTertiary}`,
+                  paddingBottom: 2,
+                }}
+              >
+                <Grid item xs={6} sx={{ marginTop: theme.spacing(2), paddingRight: 1 }}>
+                  <SelectT<CohortModuleWithProject>
+                    id='cohort'
+                    label={strings.COHORT}
+                    placeholder={strings.SELECT}
+                    options={availableCohorts}
```
tommylau523, line 260: availableCohorts here allow for duplicated selections

src/scenes/AcceleratorRouter/Modules/AddEventModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+          onClick={onClose}
+          priority='secondary'
+          key='button-1'
+        />,
+        <Button id='save' onClick={save} label={eventToEdit ? strings.UPDATE : strings.ADD} key='button-2' />,
+      ]}
+      scrolled
+    >
+      <Grid container textAlign={'left'} spacing={2}>
+        <Grid item xs={6} sx={{ marginTop: theme.spacing(2), paddingRight: 1 }}>
```
tommylau523, line 188: We should check with designs but I think we will need a way to pick timezones. I think the FE may default to browser timezone when selecting/picking a time, which will lead to some inconsistencies.

tommylau523: And similarly we probably want a way to show events time in different timezone. Maybe the events tab can have a dropdown for timezone for viewing too but likely a design question to answer

tommylau523 (commented):
Redux stuff looks good! I think I just have functional questions to raise.

src/scenes/AcceleratorRouter/Modules/AddEventModal.tsx
constanzauanini: Yes, I can add timezone in a follow up PR

src/scenes/AcceleratorRouter/Modules/ModuleView.tsx
constanzauanini: Not sure why. Will review

constanzauanini: I will fix that separately. It is not related to events.

src/scenes/AcceleratorRouter/Modules/AddEventModal.tsx
constanzauanini: Fixed

tommylau523 (approved):

PR #3278, opened 2024-10-17
SW-5940 follow up - Improve errors when uploading modules/deliverables and fix search/sort on modules

constanzauanini:


src/services/ModuleService.ts
```diff
@@ -22,6 +24,15 @@ export type GetModuleResponsePayload =
 export type ImportModuleResponsePayload =
   paths[typeof MODULES_IMPORT_ENDPOINT]['post']['responses'][200]['content']['application/json'];
 
+export type ImportProblemElement = {
+  problem: string;
+  row: number;
+};
+
+export type ImportResponsePayload = Omit<ImportModuleResponsePayload, 'problems'> & {
```
tommylau523, line 32: It reads a bit incorrect to have both modules and deliverables import depend and use this type. Perhaps we can move this type and ImportProblemElement  to ImportModal and explicitly write out 

```
message?: string,
problems: ImportProblemElement[]
```

Then the deliverable and module service can return the `ImportModulesResponsePayload` and `ImportDeliverablesResponsePayload`

tommylau523 (approved):

PR #3285, opened 2024-10-21
SW-6159 Sticky filters for project table

constanzauanini:
These changes allows using sticky filter in any TableWithSearchFilters setting the prop: stickyFilters = true

### DIFF
--- a/src/components/common/SearchFiltersWrapperV2/index.tsx
+++ b/src/components/common/SearchFiltersWrapperV2/index.tsx
@@ -6,6 +6,7 @@ import { PillList, PillListItem, Textfield } from '@terraware/web-components';
 import { FilterField } from 'src/components/common/FilterGroup';
 import strings from 'src/strings';
 import { SearchNodePayload } from 'src/types/Search';
+import { useSessionFilters } from 'src/utils/filterHooks/useSessionFilters';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
 
 import FeaturedFilters from './FeaturedFilters';
@@ -41,6 +42,8 @@ export type SearchProps = SearchInputProps & {
   setCurrentFilters: (filters: Record<string, SearchNodePayload>) => void;
   rightComponent?: ReactNode;
   onFilterApplied?: (filter: string, values: (string | number | null)[]) => void;
+  tableId?: string;
+  stickyFilters?: boolean;
 };
 
 const defaultPillValueRenderer = (values: (string | number | null)[]): string | undefined => values.join(', ');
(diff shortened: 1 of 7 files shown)


src/components/common/SearchFiltersWrapperV2/index.tsx
```diff
@@ -98,15 +111,57 @@ export default function SearchFiltersWrapperV2({
...
+        };
+      });
+
+      if (Object.keys(sessionFiltersToApply).length > 0) {
+        setCurrentFilters({
+          ...sessionFiltersToApply,
+        });
+      }
+    }
+  }, [sessionFilters, currentFilters, stickyFilters]);
```
joshbuchea, line 139: This should be a `useEffect` since no value is being returned from the `useMemo`:

```suggestion
  useEffect(() => {
    if (stickyFilters && Object.keys(sessionFilters).length > 0 && Object.keys(currentFilters).length === 0) {
      const sessionFiltersToApply: Record<string, SearchNodePayload> = {};
      const existingKeys = Object.keys(sessionFilters);
      existingKeys?.forEach((key) => {
        const valuesOfTheExistingKey = Array.isArray(sessionFilters[key])
          ? (sessionFilters[key] as any)
              .map((val: string) => val?.toString() || null)
              .filter((val: string | null) => val !== null)
          : [];

        sessionFiltersToApply[key] = {
          field: key,
          operation: 'field',
          type: 'Exact',
          values: valuesOfTheExistingKey,
        };
      });

      if (Object.keys(sessionFiltersToApply).length > 0) {
        setCurrentFilters({
          ...sessionFiltersToApply,
        });
      }
    }
  }, [sessionFilters, currentFilters, stickyFilters]);
```

src/scenes/AcceleratorRouter/ParticipantProjects/ListView.tsx
```diff
@@ -192,6 +192,7 @@ export default function ListView(): JSX.Element {
         rightComponent={actionMenus}
         rows={projects}
         title={strings.PROJECTS}
+        stickyFilters={true}
```
joshbuchea, line 195: Just a nit, but we can omit the `={true}`:

```suggestion
        stickyFilters
```

joshbuchea (approved):

src/components/common/SearchFiltersWrapperV2/index.tsx
constanzauanini: Yes, I think I was getting an infinite re-render with useEffect and that's why I changed it, but I just tested it again with useEffect and it is woking correctly now

PR #3307, opened 2024-10-30
Observations: Add indicators in Zones and Plots containing 25m sq plots

constanzauanini:
Also added "Last observation: ... " in schedule observation modal


src/scenes/ObservationsRouter/details/ObservationDetailsRenderer.tsx
```diff
@@ -14,13 +14,18 @@ const ObservationDetailsRenderer =
   (props: RendererProps<TableRowType>): JSX.Element => {
     const { column, row, value } = props;
 
+    const has25mPlots = row.plantingSubzones
+      ?.flatMap((subzone: { monitoringPlots: any[] }) => subzone.monitoringPlots.flatMap((plot) => plot.sizeMeters))
+      .some((size: number) => size.toString() === '25');
+
```
nickgraz, line 20: Perhaps this should be a static util function that receives subzones? It seems to be duplicated 3 times.

constanzauanini: It is used twice for subzones, so I added a util for that. The other checks the zone.

src/scenes/ObservationsRouter/details/index.tsx
```diff
@@ -134,6 +134,14 @@ export default function ObservationDetails(props: ObservationDetailsProps): JSX.
     }
   }, [zoneNames, searchProps.filtersProps]);
 
+  const has25mPlots = details?.plantingZones
+    .flatMap((zone) =>
```
nickgraz, line 138: Can you use the `has25mPlots` function here?

nickgraz (approved):
Looks good, one more suggestion for improvement but not absolutely necessary

PR #3308, opened 2024-10-31
SW-5791 Event Management - Follow up

constanzauanini:
- Use new datetime-picker component (mui v7)
- Add validation to end date
- Allow editing new added events, showing this string: **< ID >** on events table


src/scenes/AcceleratorRouter/Modules/EventsTable.tsx
```diff
@@ -71,48 +71,77 @@ export default function EventsTable(props: EventsTableProps): JSX.Element {
...
+    } while (generatedIds.has(newId)); // Check for uniqueness
+
+    generatedIds.add(newId);
+    return newId;
+  }
+
   const onAddEvent = (eventToAdd: ModuleEventPartial) => {
+    let newEventToAdd = { ...eventToAdd };
+    if (eventToAdd.id === -1 && !eventToAdd.feId) {
+      newEventToAdd = { ...eventToAdd, feId: generateUniqueId() };
```
nickgraz, line 89: You can you also use `Symbol()` here. [Docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol). Basically, it is always guaranteed unique

src/scenes/AcceleratorRouter/Modules/EventsTable.tsx
```diff
@@ -71,48 +71,77 @@ export default function EventsTable(props: EventsTableProps): JSX.Element {
...
+          return [...prev, newEventToAdd];
         } else {
-          return [eventToAdd];
+          return [newEventToAdd];
         }
       });
     }
   };
 
   const onEditedEvent = (editedEvent: ModuleEventPartial) => {
```
nickgraz, line 102: This function is pretty large. A couple of suggestions:
1. Try to remove unnecessary nesting. For example, this line `if (setEventsToDelete && setEventsToAdd) {` wraps the entire function contents. Instead, you can do something like
```
if (!(setEventsToDelete && setEventsToAdd)) {
  return;
}
```

2. Break out logical branches into their own functions:
For example: 
```
if (editedEvent.id === -1) { 
  addEvent(editedEvent) 
} else {
  editEvent(editedEvent)
}
```

nickgraz (approved):
This looks fine, can merge as-is but there are a couple suggestions you can implement if you want

src/scenes/AcceleratorRouter/Modules/EventsTable.tsx
constanzauanini: Thank for the suggestions!

PR #3315, opened 2024-11-05
SW-6168 Fix up types for Slate editor, attempt to fix issue with first line break from hitting enter key not working

nickgraz:
This did not end up fixing the issue, but might be an improvement we want. 

Cleans up the types, gets rid of casting. Implements `SectionElement` to contain text and variable elements.

I spent a while on this, the main area where the fix is attempted is [here](https://github.com/terraware/terraware-web/pull/3315/files#diff-f66b3b5e9a35a6d4b929206aadf4d2d5a406efd85b7f4a057f54ea5413cdc786R102), but it still does not seem to work. Even though I can see the values are being fixed (through the console), the Slate editor still does not show the line break.

### DIFF
--- a/src/components/DocumentProducer/EditableSection/Edit.tsx
+++ b/src/components/DocumentProducer/EditableSection/Edit.tsx
@@ -4,8 +4,8 @@ import { Box, Typography, useTheme } from '@mui/material';
 import { Autocomplete, Button, DropdownItem } from '@terraware/web-components';
 import { useDeviceInfo } from '@terraware/web-components/utils';
 import { isKeyHotkey } from 'is-hotkey';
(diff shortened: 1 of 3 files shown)


nickgraz:
* **#3315** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3315?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz:
We may or may not actually want this, I was able to figure out how to fix this without fixing all the types. This is nice, but the surface area is much larger.

constanzauanini (approved):
Let's merge it, it is an improvement!

PR #3328, opened 2024-11-11
SW-6233 Plants Dashboard: Add new planting site trends card

constanzauanini:
Screenshot 2024-11-11 at 6 12 08 PM


src/scenes/PlantsDashboardRouter/components/PlantingSiteTrendsCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+    width: '1px',
+    height: 'auto',
+    backgroundColor: theme.palette.TwClrBrdrTertiary,
+    marginRight: '24px',
+    marginLeft: '24px',
+  };
+
+  const plantsChartData = useMemo(() => {
+    const labels = sortedObservations.map((ob) => ob.startDate);
+    const values = sortedObservations.map(
```
constanzauanini, line 49: @sgrimm do you know if this is the correct way of calculating plants per ha? Or should I be using other value?
I'm summing up all total plants from all the zones and diving it by the totalArea of the planting site

sgrimm: The server includes `estimatedPlants` fields in the site, zone, and subzone observation results payloads, which takes the areas into account already. But that only counts subzones that have been marked "planting completed" which maybe isn't what we want here.

I'm having a hard time finding any mention of a plants-per-hectare graph in the PRD; can you point me to it? I can't give you a solid answer because I'm not actually sure what this value is supposed to mean, but hopefully the PRD has more detail.

sgrimm: If I had to guess at what the intent of that graph was, I'd say the right calculation here is maybe to total up all the live (and existing, maybe? not sure) plants across all species at the _plot_ level for the observation, then divide that by the total hectares of all the plots.

I think you don't want to use the total hectares of the site or even the zone because the observation might only be covering a small user-selected part of the site.

Calculating the value the way I just described will probably result in a graph that fluctuates wildly from observation to observation if people are observing small areas with different mixes of species but in the absence of any definition to the contrary, that's what I'd go with.

sgrimm: @alexTerraformation, does that seem right to you?

constanzauanini: I didn't find it in the PRD, but this is the design: https://www.figma.com/design/MoIzWBItJSC4t8UDSCkG95/Changeable-Plant-Monitoring?node-id=1665-32320&node-type=frame&t=JOfdniO8FDwG5Q1g-0

sgrimm: Also commented on the design, but I guess I'll take my questions to the Jira task for further discussion.

src/scenes/PlantsDashboardRouter/components/PlantingSiteTrendsCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }, [plantingSite]);
+
+  useEffect(() => {
+    if (plantingObservationsSummaryResponse?.status === 'success') {
+      setSummaries(plantingObservationsSummaryResponse.data);
+    }
+  }, [plantingObservationsSummaryResponse]);
+
+  const separatorStyles = {
```
tommylau523, line 55: Nit: we can move this into the <div> components directly

src/scenes/PlantsDashboardRouter/components/PlantingSiteTrendsCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+    backgroundColor: theme.palette.TwClrBrdrTertiary,
+    marginRight: '24px',
+    marginLeft: '24px',
+  };
+
+  const plantsChartData = useMemo(() => {
+    const labels = summaries?.map((sm) => sm.latestObservationTime);
+    const values = summaries?.map((sm) => {
+      const zone = sm.plantingZones.find((pz) => pz.plantingZoneId === selectedPlantsPerHaZone);
+      return zone?.estimatedPlants || 0;
```
tommylau523, line 67: Nit: Is it possible to drop a data point instead of returning 0? So the line chart can end instead of going to 0.

src/scenes/PlantsDashboardRouter/components/PlantingSiteTrendsCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+      labels: labels ?? [],
+      datasets: [
+        {
+          values: values ?? [],
+        },
+      ],
+    };
+  }, [summaries, selectedPlantsPerHaZone]);
+
+  const mortalityChartData = useMemo(() => {
```
tommylau523, line 80: can we add a `ChartData[]` type?

src/scenes/PlantsDashboardRouter/components/PlantingSiteTrendsCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  const separatorStyles = {
+    width: '1px',
+    height: 'auto',
+    backgroundColor: theme.palette.TwClrBrdrTertiary,
+    marginRight: '24px',
+    marginLeft: '24px',
+  };
+
+  const plantsChartData = useMemo(() => {
```
tommylau523, line 63: can we add a `ChartData[]` type?

src/scenes/PlantsDashboardRouter/components/PlantingSiteTrendsCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+        },
+      ],
+    };
+  }, [summaries, selectedPlantsPerHaZone]);
+
+  const mortalityChartData = useMemo(() => {
+    const labels = summaries?.map((sm) => sm.latestObservationTime);
+    const values = summaries?.map((sm) => {
+      const zone = sm.plantingZones.find((pz) => pz.plantingZoneId === selectedMortalityZone);
+      return zone?.mortalityRate || 0;
```
tommylau523, line 84: Same here. Drop data instead of default to 0.

tommylau523 (approved):

PR #3331, opened 2024-11-12
SW-6165 Populate accelerator orgs with ones that contain a project with an application

nickgraz:
Within the Single and Edit views for a project within the accelerator console, we need to populate organizations that contain at least one project with an application. This is because a project may have filed their application, but not yet accepted as a participant. In this state, we still need to be able to modify the project lead, which is technically the "organization TF contact", which requires us to have access to the organization even though it is not directly related to the accelerator yet.

This PR modifies the `ParticipantProjectProvider` to achieve this desired result, and fixes the issue of an Accelerator Admin (with or without any relation to any particular org) not being able to modify the project lead for a project that is not directly attached to a participant or an org with an "accelerator" internal tag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,8 @@
     "format": "prettier --write .",
     "docker:start": "./scripts/launch_server.sh",
     "docker:stop": "docker compose down",
+    "docker:remote-be:start": "docker compose -f remote-backend/docker-compose.yml up --detach",
+    "docker:remote-be:stop": "docker compose -f remote-backend/docker-compose.yml down",
     "wait-be": "wait-on -t 120000 http://localhost:8080/swagger-ui/index.html",
     "server:prepare": "./scripts/preparedb.sh",
     "server:reset": "./scripts/resetdb.sh",
--- a/src/hooks/useAcceleratorOrgs.ts
+++ b/src/hooks/useAcceleratorOrgs.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { useLocalization } from 'src/providers';
 import { requestAcceleratorOrgs } from 'src/redux/features/accelerator/acceleratorAsyncThunks';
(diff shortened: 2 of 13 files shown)


nickgraz:
* **#3331** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3331?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

package.json
```diff
@@ -21,6 +21,8 @@
     "format": "prettier --write .",
     "docker:start": "./scripts/launch_server.sh",
     "docker:stop": "docker compose down",
+    "docker:remote-be:start": "docker compose -f remote-backend/docker-compose.yml up --detach",
```
nickgraz, line 24: These are used if you are pointing your local to TW Staging or Prod.

src/hooks/useAcceleratorOrgs.ts
```diff
@@ -13,29 +13,43 @@ export type Response = {
...
+  const { includeParticipants, hasProjectApplication } = props || {};
+
   const { activeLocale } = useLocalization();
   const dispatch = useAppDispatch();
   const snackbar = useSnackbar();
 
   const [requestId, setRequestId] = useState<string>('');
   const result = useAppSelector(selectAcceleratorOrgsRequest(requestId));
 
-  useEffect(() => {
```
nickgraz, line 24: This is identical to `reload`, put that in a callback and fire in an effect

src/scenes/AcceleratorRouter/ParticipantProjects/ParticipantProjectProvider.tsx
```diff
@@ -30,7 +30,7 @@ const ParticipantProjectProvider = ({ children }: Props) => {
   const { activeLocale } = useLocalization();
   const { project, projectId } = useProjectData();
   const { participant } = useParticipant(project?.participantId ?? -1);
-  const { acceleratorOrgs, reload: reloadAll } = useAcceleratorOrgs(true);
+  const { acceleratorOrgs, reload: reloadAll } = useAcceleratorOrgs({ hasProjectApplication: true });
```
nickgraz, line 33: This provider is only used in the single and edit project view, in the admin console. In this location, we need to find all organizations that contain a project with an application, because they are not guaranteed to have an "accelerator" internal tag nor a project that is associated to a participant.

package.json
constanzauanini: Nice, I always use it but changing to remote-backend folder!

constanzauanini (approved):
Looks good! Thanks for the clarifications!

PR #3332, opened 2024-11-13
SW-6234 Plants Dashboard: Update map mortality rate overlay to us line thickness

constanzauanini:
Also updated how legends display, and added a switch to turn mortality rate on/off

Screenshot 2024-11-13 at 11 30 55 AM


joshbuchea (approved):
Looks good, nice work!

PR #3337, opened 2024-11-18
SW-6225 Use newly populated country code for planting sites location on dashboard card

constanzauanini:
Screenshot 2024-11-18 at 11 33 54 AM

### DIFF
--- a/src/api/types/generated-schema.ts
+++ b/src/api/types/generated-schema.ts
@@ -930,11 +930,23 @@ export interface paths {
     /** Gets a list of the results of observations. */
     get: operations["listObservationResults"];
   };
+  "/api/v1/tracking/observations/results/summary": {
+    /** Gets the rollup observation summary of a planting site */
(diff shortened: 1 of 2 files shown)


src/redux/features/observations/utils.ts
```diff
@@ -166,8 +166,9 @@ export const mergeObservations = (
 
 const StatusWeights: Record<MonitoringPlotStatus, number> = {
   Completed: 1,
-  InProgress: 2,
-  Outstanding: 3,
+  Claimed: 2,
+  Unclaimed: 3,
+  'Not Observed': 4,
```
nickgraz, line 171: "Claimed" and "unclaimed" make sense to me, I am not sure about "not observed" though. Maybe that is in the PRD?

constanzauanini: Couldn't find it. Maybe we should merge it like this and maybe change later if needed?

nickgraz (approved):

PR #3338, opened 2024-11-19
SW-6261 Show page message with unrecognized species

constanzauanini:
Screenshot 2024-11-19 at 11 58 23 AM


src/scenes/ObservationsRouter/details/index.tsx
```diff
@@ -108,6 +110,26 @@ export default function ObservationDetails(props: ObservationDetailsProps): JSX.
     return undefined;
   }, [activeLocale, details, observation]);
 
+  useEffect(() => {
+    const speciesWithNoIdNames = details?.species.filter((sp) => !sp.speciesId).map((sp) => sp.speciesName || '');
+    const speciesWithNoIdSet = new Set(speciesWithNoIdNames);
+    const speciesWithNoIdMap = Array.from(speciesWithNoIdSet);
```
nickgraz, line 116: If this is here to dedupe the list, you might consider using `lodash`.

Something like this might work:
```
let speciesWithNoId = _.uniqBy((details?.species || []).filter((sp) => !sp.speciesId), 'speciesName')
```

nickgraz (approved):
This looks good to me, once small suggestion, but it is not blocking.

src/scenes/ObservationsRouter/details/index.tsx
constanzauanini: Thanks!

PR #3340, opened 2024-11-19
SW-6162 Fix document upgrade modal visibility

nickgraz:
This PR fixes the component state and some checks to make sure that the upgrade modal is only visible when a yet-to-be-completed upgrade is available for the current document.

Although this can be deployed on its own, [this](https://github.com/terraware/terraware-server/pull/2612) PR needs to be merged to QA the ticket.

### DIFF
--- a/src/scenes/AcceleratorRouter/Documents/DocumentView/index.tsx
+++ b/src/scenes/AcceleratorRouter/Documents/DocumentView/index.tsx
@@ -44,7 +44,7 @@ export default function DocumentView(): JSX.Element {
     } else if (currentManifestId === latestManifestId) {
       setShowUpgradeModal(false);
     }
-  }, [document?.variableManifestId, latestManifestId]);
+  }, [currentManifestId, latestManifestId]);
 
   const onUpgradeManifest = useCallback(async () => {
     if (documentId && latestManifestId) {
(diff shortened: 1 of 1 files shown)


nickgraz:
* **#3340** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3340?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3340?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/scenes/AcceleratorRouter/Documents/DocumentView/index.tsx
```diff
@@ -31,13 +31,20 @@ export default function DocumentView(): JSX.Element {
   const [showUpgradeModal, setShowUpgradeModal] = useState(false);
   const [outlinePanelOpen, setOutlinePanelOpen] = useState(true);
 
-  const latestManifestId = useMemo(() => documentTemplate?.variableManifestId ?? -1, [documentTemplate]);
```
nickgraz, line 34: These sorts of usages of `useMemo` are actually performance robbing. If we are just calculating a property of an object, we should just set it as a regular variable, and it will be recomputed on every component refresh. This is OK because it costs less than what the `useMemo` needs to determine if the value needs to be recomputed or not.

joshbuchea: Great point! 💯

joshbuchea (approved):

PR #3343, opened 2024-11-19
SW-6263 Match Unrecognized species modal

constanzauanini:
- Added redux for calling api
- Added modal for selecting species
- Used same species dropdown selector
- Snackbar message is not showing updated names yet (separated ticket) 
- Reload observation data after saving

Screenshot 2024-11-19 at 5 35 53 PM


src/services/SpeciesService.ts
```diff
@@ -404,6 +406,19 @@ const getSpeciesProjects = (
...
 };
 
+const mergeOtherSpecies = async (
+  mergeOtherSpeciesPayload: MergeOtherSpeciesPayload,
+  observationId: number
+): Promise<Response> => {
+  const urlReplacements = {
+    '{observationId}': `${observationId}`,
+  };
+  return await httpMergeOtherSpecies.post({
```
nickgraz, line 416: If you aren't modifying the result, or attempting to catch any errors that occur with the `post`, it is not necessary to mark this function as `async` and use `await` within it.

nickgraz (approved):
This looks good, one small change suggested

PR #3346, opened 2024-11-20
SW-6264 Toast message success after matching species

constanzauanini:
Screenshot 2024-11-20 at 10 35 15 AM


src/scenes/ObservationsRouter/details/index.tsx
```diff
@@ -196,8 +205,25 @@ export default function ObservationDetails(props: ObservationDetailsProps): JSX.
...
+            const newName = allSpecies?.find((existing) => existing.id === sp.speciesId)?.scientificName;
+            return (
+              <li key={index}>
+                {sp.otherSpeciesName} &#8594; {newName}
+              </li>
+            );
+          })}
+        </ul>
+      );
+      setMergedMessage(successMessageBody);
```
nickgraz, line 222: Storing components in state is generally not advised. I think you should create and load the component into the snackbar at the time of the toast.

src/scenes/ObservationsRouter/details/index.tsx
```diff
@@ -174,9 +183,9 @@ export default function ObservationDetails(props: ObservationDetailsProps): JSX.
   }, [zoneNames, searchProps.filtersProps]);
 
   useEffect(() => {
-    if (matchResponse?.status === 'success') {
+    if (mergedMessage && matchResponse?.status === 'success') {
```
nickgraz, line 186: If `mergedMessage` is only created when `onSaveMergedSpecies` runs and `merged.length > 0`, then maybe you can check that condition here? Also, are "merge" and "match" the same thing in this file?

constanzauanini: Merge and match are the same thing, yes. I'm checking `merged.length > 0` in `onSaveMergedSpecies` because if there are no merged species, there is no need to call the api.

src/scenes/ObservationsRouter/details/index.tsx
nickgraz: Something like this could solve it: https://github.com/terraware/terraware-web/pull/3350.

But I am unsure if that code actually works with what you're attempting. What you propose, saving the names into state, would probably also work just as well. 

I also think that since this is the only place `requestMergeOtherSpecies`, a cleaner approach would be to handle the "payload creation logic" in the thunk. This way the component really just needs to dispatch the request, and the result could contain the filtered payloads with the names. 

Ideally a component doesn't need to manage to pieces of separate state for a request if the request itself can contain all the data necessary to satisfy the UI.

nickgraz: Here's an example of that: https://github.com/terraware/terraware-web/pull/3351/files

Again, this is not tested, so it may not satisfy your needs. Anyways, these are just suggestions. Let me know what you think. I don't want to make you completely redo this effort, so if what I'm asking for is off base just let me know!

nickgraz (approved):

PR #3352, opened 2024-11-20
Updated BE types and fix bugs with not being able to assign modules

tommylau523:
Mostly changes to sync with BE: https://github.com/terraware/terraware-server/pull/2617

Also fixed an issue where module list is not showing up when assigning new modules, by using new module endpoints.


src/hooks/useListModules.ts
```diff
@@ -1,60 +1,21 @@
...
-    [dispatch, setRequestId]
-  );
-
-  const listDeliverables = useCallback(() => {
-    const dispatched = dispatch(requestListDeliverables({ locale: activeLocale }));
-    setDeliverablesRequestId(dispatched.requestId);
-  }, [dispatch, setDeliverablesRequestId]);
-
-  useEffect(() => {
-    if (allDeliverablesResponse && allDeliverablesResponse.status === 'success') {
```
constanzauanini, line 39: Nice!

src/redux/features/modules/modulesAsyncThunks.ts
```diff
@@ -28,18 +19,17 @@ export const requestGetModule = createAsyncThunk(
...
-  'modules/list',
-  async (request: ListModulesRequestParam, { rejectWithValue }) => {
-    const response = await ModuleService.list(request);
+export const requestListModules = createAsyncThunk('modules/list', async (_, { rejectWithValue }) => {
+  const response = await ModuleService.list();
 
-    if (response !== null && response.requestSucceeded && response?.data?.modules !== undefined) {
-      return response.data.modules;
-    }
+  console.log(response);
```
constanzauanini, line 25: remove this log?

constanzauanini:
Thanks for all these changes Tommy! I was testing this and in the cohort edit view, I add a module and then delete it, and it doesn't get delete (I think it is just UI, because it happens after saving)

tommylau523:
addressed the module changes not getting applied issue! In general there was a bit of weirdness around having no modules. (for example from deleting all). This should have fixed it now!

constanzauanini (approved):

PR #3365, opened 2024-11-27
SW-6300 SW-6301 Select (dropdown) menu doesn't show any values for first match candidate and add clarifying text

constanzauanini:


nickgraz (approved):
Nice solution!

PR #3394, opened 2024-12-04
SW-6273 Planting site static image not loading

constanzauanini:
The issue is that if the geojson used in the static image URL is too long, the browser have a limit, and it can't load the image. The fix is to simplify the geojson, decreasing the precision.

NOTE: With this change the image of complicated planting sites, will not be exact and precise, but at least it will show.


constanzauanini:
I will close this since we are going to replace the image with a map

PR #3438, opened 2024-12-12
SW-6333 Ability to End Observation without all plots observed

constanzauanini:


nickgraz (approved):
Looks good to me!

joshbuchea (approved):
Looks great, nice work!

PR #3466, opened 2024-12-16
SW-6359 Clicking "Add an Accession" CTA while scrolled down lands user in middle of next screen

constanzauanini:


joshbuchea (approved):
This worked great for me!

PR #3470, opened 2024-12-16
SW-6360 "Please reach out to an administrator" pages for Accessions and Inventory look different

constanzauanini:


joshbuchea (approved):
Love updates that make messaging in Terraware more consistent 💯

PR #3471, opened 2024-12-16
SW-6361 SW-6364 Home Dashboard: Show Mortality Rate as percentage and save Planting Site selected

constanzauanini:


joshbuchea (approved):
Seems pretty straight forward

PR #3516, opened 2024-12-30
SW-6378 Reload organization after new application is created

tommylau523:


tommylau523:
* **#3516** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3516?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/components/emptyStatePages/NoOrgApplicationLandingPage.tsx
```diff
@@ -54,6 +57,7 @@ export default function NoOrgApplicationLandingPage(): JSX.Element {
         toastSuccess(strings.SUCCESS);
       }
       goToApplication(result.data);
+      reloadOrganizations();
```
sgrimm, line 60: Should this happen before `goToApplication`? I guess it's all async anyway so maybe it doesn't matter.

tommylau523: It is all async like you said. I think there was a problem with the OrgRouter refreshing before the redirect to applications, which was the cause of the issue before. This is now in the same effect so it should not happen. I would prefer to keep it this way just to show an order of operations logically speaking.

joshbuchea (approved):

PR #3520, opened 2025-01-07
SW-6418 Add missing dependency to home page rows

alexTerraformation:
Use memo was missing this dependency which causes some display issues in production that we didn't see in prod.

### DIFF
--- a/src/scenes/Home/TerrawareHomeView/index.tsx
+++ b/src/scenes/Home/TerrawareHomeView/index.tsx
@@ -229,6 +229,7 @@ const TerrawareHomeView = () => {
     activeLocale,
     availableSpecies,
     orgNurserySummary,
+    plantingSites,
     seedBankSummary,
     selectedOrganization,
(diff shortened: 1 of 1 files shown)


src/types/Module.ts
```diff
@@ -30,7 +30,7 @@ export const getEventType = (input: ModuleEventType): string => {
     case 'Workshop':
       return strings.WORKSHOP;
     default:
-      return `${input}`;
+      return `${input as string}`;
```
tommylau523, line 33: Is this necessary? The bracket already calls a toString. same in other places.

tommylau523 (approved):

src/types/Module.ts
alexTerraformation: Probably not necessary, but it gets rid of the lint error.  Alternatives are doing a single line disable of the linter rule, or turning off the rule altogether.  This seemed like the best option.

PR #3528, opened 2025-01-15
SW-6279 Add standard deviation to Observed Density trend chart

constanzauanini:
Screenshot 2025-01-15 at 12 22 46 PM


src/scenes/PlantsDashboardRouter/components/htmlLegendPlugin.ts
```diff
@@ -0,0 +1,89 @@
...
+      ul.firstChild.remove();
+    }
+
+    // Reuse the built-in legendItems generator
+    const generateLabelsFunction = chart.options?.plugins?.legend?.labels?.generateLabels;
+    if (generateLabelsFunction) {
+      const items = generateLabelsFunction(chart);
+      items.forEach((item) => {
+        const li = document.createElement('li');
+        li.style.alignItems = 'center';
```
tommylau523, line 41: can we use

li.style = {
    …li.style, 
    alignItems: ‘center’,
    cursor: ‘pointer,
    …
}

src/scenes/PlantsDashboardRouter/components/htmlLegendPlugin.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import strings from 'src/strings';
+
+export const getOrCreateLegendList = (chart: any, id: string) => {
+  const legendContainer = document.getElementById(id);
+  let listContainer = legendContainer?.querySelector('ul');
+
+  if (!listContainer) {
+    listContainer = document.createElement('ul');
+    listContainer.style.display = 'flex';
```
tommylau523, line 11: Same here as the li style. Can we use an object for the attributes?

src/scenes/PlantsDashboardRouter/components/htmlLegendPlugin.ts
```diff
@@ -0,0 +1,89 @@
...
+          if (item.datasetIndex !== undefined) {
+            chart.setDatasetVisibility(item.datasetIndex, !chart.isDatasetVisible(item.datasetIndex));
+            chart.update();
+          }
+        };
+
+        // Color box
+        const boxSpan = document.createElement('span');
+        if (item.text === strings.ACTUAL) {
+          boxSpan.style.background = item.strokeStyle?.toString() || '';
```
tommylau523, line 57: same here

src/scenes/PlantsDashboardRouter/components/htmlLegendPlugin.ts
```diff
@@ -0,0 +1,89 @@
...
+          boxSpan.style.background = '#B8A0D64D';
+          boxSpan.style.display = 'inline-block';
+          boxSpan.style.height = '12px';
+          boxSpan.style.marginRight = '10px';
+          boxSpan.style.width = '40px';
+        }
+
+        // Text
+        const textContainer = document.createElement('p');
+        textContainer.style.color = item.fontColor?.toString() || '';
```
tommylau523, line 72: same here

tommylau523 (approved):

src/scenes/PlantsDashboardRouter/components/PlantingSiteTrendsCard.tsx
```diff
@@ -62,21 +55,47 @@ export default function PlantingSiteTrendsCard({ plantingSiteId }: PlantingSiteT
...
     const labels = filteredSummaries?.map((sm) => sm.latestObservationTime);
     const values = filteredSummaries?.map((sm) => {
       const zone = sm.plantingZones.find((pz) => pz.plantingZoneId === selectedPlantsPerHaZone);
-      return zone?.estimatedPlants || 0;
+      return zone?.plantingDensity || 0;
+    });
+
+    const minValues = filteredSummaries?.map((sm) => {
+      const zone = sm.plantingZones.find((pz) => pz.plantingZoneId === selectedPlantsPerHaZone);
+      return (zone?.plantingDensity || 0) - (zone?.plantingDensityStdDev || 0);
```
tommylau523, line 70: if plantingDensity is null, that means that one of the zones haven’t completed planting. It could go from having values to null if a user marks one of the subzones planting incomplete. The graph shows 0 which is a bit jarring/incorrect at the moment. We should ask product whether we can do either
1) Calculate planting density for incomplete planting zones as well
2) Update the charts design to confirm this (maybe have broken up lines?).

1 is probably better

constanzauanini: I'm filtering out all the summaries that their zone doesn't have plantingDensity, so if all zones have plantingDensity null it will show an empty chart, not a 0. The ' || 0 ' value is for typescript check, but it should not happen, since summaries are filtered before

src/scenes/PlantsDashboardRouter/components/htmlLegendPlugin.ts
constanzauanini: If I do that I get this error: 

_Cannot assign to 'style' because it is a read-only property_

PR #3535, opened 2025-01-17
SW-6396 Disable clickable rows in people table

alexTerraformation:
A previous PR removed the checkboxes for non-admins, but the rows were still clickable.  Remove that functionality as well.  Checkboxes still work for admins.

### DIFF
--- a/src/scenes/PeopleRouter/PeopleListView.tsx
+++ b/src/scenes/PeopleRouter/PeopleListView.tsx
@@ -399,6 +399,7 @@ export default function PeopleListView(): JSX.Element {
                       rows={results}
                       orderBy='name'
                       Renderer={TableCellRenderer}
+                      isClickable={() => false}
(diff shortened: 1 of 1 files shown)


src/scenes/PeopleRouter/PeopleListView.tsx
```diff
@@ -399,6 +399,7 @@ export default function PeopleListView(): JSX.Element {
                       rows={results}
                       orderBy='name'
                       Renderer={TableCellRenderer}
+                      isClickable={() => false}
```
constanzauanini, line 402: Should it be clickable for admins?

alexTerraformation: Admins still have the checkbox to select it.  I don't think we need selectable rows if we have the checkboxes.

constanzauanini (approved):

PR #3544, opened 2025-01-20
SW-6445 Create biomass measurement component

constanzauanini:
- Also added empty state to the component
- Added tabs to observation view to render the new component

Screenshot 2025-01-20 at 11 59 36 AM


src/scenes/ObservationsRouter/ObservationsHome.tsx
```diff
@@ -107,23 +129,44 @@ export default function ObservationsHome(props: ObservationsHomeProps): JSX.Elem
...
           />
         ) : (
-          <Card style={{ margin: '56px auto 0', borderRadius: '24px', height: 'fit-content' }}>
-            <EmptyStateContent
-              title={strings.OBSERVATIONS_EMPTY_STATE_TITLE}
-              subtitle={[strings.OBSERVATIONS_EMPTY_STATE_MESSAGE_1, strings.OBSERVATIONS_EMPTY_STATE_MESSAGE_2]}
-            />
-          </Card>
+          <>
+            {observationsResults === undefined ? (
```
tommylau523, line 151: should we replace this whole part with `<PlantMonitoring />`?

tommylau523 (approved):
One minor thing to reduce some duplicated code, looks good otherwise!

src/scenes/ObservationsRouter/ObservationsHome.tsx
constanzauanini: This whole part is the old Observations list (when adHocObservationSupport is not Enabled). If I use PlantMonitoring it will include the selector for ad hoc observations in the table

PR #3546, opened 2025-01-20
SW-6447 Add Ad Hoc standard observations list view

constanzauanini:


src/redux/features/observations/observationsSelectors.ts
```diff
@@ -55,6 +55,20 @@ export const selectPlantingSiteObservationsResults = createCachedSelector(
...
 
+export const selectPlantingSiteAdHocObservationsResults = createCachedSelector(
+  (state: RootState, plantingSiteId: number) => selectAdHocObservationsResults(state),
+  (state: RootState, plantingSiteId: number) => plantingSiteId,
+  (observationsResults, plantingSiteId) => {
+    if (plantingSiteId === -1) {
+      return observationsResults;
+    }
+    return observationsResults?.filter((observationResults) => {
+      const matchesSite = plantingSiteId === -1 || observationResults.plantingSiteId === plantingSiteId;
```
tommylau523, line 66: can return this directly or even get rid of the functional block {}

src/redux/features/observations/utils.ts
```diff
@@ -280,3 +281,25 @@ export const has25mPlots = (
     ?.flatMap((subzone: { monitoringPlots: any[] }) => subzone.monitoringPlots.flatMap((plot) => plot.sizeMeters))
     .some((size: number) => size.toString() === '25');
 };
+
+export const mergeAdHocObservations = (
```
tommylau523, line 285: I am wondering if we should move the AdHocObservation type usage to during render time. That way we can reuse ObservationResults for the redux stuff for the most part. And probably won't need this new merge method.

src/redux/features/observations/observationsSelectors.ts
```diff
@@ -55,6 +55,20 @@ export const selectPlantingSiteObservationsResults = createCachedSelector(
   }
 )((state, plantingSiteId: number, status?: ObservationState[]) => `${plantingSiteId}_${status?.join(',')}`);
 
+export const selectPlantingSiteAdHocObservationsResults = createCachedSelector(
```
tommylau523, line 58: We could add a isAdHoc flag to the other search observation selector and slice and thunks as well to handle this. Especially since the API returns the same ObservationResult type.

I think the observationsThunk will use the flag to make different service calls. But the result can be handled the same way by the slice and the selector.

tommylau523: I think this is probably the easier way so we don't have two separate code sections handling the same data type, returned by two API endpoints.

src/redux/features/observations/observationsThunks.ts
```diff
@@ -90,3 +91,25 @@ export const requestGetPlantingSiteObservationsSummaries = createAsyncThunk(
     return rejectWithValue(strings.GENERIC_ERROR);
```
tommylau523, line 91: <p class='graphite__hidden'><i>[Re: line +63]</i></p>
we can add the `isAdHoc=false` flag here

below we can set `const response = isAdHoc ? service.listAdHoc... : service.listObservations ...`
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3546?utm_source=unchanged-line-comment">Graphite</a>.</p>

src/redux/features/observations/observationsSelectors.ts
tommylau523: We may need to update our slice a bit to cache results better.  Maybe we should rework the redux for all observations stuff since it is all quite messy at the moment.

tommylau523 (approved):
FE code all looks good. I think the redux stuff is a bit messy but that requires more effort outside of the requirements for now.

PR #3557, opened 2025-01-24
SW-6464 Mortality Rate and Observation Recency legend boxes are not showing in zero state

constanzauanini:


src/components/common/MapLegend/index.tsx
```diff
@@ -12,13 +12,17 @@ type MapLegendItem = {
...
   height?: string;
+  isDisabled?: boolean;
 };
 
 export type MapLegendGroup = {
   title: string;
   items: MapLegendItem[];
   switch?: boolean;
+  // is the legend turned off
   disabled?: boolean;
```
tommylau523, line 23: can we rename this to `checked` or `visible` or maybe `hidden`? then the other one can be `disabled`. 

to me "disable" means if it is turned off by the system or if it is unavailable. 
"hidden" or "visible" implies a setting that the user can manage at least in this context.

The distinction makes sense! I think the naming can be clarified.

constanzauanini: Yes, I think checked would be the correct name here. Thanks for the suggestion!

src/components/common/MapLegend/index.tsx
```diff
@@ -12,13 +12,17 @@ type MapLegendItem = {
...
 
 export type MapLegendGroup = {
   title: string;
   items: MapLegendItem[];
   switch?: boolean;
-  disabled?: boolean;
+  // is the legend turned on
+  checked?: boolean;
+  // the legend is disabled (can't be turned on/off)
+  isDisabled?: boolean;
```
tommylau523, line 25: should we rename this to disabled? for consistency

tommylau523 (approved):

PR #3576, opened 2025-01-30
SW-6507 Mortality rate overlay disappears after switching between satellite and outdoor map styles

constanzauanini:
When switching map style, images used in the map need to be re-added

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -110,19 +110,22 @@ export default function Map(props: MapProps): JSX.Element {
     setFirstVisible((fv) => fv || visible);
   }, [visible]);
 
-  const loadImages = (map: MapRef) => {
-    mapImages?.forEach(({ name, url }) => {
-      if (!map.hasImage(name)) {
-        map.loadImage(url, (error, image) => {
-          if (error || !image) {
-            snackbar.toastError(error?.message ?? 'Error loading map image.');
-          } else {
-            map.addImage(name, image, { sdf: true });
-          }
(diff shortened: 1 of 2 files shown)


src/components/Map/Map.tsx
```diff
@@ -109,6 +110,20 @@ export default function Map(props: MapProps): JSX.Element {
     setFirstVisible((fv) => fv || visible);
   }, [visible]);
 
+  const loadImages = (map: MapRef) => {
```
tommylau523, line 113: should this be a useCallback as well?

src/components/Map/Map.tsx
```diff
@@ -365,7 +377,7 @@ export default function Map(props: MapProps): JSX.Element {
     ));
 
     return sources;
-  }, [geoData]);
+  }, [geoData, reloadSources]);
```
tommylau523, line 380: I think this needs to have loadImages

src/components/Map/Map.tsx
```diff
@@ -117,16 +135,7 @@ export default function Map(props: MapProps): JSX.Element {
...
-                snackbar.toastError(error?.message ?? 'Error loading map image.');
-              }
-              map.addImage(name, image!, { sdf: true });
-            });
-          }
-        });
+        loadImages(map);
       }
     },
     [mapImages, snackbar]
```
tommylau523, line 141: This needs loadImages too

tommylau523 (approved):
Looks good! Some deps required

PR #3582, opened 2025-01-31
SW-6513 Planted Species chart: Not representing the top 4 species in numbers withdrawn/planted

constanzauanini:


ice1080 (commented):
Question on this, could these 2 changes be combined to a common function since they have the same content? Not sure if worth doing or not

alexTerraformation (approved):

PR #3586, opened 2025-01-31
Updates to readme for first-time setup

ice1080:
The confluence docs were out of date, and the readme was missing playwright install notes. Can also remove the link to confluence if that is preferred

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ You may see references to some private repositories in the documentation. We're 
 ## How to Run the App in Development Mode
 
 1. Copy `.env.sample` from the root directory of this repo to the same directory, with the filename `.env`
-   1. Set up the secrets using the notes in the [Frontend Onboard page](https://terraformation.atlassian.net/wiki/spaces/SW/pages/567541907/Onboarding+Plan+for+Frontend+WebApp+Developers)
+   1. If you work for Terraformation and are using existing Keycloak setup, refer to the secrets in the "Onboarding Plan for Frontend WebApp Developers" confluence page
 2. Login to Docker hub
 
 ```shell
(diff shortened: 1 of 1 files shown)


tommylau523 (commented):
Thanks for keeping this updated!

README.md
```diff
@@ -20,7 +20,8 @@ You may see references to some private repositories in the documentation. We're
 
 ## How to Run the App in Development Mode
 
-1. Configure the `.env` file using the sample file `.env.sample` located at the root directory of this repo.
+1. Copy `.env.sample` from the root directory of this repo to the same directory, with the filename `.env`
+   1. Set up the secrets using the notes in the [Frontend Onboard page](https://terraformation.atlassian.net/wiki/spaces/SW/pages/567541907/Onboarding+Plan+for+Frontend+WebApp+Developers)
```
tommylau523, line 24: We may not want to expose internal Atlassian links since this is technically an open-source repo. It is likely not really security issue, but it will not be a useful url for anyone not in the organization.

tommylau523 (approved):
Looks good to me! Thanks for adding back the playwright step. One very minor issue about an internal URL.

README.md
ice1080: Makes sense @tommylau523 . Am I good to change to something like this?
```md
   1. If you work for Terraformation and are using existing Keycloak setup, refer to the secrets in the "Onboarding Plan for Frontend WebApp Developers"
```

Or would we prefer to just remove entirely and rely on TF employees to already be following the confluence page when they come across this?

tommylau523: Either is fine! No need to split too much hairs on this. I trust your judgement here since it has been a while since I onboarded. 

FWIW I think it never hurts to add extra references and verbose in setup docs! As long as we are clear that a link is not meant to be accessible for everyone!

PR #3594, opened 2025-02-03
SW-6506 Density based on observation of less than a full zone warning is not shown when it should be

constanzauanini:


tommylau523 (approved):

src/scenes/PlantsDashboardRouter/components/PlantingSiteDensityCard.tsx
```diff
@@ -47,13 +47,15 @@ export default function PlantingSiteDensityCard({ plantingSiteId }: PlantingSite
...
       return true;
     }
 
-    // Check that each subzone has at least one completed monitoring plot
-    const allSubzones = summaries[0].plantingZones.flatMap((zone) => zone.plantingSubzones);
-    return allSubzones.every((subzone) => subzone.monitoringPlots.length > 0);
+    const allSubzones = plantingSite.plantingZones?.flatMap((zone) => zone.plantingSubzones);
+    const allSubzonesObserved = summaries[0].plantingZones.flatMap((zone) => zone.plantingSubzones);
+    return allSubzones?.every((subzone) =>
+      allSubzonesObserved.find((subzoneObv) => subzoneObv.plantingSubzoneId === subzone.id)
```
tommylau523, line 57: we may also want to check there is at least one monitoringPlot here.

PR #3596, opened 2025-02-03
SW-5639 - fixing reports missing several required fields

ice1080:
- Fixes reports missing several required fields by changing to loose null checks instead of strict
- Also updates readme a bit for running against remote backend
- Also fixes extra comma in `en.csv`

Report form after Submit before update:
image

Report form after update:
image


tommylau523 (approved):
Nice fix!

PR #3599, opened 2025-02-03
Handle mortality rate nullable

tommylau523:


tommylau523:
* **#3599** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3599?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3599?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/api/types/generated-schema.ts
ice1080, line 1: Does the `mortalityRate` on line 5492 also need to be made optional with a server change? I noticed 9 instances of an optional mortalityRate and just that one instance where it was still required on `GetNurseryV1`

tommylau523: Good catch! I think those are not supposed to be nullable as far as I am aware. At least we don't have to make that change now as part of this PR. This set of mortality rate is related to plant tracking, which is a separate set of stats from the nursery's ones. 

Although there is an argument to be made, that a nursery with no plant saplings at all should not have a mortality rate. 

I will take a look at how nursery mortality rate is used and is displayed in the future!

ice1080 (approved):
Thanks for the clarifications!

PR #3616, opened 2025-02-06
SW-6539 Use completedTime instead of start date for lastObservation on map

constanzauanini:


alexTerraformation:
I don't think this is a change we ought to make because it's inconsistent with everywhere else we use dates for observations, where we use the start date.  I'm going to push back on the Jira task for this one.

alexTerraformation:
> I don't think this is a change we ought to make because it's inconsistent with everywhere else we use dates for observations, where we use the start date. I'm going to push back on the Jira task for this one.

Okay after more investigation this change is good to go!  Sorry for holding things up!

src/scenes/PlantsDashboardRouter/components/ZoneLevelDataMap.tsx
```diff
@@ -213,15 +214,22 @@ export default function ZoneLevelDataMap({ plantingSiteId }: ZoneLevelDataMapPro
       (pz: ObservationPlantingZoneResultsWithLastObv) => {
         return {
           ...pz,
-          lastObv: lastZoneObservation(zoneObservations?.[pz.plantingZoneId])?.startDate,
-          plantingSubzones: pz.plantingSubzones.map((oldSubzone) => ({
-            ...oldSubzone,
-            lastObv: lastSubZoneObservation(
+          lastObv:
+            lastZoneObservation(zoneObservations?.[pz.plantingZoneId])?.completedTime ||
+            lastZoneObservation(zoneObservations?.[pz.plantingZoneId])?.startDate,
```
alexTerraformation, line 219: This seems counterintuitive to me since we shouldn't be showing in-progress observation stats on the plants dashboard.

alexTerraformation: Latest completed observation, right?  That's the distinction I'm making.  We shouldn't be showing in-progress data on the plants dashboard map.

src/scenes/PlantsDashboardRouter/components/ZoneLevelDataMap.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { Box, CircularProgress, Typography, useTheme } from '@mui/material';
+import { getDateDisplayValue } from '@terraware/web-components/utils';
 
 import { PlantingSiteMap } from 'src/components/Map';
 import { MapTooltip, TooltipProperty } from 'src/components/Map/MapRenderUtils';
```
tommylau523, line 7: <p class='graphite__hidden'><i>[Re: line +70]</i></p>
looks like summaries are not used anywhere
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3616?utm_source=unchanged-line-comment">Graphite</a>.</p>

src/scenes/PlantsDashboardRouter/components/ZoneLevelDataMap.tsx
constanzauanini: Yes, in the popup it's the latest completed observation. This part is used for recency and it was searching between all observations. Just added on more filter to get completed ones

src/scenes/PlantsDashboardRouter/components/ZoneLevelDataMap.tsx
constanzauanini: Yes, it is used for Observed Plants, when the zone observation is not part of the last observation (line 360)

constanzauanini: > but perhaps for the plant dashboard, we can use it to get the latest observation dates for each zone and subzone.

I though about it, but the summaries don't specify the latest observation of the zone or subzone I think

tommylau523: They do have it. Zone will have a `latestObservationTime`, subzone will only have one `completedTime`

constanzauanini: Great, I can change it to use those values! Thanks

alexTerraformation (approved):

PR #3624, opened 2025-02-10
SW-6551 Observation detail map issues

constanzauanini:


src/scenes/ObservationsRouter/map/ObservationMapView.tsx
```diff
@@ -171,6 +216,7 @@ export default function ObservationMapView({
                 />
               )
             }
+            zoneInteractive={true}
```
joshbuchea, line 219:
```suggestion
            zoneInteractive
```

joshbuchea (approved):
Just one optional nit, but otherwise these changes look good 👍

PR #3630, opened 2025-02-13
SW-6477 Plants Dashboard not showing correct geometry and plots

constanzauanini:
Also fixed geometry changed message


joshbuchea (approved):
Changes look good, nice work fixing this issue! 💯

PR #3634, opened 2025-02-14
SW-6448 Add Ad Hoc standard observations list view

constanzauanini:


src/redux/features/observations/observationsThunks.ts
```diff
@@ -38,10 +39,16 @@ export const requestObservationsResults = (organizationId: number) => {
 /**
  * Fetch planting site observation results
  */
-export const requestPlantingSiteObservationsResults = (organizationId: number, plantingSiteId: number) => {
+export const requestPlantingSiteObservationsResults = (
+  organizationId: number,
+  plantingSiteId: number,
+  adHoc?: boolean
+) => {
```
joshbuchea, line 46: I assume it simplifies the code to add an `adHoc` param rather than creating a separate set of functions for ad hoc observations? I could see an argument being made to create separate functions, but this change seems reasonable to me 👍

joshbuchea (approved):
Sorry for not reviewing this one earlier. Just one question/comment, but not a blocker. Changes look good 👍

src/redux/features/observations/observationsThunks.ts
constanzauanini: Yes, the responses are the same, that's why I used same functions for this, so we don't have a lot of repeated code

PR #3677, opened 2025-03-05
SW-6643 Variable status not always saving when updated from Document Producer

constanzauanini:


src/types/documentProducer/Variable.ts
```diff
@@ -117,6 +117,15 @@ export const VariableStatuses: VariableStatusType[] = [
   'Rejected',
 ];
 
+export const NonSectionVariableStatuses: VariableStatusType[] = [
+  'Approved',
```
ice1080, line 121: I didn't see this used, is that for a later PR?

ice1080 (approved):

src/types/documentProducer/Variable.ts
constanzauanini: Oh, thanks. I forgot to push one file

PR #3692, opened 2025-03-11
SW-6333 Standard Metrics Table

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsSettings.tsx
```diff
@@ -44,6 +60,12 @@ export default function ReportsSettings(): JSX.Element {
     }
   }, [specificMetricsResponse]);
 
+  useEffect(() => {
+    if (standardMetricsResponse && standardMetricsResponse.status === 'success') {
```
joshbuchea, line 64: This could be simplified to:

```suggestion
    if (standardMetricsResponse?.status === 'success') {
```

joshbuchea (approved):
Suggested an optional change, but feel free to ignore. This looks great!

PR #3693, opened 2025-03-11
SW-6621 Add Funding Entity to top bar

ice1080:
Add `userType` to User definition

Create FundingEntityProvider and useFundingEntity hook

Add Funding Entity to top bar content if user is Funder

Screenshot:
image

### DIFF
--- a/src/providers/FundingEntityProvider.tsx
+++ b/src/providers/FundingEntityProvider.tsx
@@ -1,10 +1,11 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
 import { APP_PATHS } from 'src/constants';
-import FundingEntityService from 'src/services/FundingEntityService';
+import { requestFundingEntity } from 'src/redux/features/funder/fundingEntitiesAsyncThunks';
+import { selectFundingEntityRequest } from 'src/redux/features/funder/fundingEntitiesSelectors';
+import { useAppDispatch, useAppSelector } from 'src/redux/store';
 import strings from 'src/strings';
-import { FundingEntity } from 'src/types/FundingEntity';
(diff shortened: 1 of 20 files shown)


src/providers/FundingEntityProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+export default function FundingEntityProvider({ children }: FundingEntityProviderProps) {
+  const { user, bootstrapped: userBootstrapped } = useUser();
+  const [fundingEntity, setFundingEntity] = useState<FundingEntity | undefined | null>();
+  const [entityAPIRequestStatus, setEntityAPIRequestStatus] = useState<APIRequestStatus>(APIRequestStatus.AWAITING);
+  const navigate = useNavigate();
+  const { isDev, isStaging } = useEnvironment();
+
+  const populateFundingEntity = useCallback(async (userId: number) => {
+    const response = await FundingEntityService.getUserFundingEntity(userId);
```
constanzauanini, line 32: Could we do this using redux? I know we have some places where we still call services directly, but we are trying to use redux in new added code

ice1080: Ya I can update it to use redux. My redux skills are definitely weaker so I didn't think of that and was modeling all of this after Organization and User Providers which didn't seem to use it, but they're probably older code

constanzauanini: Yes, probably those were written before moving to redux

ice1080: Updated to redux!

constanzauanini (approved):
Looks good! Thanks for the update!

ice1080:
### Merge activity

* **Mar 12, 2:46 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/3693).

PR #3696, opened 2025-03-12
SW-6664 Add Project Specific Metric

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/Reports/NewProjectSpecificMetric.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { useAppDispatch, useAppSelector } from 'src/redux/store';
+import strings from 'src/strings';
+import { NewMetric } from 'src/types/AcceleratorReport';
+import useForm from 'src/utils/useForm';
+
+export default function NewProjectSpecificMetric(): JSX.Element {
+  const theme = useTheme();
+  const navigate = useNavigate();
+  const pathParams = useParams<{ projectId: string }>();
+  const projectId = Number(pathParams.projectId);
```
ice1080, line 25: I'm curious why we're converting `projectId` to a number if we convert it back to a string in several of its usages? And if the answer is just "because we always convert numeric strings to numbers then that makes sense to me haha

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/NewProjectSpecificMetric.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const goToProjectReports = () => {
+    navigate(`${APP_PATHS.ACCELERATOR_PROJECT_REPORTS.replace(':projectId', projectId.toString())}?tab=settings`);
+  };
+
+  const [newMetric, , onChange] = useForm<NewMetric>({
+    component: 'Biodiversity',
+    name: '',
+    reference: '',
+    type: 'Activity',
+  });
```
ice1080, line 53: I don't see `description` in this but I see we're using it later on in the PageForm itself. Is that something that needs added here or is it null by default and we're ok with that?

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/NewProjectSpecificMetric.tsx
constanzauanini: projectId is a number in the BE, so if we needed to "work" with it, it would be a number. In this case I'm just using it in the api url and we need it as a string only, so you are right, it would be simpler to just use it as a string!

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/NewProjectSpecificMetric.tsx
constanzauanini: It can be null, that's why I didn't initialize it with a value here

joshbuchea (approved):

PR #3698, opened 2025-03-12
SW-6673 - Add planting site/zone/subzone area to planting site detail view

alexTerraformation:
In the planting site detail view:
- add the planting site area above the map
- add the zone area to the map pop-up
- add the subzone area to the map pop-up
- add the zone area to the list view
- add the subzone area to the list view


src/scenes/PlantingSitesRouter/view/GenericZoneView.tsx
```diff
@@ -153,5 +158,9 @@ const DetailsRenderer =
       return <CellRenderer {...props} value={numMonitoringPlots > 0 ? numMonitoringPlots : ''} sx={textStyles} />;
     }
 
+    if (column.key === 'areaHa') {
+      return <CellRenderer {...props} value={value ? value : ''} sx={textStyles} />;
+    }
```
ice1080, line 163: Nit - this can be simplified from `value={value ? value : ''}` to `value={value || ''}`, unless value can be 0 or false in which case you could change it to `value={value ?? ''}` (which means use `''` only if `value` is null or undefined but keeps 0 and false

src/scenes/PlantingSitesRouter/view/PlantingSiteDetailsTable.tsx
```diff
@@ -235,6 +240,10 @@ const DetailsRenderer =
       );
     }
 
+    if (column.key === 'areaHa') {
+      return <CellRenderer {...props} value={value ? value : ''} sx={textStyles} />;
```
ice1080, line 244: Same comment

src/scenes/PlantingSitesRouter/view/GenericZoneView.tsx
alexTerraformation: Good catch

src/scenes/PlantingSitesRouter/view/BoundariesAndZones.tsx
```diff
@@ -427,6 +428,7 @@ const contextRenderer =
       const zone = data.find((z) => z.id === entity.id);
       title = zone?.name;
       properties = [
+        { key: strings.AREA_HA, value: zone?.areaHa && zone?.areaHa > 0 ? zone?.areaHa : '' },
```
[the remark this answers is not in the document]
alexTerraformation: The only time we have an area of 0 is when the planting site is still in draft, in which case we do not want to display a misleading value.

ice1080 (approved):

PR #3702, opened 2025-03-12
SW-6678: FE - Reports View with Tabs [Participant]

joshbuchea:
This PR adds a new "Reports" page and nav bar item.

Additional changes:

- Rename Service: `ReportService` --> `SeedFundReportService`

## Screenshot

localhost_3000_

### DIFF
--- a/src/scenes/OrgRouter/NavBar.tsx
+++ b/src/scenes/OrgRouter/NavBar.tsx
@@ -20,7 +20,7 @@ import { useParticipantData } from 'src/providers/Participant/ParticipantContext
 import { useLocalization, useOrganization } from 'src/providers/hooks';
 import { NurseryWithdrawalService } from 'src/services';
 import DeliverablesService from 'src/services/DeliverablesService';
-import ReportService, { Reports } from 'src/services/ReportService';
+import SeedFundReportService, { Reports } from 'src/services/SeedFundReportService';
 import strings from 'src/strings';
 import { isAdmin, isManagerOrHigher } from 'src/utils/organization';
 import useDeviceInfo from 'src/utils/useDeviceInfo';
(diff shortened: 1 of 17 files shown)


src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -196,20 +197,37 @@ export default function NavBar({
...
+          label={strings.REPORTS}
+          selected={!!isReportsRoute}
+          onClick={() => {
+            closeAndNavigateTo(APP_PATHS.REPORTS);
+          }}
+          id='reports-list'
+        />
+      ) : null,
+    [closeAndNavigateTo, isReportsRoute, seedFundReports.length, selectedOrganization.canSubmitReports]
+  );
```
joshbuchea, line 215: I'm not very familiar with `selectedOrganization.canSubmitReports`. Would we still want to check this value for non seed fund reports? Or would we potentially need a new value/flag for this report type?

src/scenes/OrgRouter/index.tsx
```diff
@@ -48,6 +48,8 @@ import useDeviceInfo from 'src/utils/useDeviceInfo';
 import useEnvironment from 'src/utils/useEnvironment';
 import useStateLocation from 'src/utils/useStateLocation';
 
+import ReportsRouter from '../Reports';
+
```
ice1080, line 52: this relative path might need to be updated to `src/<something>/Reports`

src/scenes/Reports/index.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { Route, Routes } from 'react-router-dom';
```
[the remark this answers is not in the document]
ice1080: I think this comment is correct and needs updated

ice1080 (commented):
Just a few comments; I don't have enough context to be able to answer your question though

src/scenes/OrgRouter/NavBar.tsx
constanzauanini: We used to enable the possibility of submitting reports to orgs, but maybe now it will apply to all orgs? @alexTerraformation may know that answer better

constanzauanini: I think we need to check the feature flag here

joshbuchea: Oops! I completely forgot to implement the feature flag in this PR. Adding now, thanks @constanzauanini 🙌

src/scenes/OrgRouter/index.tsx
joshbuchea: Good catch, thanks! 🙌

constanzauanini (approved):
Looks good, feel free to merge once you know if we need to check organization being able to submit reports or not

PR #3707, opened 2025-03-13
SW-6561 Use custom PostgreSQL 17 image

sgrimm:
Use PostgreSQL 17, rather than 13, for the local server+database
stack that's used for Playwright tests and sometimes for local
development.

This uses a custom image that includes the extensions that will be
needed by the server for some upcoming features. The image supports
both ARM and x86, replacing the x86-only PostgreSQL 13 image we
were using; in ARM dev environments, this should give a slight
performance increase.

Note that the new version isn't compatible with data files from the
old version. People with existing database files in their local
environments will need to either dump the existing data and restore
it, or remove the data files so they start from a clean slate.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   postgres:
-    image: 'postgis/postgis:13-3.3'
+    image: 'terraware/postgres:v20250227.1'
     ports:
       - '5432:5432'
     volumes:


sgrimm:
* **#3707** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3707?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3707?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

PR #3712, opened 2025-03-17
SW-6665 Add Targets Table

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargetsCellRenderer.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { RendererProps } from 'src/components/common/table/types';
+
+export default function ReportsTargetsCellRenderer(props: RendererProps<TableRowType>): JSX.Element {
+  const { column, row, value, index } = props;
+
+  if (column.key === 'name') {
+    return <CellRenderer column={column} value={value} row={row} index={index} />;
+  }
+
+  return <CellRenderer {...props} />;
```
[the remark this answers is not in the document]
constanzauanini: No, nothing. I need the renderer for the edit, because when clicking the name it will open a modal, but it is not implemented here yet

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -1,7 +1,232 @@
...
+      const metrics: RowMetric[] = [];
+      firstReport.systemMetrics.forEach((sm) => {
+        metrics.push({ name: sm.metric, type: sm.type, reference: sm.reference, component: sm.component, id: -1 });
+      });
+      firstReport.standardMetrics.forEach((sm) => {
+        metrics.push({ name: sm.name, type: sm.type, reference: sm.reference, component: sm.component, id: sm.id });
+      });
+      firstReport.projectMetrics.forEach((pm) => {
+        metrics.push({ name: pm.name, type: pm.type, reference: pm.reference, component: pm.component, id: pm.id });
+      });
```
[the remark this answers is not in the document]
constanzauanini: I think that if a metric is added, then the api will return that metric in all the reports. But maybe @tommylau523 can confirm that

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargetsCellRenderer.tsx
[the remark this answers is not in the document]
ice1080: Ahh ok that makes sense!

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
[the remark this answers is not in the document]
tommylau523: There is a chance if a metric is added after a report is submitted. In this case, the API will throw an exception saying that the metric does not exist 

for example, if a Q1 report is already submitted, and metric A is added, it will show up for all unsubmitted reports other quarters and onwards.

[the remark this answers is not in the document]
constanzauanini: Ok, so probably the first report will not have that metric if it is submitted. Should I get the metrics from an unsubmitted report (if there is one) and if not, use the first report?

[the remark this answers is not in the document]
ice1080: To me that sounds like reports are not reliable for getting all metrics right? For an example where all reports are submitted, there are no unsubmitted reports, but the user adds a bunch of metrics, then all the past reports would no longer contain all the metrics for this if we're trying to display all metrics right?

Or am I misunderstanding and we're only displaying the metrics that are in the list of reports? If that's the case then we might have to do what you're suggesting and get either an unsubmitted report's metrics, or the most recent report if none are unsubmitted.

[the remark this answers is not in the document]
tommylau523: I think all the metrics from submitted and unsubmitted reports should be considered.

In this case, the cells in the table should be empty and not have an input for entering target, if that report does not actually have that metric. 

That’s probably the simplest way without complex logic and retinkering of UI.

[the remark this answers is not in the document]
tommylau523: We can perhaps keep “all metrics” to just be the list of metrics for the set of reports we are looking at. An example is, if we look at 2020 submitted reports, we probably just want to see all of their metric targets, and not worry about future added metrics.

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -1,7 +1,277 @@
...
+      setRequestId(request.requestId);
+    },
+    [dispatch]
+  );
+
+  const getReportsYears = useMemo(() => {
+    const avaiableYears: Set<string> = new Set();
+    allReports?.forEach((report) => {
+      const reportYear = DateTime.fromFormat(report.startDate, 'yyyy-MM-dd').year;
+      avaiableYears.add(reportYear.toString());
```
ice1080, line 241: Minor typo here - `avaiableYears` -> `availableYears`

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -1,7 +1,277 @@
...
+              metric.year = DateTime.fromFormat(report.startDate, 'yyyy-MM-dd').year;
+              if (metric.id !== -1) {
+                const foundMetric = [...report.standardMetrics, ...report.projectMetrics].find(
+                  (met) => met.id === metric.id
+                );
+                metric[quarterProp] = foundMetric?.target ?? undefined;
+              } else {
+                const foundMetric = report.systemMetrics.find((met) => met.metric === metric.name);
+                metric.q1Target = foundMetric?.target;
+              }
```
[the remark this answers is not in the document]
constanzauanini: Oh, yes, thank you!

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -1,7 +1,277 @@
...
+    }
+  }, [reportsResults]);
+
+  const getReportQuarter = (report: AcceleratorReport) => {
+    const startDate = DateTime.fromFormat(report.startDate, 'yyyy-MM-dd');
+    const startMonth = startDate.month;
+    const endDate = DateTime.fromFormat(report.endDate, 'yyyy-MM-dd');
+    const endMonth = endDate.month;
+
+    if (startMonth === 1 && endMonth === 3) {
```
tommylau523, line 131: They don't have to be 1 or 3. But they should both be between 1 and 3. Same for all others.

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -1,7 +1,277 @@
...
+      availableYears.add(reportYear.toString());
+    });
+    return Array.from(availableYears);
+  }, [allReports]);
+
+  const featuredFilters: FilterConfigWithValues[] = useMemo(() => {
+    const currentYear = DateTime.now().year;
+    return activeLocale
+      ? [
+          {
```
tommylau523, line 250: I think this should be a single select instead of multi-select, and `null` should not be allowed. I am not sure if we support not-nulls only yet

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -1,7 +1,277 @@
...
+  const dispatch = useAppDispatch();
+  const [allReports, setAllReports] = useState<AcceleratorReport[]>();
+  const [reports, setReports] = useState<AcceleratorReport[]>();
+  const { activeLocale } = useLocalization();
+  const pathParams = useParams<{ projectId: string }>();
+  const projectId = String(pathParams.projectId);
+  const [metricsToUse, setMetricsToUse] = useState<RowMetric[]>();
+
+  useEffect(() => {
+    const request = dispatch(requestListAcceleratorReports({ projectId }));
```
tommylau523, line 103: I think for the target table, we do need to set `includeFuture = true` in the API. Because we want to be able to review targets set in the future.

tommylau523: For reference, I set a config from 2025 to 2035, but I can only find 2025 targets now.

src/services/AcceleratorReportService.ts
```diff
@@ -109,6 +119,44 @@ const createProjectMetric = async (
...
+  sortOrder?: SearchSortOrder
+): Promise<Response & AcceleratorReportsData> => {
+  let searchOrderConfig: SearchOrderConfig | undefined;
+  if (sortOrder) {
+    searchOrderConfig = {
+      locale: locale ?? null,
+      sortOrder,
+    };
+  }
+  let params = { includeMetrics: 'true' };
```
tommylau523, line 135: this can be a `const`, but I also think `includeFuture`, `includeMetrics` should maybe both be function params. 

`includeFuture = true` is particularly useful for setting/viewing targets.

Some report tables don't want the metrics to be included, although it doesn't hurt to be inculded.

constanzauanini:
@tommylau523 I made all changes you suggested, let me know if that works!

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
constanzauanini: I added it as a multi-select because that is the default behaviour in table filters, but you are right, in designs looks like a single select. I will change it to be a single selector

tommylau523 (approved):
Looks great! I have some very minor suggestions that probably don't need addressing now.

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -1,7 +1,301 @@
...
+  useEffect(() => {
+    if (reportsResults?.status === 'error') {
+      return;
+    }
+    if (reportsResults?.data) {
+      setReports(reportsResults.data);
+    }
+  }, [reportsResults]);
+
+  const getReportQuarter = (report: AcceleratorReport) => {
```
tommylau523, line 127: We can probably make a util function for getQuarterFromDate. That will likely be useful for the external side too.

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -1,7 +1,301 @@
...
+      const request = dispatch(
+        requestListAcceleratorReports({ projectId, search, sortOrder, includeFuture: true, includeMetrics: true })
+      );
+      setRequestId(request.requestId);
+    },
+    [dispatch]
+  );
+
+  const currentYear = DateTime.now().year;
+  const [yearFilter, setYearFilter] = useState<string>(currentYear.toString());
```
tommylau523, line 240: This should probably be currentYear if reports include current year, or first year in the future otherwise.

constanzauanini: Oh, ok. I will add this logic in this same PR

PR #3713, opened 2025-03-17
SW-6676 Add hectares planted to home dashboard

alexTerraformation:
Screenshot 2025-03-17 at 10 12 07 AM

### DIFF
--- a/src/scenes/Home/TerrawareHomeView/StatsCardItem.tsx
+++ b/src/scenes/Home/TerrawareHomeView/StatsCardItem.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { Box, Typography, useTheme } from '@mui/material';
+import { IconTooltip } from '@terraware/web-components';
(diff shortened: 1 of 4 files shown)


constanzauanini (approved):

src/scenes/Home/TerrawareHomeView/StatsCardItem.tsx
```diff
@@ -20,6 +23,8 @@ export const StatsCardItem = ({
   linkText,
   showBorder = true,
   showLink = true,
+  showTooltip = false,
```
constanzauanini, line 26: We could have just one attribute, so if it has tooltipText we show the tooltip and if not, we don't, but this is fine too!

alexTerraformation: I considered it, but opted for following the pattern that the link text/enable already follows.

PR #3726, opened 2025-03-20
SW-6666 Edit Targets Modal

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/Reports/EditTargetsModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+  onClose: () => void;
+  reload: () => void;
+  row: RowMetric;
+}
+
+export default function EditTargetsModal(props: EditTargetsModalProp): JSX.Element {
+  const { onClose, row, reload } = props;
+  const [requestId, setRequestId] = useState<string>('');
+  const dispatch = useAppDispatch();
+  const updateReportMetricsResponse = useAppSelector(selectUpdateAcceleratorReportMetrics(requestId));
```
tommylau523, line 29: The update endpoint is intended for use by external users. An admin using this endpoint will most likely hit a Forbidden error.

You can use the review metrics endpoint for setting targets here. 
@joshbuchea can use the update metrics endpoint for setting targets on the Terraware side.

tommylau523: To help share some code, we could make this modal live under components, and take in a  dispatcher prop, so you can pass in different API requests based on needs.

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -98,6 +106,19 @@ export default function ReportsTargets(): JSX.Element {
   const pathParams = useParams<{ projectId: string }>();
   const projectId = String(pathParams.projectId);
   const [metricsToUse, setMetricsToUse] = useState<RowMetric[]>();
+  const [editOpenModal, setEditOpenModal] = useState(false);
+  const [selectedRows, setSelectedRows] = useState<TableRowType[]>([]);
+  const [selectedMetric, setSelectedMetric] = useState<RowMetric>();
+  const currentYear = DateTime.now().year;
+  const [yearFilter, setYearFilter] = useState<string>();
+  const getReportsYears = useMemo(() => {
```
tommylau523, line 114: The default yearFilter is showing empty for me. Perhaps we can add a `setYearFilter(min(availableYears))` here. 

(Or in a separate useEffect that depends on availableYears. Which may be more clear.)

tommylau523 (changes requested):
The scene code looks good. This screen will need to use a different API endpoint.

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
constanzauanini: But min(availableYears) will not get the current year, or the first year in the future 🤔

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
```diff
@@ -223,49 +266,14 @@ export default function ReportsTargets(): JSX.Element {
...
-        values: [`${yearFilter}`],
-      },
-    ];
-  }, [yearFilter, allReports]);
-
   const fuzzySearchColumns = useMemo(() => ['name'], []);
 
   useEffect(() => {
-    if ((allReports?.length || 0) > 0) {
+    if ((allReports?.length || 0) > 0 && getReportsYears.length > 0) {
```
constanzauanini, line 276: @tommylau523 Maybe I just need to add getReportsYears to the dependencies, that's why the default it's not working always...

constanzauanini: But if we have future and past years, the min will select a past year instead of a future one

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/EditTargetsModal.tsx
constanzauanini: Yes, we need a dispatcher prop to reuse this modal in the participant side. I will leave that change to @joshbuchea when he works on targets on participant side

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx
tommylau523: Which is why I think we have to filter for above the currentYear, to get rid of the past years.

constanzauanini: But designs in targets view says we want to show past years too: https://www.figma.com/design/AHpqtUqr0dPgY3Waqm3CbG/Accelerator-Reports---Assign-and-Collect?node-id=1632-33302&t=IJBaDF8An0QCUsuP-0

tommylau523: You can just replace `availableYears` in my block as `getReportYear`, but it doesn't solve the issue. 

When we calls `setYearFilter` this leads to an new API call that updates `allReports`, so this gets fired again. As a result, it always resets the year I selected.

tommylau523 (approved):

PR #3730, opened 2025-03-20
SW-6679: FE - Reports View - Reports Tab - Reports Table [Participant]

joshbuchea:
This PR includes changes to add a new Reports View (Participant).

## Screenshots

localhost_3000_reports_tab=reports organizationId=11

localhost_3000_reports_tab=reports organizationId=11 (3)


ice1080 (commented):
My guess is that search isn't working because `report` is a string column, but the value isn't a string since the it's instead the link to the report and the name itself is a built string instead of a value on the row. Perhaps @constanzauanini knows how to handle that differently though?
For the year filter, my guess is that it's similar here too. I'm guessing the `row` object doesn't have a field of `'year'` which is what the featuredFilter is attempting to find on the row, so if you select a year it probably displays nothing. The `searchAndSort` function just knows how to look for properties on the row.

Actually for both you might be able to modify the data coming in from the backend and extract the year and calculate the reportName before setting it on state, so that `searchAndSort` is able to find those? Just brainstorming

src/scenes/Reports/ReportCellRenderer.tsx
ice1080, line 1: Not sure if we'd want to memoize the createLinkToReport and the reportName calculation? Wouldn't affect much, just mentioning

src/scenes/Reports/ReportsList.tsx
```diff
@@ -1,7 +1,160 @@
...
+  }, [activeLocale, availableYears, isAcceleratorRoute]);
+
+  return (
+    <Card style={{ display: 'flex', flexDirection: 'column', paddingLeft: 0, paddingRight: 0 }}>
+      <Grid container sx={{}}>
+        <Grid item xs={12} paddingLeft={3}>
+          <Typography color={theme.palette.TwClrTxt} fontSize='20px' fontWeight={600} lineHeight='28px'>
+            {strings.REPORTS}
+          </Typography>
+        </Grid>
```
ice1080, line 141: Might be worth checking with design, but wondering if we could leave this out, since the page has a title of Reports, the tab is called Reports, and there's this card header of Reports

constanzauanini:
> My guess is that search isn't working because `report` is a string column, but the value isn't a string since the it's instead the link to the report and the name itself is a built string instead of a value on the row. Perhaps @constanzauanini knows how to handle that differently though? For the year filter, my guess is that it's similar here too. I'm guessing the `row` object doesn't have a field of `'year'` which is what the featuredFilter is attempting to find on the row, so if you select a year it probably displays nothing. The `searchAndSort` function just knows how to look for properties on the row.
> 
> Actually for both you might be able to modify the data coming in from the backend and extract the year and calculate the reportName before setting it on state, so that `searchAndSort` is able to find those? Just brainstorming

I think you should move to the FE Table, because if not the TableWithSearchFilters searches in the BE results, and in the BE results there is no "report" column. Same thing happen to me with in the Target table; there is a "metric" column that I created in the FE, so the search has to happen in the FE. And then, for the year filter, you can check what I have in my [opened PR](https://github.com/terraware/terraware-web/pull/3726/files). The year is not a normal filter, and we don't want it in the search node object. We need it in the query params of the api, so I changed the implementation to pass the selected year to the service. So the way to make that "filter" work is to just call the api each time the year changes. Also, that filter is a single select dropdown and it always needs a value (at least for the targets table it works like that, i think in reports view it would be the same) so you will always call the api with the query parameter

joshbuchea:
@ice1080 & @constanzauanini, thank you for the PR reviews 🙌

joshbuchea:
@ice1080 Good point about the report column being a rendered link. But doesn’t `searchAndSort` just use the raw data?

I thought the issue might be the missing `report` or `year` properties, but even after adding `year` to the API response, the filter still didn’t work. I might’ve missed something—I'll give it another shot.

joshbuchea:
@constanzauanini thank you for the explanation—I'll move to the FE table as you recommend since it simplifies adding data.

And I'll revisit your PR to see how you wired the year filter to the API.

Quick question: how do you make a filter use a single-select dropdown instead of multi-select?

constanzauanini:
> @constanzauanini thank you for the explanation—I'll move to the FE table as you recommend since it simplifies adding data.
> 
> And I'll revisit your PR to see how you wired the year filter to the API.
> 
> Quick question: how do you make a filter use a single-select dropdown instead of multi-select?

All featured filters are multi select, what I did for the year filter was using the extraComponent prop of the table to add a single-select. You can take a look [here](https://github.com/terraware/terraware-web/blob/main/src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportsTargets.tsx#L283)

src/scenes/Reports/ReportCellRenderer.tsx
joshbuchea: Thanks for flagging, will update 👍

src/scenes/Reports/ReportsList.tsx
joshbuchea: I agree this feels overly redundant on this tab. I think Design opted for consistency over limiting redundancy in this case, but I can check.

src/scenes/Reports/ReportCellRenderer.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const { activeLocale } = useLocalization();
+  const { isAcceleratorRoute } = useAcceleratorConsole();
+  const dispatch = useAppDispatch();
+
+  const modifiedByUser = useAppSelector(selectUser(row.modifiedBy));
+  const submittedByUser = useAppSelector(selectUser(row.submittedBy));
+
+  const createLinkToReport = useCallback(() => {
+    // TODO: update link to accelerator report views once ready
+    const reportUrl = isAcceleratorRoute ? APP_PATHS.REPORTS_VIEW : APP_PATHS.REPORTS_VIEW;
```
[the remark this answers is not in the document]
joshbuchea: This change was intentional (see todo comment above for context), but it would’ve been a good catch otherwise.

src/services/AcceleratorReportService.ts
```diff
@@ -153,7 +155,7 @@ const listAcceleratorReports = async (
       params,
```
tommylau523, line 155: <p class='graphite__hidden'><i>[Re: line +141]</i></p>
You can refer to Constanza's PR here:
https://github.com/terraware/terraware-web/pull/3726

Instead of putting year in the search node, and reverse searching for it. It is much easier to just pass it as a param in my opinion. 

The downside is that we cannot select multiple years, but that is the intended use case anyways.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3730?utm_source=unchanged-line-comment">Graphite</a>.</p>

joshbuchea:
There are just a couple of minor UI issues remaining:

- Year filter option labels are centered for some reason
- Order of table filters doesn't match designs (`extraComponent` is rendered before `featuredFilters`, perhaps this doesn't matter..?)

localhost_3000_reports_tab=reports organizationId=11 (5)

ice1080 (approved):
This all looks good to me.
@joshbuchea did the recommendations from everyone work for the fuzzy search and the year filter now?

joshbuchea:
@ice1080 yes, search and the year filter are working great now 🎉  Thanks for the review!

PR #3740, opened 2025-03-24
SW-6603 Create settings page

ice1080:
Create settings page

Fix styling of form


src/components/TopBar/FunderBreadcrumbs.tsx
ice1080, line 1: There's a few like this one. I couldn't decide if I should keep both `isFunderRoute` and the `userType` check. Decided on both for now in a few instances, and only one in other instances. Can be convinced differently if preferred, but `isFunderRoute` won't cover all cases (such as this new settings page).

src/scenes/MyAccountRouter/MyAccountForm.tsx
ice1080, line 1: Renamed this one and moved the `TfMain` wrapper to a different component, in order to make this form fit inside the settings page.

src/components/TopBar/FunderBreadcrumbs.tsx
tommylau523: I like showing the Funder portal breadcrumbs for funder user type.

tommylau523 (approved):

PR #3743, opened 2025-03-25
Show system metrics in Settings

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/Reports/SystemMetricsRenderer.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import CellRenderer, { TableRowType } from 'src/components/common/table/TableCellRenderer';
+import { RendererProps } from 'src/components/common/table/types';
+
+export default function SystemMetricsRenderer(props: RendererProps<TableRowType>): JSX.Element {
+  const { column, row, index } = props;
+
+  if (column.key === 'name') {
+    return <CellRenderer column={column} value={row.metric} row={row} index={index} />;
+  }
```
ice1080, line 11: Another option here is just to change the columns definition to something like this:
```ts
  const columns = (): TableColumnType[] => [
    {
      key: 'metric',
      name: strings.NAME,
      type: 'string',
    },
    ...
```
that way you don't need a custom renderer. Doesn't matter either way though

ice1080 (approved):

PR #3763, opened 2025-03-28
Refactor & Rename Accelerator Reports Components, Refactor Reports List/Table

joshbuchea:
This PR includes changes to rename and refactor components related to Accelerator Reports.

It also includes a change to refactor the Accelerator Reports List/Table to use a single `projectId` variable.

### DIFF
--- a/src/components/AcceleratorReports/AcceleratorReportsTable.tsx
+++ b/src/components/AcceleratorReports/AcceleratorReportsTable.tsx
@@ -16,7 +16,7 @@ import strings from 'src/strings';
 import { AcceleratorReport, AcceleratorReportStatuses } from 'src/types/AcceleratorReport';
 import { SearchSortOrder } from 'src/types/Search';
 
-import ReportCellRenderer from './ReportCellRenderer';
+import AcceleratorReportCellRenderer from './AcceleratorReportCellRenderer';
 
 type AcceleratorReportRow = AcceleratorReport & {
   reportName?: string;
@@ -28,7 +28,7 @@ const defaultSearchOrder: SearchSortOrder = {
   direction: 'Descending',
 };
 
-export default function ReportsList(): JSX.Element {
+export default function AcceleratorReportsTable(): JSX.Element {
   const dispatch = useAppDispatch();
   const { activeLocale } = useLocalization();
   const { currentParticipantProject } = useParticipantData();
@@ -181,7 +181,7 @@ export default function ReportsList(): JSX.Element {
       fuzzySearchColumns={fuzzySearchColumns}
       id='accelerator-reports-table'
       isClickable={() => false}
-      Renderer={ReportCellRenderer({ projectId })}
+      Renderer={AcceleratorReportCellRenderer({ projectId })}
       rows={acceleratorReports}
       stickyFilters
       title={strings.REPORTS}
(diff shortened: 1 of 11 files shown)


constanzauanini (commented):
Refactor looks good! I just noticed that year filter in the reports list is not working correctly, I don't see all the years in the filter, just one year.

joshbuchea:
@constanzauanini it seems the issue exists on Staging now so is unrelated to changes in this PR, but I'll still fix it.

src/components/AcceleratorReports/AcceleratorReportsTable.tsx
```diff
@@ -64,90 +106,61 @@ export default function ReportsList({ fromConsole }: ReportsListsProps): JSX.Ele
...
-  const availableYears = useMemo(() => {
-    const years = acceleratorReports.map((report) => report.startDate?.split('-')?.[0]);
-    const uniqueYears = Array.from(new Set(years)).sort((a, b) => (b || '').localeCompare(a || ''));
-    return uniqueYears;
-  }, [acceleratorReports]);
+  const yearFilterOptions = useMemo(() => {
+    const years: Set<number> = new Set();
+
+    const currentYear = new Date().getFullYear();
+    years.add(currentYear);
```
constanzauanini, line 153: I think we don't want so show the current year if there are no reports for that year. @tommylau523 told me for the targets table to select the first year in the future for that case, or the last in the past if no future years. But I'm not sure if reports table should work the same way

joshbuchea: @constanzauanini I can make that change, but would you mind preemptively approving before you leave so I can merge before the weekend? 🙏

joshbuchea: That issue also exists on Staging currently.

src/components/AcceleratorReports/AcceleratorReportsTable.tsx
```diff
@@ -174,34 +187,42 @@ export default function ReportsList({ fromConsole }: ReportsListsProps): JSX.Ele
...
+      setYearFilter(yearFilterOptions[0]);
     }
-  }, [availableYears, yearFilter]);
+  }, [yearFilterOptions, yearFilter]);
+
+  if (!projectId) {
+    return <></>;
+  }
+
+  if (listAllAcceleratorReportsRequest?.status === 'success' && !allAcceleratorReports?.length) {
```
constanzauanini, line 211: I'm not seeing this message. Maybe the condition should be allAcceleratorReports?.length === 0 ?

If not adding the current year, this wouldn't be necessary because we will always have reports for selected years

constanzauanini (approved):
Looks good!

src/components/AcceleratorReports/AcceleratorReportsTable.tsx
tommylau523: There is another piece of requirements that go into this. For future reports (end date more than 15 days in the future), we don't want to show it at all. So likely, in the participant/console side, this filter should only show current/past years. 

This is slightly different than the target able.

src/components/AcceleratorReports/AcceleratorReportsTable.tsx
joshbuchea: This change is primarily for the participant side, but just realized I need to add some styles so it doesn't look so crude 😅

localhost_3000_accelerator_projects_8_reports_tab=reports

joshbuchea: Went ahead and removed this message until we have designs for the empty state.

PR #3766, opened 2025-03-28
SW-6613 Only accelerator projects, and deal name

ice1080:


ice1080:
* **#3766** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/hooks/useAcceleratorProjects.tsx
```diff
@@ -0,0 +1,30 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (result?.status === 'success' && result?.data) {
+      setAllProjects(result?.data || []);
+      setBootstrapped(true);
+    }
+  }, [dispatch, result]);
+
+  return { allProjects, bootstrapped };
```
tommylau523, line 29: Instead of using `bootstrapped`, you can initialize `allProjects` to `null` instead of `[]`. 

Later on, you can check that `allProjects = null` to be loading.

src/hooks/useAcceleratorProjects.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const useAcceleratorProjects = () => {
+  const dispatch = useAppDispatch();
+  const [allProjects, setAllProjects] = useState<ParticipantProject[]>([]);
+  const [requestId, setRequestId] = useState<string>('');
+  const [bootstrapped, setBootstrapped] = useState<boolean>(false);
+  const result = useAppSelector(selectParticipantProjectsListRequest(requestId));
+
+  useEffect(() => {
+    if (allProjects.length === 0) {
```
tommylau523, line 16: Unlikely we will ever have a case where there is no actual `allProjects`. But, for correctness, I think you can either use `bootstrapped` here, or check for `null` instead if you decide to go that route. 

Then you can add `dispatch` and `allProjects` into the deps and still have it work correctly. I know some linters don't like not having the deps filled out even if it is intended.

tommylau523 (approved):
Looks good! Some minor suggestions.

PR #3770, opened 2025-03-28
Accelerator Reports follow-up

joshbuchea:
This PR includes a fix to remove an unnecessary "Q" prefix being applied to report quarters in the accelerator reports table.

Result: "QQ1" --> "Q1"


joshbuchea:
This is a very minor low risk change in an area that's feature flagged, I'm going to go ahead and merge this fix before we get any reports.

PR #3774, opened 2025-03-31
SW-6716 Show observations of deleted zones/subzones

sgrimm:
If a planting zone or subzone is deleted after an observation has
happened there, its ID no longer exists on the server side, but there's
still a record of it, which includes its name, in the planting site
history.

The server will now include names in observation results, and we can
use those to uniquely identify regions even after they've been deleted.

Update the observation views to use zone names rather than IDs and to
stop assuming that a zone or subzone in an observation result still
exists in the current version of the site.

This simplifies some code that needed to figure out the zone or subzone
name for an observation result. As part of that simplification, this
change also removes some code that figured out the zone boundary for an
observation result; the boundary was never actually used anywhere since
the map display code gets boundaries from the planting site history.


sgrimm:
* **#3774** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
Depends on https://github.com/terraware/terraware-server/pull/2968

src/api/types/generated-schema.ts
```diff
@@ -7260,7 +7261,7 @@ export interface components {
             /** Format: int32 */
             plantingDensityStdDev?: number;
             /** Format: int64 */
-            plantingSubzoneId: number;
+            plantingSubzoneId?: number;
```
sgrimm, line 7264: Nullability of the subzone and zone IDs hasn't actually changed in the API yet (even with https://github.com/terraware/terraware-server/pull/2968) but it will change in an upcoming server PR. I made the change here so I could verify that the zone/subzone IDs becoming nullable wouldn't cause type-checking errors on the client side.

ice1080 (commented):
Some nice cleanup as part of this, thanks. I'm not familiar enough with the zoning things to feel comfortable approving, but I went through pretty thoroughly and nothing seemed bad to me :+1:

tommylau523 (approved):

src/scenes/PlantsDashboardRouter/components/ZoneLevelDataMap.tsx
```diff
@@ -232,7 +232,7 @@ export default function ZoneLevelDataMap({ plantingSiteId }: ZoneLevelDataMapPro
     }
 
     const baseMap = MapService.getMapDataFromPlantingSite(plantingSite);
-    if (!observation) {
+    if (!observation || !plantingSiteHistory) {
       return baseMap;
```
constanzauanini, line 236: @sgrimm this change is making the recency layer to not work. If there is no plantingSiteHistory we still need to "calculate" the lastObservation for each zone and subzone to show in the map. Sorry I missed this before, I will try to make a fix in a new PR

PR #3781, opened 2025-04-01
SW-6682: FE - Report View - View Report [Participant]

joshbuchea:
This PR adds the Accelerator Report View for the Participant side.


src/scenes/Reports/AcceleratorReportView.tsx
```diff
@@ -0,0 +1,211 @@
...
+import { useAppDispatch, useAppSelector } from 'src/redux/store';
+import strings from 'src/strings';
+import {
+  AcceleratorReport,
+  ReportProjectMetric,
+  ReportStandardMetric,
+  ReportSystemMetric,
+} from 'src/types/AcceleratorReport';
+
+import ApprovedReportMessage from '../AcceleratorRouter/ParticipantProjects/Reports/ApprovedReportMessage';
```
tommylau523, line 26: Can we move these 3 things into components instead? So we are not depending on another router

src/scenes/Reports/AcceleratorReportView.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const { currentParticipantProject, setCurrentParticipantProject } = useParticipantData();
+  const theme = useTheme();
+  const dispatch = useAppDispatch();
+
+  const pathParams = useParams<{ projectId: string; reportId: string }>();
+  const reportId = String(pathParams.reportId);
+  const projectId = String(pathParams.projectId);
+
+  const [requestId, setRequestId] = useState<string>('');
+  const [reports, setReports] = useState<AcceleratorReport[]>();
```
tommylau523, line 41: I am going to add a fetchOne API for this! I think it is easier to not over fetch data.

src/scenes/Reports/AcceleratorReportView.tsx
```diff
@@ -0,0 +1,211 @@
...
+          id='editReport'
+          label={strings.EDIT}
+          onClick={() => {
+            // TODO: edit page
+          }}
+          priority='secondary'
+          size='medium'
+        />
+        <Button
+          disabled={selectedReport?.status === 'Approved'}
```
tommylau523, line 104: This should probably be `!== 'Not Submitted'` instead.

tommylau523 (changes requested):
UI looks good! Just a few minor things.

src/scenes/Reports/AcceleratorReportView.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  const reportsResults = useAppSelector(selectListAcceleratorReports(requestId));
+
+  const reload = () => {
+    if (projectId) {
+      const request = dispatch(requestListAcceleratorReports({ projectId, includeFuture: true, includeMetrics: true }));
+      setRequestId(request.requestId);
+    }
+  };
+
```
tommylau523, line 53: Adding on to the hide annual report, I think we should add a `useEffect` here, to navigate away from an annual report, until that feature is released (or hide it behind an annual report feature flag).

joshbuchea:
Thanks for the quick review @tommylau523 🙌

src/scenes/Reports/AcceleratorReportView.tsx
joshbuchea: Sounds good, I was wondering why there wasn't one available.

src/scenes/Reports/AcceleratorReportView.tsx
joshbuchea: I think perhaps the `disabled` prop should be added to the Edit button with that expression, but the `disabled` logic for the "Submit for Approval" seems correct to me 🤔

tommylau523: We can get some requirements clarification on this. But the metrics cannot be edited and resubmitted, after a submission for approval. A report may be in 'In Review', 'Needs Update' or 'Approved' after submission. 

It is unclear if "Needs Update" should be actionable here.

joshbuchea: Wouldn't we also want the submit button to be enabled for "Needs Update" status?

tommylau523: I think you are right, but I remember reading in the PRD that only Not Submitted reports are editable and submittable. 

Obviously that wouldn't make sense. So I think a requirement clarification is good for here.

src/scenes/Reports/AcceleratorReportView.tsx
tommylau523: https://github.com/terraware/terraware-server/pull/2976

tommylau523 (approved):
Looks good already! Pending a few minor clarifications.

src/scenes/Reports/AcceleratorReportView.tsx
tommylau523: Approved for now since the components all look good. We can make a different PR for adding the new redux later too.

PR #3783, opened 2025-04-01
SW-6727 Plants dashboard: Changing between planting sites causes mult…

constanzauanini:
…iple maps to get drawn


ice1080 (approved):

alexTerraformation:
Is it possible this was also the cause behind https://terraformation.atlassian.net/browse/SW-6714 ?

constanzauanini:
> Is it possible this was also the cause behind https://terraformation.atlassian.net/browse/SW-6714 ?

It could be, because it was not refreshing when zone observations changed, but we need to test. I couldn't reproduce [SW-6714](https://terraformation.atlassian.net/browse/SW-6714) before

PR #3787, opened 2025-04-02
SW-6732 Clicking on a report doesn't bring me to the right report

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportView.tsx
```diff
@@ -177,11 +177,11 @@ const ReportView = () => {
...
-  const quarterNumber = selectedReport?.startDate
-    ? Math.ceil((new Date(selectedReport?.startDate).getMonth() + 1) / 3)
-    : 0;
-  const reportName =
-    selectedReport?.frequency === 'Annual' ? `${year}` : quarterNumber ? `${year}-Q${quarterNumber}` : '';
+  const reportName = selectedReport
+    ? selectedReport.frequency === 'Annual'
+      ? `${year}`
+      : `${year}-${selectedReport.quarter}`
+    : '';
```
ice1080, line 184: Is there ever a chance that a startDate will be in the year prior to the end date? e.g. if a Q1 report was started in December due to timezone diffs or something?

ice1080 (approved):

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportView.tsx
constanzauanini: Hmm, I think it shouldn't happen, but I'm not 100% sure

PR #3792, opened 2025-04-02
SW-6744 - Ad-hoc & biomass observations - remove feature flag

constanzauanini:


src/scenes/PlantingSitesRouter/view/BoundariesAndZones.tsx
```diff
@@ -78,53 +76,51 @@ export default function BoundariesAndZones({
         <Typography fontSize='16px' fontWeight={600} margin={theme.spacing(3, 0)}>
           {strings.BOUNDARIES_AND_ZONES}
         </Typography>
-        {adHocObservationSupportEnabled && view === 'list' && (
-          <Box display={'flex'} alignItems='center'>
```
ice1080, line 82: I think this one still needs to check that `view === 'list'` right?

src/scenes/ObservationsRouter/ObservationsHome.tsx
```diff
@@ -43,9 +40,8 @@ export default function ObservationsHome(props: ObservationsHomeProps): JSX.Elem
   const { selectedOrganization } = useOrganization();
   const [selectedPlantingSite, setSelectedPlantingSite] = useState<PlantingSite>();
   const [plantsSitePreferences, setPlantsSitePreferences] = useState<Record<string, unknown>>();
-  const [view, setView] = useState<View>();
+  const [view] = useState<View>();
   const plantingSites = useAppSelector(selectPlantingSites);
```
ice1080, line 44: Is `view` used anymore in this file, or can we get rid of this `setState`?

constanzauanini: It is used in line 130, but probably it is not necessary anymore. Let me take another look

src/scenes/PlantingSitesRouter/view/BoundariesAndZones.tsx
constanzauanini: Oops, good catch! yes!

ice1080 (approved):

PR #3794, opened 2025-04-02
SW-668: FE - Report View - Targets Not Set [Participant]

joshbuchea:
This PR adds a warning message in the Report View that shows when report targets must be set.

## Screenshot

localhost_3000_reports_28_213_organizationId=28


src/scenes/Reports/AcceleratorReportView.tsx
```diff
@@ -86,6 +87,18 @@ const AcceleratorReportView = () => {
...
+    const allMetrics = [
+      ...(selectedReport?.systemMetrics || []),
+      ...(selectedReport?.projectMetrics || []),
+      ...(selectedReport?.standardMetrics || []),
+    ];
+
+    const allMetricsHaveTargets = allMetrics.every((metric) => metric.target !== undefined && metric.target !== null);
+
+    return !!allMetrics.length && allMetricsHaveTargets;
+  }, [selectedReport]);
```
joshbuchea, line 100: I'm guessing this logic isn't correct. I started with a very simple set of criteria until I can get more clarity around this requirement:

- All metrics have corresponding targets set
- At least one metric is defined

Questions:

- Should we be checking all three types of metrics or only a subset?
- What other criteria should be included?

src/strings/csv/en.csv
```diff
@@ -706,6 +706,7 @@ GO_TO_DASHBOARD,Go to Dashboard,
 GO_TO_NURSERIES,Go to Nurseries,
 GO_TO_PLANTING_SITES,Go to Planting Sites,
 GO_TO_PROFILE,Go to Profile,
+GO_TO_REPORTS_AND_THEN_TO_TARGETS_TAB_TO_SET_UP_TARGETS_FOR_EACH_METRIC,Go to Reports and then to the Targets tab to set up targets for each metric.,
 GO_TO_SEED_BANKS,Go to Seed Banks,
```
ice1080, line 710: What about something like `GO_TO_REPORTS_TARGETS_FOR_METRIC`?

src/scenes/Reports/AcceleratorReportView.tsx
```diff
@@ -86,6 +87,18 @@ const AcceleratorReportView = () => {
...
 
+  const targetsAreSet = useMemo(() => {
+    const allMetrics = [
+      ...(selectedReport?.systemMetrics || []),
+      ...(selectedReport?.projectMetrics || []),
+      ...(selectedReport?.standardMetrics || []),
+    ];
+
+    const allMetricsHaveTargets = allMetrics.every((metric) => metric.target !== undefined && metric.target !== null);
+
```
ice1080, line 98: NIT you can change this to just `metric.target != null` which will check both `undefined` and `null` in one call

src/scenes/Reports/AcceleratorReportView.tsx
ice1080: Will the updated logic be part of this PR or come later?

src/scenes/Reports/AcceleratorReportView.tsx
```diff
@@ -101,7 +114,9 @@ const AcceleratorReportView = () => {
           size='medium'
         />
         <Button
-          disabled={selectedReport?.status !== 'Not Submitted' && selectedReport?.status !== 'Needs Update'}
+          disabled={
+            (selectedReport?.status !== 'Not Submitted' && selectedReport?.status !== 'Needs Update') || !targetsAreSet
+          }
```
ice1080, line 119: We might want to create a list of statuses somewhere for updateable components in this view, since the next PR will also have several. Could do something like this?
```ts
const UPDATABLE_STATUSES = ['Not Submitted', 'Needs Update']
...
disabled={UPDATABLE_STATUSES.includes(selectedReport?.status}
```

This isn't required for approval or anything, just commenting on it.

src/scenes/Reports/AcceleratorReportView.tsx
tommylau523: The report will only include metrics that are included in this report. I believe the error message should appear if any of the included metrics are missing targets.

tommylau523: For submission, another check is that, each metric (except for system metrics), must have a value inputted.

src/strings/csv/en.csv
joshbuchea: Sure

src/scenes/Reports/AcceleratorReportView.tsx
joshbuchea: I prefer using strict over loose equality for comparisons.

src/scenes/Reports/AcceleratorReportView.tsx
joshbuchea: @ice1080 updated logic will be part of this PR.

joshbuchea: Thanks for clarification around requirements @tommylau523 🙌

src/scenes/Reports/AcceleratorReportView.tsx
joshbuchea: I think that's a good optimization we should probably make. Although it's possible the disabled logic changes after feedback from QA, so I think handling it in the next PR makes sense.

src/scenes/Reports/AcceleratorReportView.tsx
joshbuchea: @tommylau523 should we be checking there is some number of metrics or is that unnecessary? Can we assume metrics will always be populated?

tommylau523: I asked for clarification on this just now on the PRD.

joshbuchea: I just asked for a message copy update in Figma.

joshbuchea:
Closing this since we're not going to require targets before report submission for launch.

PR #3797, opened 2025-04-03
SW-6707 Match "Other" species on ad-hoc observations

sgrimm:
For assigned observations, we let users match the "Other" species
with ones on the organization's species list. Add the same
functionality for ad-hoc monitoring observations.

### DIFF
--- a/src/scenes/ObservationsRouter/plot/AdHocObservationDetails.tsx
+++ b/src/scenes/ObservationsRouter/plot/AdHocObservationDetails.tsx
@@ -1,27 +1,49 @@
-import React, { useEffect, useMemo } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { useNavigate, useParams } from 'react-router-dom';
 
 import { Box, Grid, Typography, useTheme } from '@mui/material';
-import { Textfield } from '@terraware/web-components';
+import { Button, Message, Textfield } from '@terraware/web-components';
 import { getDateDisplayValue } from '@terraware/web-components/utils';
+import _ from 'lodash';
 
 import Card from 'src/components/common/Card';
+import OptionsMenu from 'src/components/common/OptionsMenu';
 import { APP_PATHS } from 'src/constants';
-import { useLocalization } from 'src/providers';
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#3797** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3797?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3797?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
This is pretty much copy-pasted from the assigned observation details page. I don't love the code duplication but I don't know the right way to abstract out a bunch of `useEffect` calls and such. Happy to revise this if there's an existing pattern I can follow for that.

In case it's relevant, the bulk of SW-6707 is actually about adding species matching for biomass observations. That's going to be a bigger change than this since the data models are different, but if I deduplicate the code here, I'll probably want to do it in such a way that I can reuse some/all of it for biomass observations too.

src/scenes/ObservationsRouter/plot/AdHocObservationDetails.tsx
```diff
@@ -101,12 +134,123 @@ export default function AdHocObservationDetails(): JSX.Element {
...
+      setShowPageMessage(true);
+    } else {
+      setShowPageMessage(false);
+    }
+  }, [monitoringPlot]);
+
+  useEffect(() => {
+    if (matchResponse?.status === 'success' && matchResponse?.data && matchResponse.data.length > 0) {
+      // Force reload page to show updated data
+      window.location.reload();
```
constanzauanini, line 160: Instead of doing this reload you can pass the reload function as an attribute and call it here.
In src/scenes/ObservationsRouter/index.tsx:

`      <Route
        path={'/:plantingSiteId/results/:observationId/adHocPlot/:monitoringPlotId'}
        element={<AdHocObservationDetails reload={reload}/>}
      />`

constanzauanini (approved):

PR #3798, opened 2025-04-03
SW-6734 Clicking back to Reports page from a Report should keep table filters

constanzauanini:
The status filter was already kept. The year filter, since it's not a real filter, was not. So added a query param to keep it


src/components/AcceleratorReports/AcceleratorReportsTable.tsx
ice1080, line 1: Should this component also set the `year` query param once it's been determined from the current year? Or do we just want to add it when navigating back to this page?

ice1080 (approved):

src/components/AcceleratorReports/AcceleratorReportsTable.tsx
constanzauanini: I would say just when navigating back, it wouldn't change anything if we set it here

PR #3804, opened 2025-04-04
SW-6702 Can't edit link in Project Report Settings

constanzauanini:


src/redux/features/reports/reportsThunks.ts
```diff
@@ -49,12 +49,17 @@ export const requestCreateReportConfig = createAsyncThunk(
...
-      return response.data;
+    if (request.logframeUrl) {
+      promises.push(AcceleratorReportService.updateLogframeUrl(request.logframeUrl, request.projectId));
     }
 
+    const results = await Promise.all(promises);
+
+    if (results.every((result) => result && result.requestSucceeded)) {
+      return true;
+    }
```
ice1080, line 62: If the config updates successfully but the logframe fails, would we want to reject with a different value? Or are we ok with failing both even if the config updated? What would the UI do in that case, since we're not reloading or redirecting?

constanzauanini: If one fails, it remains in the same page and shows "there was an error". Not sure if it the best, but the UI doesn't know there are two different apis for this

tommylau523: https://github.com/terraware/terraware-server/pull/2993

This makes the APIs simpler!

tommylau523 (approved):

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/EditSettings.tsx
```diff
@@ -131,9 +138,8 @@ export default function EditSettings(): JSX.Element {
                   id='logframe'
                   label={strings.LOG_FRAME_AND_ME_PLAN_URL}
                   type='text'
-                  onChange={(value) => onChange('logframe', value)}
-                  disabled={true}
-                  value={''}
+                  onChange={(value) => onChange('logframeUrl', value)}
+                  value={newConfig.logframeUrl}
```
tommylau523, line 142: Not for the edit screen, but on the view screen, the URL should be a clickable link.

PR #3812, opened 2025-04-05
SW-6683: FE - Report View - Edit Report [Participant]

joshbuchea:
This PR adds the Edit View for Accelerator Reports on the Participant side.


ice1080 (commented):
I noticed in the thunks and service and slice etc you left all the existing endpoint calls even though they don't exist anymore. Is the plan to remove them in a future PR?

src/scenes/Reports/AcceleratorReportEdit.tsx
```diff
@@ -0,0 +1,93 @@
...
+      return;
+    }
+    if (reportsResults?.data) {
+      setReports(reportsResults.data);
+    }
+  }, [reportsResults]);
+
+  useEffect(() => {
+    if (reports) {
+      const reportSelected = reports.find((report) => report.id.toString() === reportId);
```
tommylau523, line 59: We should replace this with the get one report API

src/scenes/Reports/AcceleratorReportEdit.tsx
```diff
@@ -0,0 +1,93 @@
...
+      setSelectedReport(reportSelected);
+    }
+  }, [reportId, reports]);
+
+  const year = useMemo(() => {
+    return selectedReport?.startDate.split('-')[0];
+  }, [selectedReport]);
+
+  const reportName =
+    selectedReport?.frequency === 'Annual' ? year : selectedReport?.quarter ? `${year}-${selectedReport?.quarter}` : '';
```
tommylau523, line 69: `${year}` instead of year can make sure that reportName has a string type.

src/scenes/Reports/AcceleratorReportEditForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+          {['system', 'project', 'standard'].map((type) => {
+            const metrics =
+              type === 'system'
+                ? record.systemMetrics
+                : type === 'project'
+                  ? record.projectMetrics
+                  : record.standardMetrics;
+
+            return metrics?.map((metric, index) => (
+              <MetricBox
```
tommylau523, line 130: For system metrics, a participant should not be able to override metrics, or reset metrics.

src/scenes/Reports/AcceleratorReportEditForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }, [projectId, reportId, saveReportResponse]);
+
+  useEffect(() => {
+    if (projectId !== currentParticipantProject?.id?.toString()) {
+      setCurrentParticipantProject(projectId);
+    }
+  }, [currentParticipantProject?.id, projectId]);
+
+  const onChangeMetric = useCallback(
```
tommylau523, line 69: I am wondering if it is easier to keep 3 separate `onChangeMetrics` methods, instead of combining the calls, looking up by types, and then having to deconstruct it back into 3 fields for the API.

tommylau523 (commented):
Some minor feedback!

src/scenes/Reports/AcceleratorReportEditForm.tsx
joshbuchea: In this case, I think it's simpler using a single function.

src/scenes/Reports/AcceleratorReportEditForm.tsx
joshbuchea: What needs to change in the UI?

src/scenes/Reports/AcceleratorReportEditForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+          )}
+          {['system', 'project', 'standard'].map((type) => {
+            const metrics =
+              type === 'system'
+                ? record.systemMetrics
+                : type === 'project'
+                  ? record.projectMetrics
+                  : record.standardMetrics;
+
+            return metrics?.map((metric, index) => (
```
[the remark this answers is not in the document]
joshbuchea: Let's confirm/clarify this and we can handle updating all instances where accelerator report metrics are rendered.

src/scenes/Reports/AcceleratorReportEditForm.tsx
tommylau523: The system metrics should not have the reset button, or the input box for values.

ice1080 (commented):
No issues on my end! I'll let @tommylau523 do the final approval since he had more feedback

src/scenes/Reports/AcceleratorReportEditForm.tsx
joshbuchea: Got it, thanks @tommylau523!

joshbuchea: @tommylau523 I updated the UI to hide the reset button for system metrics on the participant side, but waned to confirm the change to hide the input box and prevent overriding metrics:

Should the result be that there's no edit button button in the UI to open the modal containing the input box?

tommylau523 (approved):

src/scenes/Reports/AcceleratorReportEditForm.tsx
tommylau523: That's correct, no button for the modal.

tommylau523 (approved):
Looks good! Other than needing to hide the edit button.

PR #3820, opened 2025-04-08
SW-6242 Add CSV export of biomass observations

sgrimm:
Allow users to export CSV files from the biomass observation list
view and from the biomass observation details page (where there are
three different export options).

`SearchFiltersWrapper` now supports an `onExport` property which,
if present, causes an export button to be added to the right of
the search bar.


sgrimm:
* **#3820** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3820?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
Depends on https://github.com/terraware/terraware-server/pull/2999

src/scenes/ObservationsRouter/BiomassMeasurementsDetails.tsx
ice1080, line 1: Perhaps more of a question for product, but it looks like we have an `ExportCsvModal` for this type of functionality. Is there a reason we're not using that here?

src/scenes/ObservationsRouter/BiomassMeasurementsDetails.tsx
```diff
@@ -138,8 +141,65 @@ export default function BiomassMeasurementsDetails(): JSX.Element {
...
+    await downloadCsv(strings.BIOMASS_MEASUREMENTS, ObservationsService.exportBiomassDetailsCsv);
+  }, [observation]);
+
+  const exportSpeciesCsv = useCallback(async () => {
+    await downloadCsv(strings.SPECIES, ObservationsService.exportBiomassSpeciesCsv);
+  }, [observation]);
+
+  const exportTreesShrubsCsv = useCallback(async () => {
+    await downloadCsv(strings.TREES_AND_SHRUBS, ObservationsService.exportBiomassTreesShrubsCsv);
+  }, [observation]);
```
ice1080, line 174: I think these will need to also be dependent on `plantingSite`, otherwise it might always be null.

src/scenes/ObservationsRouter/BiomassMeasurementsDetails.tsx
sgrimm: The Jira task didn't say one way or another, but it suggested that this should be modeled after the CSV export from the species list page, which doesn't use that dialog.

ice1080 (approved):
Looks good! I think at some point in the future we should ask product if it makes sense to combine our export functionality to be consistent, but if there's already multiple then this isn't the PR for it.

PR #3824, opened 2025-04-09
Console changes for publishing report

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/Reports/EditMetricModal.tsx
```diff
@@ -131,6 +131,16 @@ export default function EditMetricModal(props: EditMetricModalProps): JSX.Elemen
...
+        <Grid item xs={12}>
+          <Checkbox
+            disabled={false}
+            id={'isPublishable'}
+            name={'isPublishable'}
+            label={strings.PUBLISH_TO_FUNDER_PORTAL}
+            value={record.isPublishable}
+            onChange={(newValue: boolean) => onChange('isPublishable', newValue)}
+          />
+        </Grid>
```
ice1080, line 143: Does this need to be hidden for console? Or is it available for both console and participants?

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportView.tsx
```diff
@@ -200,6 +238,7 @@ const ReportView = () => {
 
   return (
     <>
+      {showPublishModal && <PublishModal onClose={() => setShowPublishModal(false)} onSubmit={publishReport} />}
```
ice1080, line 241: Is the `open` prop in the `confirm` component essentially the same thing as doing `{showPublishModal && <...`? Looks like we've done it both ways so no need to change, just curious

ice1080 (approved):

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/EditMetricModal.tsx
constanzauanini: I think it's available for both, but will check

src/scenes/AcceleratorRouter/ParticipantProjects/Reports/ReportView.tsx
constanzauanini: We initially used the "open" prop, but because the state does not clear if you close the modal we moved to the other approach. With the `{showPublishModal && <...` it's a new render every time so if you close the modal and open again, there will be no old data. In this case we don't have fields to fill, so it would be the same

ice1080: Ah that makes sense!

PR #3826, opened 2025-04-09
SW-6773 Update to latest react-router 6, add startTransition future flag

ice1080:
Update to latest version of react-router 6.

Add future flag of `v7_startTransition` to prepare for react-router 7. 
Remove `React.lazy` imports since react-router 7 does not support them with it's useTransition logic - [Upgrade Docs](https://reactrouter.com/upgrading/v6#v7_starttransition) and [Changelog](https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#futurev7_starttransition)

Note that this PR should not be merged until right after a release so that it sits in staging for a week for issues to be found.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -22,7 +22,12 @@ root.render(
   // of the HTML which is later parsed by PagedJS
   // <React.StrictMode>
   <React.Suspense fallback={strings.LOADING}>
-    <Router>
+    <Router
+      future={{
+        // v7_relativeSplatPath: true,
+        v7_startTransition: true,
+      }}
+    >
       <Routes>
         <Route
           path={APP_PATHS.ERROR}
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "react-multi-carousel": "^2.8.2",
     "react-redux": "^9.0.0",
     "react-responsive": "^10.0.0",
-    "react-router-dom": "^6.20.0",
+    "react-router-dom": "6",
     "sanitize-filename": "^1.6.3",
     "slate": "^0.112.0",
     "slate-dom": "^0.112.0",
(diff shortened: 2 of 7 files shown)


ice1080:
* **#3826** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3826?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3826?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080:
Not sure why there's different failures in the actions than locally; they're all passing locally

src/index.tsx
```diff
@@ -22,7 +22,12 @@ root.render(
   // of the HTML which is later parsed by PagedJS
   // <React.StrictMode>
   <React.Suspense fallback={strings.LOADING}>
-    <Router>
+    <Router
+      future={{
+        // v7_relativeSplatPath: true,
```
constanzauanini, line 27: Why is this commented?

constanzauanini:
Changes looks good! Not sure why the test is failing, looks like the species scientific name is different? I don't see why that could be happening

src/index.tsx
ice1080: That's the other necessary piece needed to be updated before upgrading to react-router v7. When you add that future flag, you're saying that you want to use the v7 behavior instead of v6, which causes breaking changes. I left it in to help us remember what else is needed, but will work on the changes necessary for that flag in a future PR.

ice1080:
Ya agreed. I can't seem to figure out why it's only failing in the action too. I've locally run the following:
- `yarn server:reset`
- `yarn test:e2e`

alexTerraformation:
This is the failure screenshot FWIW: 
dfeb372a1f60810e13477c7c088738723058e97a

ice1080:
Ya it's weird, there's actually 2 retries too and two of the screenshots look like that, but the third one loaded it. In all 3 cases playwright somehow found 2 elements with the same text

alexTerraformation:
> Ya it's weird, there's actually 2 retries too and two of the screenshots look like that, but the third one loaded it. In all 3 cases playwright somehow found 2 elements with the same text

The second one is the display text from the species details page.
Screenshot 2025-04-09 at 2 10 55 PM

constanzauanini (approved):

PR #3827, opened 2025-04-09
SW-6684: FE - Report View - Submit Report [Participant]

joshbuchea:
This PR adds the ability to submit accelerator reports from the participant side.


ice1080 (approved):
👍

PR #3840, opened 2025-04-11
Funder Report View

constanzauanini:
This will show the same report for all funders, since the report is mocked.


ice1080 (approved):
Looks good to me!

PR #3844, opened 2025-04-15
SW-6799 Funder Reports - Add BE api calls to UI

constanzauanini:


src/redux/features/funder/fundingEntitiesSelectors.ts
ice1080, line 1: I'm still trying to come up to speed on redux and remember everything with it. 
If we changed `requestId` to `projectId`, then would that allow us to not have to request again if we navigate back to the page from somewhere else?

constanzauanini: Yes, I think you are right. It make sense to use the projectId here

ice1080 (commented):
Looks good! Is the graphite comment valid too?

tommylau523 (commented):
We still need to add metrics to the endpoint. Those will follow.

ice1080 (approved):

PR #3846, opened 2025-04-15
SW-6761: Reports page should not be accessible by Contributors & Managers (Follow-ups)

joshbuchea:
This PR includes some follow-up updates to the initial PR:

- Adjust permissions required to read reports
- Only render routes for Accelerator Reports if feature flag is enabled and user has required permission, otherwise redirect user to Home

### DIFF
--- a/src/scenes/Reports/index.tsx
+++ b/src/scenes/Reports/index.tsx
@@ -1,11 +1,27 @@
 import React from 'react';
-import { Route, Routes } from 'react-router-dom';
+import { Navigate, Route, Routes } from 'react-router-dom';
+
+import { APP_PATHS } from 'src/constants';
+import isEnabled from 'src/features';
+import { useUser } from 'src/providers';
 
 import AcceleratorReportEdit from './AcceleratorReportEdit';
 import AcceleratorReportView from './AcceleratorReportView';
 import AcceleratorReportsView from './AcceleratorReportsView';
 
 const AcceleratorReportsRouter = () => {
+  const { isAllowed } = useUser();
+  const isReportsEnabled = isEnabled('Assigning and Collecting Reports');
+  const isAllowedReadReports = isAllowed('READ_REPORTS');
+
+  if (!(isReportsEnabled && isAllowedReadReports)) {
+    return (
+      <Routes>
+        <Route path='*' element={<Navigate to={APP_PATHS.HOME} />} />
+      </Routes>
+    );
+  }
+
   return (
     <Routes>
       <Route path='/*' element={<AcceleratorReportsView />} />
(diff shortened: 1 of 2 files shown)


src/utils/acl.ts
```diff
@@ -133,6 +133,19 @@ const isAllowedReadDeliverable: PermissionCheckFn<ReadSubmissionMetadata> = (
...
+ */
+type ReadReportsMetadata = { organization: Organization };
+const isAllowedReadReports: PermissionCheckFn<ReadReportsMetadata> = (
+  user: User,
+  _: GlobalRolePermission,
+  metadata?: ReadReportsMetadata
+): boolean => {
+  return isAcceleratorAdmin(user) || isAdmin(metadata?.organization);
+};
+
```
joshbuchea, line 148: The ticket requests making accelerator reports visible to all admins, but not to managers or contributors—this conflicts with the role permissions outlined in the [PRD](https://terraformation.atlassian.net/wiki/spaces/PM/pages/1547337737/PRD+Assigning+and+Collecting+Reports). I've asked for clarification on which roles should have access and will open the PR once I hear back.

joshbuchea: Going to go ahead and open this for review for feedback, but will wait to merge until I hear back from Product.

tommylau523: Just saw this. Is this a requirement change? I will go ahead and make the BE change if so.

src/scenes/Reports/index.tsx
```diff
@@ -1,11 +1,27 @@
...
+  const isReportsEnabled = isEnabled('Assigning and Collecting Reports');
+  const isAllowedReadReports = isAllowed('READ_REPORTS');
+
+  if (!(isReportsEnabled && isAllowedReadReports)) {
+    return (
+      <Routes>
+        <Route path='*' element={<Navigate to={APP_PATHS.HOME} />} />
+      </Routes>
+    );
+  }
```
ice1080, line 23: This could also be a useEffect with a navigate to home in it, similar to `TerrawareRouter`. Not sure if one is better than the other though

src/utils/acl.ts
joshbuchea: @tommylau523 so there's a [comment](https://terraformation.atlassian.net/browse/SW-6761?focusedCommentId=24164) in this QA ticket [SW-6761](https://terraformation.atlassian.net/browse/SW-6761) requesting changes that don't seem to align with the [permissions table in the PRD](https://terraformation.atlassian.net/wiki/spaces/PM/pages/1547337737/PRD+Assigning+and+Collecting+Reports#User-Permissions-for-Report-Data). I asked for clarification from Léann, but haven't heard back yet.

src/scenes/Reports/index.tsx
joshbuchea: I chose this approach because it felt clean in terms of UX, security, and efficiency—no need to request resources or render routes the user shouldn’t access.

PR #3849, opened 2025-04-16
SW-6762: In TW, Reports nav item should not show up for Projects that have no report start & end date

joshbuchea:
This PR includes changes to use the new organization features API for dynamically rendering select nav items.

### DIFF
--- a/src/scenes/OrgRouter/NavBar.tsx
+++ b/src/scenes/OrgRouter/NavBar.tsx
@@ -59,14 +59,14 @@ export default function NavBar({
   const isAccessionDashboardRoute = useMatch({ path: APP_PATHS.SEEDS_DASHBOARD + '/', end: false });
   const isAccessionsRoute = useMatch({ path: APP_PATHS.ACCESSIONS + '/', end: false });
   const isApplicationRoute = useMatch({ path: APP_PATHS.APPLICATIONS + '/', end: false });
-  const isCheckinRoute = useMatch({ path: APP_PATHS.CHECKIN + '/', end: false });
+  const isCheckInRoute = useMatch({ path: APP_PATHS.CHECKIN + '/', end: false });
   const isDeliverablesRoute = useMatch({ path: APP_PATHS.DELIVERABLES + '/', end: false });
   const isDeliverableViewRoute = useMatch({ path: APP_PATHS.DELIVERABLE_VIEW + '/', end: false });
   const isHomeRoute = useMatch({ path: APP_PATHS.HOME + '/', end: false });
   const isPeopleRoute = useMatch({ path: APP_PATHS.PEOPLE + '/', end: false });
   const isSpeciesRoute = useMatch({ path: APP_PATHS.SPECIES + '/', end: false });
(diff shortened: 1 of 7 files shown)


src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -214,12 +190,12 @@ export default function NavBar({
...
         />
       ) : null,
-    [closeAndNavigateTo, isReportsEnabled, isReportsRoute]
+    [activeLocale, closeAndNavigateTo, isReportsEnabled, isReportsRoute, orgFeatures?.reports?.enabled]
   );
 
   const seedFundReportsMenu = useMemo<JSX.Element | null>(
     () =>
-      seedFundReports.length > 0 && selectedOrganization.canSubmitReports && activeLocale ? (
+      selectedOrganization.canSubmitReports && !!orgFeatures?.seedFundReports?.enabled && activeLocale ? (
```
joshbuchea, line 198: Assumed we still need to check `selectedOrganization.canSubmitReports`

src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -234,14 +210,17 @@ export default function NavBar({
...
     ]
   );
 
   const modulesMenu = useMemo<JSX.Element | null>(
     () =>
-      currentParticipantProject && orgHasModules && isManagerOrHigher(selectedOrganization) ? (
+      currentParticipantProject &&
+      !!orgFeatures?.modules?.enabled &&
+      isManagerOrHigher(selectedOrganization) &&
+      activeLocale ? (
```
joshbuchea, line 223: Also assuming we still need to check `currentParticipantProject` & `isManagerOrHigher(selectedOrganization` unless the new org features endpoint handles any of those checks?

tommylau523 (approved):
All looks good component-wise. I can review again for Redux changes or it can be a follow up PR as well.

tommylau523:
Another follow-up is likely all the project drop-downs in each route. So we can only show projects with the feature. That one is certainly going to be a follow-up.

src/services/OrganizationService.ts
```diff
@@ -300,15 +346,16 @@ const getAllOrganizationsInternalTags = async (): Promise<OrganizationsInternalT
...
-  initializeTimeZone,
+  getAllOrganizationsInternalTags,
+  getOrganizationFeatures,
   getOrganizationInternalTags,
+  getOrganizationRoles,
+  getOrganizations,
+  initializeTimeZone,
+  updateOrganization,
   updateOrganizationInternalTags,
-  getAllOrganizationsInternalTags,
```
joshbuchea, line 311: Aside from adding `getOrganizationFeatures`, only change here was to sort exports

src/scenes/OrgRouter/NavBar.tsx
tommylau523: New org feature endpoint will only be allowed for Manager or higher.  We should add a check for that before dispatching/making an API call.

tommylau523: `currentParticipantProject` seems unnecessary.

PR #3851, opened 2025-04-17
SW-6400 follow-up When changing orgs, planting site data is not updated correctly

constanzauanini:


src/redux/features/tracking/trackingThunks.ts
```diff
@@ -29,19 +29,31 @@ export const requestPlantingSite = (plantingSiteId: number, locale?: string | nu
...
+  async (organizationId: number, { dispatch, getState, rejectWithValue, fulfillWithValue }) => {
     try {
-      const response = await TrackingService.listPlantingSites(organizationId, true, locale);
-      const { error, sites } = response;
-      dispatch(setPlantingSitesAction({ error, plantingSites: sites }));
+      const existingRequest = (getState() as RootState).tracking;
+
+      if (['success'].includes(existingRequest?.data?.status)) {
+        return fulfillWithValue(existingRequest?.data?.data?.plantingSites);
+      }
```
ice1080, line 40: Would we also want to handle a case of status=pending, so that we don't send duplicate requests?
Or is that not possible because we're doing an `await` below?

src/redux/features/observations/observationsThunks.ts
ice1080, line 1: Same question on this one as trackingThunks

src/redux/features/tracking/trackingThunks.ts
constanzauanini: Would that case be just a return?

`      if (['pending'].includes(existingRequest?.status)) {
        return;
      }`

I think that is possible to avoid duplicate requests

ice1080: I _think_ so? But I'm not completely sure. Some of that is logic that is buried in the `createAsyncThunk` definition, and isn't super clear to me. It might not hurt to add that logic, even if it never gets used

constanzauanini: Ok, yes, I will add it just in case

ice1080 (approved):

PR #3862, opened 2025-04-22
SW-6761: Reports page should not be accessible by Contributors & Managers (Follow-ups)

joshbuchea:
This PR includes follow-up changes to enable Reports for Org admins.

### DIFF
--- a/src/scenes/OrgRouter/NavBar.tsx
+++ b/src/scenes/OrgRouter/NavBar.tsx
@@ -184,7 +184,10 @@ export default function NavBar({
 
   const reportsMenu = useMemo<JSX.Element | null>(
     () =>
-      isReportsEnabled && isAllowed('READ_REPORTS') && !!orgFeatures?.data?.reports?.enabled && activeLocale ? (
+      isReportsEnabled &&
+      isAllowed('READ_REPORTS', { organization: selectedOrganization }) &&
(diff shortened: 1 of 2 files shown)


src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -184,7 +184,10 @@ export default function NavBar({
 
   const reportsMenu = useMemo<JSX.Element | null>(
     () =>
-      isReportsEnabled && isAllowed('READ_REPORTS') && !!orgFeatures?.data?.reports?.enabled && activeLocale ? (
+      isReportsEnabled &&
```
tommylau523, line 187: This will be functionally correct I think, but we can potentially run into some issues like if selectedOrganization and orgFeatures don't actually refer to the same org. One fix I can think of is to put orgFeatures into the `OrganizationProvider` as well.

tommylau523 (approved):
Approved since this is correct! But I think there is an assumption we made that is not explicitly clear.

tommylau523:
> Approved since this is correct! But I think there is an assumption we made that is not explicitly clear. 

Specifically, this is correct because our `orgFeatures` state actually depends on `selectedOrganization` when we call the dispatch. But this is an assumption we do have to try and maintain every where. We can also fold `orgFeatures` into the same provider as well.

tommylau523 (approved):

PR #3867, opened 2025-04-23
SW-6818 Add query param to view one report

constanzauanini:


src/scenes/FunderReport/FunderReportView.tsx
```diff
@@ -46,9 +48,14 @@ const FunderReportView = () => {
 
   useEffect(() => {
     if (!selectedReport && reports?.length) {
-      setSelectedReport(reports[0]);
+      if (query.get('reportId')) {
+        const found = reports?.find((r) => r.reportId.toString() === query.get('reportId'));
+        setSelectedReport(found || reports[0]);
+      } else {
+        setSelectedReport(reports[0]);
+      }
```
ice1080, line 56: Would this mean that if the query params had `reportId=123` and then they changed the current report to 456, that the param will still say 123?
Not saying we have to change it, just curious if that's how this would work

ice1080: Alternatively, instead of keeping the query param in sync, we could also just grab the reportId from the query params if it exists and then delete it after

PR #3872, opened 2025-04-24
SW-6809 Don't call features api for contributors

constanzauanini:


ice1080 (approved):
Nice, that should fix some server errors

PR #3875, opened 2025-04-24
Fix node warning in console: `util._extend` API is deprecated

joshbuchea:
This PR uses `overrides` in `package.json` to use the commit from the [PR that was submitted to the http-proxy package](https://github.com/http-party/node-http-proxy/pull/1666).

Summary:

- Root issue is a deprecated Node API (`util._extend`) is being used by a dependency that hasn't been updated in multiple years ([http-proxy](https://github.com/http-party/node-http-proxy/))
- A [PR was submitted to the http-proxy package](https://github.com/http-party/node-http-proxy/pull/1666) but not been merged yet
- [Open issue in http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware/issues/1017) discusses the issue and links to the PR in `http-proxy`

PR #3881, opened 2025-04-24
Fix node warning in console: `util._extend` API is deprecated (Follow-up)

joshbuchea:
This PR reverts the change made in the [previous PR for this issue](https://github.com/terraware/terraware-web/pull/3875) since it didn't actually silence the node warning and instead directly patches the `http-proxy` package.


ice1080 (approved):
Nice, even cleaner than the other one and fixes all the renovate stuff ha.

PR #3885, opened 2025-04-25
SW-6821 Add validation to subset count by count

constanzauanini:
Previously the validation was added only when the quantity was set by weight. Added to both cases and also add validation for negative numbers in subset count as in subset weight field

### DIFF
--- a/src/scenes/AccessionsRouter/edit/QuantityModal.tsx
+++ b/src/scenes/AccessionsRouter/edit/QuantityModal.tsx
@@ -104,8 +104,8 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
       setRemainingQuantityNotesError(false);
     }
 
-    if (quantityChanged && isByWeight) {
-      if (!record.subsetWeight) {
+    if (isByWeight || isSubsetOpen) {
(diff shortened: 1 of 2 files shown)


src/scenes/AccessionsRouter/edit/QuantityModal.tsx
```diff
@@ -114,14 +114,16 @@ export default function QuantityModal(props: QuantityModalProps): JSX.Element {
...
-      setSubsetWeightError('');
     }
 
-    if (quantityChanged && isByWeight && !record.subsetCount) {
-      setSubsetCountError(strings.REQUIRED_FIELD);
-    } else {
-      setSubsetCountError('');
+    if (isByWeight || isSubsetOpen) {
+      if (!record.subsetCount) {
+        setSubsetCountError(strings.REQUIRED_FIELD);
```
ice1080, line 121: This is the same initial if-check as above, should it just be added to the same conditional?

constanzauanini: I was removing onditions and didn't realize about that. Thanks!

ice1080 (approved):

PR #3888, opened 2025-04-26
SW-6812: Add Deliverables tab to Project Profile in Console

joshbuchea:
This PR adds the "Deliverables" tab in the new Project Profile.

## Screenshots

localhost_accelerator_projects_52_tab=projectDeliverables


src/components/DeliverablesTable/DeliverableCellRenderer.tsx
```diff
@@ -19,7 +19,7 @@ export default function DeliverableCellRenderer(props: RendererProps<TableRowTyp
     const deliverableUrl = isAcceleratorRoute ? APP_PATHS.ACCELERATOR_DELIVERABLE_VIEW : APP_PATHS.DELIVERABLE_VIEW;
     const to = deliverableUrl.replace(':deliverableId', `${row.id}`).replace(':projectId', `${row.projectId}`);
     return (
-      <Link to={to}>
+      <Link to={`${to}?source=${window.location.pathname}`}>
```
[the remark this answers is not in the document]
joshbuchea: This would be helpful if the possible URL values were unknown, but in this case we know the values in advance and they don't include query params.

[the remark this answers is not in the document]
joshbuchea: Although it's unnecessary now, I suppose it future-proofs the code and prevents a potential regression if the URLs are updated to include query params.

src/scenes/AcceleratorRouter/Deliverables/DeliverableView.tsx
```diff
@@ -163,15 +179,25 @@ const DeliverableView = () => {
...
-    [activeLocale]
-  );
+  const crumbs: Crumb[] = useMemo(() => {
+    if (!activeLocale) {
+      return [];
+    }
+
+    return source?.startsWith('/accelerator/projects/')
+      ? [
+          {
```
ice1080, line 189: I'm wondering if there's a better way we can do this in case the path ever changes, but I'm not thinking of anything right now 🤔

src/scenes/AcceleratorRouter/ParticipantProjects/ProjectDeliverablesView.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  return (
+    <Card
+      style={{
+        borderRadius: theme.spacing(1),
+        padding: 0,
+      }}
+    >
+      <Box marginX={`-${theme.spacing(4)}`}>
+        <DeliverablesList projectId={projectId} />
```
ice1080, line 23: Is it common for us to use negative margins?

src/scenes/AcceleratorRouter/Deliverables/DeliverableView.tsx
joshbuchea: Yeah I didn't love this either... Just thought of a simple method to make this more resilient, will fix.

src/scenes/AcceleratorRouter/ParticipantProjects/ProjectDeliverablesView.tsx
joshbuchea: There's inconsistent margin/padding throughout the app, often applied in components when it should be on the parent. Specifically, the `Page` and `Tabs` components have padding, and removing or making it optional will require adjustments wherever these components are used.

Wasn't sure if I should tackle it in this PR, but I can.

src/scenes/AcceleratorRouter/Deliverables/DeliverableView.tsx
```diff
@@ -163,15 +179,25 @@ const DeliverableView = () => {
...
-    ],
-    [activeLocale]
-  );
+  const crumbs: Crumb[] = useMemo(() => {
+    if (!activeLocale) {
+      return [];
+    }
+
+    return source?.startsWith(APP_PATHS.ACCELERATOR_PROJECT_VIEW.replace(':projectId', ''))
+      ? [
```
ice1080, line 188: Approving, though will mention you could also add a new constant for this, like `ACCELERATOR_PROJECT_BASE` so that you don't have to do a replace here

ice1080 (approved):

src/scenes/AcceleratorRouter/Deliverables/DeliverableView.tsx
joshbuchea: Good suggestion, will update.

joshbuchea: I opened a [PR in web-components](https://github.com/terraware/web-components/pull/661) that modifies the `Tabs` component to allow customizing tab styles. I'll remove this negative margin in a follow-up PR.

PR #3893, opened 2025-04-28
Remove unnecessary supported locales to numberFormatter

ice1080:
The `useNumberFormatter` hook added `supportedLocales` inside it a while back, but some usages of `numberFormatter` were still passing in `supportedLocales`. The function doesn't use the passed in value anyway, so this PR removes it.


constanzauanini (approved):
Nice! Thanks for cleaning this out

PR #3919, opened 2025-04-30
SW-6706  Surface Plot Location coordinates in Detail screens

constanzauanini:


src/scenes/ObservationsRouter/adhoc/AdHocObservationDetails.tsx
```diff
@@ -76,6 +76,9 @@ export default function AdHocObservationDetails(props: AdHocObservationDetailsPr
   const data: Record<string, any>[] = useMemo(() => {
     const handleMissingData = (num?: number) => (!monitoringPlot?.completedTime && !num ? '' : num);
 
+    const swCoordinatesLat = monitoringPlot?.boundary?.coordinates?.[0]?.[0]?.[0];
```
constanzauanini, line 79: I don't know the South West coordinate index in the boundary coordinates array. @tommylau523 @sgrimm

src/scenes/ObservationsRouter/adhoc/index.tsx
```diff
@@ -97,6 +100,11 @@ export default function ObservationMonitoringPlot(): JSX.Element {
         value: monitoringPlot?.conditions.map((condition) => getConditionString(condition)).join(', ') || '- -',
       },
       { label: strings.FIELD_NOTES, value: monitoringPlot?.notes || '- -', text: true },
+      {
+        label: strings.PLOT_LOCATION,
+        value: `${String(strings.formatString(String(strings.SW_CORNER_LATITUDE), String(swCoordinatesLat)))}\n${String(strings.formatString(String(strings.SW_CORNER_LONGITUDE), String(swCoordinatesLong)))}`,
+        text: true,
```
ice1080, line 106: Just curious, did you add the `String()` around all the `strings.<bla>` to fix linting issues?

constanzauanini: Yes, it was a lint issue that was avoiding me: `Invalid type "string | string[]" of template literal expression`

ice1080: I noticed the same in mine. I think I'll try to talk about linting in roundtable. I'd like to find a better way to handle some of this than changing how we'd normally write our code just to fix linting issues

src/scenes/ObservationsRouter/adhoc/AdHocObservationDetails.tsx
tommylau523: This looks correct

sgrimm: When we generate plots on the server side, the coordinates are always the same order: SW, SE, NE, NW. The server depends on this in a few places too, so it's safe to rely on.

src/scenes/ObservationsRouter/adhoc/index.tsx
sgrimm: I ran into this when I was adding the i18n logic initially. The problem is that `strings.formatString()` can take non-string values for placeholders, like `strings.formatString('hello {0}', <span>world</span>)`, and it returns an array in that case.

In retrospect, rather than using `as string` I should probably have added a wrapper that only accepts and returns strings.

src/scenes/ObservationsRouter/adhoc/AdHocObservationDetails.tsx
constanzauanini: Thank you!

src/scenes/ObservationsRouter/adhoc/index.tsx
tommylau523: We can add this wrapper in our `strings/index.tsx` file. afaik we never ever format String with a JSX value.

ice1080: Just did a quick check, and looks like we unfortunately do. First example that turned up in my search was `PlantsDashboardView.tsx` line 197, which calls `getLatestObservationLink()` which returns a `<Link>` element

tommylau523: I wonder if it still makes sense to make two different wrappers. One can return a string, another can return a JSX element. We can make different calls as required.

tommylau523: Or we can always return a JSX fragment and wrap our string in `<>`. We do lose some type safety like that

ice1080 (approved):

src/scenes/ObservationsRouter/adhoc/index.tsx
constanzauanini: I will merge this PR since the new wrapper change is not related. We can do that on a separate PR

PR #3930, opened 2025-05-02
SW-6860 Report data provided by Partner and TF user not appearing in published Funder Report

constanzauanini:


src/components/AcceleratorReports/ChallengesMitigationBox.tsx
```diff
@@ -133,6 +133,10 @@ const ChallengesMitigationBox = (props: ReportBoxProps) => {
   const updateReportResponse = useAppSelector(selectReviewAcceleratorReport(requestId));
   const snackbar = useSnackbar();
 
+  useEffect(() => {
+    setChallengeMitigations(report?.challenges || []);
```
constanzauanini, line 137: @ice1080 this is similar to line 149. But without this funder report chllanges are not updated correctly, because getNonEmptyChallenges are not updated. Maybe you have a better idea on how to fix this?

constanzauanini: Maybe we can remove that useEffect and just use this one?

ice1080: Hmm ya I'm not sure the best way to do this. The problem is line 170, where modifying `challengeMitigations` calls `onChange` which modifies `report?.challenges`, causing an infinite loop (only in participant edit view)

src/components/AcceleratorReports/ChallengesMitigationBox.tsx
```diff
@@ -143,11 +143,14 @@ const ChallengesMitigationBox = (props: ReportBoxProps) => {
...
   useEffect(() => {
+    if (!editing) {
+      setChallengeMitigations(report?.challenges || []);
+    }
     // For participant editing, react can't keep up with setting challengeMitigations, then calling onChange on the
     // report, and having this useEffect update challengeMitigations again. This check ensures we're only setting it
     // from report when needed
-    if ((!getNonEmptyChallenges() || getNonEmptyChallenges().length === 0) && report?.challenges) {
-      setChallengeMitigations(report.challenges || []);
+    if (((editing && !getNonEmptyChallenges()) || getNonEmptyChallenges().length === 0) && report?.challenges) {
```
constanzauanini, line 152: I added this here and seems to be working. Let me know what you think @ice1080

ice1080: That seems like a good option!

ice1080 (approved):

PR #3933, opened 2025-05-02
Create PhotoDragDrop component in prep for new selector, update description text

ice1080:


src/components/Photo/PhotoDragDrop.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }
+    if (event.currentTarget.files) {
+      addFiles(event.currentTarget.files);
+    }
+  };
+
+  const onChooseFileHandler = () => {
+    inputRef.current?.click();
+    document.querySelectorAll<HTMLElement>('.photo-button').forEach((el) => el.blur());
+  };
```
ice1080, line 72: I couldn't think of a better way to do this, so open to suggestions. Previously we were calling `divRef.current?.focus();` which was just some random other element, that unfortunately doesn't exist in this smaller component. I tried with several other components in this one and couldn't get it to work (including adding a `ref={ref}` to the Button element in web-components).

constanzauanini: I don't remember why we needed that blur

constanzauanini (approved):

src/components/Photo/PhotoDragDrop.tsx
ice1080: I'm guessing it's just because the button has a focus border around it after selecting, and it shouldn't really anymore?

ice1080:
### Merge activity

* **May 5, 11:19 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/3933).

PR #3956, opened 2025-05-06
SW-6832 Use -2 id for rolled up view

constanzauanini:


src/components/PlantsPrimaryPage/PlantsPrimaryPageView.tsx
```diff
@@ -25,7 +25,7 @@ export type PlantsPrimaryPageViewProps = {
   actionButton?: ButtonProps;
   children: React.ReactNode; // primary content for this page
   isEmptyState?: boolean; // optional boolean to indicate this is an empty state view
-  onSelect: (plantingSite: PlantingSite) => void; // planting site selected, id of -1 refers to All
+  onSelect: (plantingSite: PlantingSite | undefined) => void; // planting site selected, id of -1 refers to All
   plantingSites: PlantingSite[] | undefined;
```
ice1080, line 29: I'm still not super great with typescript. What are the benefits of doing `plantingSite: PlantingSite | undefined` instead of just `plantingSite?: PlantingSite`?

src/components/PlantsPrimaryPage/PlantsPrimaryPageView.tsx
```diff
@@ -54,17 +54,30 @@ export default function PlantsPrimaryPageView({
...
 
-  const options = useMemo(
-    () => plantingSites?.map((site) => ({ label: site.name, value: site.id })) ?? [],
-    [plantingSites]
-  );
+  const options = useMemo(() => {
+    const optionsToReturn = plantingSites?.map((site) => ({ label: site.name, value: site.id })) ?? [];
+    if (projectId) {
+      optionsToReturn.unshift({ label: strings.ALL_PLANTING_SITES, value: -2 });
+    }
```
ice1080, line 74: Should this only `unshift` if `optionsToReturn.length !== 1`?

constanzauanini: Yes, if we only have one site we shouldn't add the rolled up view. I was not including that condition yet, but I can add it

src/components/PlantsPrimaryPage/PlantsPrimaryPageView.tsx
constanzauanini: No benefit here, I guess. I was getting an error of undefined not being an option in that function and just added like that. But I will add it as `plantingSite?: PlantingSite` since that will also allow nullable (null) options

ice1080 (approved):

PR #3966, opened 2025-05-08
SW-6832 Total plants and species for rolled up view

constanzauanini:


src/scenes/PlantsDashboardRouter/components/PlantsAndSpeciesCard.tsx
```diff
@@ -17,29 +22,65 @@ import PlantsReportedPerSpeciesCard from './PlantsReportedPerSpeciesCard';
...
 export default function PlantsAndSpeciesCard({
   plantingSiteId,
   hasReportedPlants,
+  organizationId,
+  projectId,
 }: PlantsAndSpeciesCardProps): JSX.Element {
   const siteReportedPlants = useAppSelector((state) => selectSiteReportedPlants(state, plantingSiteId));
   const theme = useTheme();
   const plantings = useAppSelector((state) => selectPlantingsForSite(state, plantingSiteId));
+  const allPlantings = useAppSelector((state) => selectPlantings(state));
```
ice1080, line 38: What is the benefit of passing `state` in here explicitly? I thought that already does that by default?

ice1080 (approved):

src/scenes/PlantsDashboardRouter/components/PlantsAndSpeciesCard.tsx
constanzauanini: No benefit, just copied from the other selector that was passing the plantingSiteId. Will fix it

PR #3968, opened 2025-05-08
SW-6893 Add Funder Portal Project Profile view

ice1080:
Refactor ProjectProfile view to use `projectDetails` prop which has common elements of both `ParticipantProject` and `FunderProjectDetails`.

Add Project Profile view to funder portal.

Can't be merged until after [the server code](https://github.com/terraware/terraware-server/pull/3096) is merged

### DIFF
--- a/src/redux/features/funder/projects/funderProjectsSelectors.ts
+++ b/src/redux/features/funder/projects/funderProjectsSelectors.ts
@@ -1,0 +1,3 @@
+import { RootState } from 'src/redux/rootReducer';
+
+export const selectFunderProjectRequest = (projectId: number) => (state: RootState) => state.funderProject[projectId];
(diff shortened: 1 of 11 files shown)


src/redux/features/funder/projects/funderProjectsSelectors.ts
```diff
@@ -0,0 +1,4 @@
+import { RootState } from 'src/redux/rootReducer';
+
+export const selectFunderProjectRequest = (projectId?: number) => (state: RootState) =>
```
constanzauanini, line 3: Not sure if it has any advantage or not, but I think we always make the param here (in the selector) required, and where the selector is called we pass -1 if there is no selection (no projectId selected in this case)

constanzauanini (approved):

ice1080:
### Merge activity

* **May 14, 1:09 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/3968).

PR #3973, opened 2025-05-09
SW-6807: Project Org Admin should be able to edit metric targets (Follow-up #2)

joshbuchea:
This PR removes unnecessary data from requests when updating report targets.

### DIFF
--- a/src/components/AcceleratorReports/EditAcceleratorReportTargetsModal.tsx
+++ b/src/components/AcceleratorReports/EditAcceleratorReportTargetsModal.tsx
@@ -140,41 +140,17 @@ export default function EditAcceleratorReportTargetsModal({
         const reportClone = _.cloneDeep(report);
 
         ['projectMetrics', 'standardMetrics', 'systemMetrics'].forEach((metricType) => {
+          const reportMetrics = reportClone?.[metricType as keyof AcceleratorReport] as (
+            | ReportProjectMetricEntries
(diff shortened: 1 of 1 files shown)


src/components/AcceleratorReports/EditAcceleratorReportTargetsModal.tsx
```diff
@@ -140,41 +140,17 @@ export default function EditAcceleratorReportTargetsModal({
...
-                : reportMetrics.findIndex(
-                    (m) => (m as ReportStandardMetricEntries).id === (metric as ReportStandardMetricEntries).id
-                  );
-            const reportMetric = reportMetrics[reportMetricIndex];
-            if (typeof reportMetric !== 'object' || typeof metric !== 'object') {
-              return;
-            }
-
-            const reportMetricUpdate = {
-              ...reportMetric,
```
constanzauanini, line 168: Was this extra data the cause of org admins not being able to edit?

joshbuchea: Sorry, the description was pretty vague. When the ticket was originally created it was because the edit targets modal wasn't accessible for certain user roles/types, but there have now been a couple of follow-ups related to errors when trying to save.

The extra data was causing errors when saving in some cases.

constanzauanini (approved):

PR #3974, opened 2025-05-09
Rolled up view - Map and species charts

constanzauanini:
This needs an API for getting all populations of an org, so making it draft until that


src/scenes/PlantsDashboardRouter/components/MultiplePlantingSiteMap.tsx
```diff
@@ -0,0 +1,61 @@
...
+  organizationId,
+  hideAllControls,
+  style,
+}: MultiplePlantingSiteMapProps): JSX.Element {
+  const { activeLocale } = useLocalization();
+  const [plantingSites, setPlantingSites] = useState<PlantingSite[]>();
+
+  useEffect(() => {
+    const populatePlantingSites = async () => {
+      const serverResponse = await TrackingService.listPlantingSites(organizationId, undefined, activeLocale);
```
constanzauanini, line 28: Planting sites should be get from new planting site provider, will modify when that is ready

constanzauanini:
Adding map and charts in https://github.com/terraware/terraware-web/pull/4003

PR #3976, opened 2025-05-09
SW-6840 Export observation results

sgrimm:
Add the ability to export the details of an observation, including
some subtotals that aren't shown in the UI at the moment. This data
can be used to generate statistics about observations beyond what
Terraware already calculates.

The export generates a zipfile containing two CSV files, one with
data rolled up per monitoring plot and the other with per-plot,
per-species data.

Unlike other CSV exports in the app, the files are generated locally
in the client code rather than downloaded from the server. This is
mostly because some of the columns in the CSVs (e.g., the link to the
plot details page) would be awkward to calculate using the search
API.


sgrimm:
* **#3976** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3976?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/scenes/ObservationsRouter/org/OrgObservationsRenderer.tsx
```diff
@@ -21,11 +21,13 @@ const OrgObservationsRenderer =
     goToRescheduleObservation: (observationId: number) => void,
     exportObservationCsv: (observationId: number) => void,
     exportObservationGpx: (observationId: number) => void,
+    exportObservationResults: (observationId: number) => void,
     openEndObservationModal: (observationId: Observation) => void
   ) =>
   // eslint-disable-next-line react/display-name
   (props: RendererProps<TableRowType>): JSX.Element => {
     const { column, row, value } = props;
+    const observationId = Number(row.observationId);
```
ice1080, line 30: what happens if `Number(row.observationId)` returns NaN? Is that a concern here?

sgrimm: I don't think that should be possible in practice, but I'll add some code to guard against it.

ice1080 (approved):

PR #3982, opened 2025-05-12
Disable template expression rules

ice1080:
The rules `no-base-to-string` and `restrict-template-expressions` are disabled in `package.json`'s `eslintConfig` property, so disabling them in eslint's config too. This brings the failure count from ~181 to ~142

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -206,6 +206,9 @@ export default fixupConfigRules([
       '@typescript-eslint/no-unsafe-call': 'off',
       '@typescript-eslint/no-unsafe-member-access': 'off',
 
+      '@typescript-eslint/no-base-to-string': 'off',
+      '@typescript-eslint/restrict-template-expressions': 'off',
+
       'arrow-parens': ['error', 'always'],
       'brace-style': ['off', 'off'],
       complexity: 'off',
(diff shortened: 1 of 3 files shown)


ice1080:
* **#3982** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3982?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3982?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

tommylau523:
I think these are potentially helpful rules for us to have. We do want to make sure our `toString()` calls are safe. But it does depend on us using less of `unknown` and `any` types first.

ice1080:
That's fine with me, I'm happy to close this PR if we're in favor of keeping these rules!

ice1080:
@tommylau523 are you thinking we should keep both or just the `no-base-to-string`?

tommylau523:
I think `restrict-template-expressions` is actually more valuable since it prevents "[object Object]"

`no-base-to-string` is probably less applicable to us since we almost never override the `toString()` function afaik. 

No strong opinion on keeping these enabled/disabled either way. My approach would be to disable both of these rules for now, and once we are able to get our types to be stronger (no anys/unknowns), we can look into re-enabling them. But really not a strong opinion at all!!!

ice1080:
> My approach would be to disable both of these rules for now, and once we are able to get our types to be stronger (no anys/unknowns), we can look into re-enabling them. But really not a strong opinion at all!!!

I like the idea of eventually getting to that, but since there are several hundred failures related to no anys/unknowns and the recent consensus was to disable all of these, I don't see that actually happening anytime soon. I'll close this PR and publish a different one that fixes several of these and disables a few of them

ice1080:
This PR to be replaced by #3987

PR #3986, opened 2025-05-13
SW-6807: Project Org Admin should be able to edit metric targets (Follow-up #3)

joshbuchea:
This PR includes the following changes:

- Determine `disabled` status for each report target input field and memoize the values
- Revert [change in previous PR](https://github.com/terraware/terraware-web/pull/3973) to exclude unchanged metric data

### DIFF
--- a/src/components/AcceleratorReports/EditAcceleratorReportTargetsModal.tsx
+++ b/src/components/AcceleratorReports/EditAcceleratorReportTargetsModal.tsx
@@ -151,7 +151,26 @@ export default function EditAcceleratorReportTargetsModal({
             return;
           }
 
-          reportMetrics.splice(0, reportMetrics.length, ...requestMetrics);
+          requestMetrics.forEach((metric) => {
+            const reportMetricIndex =
+              metricType === 'systemMetrics'
+                ? reportMetrics.findIndex(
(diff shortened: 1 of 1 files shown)


src/components/AcceleratorReports/EditAcceleratorReportTargetsModal.tsx
```diff
@@ -167,6 +187,29 @@ export default function EditAcceleratorReportTargetsModal({
     }
   };
 
+  const disabledFields = useMemo(() => {
+    const isFieldDisabled = (reportId?: number) => {
```
tommylau523, line 191: We can remove a lot of boilerplating here. 

`isFieldDisabled` can be a `useCallback` at the top level. 

`disabledFields` can be removed and calculated at render time instead. 

Below, when rendering `TextFields`, we can use `disabled={isFieldDisabled(record. q1ReportId)}`, `disabled={isFieldDisabled(record. q2ReportId)}` etc.

tommylau523:
```
isFieldDisabled = useCallback((reportId: number) => {
    const report = reports.find((report) => report.id === reportId)
    if (!report) {
        return true;
    } 
    return report.status === 'Not Submitted' || 'Needs Update';
}, [reports])
```

tommylau523 (approved):

PR #3990, opened 2025-05-13
SW-6807: Project Org Admin should be able to edit metric targets (Follow-up #4)

joshbuchea:
This PR includes a follow-up change to only send requests for modified reports.

Most of the diff is indentation, the only change is the addition of the `.filter` and if/else clause to handle closing the modal when there are no changes.


ice1080 (approved):

joshbuchea:
@ice1080 thanks for the review 🙌

I noticed the reports in the modal weren't updating after changes, so I fixed that and replaced instances of `.find` with `reportsById`.

ice1080 (approved):
Nice, looks good!

PR #3992, opened 2025-05-14
Fix some lint issues

alexTerraformation:


src/scenes/AcceleratorRouter/Documents/PreviewView/util/markdown-table.ts
```diff
@@ -7,6 +7,7 @@ import {
 } from 'src/types/documentProducer/VariableValue';
 
 export const hasMarkdownTableHeaderSeparatorRow = (input: string | undefined): boolean =>
+  /* eslint-disable-next-line no-useless-escape */
   !!/\|\s*\-\-\-+\s*\|/g.exec(input || '');
```
ice1080, line 11: I think for this one we can actually change `\-\-\-` to just `---`. In regex, the `-` character doesn't need to be escaped if it's not inside square brackets `[]`. I tested locally and the tests still pass

src/providers/Application/Context.tsx
```diff
@@ -18,9 +18,9 @@ export const ApplicationContext = createContext<ApplicationData>({
   applicationSections: [],
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   getApplicationByProjectId: (projectId: number) => undefined,
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
   setSelectedApplication: (applicationId: number) => {},
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
   reload: (onReload?: () => void) => {},
```
ice1080, line 24: Another option for these is something like this:
```ts
{
  // no-op
  return;
}
```
But I don't think that it necessarily provides benefit over just disabling it for one line

ice1080: Other alternative is disabling and enabling similar to `ParticipantProjectSpeciesContext.tsx`

src/scenes/AcceleratorRouter/Documents/PreviewView/util/markdown-table.ts
alexTerraformation: Yeah, I considered that.  But since character classes could cause `-` to need an escape I opted against it and it doesn't hurt anything.

ice1080 (approved):

PR #3997, opened 2025-05-14
SW-6893 Cleanup Funder Project Profile, strip country from deal name

ice1080:


ice1080:
* **#3997** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3997?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/3997?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/scenes/AcceleratorRouter/ParticipantProjects/ProjectProfileView.tsx
```diff
@@ -145,6 +145,14 @@ const ProjectProfileView = ({
     [funderView, lastPublishedReport, lastSubmittedReport]
   );
 
+  const strippedDealName = useMemo(() => {
+    if (projectDetails?.dealName?.startsWith(`${projectDetails?.countryAlpha3}_`)) {
```
tommylau523, line 149: projects can have XXX_ if there is no country selected. I think it may be more straightforward to match `^[A-Z]{3}_` and remove that.

tommylau523: But taking a step back, I am hoping to have country code removed from deal name entirely in the database. I think we will hear back from Alex/product today.

ice1080: I went ahead and changed to regex instead for now; we can cleanup later if we update those requirements.

As an aside, if we remove country code from deal names deal names, won't that mean that all of those projects will have a deal name that is just the organization name? And won't that mean that orgs with multiple projects might change to having all projects with the same deal name?

tommylau523: Yes I think part of this change will be to enforce constraints we need to have on deal name. 

Even now, an organization can submit multiple applications and they will end up with the same deal name, with the exact same country code. I think when it was a column in the applications table only, we added a `_{number}` suffix. But this became difficult when deal names are editable by accelerator admins and got updated to be a variable.

constanzauanini (approved):

PR #4002, opened 2025-05-14
SW-6807: Project Org Admin should be able to edit metric targets (Follow-up #5 🫠)

joshbuchea:
This PR fixes an issue where requests were sent for unchanged targets with no previous value.

If `target` was `undefined` before and after edits, it was omitted in the API response but still set in the report copy, causing the `isEqual` check to return false despite both copies having `undefined` for `target`.


tommylau523 (approved):
This works! Thanks for the fix. Hopefully once we have the merge metric target API, this all will look a lot simpler!

ice1080 (approved):

PR #4021, opened 2025-05-19
SW-6932 Add loading spinner to plants dashboard

constanzauanini:


tommylau523 (approved):
Looks great! This actually prevents the charts from re-rendering too during reloads which fixes SW-6836.

PR #4038, opened 2025-05-20
SW-6249 Fix accessions back button

ice1080:
Fix browser back button on View Accession page by adding an optional `replace: true` to `useStickyTabs`


constanzauanini:
This is working better now, but when I was testing this a while ago, I found that you can't go back to previous pages after going to an accession details (for example if you go to species, then go to accessions and then to an accession details; If you click back after back in accessions page, you can't go to species).

Here is a recording of what I tested:

https://github.com/user-attachments/assets/0a2bf29e-eed1-41dd-bf20-bcb12967cfd5

ice1080:
Ya that's what @alexTerraformation noticed as well. I pointed out in jira that we have several other pages that do that as well, and it should probably be investigated separately from this as I think they are unrelated issues.

constanzauanini:
> Ya that's what @alexTerraformation noticed as well. I pointed out in jira that we have several other pages that do that as well, and it should probably be investigated separately from this as I think they are unrelated issues.

Oh, I didn't see the Jira comment. Ok, sound goods!

constanzauanini (approved):

PR #4043, opened 2025-05-20
SW-6932 Improve spinners on dashboard

constanzauanini:


src/providers/Tracking/PlantingSiteContext.tsx
```diff
@@ -27,6 +27,7 @@ export type PlantingSiteData = {
   latestResult?: ObservationResultsPayload;
 
   isLoading: boolean;
+  isPlantingSiteSet: boolean;
```
ice1080, line 30: What's the difference between `isPlantingSiteSet` and just checking `isLoading`?

constanzauanini: So I did this because the plantsDashboards loads with no planting site selected at the begging, and isLoading returns false, then a planting site is selected (automatically from the planting sites list) and the isLoading goes back to true, and that make the dashboard to show for some miliseconds and then show the spinner again, and then show the dashboard again when the isLoading finished. 

With the isPlantingSiteSet the spinner shows one time only. Not sure if it the best approach though.

ice1080: I haven't spent much time looking at the provider, but is there any way to change when we set `isLoading` to false so that it doesn't do that? There might not be, just wanted to check

tommylau523: We can keep `isLoading` for all the single planting site operations by removing `plantingSitesResults`. And then keep a separate `isInitiated` to note if all planting sites were loaded. 

I also think if we keep this, `isPlantingSiteSet` should live in the `PlantDashbaordView` or the `PlantsPrimaryPageView`, since this provider is used for other observations/planting site routes as well and that can get confusing.

constanzauanini: That is true. I will try with your suggestion Tommy

ice1080 (approved):
Looks good to me, not sure if you want @tommylau523 to also review since he has more context with plants dashboard

src/scenes/PlantsDashboardRouter/PlantsDashboardView.tsx
```diff
@@ -399,8 +399,7 @@ export default function PlantsDashboardView({
       projectId={projectId}
       onSelectProjectId={(newProjectId: number) => setProjectId(newProjectId === -1 ? undefined : newProjectId)}
       organizationId={organizationId}
-      isEmptyState={isLoading ? false : plantingSite === undefined}
-      isLoading={isLoading}
+      isEmptyState={isLoading ? false : (allPlantingSites?.length ?? 0) <= 1}
```
tommylau523, line 402: should this use `isInitiated`?

constanzauanini: Not necessarily. The component internally is using isInitiated to show the empty state, but here we just need to know if there are more than one planting site (not necessary a planting site set)

tommylau523: Makes sense!

PR #4046, opened 2025-05-21
SW-6936 Separate planting site tables and draft planting site tables, and show only current info

tommylau523:
This ends up being a massive PR because the types are very entangled which had a lot of complexity as a result. It was also starting to break as planting sites and draft planting sites diverge in data. So key changes below:
1) Fetch latest observations + monitoring plot data from the BE.
2) Replaced not functioning selectors for planting site selectors with planting site provider in the planting site details table
3) Separated out draft planting site data into a separate table and code path, same as zone data. Removed unneeded columns and observation selections for draft planting sites since they will never have results
4) Removed selectors that are no longer used
5) Removed complex aggregate types


tommylau523:
* **#4046** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4046?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4046?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "license": "Apache-2.0",
   "scripts": {
-    "start": "craco start",
+    "start": "craco --max_old_space_size=8192 start",
     "build": "craco --max_old_space_size=8192 build",
```
ice1080, line 8: Won't this negatively affect local development?

tommylau523: Ah didn’t mean to commit this line.

tommylau523: But on this note, I always add this line when starting up locally or else I run into OOM.

tommylau523: I run in in terminal.

src/scenes/PlantingSitesRouter/edit/PlantingSiteCreate.tsx
```diff
@@ -191,15 +176,13 @@ export default function CreatePlantingSite(props: CreatePlantingSiteProps): JSX.
                     setPlantingSeasons={setPlantingSeasons}
                     setRecord={setRecord}
                   />
-                  {record?.plantingZones && (
+                  {plantingSite?.plantingZones && (
                     <BoundariesAndZones
```
ice1080, line 180: Should this one still be record?

ice1080 (commented):
I'm still going through this large PR, but based on the thread in slack, I want to make sure that this is still a valid PR and shouldn't be converted to draft while product figures out what changes will be needed around all of this?

tommylau523:
This is a release blocker so I think it is a valid PR. This only changes the planting site router fwiw.

ice1080:
> This is a release blocker so I think it is a valid PR. This only changes the planting site router fwiw. The uncertain parts, which has to do with what to show in the historic map view under planting sites, and what to do with the subzone table, are both resolved by these components removed: the map date slider, the PlantingSiteSubzoneView.

Sounds good. With the massiveness of this PR and the areas that it touches that I'm not familiar with, I'll do a thorough review of the code itself, but I think @constanzauanini or @joshbuchea should do the actual approval

alexTerraformation:
One thing I noticed from the vercel deployment is that the legend for dfatft planting site maps is in the wrong spot.  Probably a merge issue with @constanzauanini's changes that went in today.
Screenshot 2025-05-21 at 5 11 07 PM

alexTerraformation:
Worth asking at this point too -- if we are removing the monitoring data from the planting site view do we still need this PR for the release?  Feels pretty risky to include a 2000+ line PR change right before a release when we've been battling regressions for 2 weeks.

tommylau523:
I will try to rebase the other PR onto main instead. So we can merge them separately.

tommylau523:
I am running into a ton of types issues because of all the intertwined types used in the planting site router. I can keep on rebasing it but I think that will cause some more issues. Part of what makes the API type changes downstream possible is the removal of the planting site details selector and all the aggregates types. 

Is it worth our effort to resolve all the type conflicts for data types that we don't need/want to replace?

alexTerraformation:
Fair enough.  I didn't realize the other was dependent on this one!

constanzauanini (approved):
Added a comment for the legends of the map, but in general looks correct to me. I can create a separate PR for that style if you prefer

PR #4069, opened 2025-05-23
Map Legend: Observation Events - copy update and tooltip

constanzauanini:


src/components/common/MapLegend/index.tsx
```diff
@@ -74,13 +76,20 @@ export default function MapLegend({ legends, setLegends }: MapLegendProps): JSX.
...
-                fontSize='14px'
+                fontSize='16px'
                 fontWeight={600}
                 width={isMobile ? '100%' : undefined}
                 marginRight={isMobile ? 0 : theme.spacing(4)}
                 paddingLeft={legend.switch ? theme.spacing(1) : 0}
               >
-                {legend.title}
+                {legend.title}{' '}
+                {legend.tooltip && (
```
ice1080, line 86: Is the space necessary after the title? `{' '}`

constanzauanini: We need space, but I removed it and added a margin

src/components/common/PlantingSiteMapLegend.tsx
```diff
@@ -89,6 +89,8 @@ export default function PlantingSiteMapLegend({ options }: PlantingSiteMapLegend
     theme.palette.TwClrBaseBlue300,
     theme.palette.TwClrBasePink300,
     theme.palette.TwClrBaseYellow300,
+    theme.palette.TwClrBaseOrange300,
+    hasAdHocPlot,
```
ice1080, line 93: NIT You can replace all the `theme.palette.` dependencies in this `useMemo` with just `theme`. Just a preference though

ice1080 (approved):
Approving, but not sure if the graphite comment is worth noting or not

constanzauanini:
> Approving, but not sure if the graphite comment is worth noting or not

We have all our tooltips inside a box

> Approving, but not sure if the graphite comment is worth noting or not

I will merge it since all ours tooltips with info icons are like this, but we can try to replace them with a focusable element later. I tested with IconButton real quick but it adds margin and hover effects that I don't think we want by design

PR #4086, opened 2025-05-27
SW-6994 Follow-up Last observation styles in popup

constanzauanini:


tommylau523 (approved):

tommylau523 (commented):
this popup may be used for other purposes like overdue observation plots. Can we make the color a prop or parameter?

PR #4087, opened 2025-05-27
SW-7006 Spinner doesn't go after creating planting site

constanzauanini:
After creating a planting site, we need to reload the allPlantingSites in the provider before redirecting to the planting site page


tommylau523 (approved):
Looks good! One minor deps issue.

PR #4089, opened 2025-05-27
SW-6997 Do not dispatch species request for orgId below 1

tommylau523:


ice1080 (approved):
This will be nice to get rid of the server errors we've been getting :tada:

PR #4092, opened 2025-05-28
SW-7016 Include unknown species in observation CSVs

sgrimm:
By default, observation results don't include "species"-level data for plants of
unknown species, which is fine since we don't want them to appear anywhere in the
UI. However, they need to be included in exported CSV files.

Make the CSV export code fetch a version of the results that includes the unknown
species, and use that to generate the CSVs for export.

### DIFF
--- a/src/scenes/ObservationsRouter/details/exportObservationResults.ts
+++ b/src/scenes/ObservationsRouter/details/exportObservationResults.ts
@@ -3,12 +3,19 @@ import { asBlob, generateCsv, mkConfig } from 'export-to-csv';
 import { AcceptedData, ColumnHeader } from 'export-to-csv/output/lib/types';
 
 import { APP_PATHS } from 'src/constants';
+import { mergeObservations } from 'src/redux/features/observations/utils';
+import ObservationsService from 'src/services/ObservationsService';
 import strings from 'src/strings';
 import { ObservationResults, getPlotStatus } from 'src/types/Observations';
+import { Species } from 'src/types/Species';
+import { PlantingSite } from 'src/types/Tracking';
 import downloadZipFile from 'src/utils/downloadZipFile';
 
 interface UseExportObservationResultsParams {
-  observationResults: ObservationResults;
+  observationId: number;
+  defaultTimeZone: string;
+  plantingSites: PlantingSite[];
+  species: Species[];
 }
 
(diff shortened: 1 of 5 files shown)


sgrimm:
* **#4092** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4092?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4092?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
Depends on https://github.com/terraware/terraware-server/pull/3138

src/scenes/ObservationsRouter/details/exportObservationResults.ts
```diff
@@ -185,23 +192,33 @@ function makePlotSpeciesCsv(observationResults: ObservationResults): Blob {
...
-        content: makePlotSpeciesCsv(observationResults),
-      },
-    ],
-    suffix: '.csv',
-  });
+export default async function exportObservationResults(params: UseExportObservationResultsParams) {
+  const { observationId, defaultTimeZone, plantingSites, species } = params;
+
+  const resultsWithUnknown = await ObservationsService.getSingleObservationResults(observationId, true);
+  if (resultsWithUnknown.requestSucceeded && resultsWithUnknown.data?.observation) {
```
sgrimm, line 199: We should probably show an error if the request fails. On the other hand, this code doing nothing is an improvement over the other CSV export buttons in the app, whose reaction to a failed API request is to download a CSV file whose contents are the string `undefined`.

src/services/ObservationsService.ts
```diff
@@ -231,6 +236,18 @@ const exportGpx = async (observationId: number): Promise<any> => {
   }
 };
 
+const getSingleObservationResults = async (
+  observationId: number,
+  includeUnknown: boolean = false
+): Promise<Response2<ObservationResultsSingleResponsePayload>> => {
+  const response = await httpObservationResultsSingle.get2<ObservationResultsSingleResponsePayload>({
```
sgrimm, line 243: I assume `get2` is what we should be using for new code. There's a mix of `get` and `get2` in this file but `get2` seems cleaner to me.

sgrimm:
Abandoning this; there's a simpler approach I'll use instead.

PR #4109, opened 2025-06-03
SW-7023 Stop using organization ID -1 as sentinel

sgrimm:
Currently, when there's no organization selected, the
OrganizationProvider sets the selected organization to a dummy one
with an organization ID of -1, which is treated as a sentinel value
in the code. But sometimes we forget to check for it and the -1
value ends up in API requests, which then sometimes cause server
errors.

* Change the provider's context to make the selected organization
  object optional; it will be undefined when there's no organization
  selected.
* Update all the places where we were checking for organization -1
  to instead check whether the selected organization object exists.
* Fix the TypeScript errors that indicated where we were using the
  organization ID without checking whether it was valid.

There may still be places where we're missing a required check, but
there should at least be fewer of them than before.

### DIFF
--- a/src/components/OrganizationsDropdown.tsx
+++ b/src/components/OrganizationsDropdown.tsx
@@ -60,7 +60,7 @@ export default function OrganizationsDropdown(): JSX.Element {
         onSuccess={(organization) => void handleSuccess(organization)}
       />
       <PopoverMenu
-        anchor={<p style={{ fontSize: '16px' }}>{selectedOrganization.name}</p>}
+        anchor={<p style={{ fontSize: '16px' }}>{selectedOrganization?.name}</p>}
         menuSections={[
           organizations?.map((organization) => ({ label: organization.name, value: organization.id.toString() })) || [],
           [{ label: strings.CREATE_NEW_ORGANIZATION, value: '0' }],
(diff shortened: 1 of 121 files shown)


src/components/OrganizationsDropdown.tsx
```diff
@@ -60,7 +60,7 @@ export default function OrganizationsDropdown(): JSX.Element {
         onSuccess={(organization) => void handleSuccess(organization)}
       />
       <PopoverMenu
-        anchor={<p style={{ fontSize: '16px' }}>{selectedOrganization.name}</p>}
+        anchor={<p style={{ fontSize: '16px' }}>{selectedOrganization?.name}</p>}
```
sgrimm, line 63: I was torn on whether to turn this kind of thing into a higher-level guard clause vs. a null-safe reference.

src/components/common/SpeciesSelector/index.tsx
```diff
@@ -44,11 +45,11 @@ export default function SpeciesSelector<T extends { speciesId?: number } | undef
         setSpeciesList(response.sort((a, b) => a.scientificName.localeCompare(b.scientificName)));
       }
     },
-    [selectedOrganization.id]
+    [selectedOrganization, id]
```
sgrimm, line 48: At first I thought this was a malformed edit that replaced a period with a comma, but turns out that `id` is a property of the component and was missing as a dependency here.

constanzauanini:
This looks good to me, but since it's a big change I'm not sure how we can make sure everything works as expected before merging it

constanzauanini (approved):

PR #4114, opened 2025-06-04
SW-7032 Remove Project Profile Feature Flag from dump file too

ice1080:
#4113 removed the project profile feature flag, this PR just removes the artifact of it from the dump file (and playwright test comment).

Not sure why the file was CRLF before, my editor changed it to LF automatically, but I can change it back to CRLF if we'd prefer to not have this be a large change according to git.


constanzauanini (approved):
Looks good to me! probably it was my editor that changed to CRLF, not sure

PR #4117, opened 2025-06-04
SW-6968 Mobile: Challenges and Mitigations should stack

constanzauanini:


src/components/AcceleratorReports/ChallengesMitigationBox.tsx
```diff
@@ -35,8 +37,10 @@ const ChallengeMitigationPlan = ({
   onRemove: () => void;
   validateFields: boolean;
   noTitle?: boolean;
+  funderReportView?: boolean;
 }) => {
   const theme = useTheme();
+  const { isMobile } = useDeviceInfo();
```
ice1080, line 43: If `noTitle` is defined in the parent as `!isMobile`, does it still need to be a separate prop or could we just use `isMobile` in here instead?

constanzauanini: We want titles for mobile only in the funder portal report view. But in EditReport this component is also used, and I think we want titles there for both mobile and desktop

constanzauanini: But that property was intended to show/hide titles for different conditions (not just mobile or not). I can change it or we could keep it like this, in case we need to hide titles for other use cases

ice1080: I'll approve the PR and let you make the decision. Idk if this component has many other use cases, but I haven't been around long enough yet to see old components get reused much

constanzauanini: That's right, I will change it for now, and we could always add the property in the future if we need it

PR #4121, opened 2025-06-04
SW-7021 Don't download strings until locale is known

sgrimm:
If the user's locale is something other than English, only
download the strings for that locale rather than always starting
off by downloading English strings and only loading the correct
strings afterwards.

In addition to saving the user some bandwidth, this prevents
some races where we use the currently-loaded strings to
construct API requests, which turn out to fail because the
strings don't match the user's locale. SW-7036 is one such
example. This doesn't completely fix that issue but fixes one
cause of it.


sgrimm:
* **#4121** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4121?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4121?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/providers/LocalizationProvider.tsx
```diff
@@ -48,54 +48,62 @@ export default function LocalizationProvider({
...
+      bootstrapped: !!activeLocale,
+      selectedLocale,
+      setSelectedLocale,
+      supportedTimeZones: timeZones,
+    };
 
-  return <LocalizationContext.Provider value={context}>{children}</LocalizationContext.Provider>;
+    return <LocalizationContext.Provider value={context}>{children}</LocalizationContext.Provider>;
+  } else {
+    return null;
```
sgrimm, line 107: I have another version of this change where I updated `ProvidedLocalizationData` such that the selected locale could be null, and then added `if (selectedLocale)` blocks in various places in the code. But this seemed both simpler and safer: we can't really show any UI at all until we know the locale, so there's no point even rendering the child objects before we could display any text.

If there are bad side effects of this approach, though, I can switch to the other one.

sgrimm:
Playwright failure was legit; I wasn't handling users whose locale wasn't set to anything.

joshbuchea (approved):

src/providers/LocalizationProvider.tsx
sgrimm: We default to English if the user doesn't have a locale. In practice, the locale gets set right away when the user logs in, but it needs to be nullable in the user details because when a new user is invited to an organization, we create a user record for them but we don't know their locale preference yet.

PR #4128, opened 2025-06-05
SW-7036 Use raw enum values for searches

sgrimm:
If you load the Nurseries or Seed Banks page and switch locales,
a new search request is triggered to refresh the table view with
the new locale. But it's possible for that request to be sent
before the new locale's strings have finished loading, in which
case we were using the old locale's strings to construct the
search API requests. This was resulting in a server error because
the server expects searches of enum fields to use the localized
enum values.

Change the searches to filter on the "raw" search fields instead;
they don't get localized so there's no issue if strings haven't
loaded yet when the search request is constructed.


ice1080 (approved):
Definitely had to double-take on this PR and re-read the description several times ha

PR #4134, opened 2025-06-06
Allow updating tabs after initialization

constanzauanini:


ice1080 (changes requested):
I think this is on the right track but needs a bit more. It seems that the sticky tabs are not working; I tried changing tabs and refreshing and it went back to the first tab

joshbuchea:
I'm currently working on SW-6955, which as it turns it out is an issue caused by `useStickyTabs` and `useSessionFilters`. I'm planning to update the `Tabs` component to include the "sticky tabs" functionality and remove the custom hook.

constanzauanini:
> I'm currently working on [SW-6955](https://terraformation.atlassian.net/browse/SW-6955), which as it turns it out is an issue caused by `useStickyTabs` and `useSessionFilters`. I'm planning to update the `Tabs` component to include the "sticky tabs" functionality and remove the custom hook.

I committed a fix for what @ice1080 mentioned that was not working with the change, but I'm fine with closing this PR if it doesn't make sense because of what you are working on @joshbuchea

ice1080:
Ooh nice ya let's wait and see what comes out of Josh's work

PR #4139, opened 2025-06-07
SW-6955: Browser back button doesn't work on some pages

joshbuchea:
This PR fixes an issue where the back button failed in some scenarios.

Previously, active tab state was managed by both React state and query params. When navigating back to a page with sticky tabs, replacing the URL triggered a new history stack, causing the previous history to be lost.

Note: The `tab` query param can still  beused to indicate a tab to load and if set, will override the tab stored from the previous session.

### DIFF
--- a/src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentVariablesTab.tsx
+++ b/src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentVariablesTab.tsx
@@ -95,7 +95,6 @@ const tableCellRenderer = (props: RendererProps<any>): JSX.Element => {
 
 export type DocumentVariablesProps = {
   projectId?: number;
-  setSelectedTab?: (tab: string) => void;
 };
 
 const filterSearch =
@@ -112,7 +111,7 @@ const filterSearch =
     );
   };
 
-const DocumentVariablesTab = ({ projectId: projectIdProp, setSelectedTab }: DocumentVariablesProps): JSX.Element => {
+const DocumentVariablesTab = ({ projectId: projectIdProp }: DocumentVariablesProps): JSX.Element => {
   const activeLocale = useLocalization();
   const { isAllowed } = useUser();
   const { allVariables, documentSectionVariables, getUsedSections, projectId, reload } = useDocumentProducerData();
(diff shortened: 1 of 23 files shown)


src/utils/filterHooks/useSessionFilters.ts
ice1080, line 1: Nice catch on this one

src/utils/useStickyTabs.ts
```diff
@@ -33,55 +31,48 @@ const writeTabToSession = (viewIdentifier: string, tab: string): void => {
...
+  const onChangeTab = useCallback(
     (newTab: string) => {
-      query.set('tab', newTab);
-      const emptyQuery = tab === newTab ? query.toString() : new URLSearchParams(`tab=${newTab}`);
-      navigate(getLocation(location.pathname, location, keepQuery ? query.toString() : emptyQuery.toString()), {
-        replace: !tabsAreNewPages,
-      });
+      if (queryTab) {
+        navigate(getLocation(location.pathname, location), { replace: true });
+      }
```
ice1080, line 46: Does this mean that the current tab is only stored in either sessionStorage or in the url, but not both?
I suppose that this also means that every page with tabs now supports navigation to a specific tab just using the url, right?

src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentVariablesTab.tsx
```diff
@@ -159,22 +158,21 @@ const DocumentVariablesTab = ({ projectId: projectIdProp, setSelectedTab }: Docu
...
-    },
-    [setSelectedTab]
-  );
+  const onSectionClicked = useCallback((sectionNumber: string) => {
+    setOpenEditVariableModal(false);
+    setSelectedVariable(undefined);
+
+    const documentTabElement: HTMLButtonElement | null = document.querySelector('button[role="tab"]:first-child');
+    if (documentTabElement) {
+      documentTabElement.click();
```
ice1080, line 167: I think this is my only qualm with this PR - the `querySelector` and `first-child`. And I don't necessarily know of a better way to do it. 
The only other thing I can think of is doing a `navigate` with `replace: true` and just modifying the query params. Not sure if that would be considered better or worse though

src/utils/useStickyTabs.ts
ice1080: That's great!

src/scenes/AcceleratorRouter/Documents/DocumentView/DocumentVariablesTab.tsx
joshbuchea: I'm not a fan of this either, but it eliminates the circular dependency and we're already accessing the DOM within this function. The selector still works without `:first-child` in the selector, I added it to make it more specific. But perhaps it's too specific, will remove it for now.

joshbuchea: I agree this can be cleaned up further. I'll go ahead and merge this to prevent merge conflicts in the PR and to fix the back button behavior across Terraware. And I'll try to circle back today to improve this code, but I also wouldn't be opposed to you @ice1080 or @constanzauanini making the improvements if you get the itch to.

Thanks for not holding up this PR over this change 🙌

ice1080: Oh definitely, and this isn't the only usage of `document.querySelector` anyway, and also I feel it should generally be treated as a guideline and not a hard rule. I would say don't worry about it, I just wanted to mention it.

PR #4146, opened 2025-06-10
Disable no-console

joshbuchea:
I'd like to disable the `no-console` rule for now. We should catch console statements in CI or commit hooks instead of showing editor errors. I use `console.log` heavily during development and would prefer to not deal with constant red squiggles or repeatedly disabling the rule.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -249,7 +249,7 @@ export default fixupConfigRules([
       'no-class-assign': 'error',
       'no-compare-neg-zero': 'error',
       'no-cond-assign': 'error',
-      'no-console': 'off',
+      'no-console': 'warn',
       'no-const-assign': 'error',
       'no-constant-binary-expression': 'error',
       'no-constant-condition': 'error',


eslint.config.mjs
ice1080, line 1: Would warn make sense instead of off? e.g. are yellow squiggles acceptable?

joshbuchea: I should have been clearer initially. While the inline red squiggles are an annoyance, the red highlighting of file tabs and filenames indicating an error is difficult to ignore.

I still prefer catching console statements in CI or commit hooks, but using warn would reduce most of my frustration and I can deal with yellow squiggles 🙂

ice1080: Oh you're totally good to use off instead of warn. I won't be able to put it into ci for a bit while I finish up some other stuff, but I'm fine with that. I probably need to setup a commit hook because I mostly just don't like CI catching it instead of sometime pre-commit. So keep it as off if you want, I just wanted to ask the question.

ice1080 (approved):

eslint.config.mjs
ice1080: Approved, but change it to off if you prefer!

PR #4157, opened 2025-06-12
SW-7071: Zone & Plot level observation detail views are not accessible

joshbuchea:
This PR fixes an infinite render issue in the Observation Detail View.

### DIFF
--- a/src/scenes/ObservationsRouter/details/index.tsx
+++ b/src/scenes/ObservationsRouter/details/index.tsx
@@ -72,15 +72,17 @@ export default function ObservationDetails(props: ObservationDetailsProps): JSX.
   }, [query]);
 
   const observationsResults = useAppSelector((state) =>
-    searchObservations(
-      state,
-      plantingSiteId,
-      selectedOrganization,
-      defaultTimeZone.get().id,
-      searchProps.search,
-      searchProps.filtersProps?.filters?.zone?.values ?? [],
-      status
-    )
+    !selectedOrganization?.id
+      ? undefined
+      : searchObservations(
+          state,
+          plantingSiteId,
+          selectedOrganization.id,
+          defaultTimeZone.get().id,
+          searchProps.search,
+          searchProps.filtersProps?.filters?.zone?.values ?? [],
+          status
+        )
   );
 
   const selectedObservationResults = useMemo(() => {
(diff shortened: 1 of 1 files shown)


src/scenes/ObservationsRouter/details/index.tsx
```diff
@@ -72,15 +72,17 @@ export default function ObservationDetails(props: ObservationDetailsProps): JSX.
...
-    searchObservations(
-      state,
-      plantingSiteId,
-      selectedOrganization,
-      defaultTimeZone.get().id,
-      searchProps.search,
-      searchProps.filtersProps?.filters?.zone?.values ?? [],
-      status
-    )
+    !selectedOrganization?.id
```
sgrimm, line 75: You should be able to use `!selectedOrganization` here -- the ID is guaranteed to be valid if the organization object exists at all (that was the idea with my change in commit a19055154).

src/scenes/ObservationsRouter/details/index.tsx
```diff
@@ -132,7 +136,9 @@ export default function ObservationDetails(props: ObservationDetailsProps): JSX.
   const plantingSite = useAppSelector((state) => selectPlantingSite(state, plantingSiteId));
   const observation = useAppSelector((state) => selectObservation(state, plantingSiteId, observationId));
   const zoneNames = useAppSelector((state) =>
-    selectDetailsZoneNames(state, plantingSiteId, observationId, selectedOrganization)
+    !selectedOrganization?.id
+      ? []
+      : selectDetailsZoneNames(state, plantingSiteId, observationId, selectedOrganization.id)
```
sgrimm, line 141: Passing the organization object instead of the ID seems like something the type system should have caught. (Fixing that is out of scope for this PR, of course.)

joshbuchea: I agree and was surprised it didn't catch it.

sgrimm (approved):

src/scenes/ObservationsRouter/details/index.tsx
sgrimm: I took a quick look and it seems like the answer is that `createCachedSelector` doesn't preserve argument type information.

PR #4166, opened 2025-06-12
Add type guards to org membership functions

sgrimm:
The helper functions to check the user's role in an organization
accept undefined organizations and return false in that case, so
there's no need for calling code to also check that the
organization is defined.


joshbuchea (approved):
Nice improvement

PR #4169, opened 2025-06-13
Convert strings CSV files for all locales on edit

sgrimm:
When a strings CSV file changes while the dev server is running,
we automatically regenerate its JavaScript file to avoid having to
restart the server. Unfortunately, we still have to restart the
server if strings are being added to the English CSV, because the
other locales' JavaScript files are missing those strings and
thus the TypeScript types don't match.

Change the regeneration code to always regenerate all the JS
files if any of the CSV files change. Then we can freely edit
the English strings file (or switch between git branches that have
different edits to the strings file) without restarting the dev
server.

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -1,7 +1,7 @@
 const chokidar = require('chokidar');
-const { convertCsvFile } = require('./src/strings/export');
 const fs = require('fs');
 const { whenDev } = require('@craco/craco');
+const { convertAllLocales } = require('./src/strings/export');
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#4169** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #4170, opened 2025-06-13
SW-7020 Add tooltip icon to live-plants titles

sgrimm:
Commit 6bacecb9 added a tooltip to the titles of the live
plants graphs in observation detail views, but it should have
put the tooltip on an icon, not on the text of the titles.


sgrimm:
Ah, hadn't noticed that component. Thanks. Will give it a try.

joshbuchea (approved):

PR #4195, opened 2025-06-18
Remove decimal when total planted under 999 in project profile

ice1080:
We previously had a decimal count of 1 for Total Planted because I assumed it would be over 1000 most of the time, and chose to display that as 1.2k, 2.4m, etc. Remove the decimal when count is under 999 but leave it for 1000+.


constanzauanini (approved):

ice1080:
### Merge activity

* **Jun 20, 3:48 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/4195).

PR #4198, opened 2025-06-20
SW-6885: Deploy to Vercel using CI build

joshbuchea:
## Tasks

- [ ] Before merging PR, remove current branch name from `branches` (`josh/SW-6885-vercel-deploy`)
- [ ] When PR is merged, disable automatic builds in Vercel


.github/workflows/workflow.yml
```diff
@@ -8,6 +8,7 @@ on:
     branches:
       - main
       - qa
+      - josh/SW-6885-vercel-deploy
```
joshbuchea, line 11: Will remove this prior to merging

joshbuchea:
> Looks great!
> 
> This is a suggestion for the future. I am wondering if it makes sense to have a separate workflow file, so that we can have another trigger (for example a PR comment), so we can deploy to Vercel on branches without opening a pull request. This can also apply to E2E tests too since it is not always easy to run locally.

Great idea!

PR #4209, opened 2025-06-30
SW-7092 Don't save invalid substrate type

sgrimm:
The lists of substrates for lab and nursery viability tests aren't
the same.

When you're adding a viability test to an accession, if you select
a substrate that's only valid for the current test type then switch
to the other test type, we clear out the substrate selection in the
UI. But we remember the old selection so that if you switch back
to the initial test type again, your previously-selected substrate
will reappear.

But if you selected a substrate that's only valid for the current
test type, switch test types, and then save the viability test, we
were trying to create the test with an invalid substrate for its
type, which caused the request to be rejected by the server.

Change the "save" logic so it reflects the UI's behavior, clearing
out the substrate type if it's not valid.

### DIFF
--- a/src/scenes/AccessionsRouter/viabilityTesting/NewViabilityTestModal.tsx
+++ b/src/scenes/AccessionsRouter/viabilityTesting/NewViabilityTestModal.tsx
@@ -340,6 +340,10 @@ export default function NewViabilityTestModal(props: NewViabilityTestModalProps)
       if (testCompleted && !readOnly) {
         record.endDate = getTodaysDateFormatted(tz.id);
       }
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#4209** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4209?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4209?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #4224, opened 2025-07-01
Fix max-classes-per-file lint error

ice1080:


ice1080:
* **#4224** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4224?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4224?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080:
### Merge activity

* **Jul 1, 5:57 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-web/4224).

tommylau523 (commented):
Can we move each ToDo class into a todo folder under types?

ice1080:
> Can we move each ToDo class into a todo folder under types?

@tommylau523 I did consider this. I ended up leaving EventToDo in the ProjectToDo file because it is then used in the `compareToDoItems` function and I didn't want to create a circular dependency. 
Probably a better option would be to move `compareToDoItems` to a different folder since it probably shouldn't be under `src/types`. I can do this in another PR, is there a good spot for this? `src/utils` maybe?

tommylau523:
Ah that makes sense. I probably prefer the function to be under types as well since they are very much related.

We probably need a src/entities folder or something else to capture more complex objects than just types.

I think just leave it as is for now! We don’t have lots of classes anyways.

PR #4227, opened 2025-07-01
SW-7109 Add optional unit field to project specific metrics and report displays

ice1080:


src/scenes/FunderReport/MetricBox.tsx
```diff
@@ -71,6 +71,7 @@ const MetricBox = ({ metric, index, year, quarter, length }: MetricBoxProps) =>
           </Typography>
           <Typography fontSize={'24px'} fontWeight={600}>
             {metric.target !== undefined && numberFormatter.format(metric.target)} {metric.target ? addPercentSign : ''}
+            {metric.target !== undefined ? metric.unit : ''}
```
constanzauanini, line 74: Should this be `metric.unit !== undefined` ? Same in line 83

ice1080: No in this case I'm only wanting to display the unit if the target is defined, because otherwise the text would just be the unit which could be confusing!

PR #4260, opened 2025-07-10
SW-7139: "Refresh to use the latest version" does not go away

joshbuchea:
`vercel build` resets `REACT_APP_TERRAWARE_FE_BUILD_VERSION` and regenerates `build-version.txt`. This step was unintentionally modifying both before the release build. Moving the Vercel steps to the end of the job avoids blocking the release; a separate job would be a longer-term fix. Will address the longer-term fix in a follow-up PR.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -8,7 +8,6 @@ on:
     branches:
       - main
       - qa
-      - josh/SW-7139-refresh-banner-always-shows
     tags:
       # Semver (these are glob-like patterns, not regexes; the "." has no special meaning)
       - v[0-9].[0-9]+.[0-9]+
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -109,43 +109,6 @@ jobs:
           path: playwright-report/
           retention-days: 30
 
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow.yml
```diff
@@ -8,6 +8,7 @@ on:
     branches:
       - main
       - qa
+      - josh/SW-7139-refresh-banner-always-shows
```
ice1080, line 11: I think you'll want to remove this extra branch when you get a sec

joshbuchea:
Vercel deploy still works 🎉

ice1080 (approved):

sgrimm:
Doing the Vercel build in a separate workflow would have avoided the problem too, I think.

ice1080:
> Doing the Vercel build in a separate workflow would have avoided the problem too, I think.

@sgrimm That's what Josh said as well, but didn't want that work to hold up the release anymore, so this was the short term solution

sgrimm:
Ah, so he did. Serves me right for not reading to the end of the PR description before hopping to the code.

joshbuchea:
Will get the vercel steps moved into a separate workflow soon, thanks y'all! 🙌

PR #4261, opened 2025-07-11
SW-7145: Move Vercel deploy in CI to separate workflow

joshbuchea:
This PR moves the Vercel deployment steps into a separate workflow file.

**Notes:**

- I included checks (formatting, linting, types, unit tests) in the Vercel workflow since they’re quick (<2 min) and the full workflow takes \~5 min. Happy to remove them if preferred.
- This might be a good chance to rename `workflow.yml` to something clearer like `deploy.yml` or `deployment.yml`. Open to thoughts.


.github/workflows/workflow.yml
```diff
@@ -7,7 +7,6 @@ on:
   push:
     branches:
       - main
-      - qa
```
joshbuchea, line 10: Removed `qa` branch since the QA tier was removed from `set-environment.sh`

PR #4272, opened 2025-07-15
Fix Vercel deploys become inactive

joshbuchea:
Disable `auto-inactive` option for vercel deploys:

> Adds a new inactive status to all prior non-transient, non-production environment deployments with the same repository and environment name as the created status's deployment. An inactive status is only added to deployments that had a success state.

The option is `true` by default.


constanzauanini (approved):

ice1080:
Do we need any setting to tell Vercel how long to keep deployments? E.g. should they become inactive after 2 weeks or something? Or is that a setting inside Vercel?

PR #4276, opened 2025-07-15
SW-7084: Planting Sites: Maps - remove "Monitoring Plots" from the zone and subzone popovers

joshbuchea:
This PR removes "Monitoring Plots" information from the Zone & Subzone popovers that display in the Planting Site map.


tommylau523 (approved):
Thanks!!

PR #4294, opened 2025-07-22
SW-7186 Change Map corner radius 8px

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/ProjectProfileGisMaps.tsx
ice1080, line 1: Should this instead be based on `theme.spacing`?

PR #4324, opened 2025-08-05
SW-7218 Planting Site name/label missing in Zones and Subzones view of Planting Site

constanzauanini:


src/scenes/AcceleratorRouter/ParticipantProjects/ProjectProfileGisMaps.tsx
```diff
@@ -442,7 +442,7 @@ const ProjectProfileGisMaps = () => {
             <PlantingSiteMap
               mapData={filteredSiteData}
               style={{ width: '100%', borderRadius: theme.spacing(1) }}
-              layers={[selectedLayer || 'Planting Site']}
+              layers={[selectedLayer || 'Planting Site', 'Planting Site']}
```
joshbuchea, line 445: Is this change intentional?

joshbuchea (approved):
One question, but otherwise looks good!

src/scenes/AcceleratorRouter/ParticipantProjects/ProjectProfileGisMaps.tsx
constanzauanini: Yes, I wanted to always include the planting site layer (in addition to the selected layer) so that I can display the site name also in zones and subzones

PR #4325, opened 2025-08-05
SW-7220 Popover persists even when selected boundary is changed

constanzauanini:


joshbuchea (commented):
The code might try to remove an event listener that was never added, since it's only registered when `popupInfo` exists, but always removed in cleanup.

I think we could simplify this by always adding the event listener and checking for `popupInfo` inside `handleClickOutside`.

PR #4339, opened 2025-08-14
SW-7211: FE - Add "Hardening Off" to batch detail view and Batch Create/Edit modals (PART #2)

joshbuchea:
This PR includes an additional set of changes for SW-7211, covering a number of instances overlooked in the previous PR.


src/api/types/generated-schema.ts
```diff
@@ -4550,7 +4550,7 @@ export interface components {
             /** Format: int32 */
             germinatingQuantityAdded: number;
             /** Format: int32 */
-            hardeningOffQuantity: number;
+            hardeningOffQuantityAdded: number;
```
joshbuchea, line 4553: @ice1080 I think this property needs the `Added` suffix, I manually changed it after regenerating the types.

ice1080: Server PR to fix this, sorry about that 😞 
https://github.com/terraware/terraware-server/pull/3295

ice1080 (approved):
Approving, pending the merge of the Server PR to fix the property name

PR #4351, opened 2025-08-18
Update useForm to return both a callback and a direct function

ice1080:
The eslint rule `react/jsx-no-bind` is a helpful tool for finding arrow functions that should be memoized callbacks. Unfortunately it also identifies all arrow functions even if they won't cause the issues that the rule was originally added to mitigate. This causes many instances of `useCallback` being added for simple changes; one of which is all of the `onChange` calls from `useForm`. To help with this, do the following:

- Update `useForm` to return both the original `onChange`​ function as well as a callback that takes the id to be updated on the record
- Update usages of `useForm/onChange` to start reducing the excess callbacks, as well as reducing the instances of unnecessary arrow functions now that `onChangeCallback` exists.

This is not exhaustive by any means, it is just the lowest hanging fruit which includes any instances of callbacks created just to call `onChange` with an id, and any instances of arrow functions that were just to call `onChange` with an id.

All pages updated by this were tested manually to ensure no regression.

Here is the form that caused infinite render, in case someone ever wants to attempt this again. The infinite render occurred on the Edit Species page in Console -> Deliverables -> Species Deliverable -> \<Species>.

```
import React, { Dispatch, SetStateAction, useCallback } from 'react';

export interface FormChangeHandler {
  (id: string): (value: unknown) => void; // Curried form
  (id: string, value: unknown): void; // Direct form
}

const useForm = <T>(originalRecord: T): [T, Dispatch<SetStateAction<T>>, FormChangeHandler] => {
  const [record, setRecord] = React.useState(originalRecord);

  const onChange = useCallback(
    // eslint-disable-next-line
    function onChangeInternal(id: string, value?: unknown) {
      // If value is provided, execute immediately (direct form)
      if (arguments.length === 2) {
        setRecord((previousRecord: T): T => {
          return { ...previousRecord, [id]: value };
        });
        return;
      }

      // If only id is provided, return a function (curried form)
      return (newValue: unknown) => {
        setRecord((previousRecord: T): T => {
          return { ...previousRecord, [id]: newValue };
        });
      };
    } as FormChangeHandler,
    []
  );

  return [record, setRecord, onChange];
};

export default useForm;
```

### DIFF
--- a/src/components/AcceleratorDeliverableView/QuestionsDeliverableCard.tsx
+++ b/src/components/AcceleratorDeliverableView/QuestionsDeliverableCard.tsx
@@ -357,9 +357,7 @@ const QuestionBox = ({
                 type='textarea'
                 label={strings.INTERNAL_COMMENTS}
                 id='internalComment'
-                onChange={(value) => {
-                  onChange('internalComment', value as string);
-                }}
+                onChange={onChange('internalComment')}
                 sx={{ marginTop: theme.spacing(1) }}
                 value={workflowDetails.internalComment}
               />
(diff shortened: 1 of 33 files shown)


constanzauanini (approved):
Thanks for doing all this! Looks good to me! 💯

PR #4352, opened 2025-08-18
Use the new map component in place of planting dashboard map

tommylau523:
[Screen Recording 2025-08-18 at 3.41.18 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/user-attachments/video/b7c73c79-5c24-4e2a-a263-f5ccc9e56aab.mov)

The `web-components` package was having issues with exporting Map component correctly. The root cause is transpiring from esm to cjs and losing some properties as a result.

This is a temporary workaround to get Plant Dashbaord map to QA first.

### DIFF
--- a/src/scenes/PlantsDashboardRouter/components/PlantDashboardMap.tsx
+++ b/src/scenes/PlantsDashboardRouter/components/PlantDashboardMap.tsx
@@ -1,7 +1,9 @@
 import React, { useCallback, useMemo } from 'react';
 
+import { useTheme } from '@mui/material';
+
 import MapComponent, { MapFeatureSection } from 'src/components/NewMap';
-import { MapLayer, MapLayerFeatureId, MapMarker } from 'src/components/NewMap/types';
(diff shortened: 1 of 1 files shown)


tommylau523 (commented):
Everything in the NewMap folder is copied over from the web-components library.

src/scenes/PlantsDashboardRouter/components/PlantDashboardMap.tsx
```diff
@@ -0,0 +1,358 @@
...
+                type: 'MultiPolygon',
+                coordinates: site.boundary?.coordinates ?? [],
+              },
+            },
+          ],
+          label: strings.SITE,
+          layerId: 'sites',
+          style: {
+            borderColor: '#41C07F',
+            fillColor: '#41C07F',
```
ice1080, line 38: Would it be worth changing these to theme colors?

PR #4362, opened 2025-08-20
SW-7269 Non Variable columns should show up under manage columns

constanzauanini:


src/scenes/AcceleratorRouter/MatrixView/ColumnsModal.tsx
```diff
@@ -282,9 +282,14 @@ export default function ColumnsModal(props: ColumnsModalProps): JSX.Element {
     [table]
   );
 
+  const baseColumns = useMemo(
+    () => ['projectName', 'participantCohortPhase', 'elegibleLand', 'countryName', 'projectLead'],
+    []
+  );
```
ice1080, line 288: Since this is static, could also move to above the component
```ts
const BASE_COLUMNS = ['projectName', 'participantCohortPhase', 'elegibleLand', 'countryName', 'projectLead'];
```

src/scenes/AcceleratorRouter/MatrixView/ColumnsModal.tsx
```diff
@@ -465,34 +470,64 @@ export default function ColumnsModal(props: ColumnsModalProps): JSX.Element {
...
-                ))}
+                <>
+                  {baseColumns.map((col) => (
+                    <ListItem
+                      key={col}
+                      sx={{
+                        px: 0,
+                        py: 0.5,
+                        '&:hover': { backgroundColor: 'action.hover' },
+                        borderRadius: 1,
```
ice1080, line 481: Might be worth a separate component for this (even if in the same file), since this is the same as what's rendered for `filteredVariables`

PR #4368, opened 2025-08-21
SW-7273 Reset Order button deletes all selected columns instead of resetting the order

constanzauanini:


playwright/e2e/suites/matrixView.spec.ts
ice1080, line 1: Might be worth adding tests at some point for resetting the columns and/or resetting the order

ice1080 (approved):

PR #4375, opened 2025-08-22
SW-7231: FE: Add/Remove Project Internal Users in Edit Project View

joshbuchea:
This PR includes changes to enable adding/removing Project Internal Users in the Profile Profile Edit View.


src/scenes/AcceleratorRouter/ParticipantProjects/EditView/ProjectProfileEdit.tsx
```diff
@@ -139,17 +202,22 @@ const ProjectProfileEdit = () => {
     if (
       isComplete(participantProjectUpdateResponse?.status, initiatedRequests.participantProject) &&
       isComplete(assignContactResponse?.status, initiatedRequests.assignTfContact) &&
+      isComplete(updateInternalUsersResponse?.status, initiatedRequests.updateInternalUsers) &&
```
ice1080, line 205: Do we need any of the assign contact stuff anymore now that this is replacing it?

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/ProjectProfileEdit.tsx
```diff
@@ -235,16 +324,46 @@ const ProjectProfileEdit = () => {
...
+        requestProjectInternalUsersUpdate({
+          projectId,
+          usersToAssign: (internalUsers || [])
+            .filter((user) => !!user.userId)
+            .map((user) => ({
+              role: user.role,
+              roleName: user.roleName,
+              userId: user.userId as number,
+            })),
+          usersToRemove: listInternalUsersRequest?.data?.users || [],
```
ice1080, line 339: I think this may cause a race condition if a user is not being added or removed. This is saying to remove all users and then reassign some of them, right? And in the thunk we're calling all the removal endpoints and then the assign endpoints, so it seems like it's possible that a user could be re-assigned and then removed if the BE processes them in varying speeds. Am I thinking about that right?

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/ProjectProfileEdit.tsx
```diff
@@ -447,6 +613,112 @@ const ProjectProfileEdit = () => {
                 />
               </Box>
             </Grid>
+
+            <Grid item md={12}>
+              <Box border='1px solid gray' borderRadius='8px' marginX={theme.spacing(2)} padding={theme.spacing(2)}>
+                <Box borderBottom='1px solid gray' marginBottom='16px' paddingBottom='8px'>
+                  {/* TODO: update the tooltip copy below */}
+                  <Typography fontSize='16px' fontWeight={600} lineHeight='24px'>
```
ice1080, line 621: I think this TODO can be removed now

src/types/Project.ts
```diff
@@ -8,3 +10,54 @@ export type ProjectMeta = {
...
+  'Consultant',
+  'GIS Lead',
+  'Legal Lead',
+  'Phase Lead',
+  'Project Lead',
+  'Project Finance Lead',
+  'Regional Expert',
+  'Restoration Lead',
+  'Social Lead',
+];
```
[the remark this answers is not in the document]
joshbuchea: This is a pattern we use throughout the app, I think this is fine.

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/ProjectProfileEdit.tsx
joshbuchea: Good point, will remove.

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/ProjectProfileEdit.tsx
joshbuchea: The requests are initiated in the proper order in the FE, if it's possible for the BE to complete the work out-of-order perhaps we need to re-think the API for this to require fewer requests?

joshbuchea: In the case that only the role is changed for an existing user, the FE would still need to remove and re-add the modified user. Let's go with option 2.

ice1080: Ok. I'm deep in survival rate stuff so it'll be a bit before I can switch back to that. Will let you know once it's merged

joshbuchea: I'm not sure I follow your point about requests being initiated without waiting. From what I see, the FE waits for all remove requests to complete before starting the assign requests.

joshbuchea: Ah you're right, sorry. Rather than having you rewrite the API, I can separate these so the remove requests complete before starting the assign requests.

joshbuchea: @ice1080 I went ahead and cleaned this function up because it was bugging me and  it only took a couple of minutes. I'll plan on updating the FE once the API changes are ready. No rush on my end at all, I have other tasks I can focus on.

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/AddInternalUserRoleModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+    onClose();
+  }, [addInternalUserRole, newInternalUserRole, onClose]);
+
+  return (
+    <DialogBox
+      onClose={onClose}
+      open
+      size='medium'
+      title='New Contact Type'
```
constanzauanini, line 43: This should use the strings file

joshbuchea: Great catch, thanks! 🙌

PR #4394, opened 2025-08-27
Remove unused strings from strings files

sgrimm:
We've gradually accumulated unused strings over time as we've
changed the UI to stop using them. Removing the strings from
the strings files didn't work before because Phrase would add
them back unless we remembered to remove them there too, but
now that we're using llm-autotranslate, removing them from
the files will work.

This is based on the output of `scripts/unused-strings.js`.


joshbuchea (approved):
This is awesome, thanks @sgrimm!

PR #4395, opened 2025-08-27
SW-7309: Internal Leads field: Show Row by default

joshbuchea:


ice1080 (approved):
It might not matter, but when removing the last Internal User, it goes back to no rows shown. Not sure if we care or not, since it shows back up once you save and edit again

joshbuchea:
@ice1080 I'm not sure if we do care or not, but went ahead and updated anyways.

PR #4400, opened 2025-08-29
SW-7331: Internal Leads field: Global User in Person field can be selected multiple times

joshbuchea:


src/scenes/AcceleratorRouter/ParticipantProjects/EditView/ProjectProfileEdit.tsx
```diff
@@ -505,10 +505,18 @@ const ProjectProfileEdit = () => {
     setAddInternalUserRoleModalOpen(false);
   }, [setAddInternalUserRoleModalOpen]);
 
-  const globalUsersWithNoOwner = useMemo(() => {
-    const ownerId = organizationUsers?.find((orgUsr) => orgUsr.role === 'Owner')?.id;
-    return globalUsersOptions?.filter((opt) => opt.value.toString() !== ownerId?.toString()) || [];
-  }, [organizationUsers, globalUsersOptions]);
+  const ownerId = useMemo(() => organizationUsers?.find((orgUsr) => orgUsr.role === 'Owner')?.id, [organizationUsers]);
```
ice1080, line 508: Can't there be more than one Owner in an org?

ice1080 (approved):

src/scenes/AcceleratorRouter/ParticipantProjects/EditView/ProjectProfileEdit.tsx
joshbuchea: I'm not sure, can there be? The single owner id check is pre-existing, but if it's possible for an org to have more than one user, I agree we should update this.

PR #4404, opened 2025-08-30
SW-7327: Project Profile: Other users in the Internal Contacts field are not appearing in the Detail View

joshbuchea:


src/components/ProjectField/InlineMeta.tsx
ice1080, line 1: NIT - could rename to "moreTooltip" since there's nothing tying it to users in this file

joshbuchea: Good call, will update.

PR #4456, opened 2025-09-17
SW-7206 - Update Spanish translations for matching unrecognized species

alexTerraformation:


alexTerraformation:
> We just have to leave the hash untouched for the autotranslate to not update it right?

I believe that's correct!

PR #4468, opened 2025-09-19
SW-7392 Additional Spanish translation updates

alexTerraformation:


ice1080 (approved):
If these updates are happening because there are some English words that are translated better using certain spanish words, then we should also be updating the autotranslate spanish instructions

alexTerraformation:
> If these updates are happening because there are some English words that are translated better using certain Spanish words, then we should also be updating the autotranslate spanish instructions

In this case, the original translations all pre-date the autotranslate tool and so all that I did here was delete the old translations and let the autotranslate tool re-generate them. The new translations matched the request from Diana, so I think the instructions are good.

PR #4494, opened 2025-09-24
SW-7430 Survival Rates are not showing in the Observation table or in the Observation Plot Details screen

constanzauanini:


src/scenes/ObservationsRouter/org/OrgObservationsListView.tsx
```diff
@@ -123,6 +69,66 @@ export default function OrgObservationsListView({
   const [requestId, setRequestId] = useState('');
   const abandonObservationResponse = useAppSelector((state) => selectAbandonObservation(state, requestId));
   const snackbar = useSnackbar();
+  const isSurvivalRateCalculationEnabled = isEnabled('Survival Rate Calculation');
+
+  const defaultColumns = useCallback(
+    () =>
```
ice1080, line 75: Could this one change to a `useMemo`, or does the table need a callback?

constanzauanini: Yes, this can be a useMemo

PR #4524, opened 2025-10-01
SW-7450 Make map location sticky and map view style sticky

tommylau523:
This PR implements live update of lat/lng/zoom on search params when using the map on activity views. This allows further navigation to keep the map view state. An added benefit is that user is now able to share exactly the map location to other users via the URL as well.

This also implemented sticky map view style using session storage for a more seamless experience.

A note is that, all activity navigation needs to implement preserving search params to ensure the sticky location works correctly.

Also fixed another issue with token not being defined yet, causing a brief error.

### DIFF
--- a/src/components/ActivityLog/MapSplitView.tsx
+++ b/src/components/ActivityLog/MapSplitView.tsx
@@ -167,7 +167,7 @@ export default function MapSplitView({
       params.set('lat', view.viewState.latitude.toString());
       params.set('lng', view.viewState.longitude.toString());
       params.set('zoom', view.viewState.zoom.toString());
-      setSearchParams(params);
+      setSearchParams(params, { replace: true });
     },
     [searchParams, setSearchParams]
   );
--- a/src/components/ActivityLog/MapSplitView.tsx
+++ b/src/components/ActivityLog/MapSplitView.tsx
@@ -118,11 +118,6 @@ export default function MapSplitView({
     );
   }, [plantingSites]);
 
-  const boundingBox = useMemo(() => {
-    const multipolygons = siteFeatures.map((feature) => feature.geometry);
-    return getBoundingBox(multipolygons);
-  }, [siteFeatures]);
-
   const mapFeatures = useMemo((): MapFeatureSection[] => {
     return [
       {
@@ -158,11 +153,13 @@ export default function MapSplitView({
   }, [currentProjectId, projectId, setSearchParams]);
 
   useEffect(() => {
-    if (!initialMapViewState && boundingBox) {
+    if (!initialMapViewState && siteFeatures.length > 0) {
       // If no map view state is provided, move map to features
+      const multipolygons = siteFeatures.map((feature) => feature.geometry);
+      const boundingBox = getBoundingBox(multipolygons);
       fitBounds(boundingBox);
     }
-  }, [boundingBox, fitBounds, initialMapViewState]);
+  }, [fitBounds, initialMapViewState, siteFeatures]);
 
   const onMapMoveCallback = useCallback(
     (view: ViewStateChangeEvent) => {
--- a/src/components/ActivityLog/MapSplitView.tsx
+++ b/src/components/ActivityLog/MapSplitView.tsx
@@ -174,7 +174,6 @@ export default function MapSplitView({
     },
     [searchParams, setSearchParams]
   );
-  );
 
   return (
     <Box display='flex' flexDirection='column' flexGrow={1}>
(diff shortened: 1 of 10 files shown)


tommylau523:
* **#4524** <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4524?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-web/4524?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/hooks/useNavigateTo.ts
```diff
@@ -28,11 +32,15 @@ export default function useNavigateTo() {
...
           ),
+          search: location.search,
         });
       },
 
       goToAcceleratorActivityLog: () => {
-        navigate({ pathname: APP_PATHS.ACCELERATOR_ACTIVITY_LOG });
+        navigate({
+          pathname: APP_PATHS.ACCELERATOR_ACTIVITY_LOG,
+          search: location.search,
```
tommylau523, line 42: @joshbuchea tagging you in this file because I am passing down the search params (which include lat,lng,zoom) in order to preserve map view state. New routes should follow the same pattern so that going back and forth will work accordingly.

src/components/ActivityLog/MapSplitView.tsx
```diff
@@ -101,6 +101,28 @@ export default function MapSplitView({
     });
   }, [activities, activityMarkerHighlighted, onActivityMarkerClickCallback]);
 
+  const { plantingSites } = useProjectPlantingSites(projectId);
+
```
ice1080, line 105: NIT it might just be a personal preference, but it feels weird to me to have useCallbacks defined above variables that are being used in the component (such as `plantingSites`)

src/components/ActivityLog/MapSplitView.tsx
```diff
@@ -129,8 +151,29 @@ export default function MapSplitView({
...
+    if (projectId !== currentProjectId) {
+      setInitialMapViewState(undefined);
+      setCurrentProjectId(projectId);
+    }
+  }, [currentProjectId, projectId, setSearchParams]);
+
+  useEffect(() => {
+    if (!initialMapViewState) {
+      // If no map view state is provided, move map to features
+      fitBounds(boundingBox);
```
ice1080, line 163: If bounding box is not used anywhere else, perhaps it could be inside the `if` check so that it's not calculated when not needed

src/hooks/useNavigateTo.ts
ice1080, line 42: Is it possible to call this on a page that has search params that _shouldn't_ be passed to the accelerator activity log? If so, perhaps these could be defined like this:
```ts
goToAcceleratorActivityLog: (includeSearch: boolean = true) => {
...
```

tommylau523: I think the best solution is to include a list of params to include. Perhaps this can be a follow up.

joshbuchea (changes requested):
The way in which query params are implemented in this PR break back button behavior. It may still be possible to use query params, but this implementation will need to be changed.

tommylau523:
> > @tommylau523 back button behavior is still broken with the latest changes.
> > 
> > https://github.com/user-attachments/assets/a065ac5f-e007-42c5-be39-618df0625055
> > 
> > Tracking state in the URL can break history and back button behavior. A better approach is to pass state via query params before navigation, then consume and clear them in the target view.
> > 
> > 
> 
> Ah I misunderstood the error previously. Your suggestion sounds good and I will implement it on Monday.

@joshbuchea I opted to implement replacing history within the stack instead, so now back button behavior is preserved. I opted to leave it the params instead of clearing so that users have an easy way to copy and paste exactly what they are viewing on the map. I suspect this will be helpful, but I will admit it makes our URL path looks messy.

joshbuchea:
Tested this on the Vercel preview—it works well 🎉 My Activity Detail PR (just merged) adds an `activityId` URL param for reload and back-button support. @tommy please update this with `main` when you can so we can confirm both of our URL param changes work together.

ice1080 (commented):
The rest looks good to me. I'll let @joshbuchea do the approval since you all have been working on the param stuff

tommylau523:
> @tommy I appreciate your effort on this. Unfortunately, there's still an issue during map zoom animations — when the map continuously updates the URL query params, any UI element that triggers navigation becomes unresponsive until the animation finishes.
> 
> Possible fixes:
> 
> - Remove the map zoom animation
> - Only update map URL params on navigation
> 
> https://github.com/user-attachments/assets/65d52dab-5bd4-4be7-a24d-40ae1ba750cc
> 
> 

No worries at all! I removed the animations in the latest patch.

PR #4536, opened 2025-10-04
SW-7316: FE: Activity Details View (Part 1)

joshbuchea:
This PR includes the initial implementation of the Activity Detail View.

NOTE: Edit activity button and minor style adjustments will come in a follow-up PR.

## Screenshots

localhost_accelerator_projects_15_projectId=15 activityId=44


tommylau523 (approved):

src/components/ActivityLog/ActivitiesListView.tsx
```diff
@@ -257,13 +299,19 @@ const ActivitiesListView = ({ projectId }: ActivitiesListViewProps): JSX.Element
 
   return (
     <MapSplitView
-      activities={activities} // TODO: Use visible activites after pagination/filtering
+      activities={activitiesVisibleOnMap} // TODO: Use visible activities after pagination/filtering
       activityMarkerHighlighted={activityMarkerHighlighted}
```
tommylau523, line 303: If I understand correctly, both the log view and the one activity detail view use this listView component. If that is the case, we need to handle both the marker onClick and the highlighted behaviors differently. This can be done by adding functionalities to these two callback functions.

joshbuchea: Ok, thanks for calling this out.

PR #5550, opened 2026-04-03
SW-8254: Species appearing in Observation pages (tables and cards) when there is no data yet (FOLLOW-UP)

joshbuchea:
This PR fixes an issue in the Species Survival Rate Chart:

When all species have a `survivalRate` of `0`, the previous check (`values.length > 0`) treated that as having data — so species names rendered on the x-axis with invisible zero-height bars and no overlay. Changed to `values.some(v => v > 0)` so all-zero survival rates are treated as "no data," clearing the labels and showing the "Data is not yet available" overlay.


joshbuchea:
@ice1080 thanks for the review and for flagging the issue with the current changes 🙌 I think the issue in this ticket may have already been addressed by @constanzauanini in SW-8261 (#5556). I'm waiting for confirmation before closing this PR.

PR #5579, opened 2026-04-10
SW-8268 SW-8269 Virtual Walkthrough: Add Card on home page and Create/upload modals

constanzauanini:


joshbuchea (approved):
Looks great!

joshbuchea:
You can check with Product and/or Design, but I think we want to replace `video-diagram.svg` with the circular version from the designs. I have the same "X" image in the mobile app right now, but I believe we now prefer people to walk in a circular fashion rather than crossing the plot in the shape of an X.

constanzauanini:
> You can check with Product and/or Design, but I think we want to replace `video-diagram.svg` with the circular version from the designs. I have the same "X" image in the mobile app right now, but I believe we now prefer people to walk in a circular fashion rather than crossing the plot in the shape of an X.

Thanks for flagging this! I will ask in the designs

PR #5608, opened 2026-04-21
SW-8300 Add Virtual Walkthroughs position in map

constanzauanini:


src/scenes/VirtualWalkthrough/VirtualWalkthroughsMap.tsx
```diff
@@ -57,6 +64,53 @@ export default function VirtualWalkthroughsMap({
...
+      if (!pendingPlacementFile) {
+        return;
+      }
+      void updateMedia({
+        organizationId,
+        fileId: pendingPlacementFile.fileId,
+        updateOrganizationMediaRequestPayload: {
+          gpsCoordinates: { type: 'Point', coordinates: [e.lngLat.lng, e.lngLat.lat] },
+        },
+      });
```
ice1080, line 108: Confirming that this gets updated immediately whenever the user finishes dragging? There's no save button in case the user moves it a few times?

ice1080 (commented):
Is this supposed to merge into main?

src/scenes/VirtualWalkthrough/VirtualWalkthroughsMap.tsx
constanzauanini: It is immediately, no save button. We can maybe improve it later. Now, if the user wants to change it after releasing, he has to click "update location"

constanzauanini:
> Is this supposed to merge into main?

It is using a lot of code (apis calls) from my other opened PR, so I want to merge that one first and then this one

ice1080 (approved):

PR #5609, opened 2026-04-22
SW-8284 Add screenshot compares to plants dashboard e2e tests

constanzauanini:


playwright/README.md
```diff
@@ -19,16 +19,16 @@ Replace `<version>` below with that value (e.g. `1.49.1`).
...
+docker run --rm --ipc=host --platform linux/amd64 -v "$(pwd):/work" -w /work -e PLAYWRIGHT_BASE_URL=http://host.docker.internal:3000 -e TIMEOUT=60000 -e WORKERS=1 mcr.microsoft.com/playwright:v<version>-jammy npx playwright test e2e/suites/screenshots --project=dev --update-snapshots
 ```
 
 **Prod snapshots** (prod stack must be running on port 3001 via `yarn build && yarn server:reset`):
 
 ```bash
-docker run --rm --ipc=host --platform linux/amd64 -v "$(pwd):/work" -w /work -e PLAYWRIGHT_BASE_URL=http://host.docker.internal:3001 -e TIMEOUT=60000 -e WORKERS=1 mcr.microsoft.com/playwright:v<version>-jammy npx playwright test observationDetailsScreenshots --project=prod --update-snapshots
+docker run --rm --ipc=host --platform linux/amd64 -v "$(pwd):/work" -w /work -e PLAYWRIGHT_BASE_URL=http://host.docker.internal:3001 -e TIMEOUT=60000 -e WORKERS=1 mcr.microsoft.com/playwright:v<version>-jammy npx playwright test e2e/suites/screenshots --project=prod --update-snapshots
```
ice1080: Should we add a script for this to package.json?

ice1080 (commented):
I'm starting to question the benefit of having separate dev and prod versions of every image. Is there no way to make them the same, or force these tests to only run if in prod or something? Many of these are identical

playwright/README.md
constanzauanini: Yes, that is a good idea!

constanzauanini:
> I'm starting to question the benefit of having separate dev and prod versions of every image. Is there no way to make them the same, or force these tests to only run if in prod or something? Many of these are identical

Yes, we can run them in prod only. That would be enough I think

.buildkite/scripts/run-e2e-tests.sh
```diff
@@ -14,22 +11,43 @@ docker pull "mcr.microsoft.com/playwright:v${PLAYWRIGHT_VERSION}-jammy"
...
+# that enables the Reports nav item).
+echo "--- :docker: Reset backend for screenshot tests"
+yarn server:reset
+run_tests "Run screenshot tests" "playwright/e2e/suites/screenshots"
+
+# Reset again before the remaining tests so they also start from a known state.
+echo "--- :docker: Reset backend for remaining e2e tests"
+yarn server:reset
+# suites directory, excluding the screenshots subdirectory.
+run_tests "Run remaining end-to-end tests" "playwright/e2e/suites/*.spec.ts"
```
ice1080: I wonder if we could instead update the test ordering so that screenshot tests run first then the rest of them. That would save the extra 1-2 minutes that the server reset takes

ice1080: I don't have a strong preference for this, just mentioning that this change slows down CI by an extra 1-2 mins for the additional server reset

constanzauanini: Yes, I think we can. Let me try that

ice1080 (approved):

PR #5617, opened 2026-04-23
SW-8303 Restrict exploration of Virtual Walkthroughs

constanzauanini:
Computes bounds as a cube centered on originPosition with a radius of 0,5x the initial camera-to-origin distance, so the user has reasonable room to explore but can't wander arbitrarily far. The 0,5× multiplier is a default but we can adjust it if we want tighter or looser restrictions


src/components/GaussianSplat/camera-bounds.ts
ice1080, line 1: I think `camera-bounds.ts` and `CameraBounds.tsx` are no longer needed, right?

src/scenes/VirtualWalkthrough/VirtualWalkthroughModal.tsx
```diff
@@ -73,7 +85,13 @@ const VirtualWalkthroughViewer = ({ fileId, observationId, organizationId }: Vir
       <Entity name='camera-root'>
         <Entity name='camera'>
           <Camera clearColor='#EAF8FF' fov={60} />
-          <Script script={CameraControls} moveSpeed={0.3} moveFastSpeed={0.5} moveSlowSpeed={0.15} rotateSpeed={0.1} />
+          <Script
+            script={WalkthroughCamera}
+            pitchMin={-30}
+            pitchMax={5}
```
ice1080: I didn't even notice this when I was using it. We may need to remove this restriction, but I'll leave it up to someone else to make that decision.

src/scenes/VirtualWalkthrough/VirtualWalkthroughModal.tsx
```diff
@@ -58,6 +59,17 @@ const VirtualWalkthroughViewer = ({ fileId, observationId, organizationId }: Vir
     [data]
   );
 
+  // Circular exploration radius — half the camera-to-origin distance.
+  const boundsXZRadius = useMemo(() => {
+    const dx = cameraPosition[0] - origin[0];
+    const dy = cameraPosition[1] - origin[1];
+    const dz = cameraPosition[2] - origin[2];
```
ice1080, line 66: Just a note, I believe the starting camera position is currently hardcoded in `VirtualMonitoringPlot.tsx` as `DEFAULT_POSITION`. I added the ability to override that, but it's only been overridden manually in the DB for a few plots so far. I think eventually we may want to add the ability to override a default model radius, but for now this should be fine for most cases (just know that the radius will change if the starting camera position ever changes).

src/scenes/VirtualWalkthrough/VirtualWalkthroughModal.tsx
constanzauanini: I just asked in slack about this

ice1080:
All looks good, though I still think we don't need those other 2 files

src/components/GaussianSplat/camera-bounds.ts
constanzauanini: True, we are not using them anymore

src/scenes/VirtualWalkthrough/VirtualWalkthroughModal.tsx
```diff
@@ -58,6 +59,17 @@ const VirtualWalkthroughViewer = ({ fileId, observationId, organizationId }: Vir
...
+    return Math.sqrt(dx * dx + dy * dy + dz * dz) * 0.5;
+  }, [cameraPosition, origin]);
```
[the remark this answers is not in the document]
ice1080: @constanzauanini I think the description may need updated, not sure

ice1080 (approved):

src/scenes/VirtualWalkthrough/VirtualWalkthroughModal.tsx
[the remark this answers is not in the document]
constanzauanini: Oh, yes, I wrote that at the begginig

PR #5636, opened 2026-04-27
Allow for seeding the random chart color section during e2e tests

alexTerraformation:
Use a seedable rng (mulberry32) when running e2e tests to keep chart.js colors consistent between test runs.


constanzauanini:
Great! Did you test adding a screenshot of a chart?

PR #5637, opened 2026-04-28
SW-8306 Fix unreleased.log generation in buildkite CI pipeline

constanzauanini:


.buildkite/scripts/release-notes.sh
```diff
@@ -20,18 +20,6 @@ LAST_VERSION=$(git tag --list --sort=creatordate 'v[0-9]*' | tail -n2 | head -n1
 .buildkite/scripts/lib/fetch-tag.sh "$THIS_VERSION"
 
 CHANGELOG=$(git log "$LAST_VERSION".."$THIS_VERSION" --pretty=format:"%s")
-LAST_VERSION=$(echo "$TAGS" | tail -n2 | head -n1)
```
constanzauanini, line 23: This was referencing $TAGS which was never defined. LAST_VERSION is already defined in line 15. This is not necessary. fetch-tag.sh and CHANGELOG is also done before (line 17-22), so all this can be removed

ice1080 (approved):

PR #5638, opened 2026-04-28
SW-8313 Not all thumbnails showing up in virtual walkthroughs table

constanzauanini:


src/scenes/VirtualWalkthrough/VirtualWalkthroughsTable.tsx
```diff
@@ -280,13 +288,20 @@ export default function VirtualWalkthroughsTable({
...
+            observationId={walkthroughModalFile.observationId}
+            fileId={walkthroughModalFile.fileId}
+            onClose={() => setWalkthroughModalFile(undefined)}
+          />
+        ) : (
+          <VirtualWalkthroughModal
+            organizationId={organizationId}
+            fileId={walkthroughModalFile.fileId}
+            onClose={() => setWalkthroughModalFile(undefined)}
+          />
```
ice1080: Could we update `VirtualWalkthroughModal` here so that this doesn't have to be a conditional? E.g.

```
          <VirtualWalkthroughModal
            observationId={walkthroughModalFile.observationId}
            organizationId={organizationId}
            fileId={walkthroughModalFile.fileId}
            onClose={() => setWalkthroughModalFile(undefined)}
          />
```

And then inside `VirtualWalkthroughModal` it checks for observation id and if that is undefined then it checks for organization id?

ice1080 (approved):

PR #5645, opened 2026-04-29
SW-7693: Planting Progress: Map UI/UX updates

joshbuchea:
## Summary

**NOTE: Changes in this PR were assisted by Claude Opus 4.7. I'm not super familiar with the new map components, so I'm happy to make adjustments if anything was implemented in an undesirable way.**

- Replaced the legacy planting progress map with the new `MapComponent`, including a custom drawer for site/stratum/substratum details.
- Added an "All Planting Sites" option to the map's planting site selector that aggregates strata across the org.
- "See Withdrawal History" link in the drawer navigates to the withdrawal history table pre-filtered by substratum (all substrata in the stratum, or just the selected substratum).
- Aggregated "Planting Site Area" and "Planting Complete Area" totals above the map now reflect all sites when "All Planting Sites" is selected.
- Avoided the `-1` sentinel previously used for "All Planting Sites" by introducing a string sentinel internal to `PlantingSiteSelector` that's never persisted to user preferences.
- Made the All-Sites fetch lazy: the org-wide site list is only fetched when "All Planting Sites" is selected, otherwise only the selected single site is fetched.


tommylau523 (approved):
Looks great! Love the planting site selector change as well.

PR #5646, opened 2026-04-29
SW-8321 Clicking "Mark Needs Attention" for observation models doesn't do anything

constanzauanini:


src/queries/generated/observationSplats.ts
```diff
@@ -47,6 +47,16 @@ const injectedRtkApi = api.injectEndpoints({
         url: `/api/v1/tracking/observations/${queryArg.observationId}/splats/${queryArg.fileId}/info`,
       }),
     }),
+    setObservationSplatNeedsAttention: build.mutation<
+      SetObservationSplatNeedsAttentionApiResponse,
```
tommylau523, line 51: We should add the provided tag to the extension for splats

src/scenes/VirtualWalkthrough/VirtualWalkthroughsTable.tsx
```diff
@@ -180,39 +182,30 @@ export default function VirtualWalkthroughsTable({
   const FlagCell = useCallback(
     ({ cell }: { cell: MRT_Cell<OrganizationVirtualWalkthrough> }) => {
       const file = cell.row.original;
+      const setNeedsAttention = (needsAttention: boolean) => {
```
tommylau523: Can we make this into a separate useCallback? This is fine too, just for readability to have less nested functions.

src/queries/extensions/observationSplats.ts
```diff
@@ -11,6 +11,9 @@ api.enhanceEndpoints({
     setObservationSplatAnnotations: {
       invalidatesTags: (_results, _error, payload) => [{ type: QueryTagTypes.ObservationSplats, id: payload.fileId }],
     },
+    setObservationSplatNeedsAttention: {
+      invalidatesTags: [{ type: QueryTagTypes.OrganizationMedia, id: 'LIST' }],
```
tommylau523: Probably should use ObservationSplats instead. But we likely should combine ObservationSplats and OrganizationMedia into just the same Splats tag as well.

tommylau523 (approved):

PR #5648, opened 2026-04-29
SW-8306 Fix Could not find history error

constanzauanini:
In `git fetch --depth=$DEPTH origin "$HEAD_REF"` GitHub doesn't support fetching arbitrary SHAs by default. So the history never actually deepens beyond the initial 1 commit from the shallow clone, and merge-base --is-ancestor keeps failing regardless of how many loop iterations run.

By changing to `git fetch --depth=$DEPTH` the history actually deepens

### DIFF
--- a/.buildkite/scripts/lib/fetch-tag.sh
+++ b/.buildkite/scripts/lib/fetch-tag.sh
@@ -16,8 +16,6 @@ MAX_DEPTH=500
 STEP=50
 DEPTH=0
 
-HEAD_REF="$(git rev-parse HEAD)"
-
 while true; do
     if git merge-base --is-ancestor "$SEARCH_TAG" HEAD >/dev/null 2>&1; then
         echo "Found history from $SEARCH_TAG to HEAD"
@@ -32,5 +30,5 @@ while true; do
     DEPTH=$((DEPTH + STEP))
 
     echo "Fetching with depth=$DEPTH"
-    git fetch --depth=$DEPTH origin "$HEAD_REF"
+    git fetch --depth=$DEPTH
 done


ice1080:
I'm confused here. This same script works fine in terraware-server, so how could it be a git or GitHub issue?

ice1080:
Could the issue with `unreleased.log` be the difference with `deploy-docs.sh`? I see in terraware-server we have
```
if [[ -n "$LAST_TAG" ]]; then
    .buildkite/scripts/lib/fetch-tag.sh "$LAST_TAG"
    git log "${LAST_TAG}..HEAD" --oneline > docs/unreleased.log
else
    git log HEAD --oneline > docs/unreleased.log
fi
```

but in terraware-web it's only
```
.buildkite/scripts/lib/fetch-tag.sh "$LAST_TAG"
git log "${LAST_TAG}..HEAD" --oneline > docs/unreleased.log
```

constanzauanini:
> Could the issue with `unreleased.log` be the difference with `deploy-docs.sh`? I see in terraware-server we have
> 
> ```
> if [[ -n "$LAST_TAG" ]]; then
>     .buildkite/scripts/lib/fetch-tag.sh "$LAST_TAG"
>     git log "${LAST_TAG}..HEAD" --oneline > docs/unreleased.log
> else
>     git log HEAD --oneline > docs/unreleased.log
> fi
> ```
> 
> but in terraware-web it's only
> 
> ```
> .buildkite/scripts/lib/fetch-tag.sh "$LAST_TAG"
> git log "${LAST_TAG}..HEAD" --oneline > docs/unreleased.log
> ```

In terraware-web I see in deploy-docs.sh the same "if block" that we have in the server (line 24 to 29)

constanzauanini:
> I'm mainly hesitant to approve because `git fetch --depth=$DEPTH` will fetch the entire history, as opposed to the previous version which was in a slowly expanding search loop specifically so that the whole history wasn't fetched

According to Claude the --depth caps the depth regardless of whether a refspec is given. Removing `origin "$HEAD_REF" `allows handling detached HEAD / tag-triggered builds where there's no branch refspec

ice1080:
> In terraware-web I see in deploy-docs.sh the same "if block" that we have in the server (line 24 to 29)

That's totally my fault, I had them backwards. 

> According to Claude the --depth caps the depth regardless of whether a refspec is given. Removing `origin "$HEAD_REF" `allows handling detached HEAD / tag-triggered builds where there's no branch refspec

That makes sense, I just don't know why it works fine in terraware-server but not in terraware-web

ice1080 (approved):

PR #5651, opened 2026-04-30
Use deepen on fetch-tag to fix unreleased-log

constanzauanini:
Even after fetching to depth=500, git log origin/main -5 showed the same 5 commits every single iteration. The history wasn't growing because In Buildkite, the peakon/git-shallow-clone plugin checks out the build commit in detached HEAD with a shallow boundary right there. When git fetch --depth=50 runs, git goes to the remote, counts 50 commits back from the remote's tip, and fetches those. But the fetched objects only get wired to a tracking ref if the fetch refspec explicitly maps them there. Without that mapping, the objects land in the object store but the local refs (like origin/main) don't move, they stay pointing at the original depth-1 shallow boundary. So git merge-base --is-ancestor tag HEAD keeps checking against the same stale shallow history on every loop iteration, and the tag is never found.
-deepen=50 fix all of this because it doesn't talk about the remote tip at all, it just says "take whatever   
  shallow boundary you have right now and extend it by 50 commits." Git correctly updates the local shallow file and the history grows from HEAD outward, which is exactly what the loop needs.


ice1080 (approved):
Super interesting. I still don't understand why server works, but at least this is working

PR #5663, opened 2026-05-05
SW-8342 Combine multiple model viewers

constanzauanini:


ice1080:
@constanzauanini can you merge in main now that #5666 is fixed so I can compare the old and new viewers?

ice1080 (commented):
Some differences between viewers that I noticed:
- No full screen button when opened from Virtual Walkthroughs page
- No edit button when opened from Virtual Walkthroughs page
- No observation information below plot when opened from Virtual Walkthroughs page
- Full screen button no longer works correctly when opened from observations modal
- When opened from Virtual Walkthroughs page, the url doesn't contain the ids needed to send a link to someone that opens the same file

Regarding the code, I'm glad we got rid of the `VirtualPlotModal`, but I wonder if we could go even further and combine the `VirtualMonitoringPlotPage` and the `VirtualWalkthroughModal`. I don't know for sure if this would work, but it could help reduce the differences.

These are just things I noticed and have suggestions on, but I'm happy to talk through things if you think any of this isn't the right approach, or want to talk with Alex about keeping them different!

constanzauanini:
> Some differences between viewers that I noticed:
> 
> * No full screen button when opened from Virtual Walkthroughs page
> * No edit button when opened from Virtual Walkthroughs page
> * No observation information below plot when opened from Virtual Walkthroughs page
> * Full screen button no longer works correctly when opened from observations modal
> * When opened from Virtual Walkthroughs page, the url doesn't contain the ids needed to send a link to someone that opens the same file
> 
> Regarding the code, I'm glad we got rid of the `VirtualPlotModal`, but I wonder if we could go even further and combine the `VirtualMonitoringPlotPage` and the `VirtualWalkthroughModal`. I don't know for sure if this would work, but it could help reduce the differences.
> 
> These are just things I noticed and have suggestions on, but I'm happy to talk through things if you think any of this isn't the right approach, or want to talk with Alex about keeping them different!

I can try combining `VirtualMonitoringPlotPage` and the `VirtualWalkthroughModal`. And also I will try to fix all the other differences.
Edit button is for annotations, right? I wasn't sure if we had a way to save annotations in Virtual Walthroughs

ice1080:
> I can try combining `VirtualMonitoringPlotPage` and the `VirtualWalkthroughModal`. And also I will try to fix all the other differences. Edit button is for annotations, right? I wasn't sure if we had a way to same annotations in Virtual Walthroughs

Correct, edit button is for annotations. In the database there's actually no difference between observation models and organization models, but the BE has slightly different behavior. There is a different endpoint for saving splat annotations, so the annotation code would need to account for that. 

@constanzauanini if it's easier to do the work I mentioned in a different PR, that's also fine and I can review the changes as-is. Let me know what you'd prefer!

constanzauanini:
> > I can try combining `VirtualMonitoringPlotPage` and the `VirtualWalkthroughModal`. And also I will try to fix all the other differences. Edit button is for annotations, right? I wasn't sure if we had a way to same annotations in Virtual Walthroughs
> 
> Correct, edit button is for annotations. In the database there's actually no difference between observation models and organization models, but the BE has slightly different behavior. There is a different endpoint for saving splat annotations, so the annotation code would need to account for that.
> 
> @constanzauanini if it's easier to do the work I mentioned in a different PR, that's also fine and I can review the changes as-is. Let me know what you'd prefer!

I will fix in this same PR, so we can merge all the changes together

src/components/GaussianSplat/walkthrough-camera.ts
```diff
@@ -176,6 +176,36 @@ export class WalkthroughCamera extends Script {
...
+    return this._targetYaw;
+  }
+
+  get currentPitch(): number {
+    return this._targetPitch;
+  }
+
+  /**
+   * Orbit the camera by `yawDelta` degrees around `boundsCenter`.
+   * Called by AutoRotator when WalkthroughCamera is active instead of CameraControls.
```
ice1080: Can just change to `Called by AutoRotator` since we won't be using `CameraControls` anymore

ice1080 (commented):
There's a bug right now where adding an annotation to an org splat and typing in the Title field causes the splat to disappear, but there's a chance that fixing the graphite issue fixes this too

constanzauanini:
> There's a bug right now where adding an annotation to an org splat and typing in the Title field causes the splat to disappear, but there's a chance that fixing the graphite issue fixes this too

I don't think that bug is related to the changes. It is also happening on staging on observations splats

ice1080 (approved):

ice1080:
> I don't think that bug is related to the changes. It is also happening on staging on observations splats

Weird, I thought I saw it working for observation splats. It's a regression either way, and can be addressed in a later PR

PR #5675, opened 2026-05-08
SW-8345 Reject video uploads if length is greater than 6 minutes

constanzauanini:


src/strings/csv/en.csv
```diff
@@ -1174,13 +1174,13 @@ NO_NURSERIES_NON_ADMIN_MSG,"Before you can add and manage your nursery inventory
...
 NO_PLANTING_SITES_DESCRIPTION,Start by adding a planting site. Adding a planting site will allow you to track where seedlings go after they leave the nursery.,
-NO_PLANTING_SITES_NON_ADMIN_MSG,"Only admins can add planting sites; please reach out to yours for assistance.",
+NO_PLANTING_SITES_NON_ADMIN_MSG,Only admins can add planting sites; please reach out to yours for assistance.,
 NO_PLANTING_SITES_TITLE,"Just a moment, let’s add a planting site first.",
 NO_PLANTS,No Plants,
 NO_PROJECT,No Project,
 NO_PROJECT_SELECTED,No Project Selected,
 NO_PROJECT_SPECIFIC_METRICS_TO_SHOW,No Project Specific Metrics to show.,
-NO_PROJECTS_NON_ADMIN_MSG,"Only admins can add projects; please reach out to yours for assistance.",
+NO_PROJECTS_NON_ADMIN_MSG,Only admins can add projects; please reach out to yours for assistance.,
```
ice1080, line 1183: Are these used in a later PR? They're also not translated yet

ice1080 (approved):

src/strings/csv/en.csv
constanzauanini: Maybe the quotes "" were removed when I ran yarn:translate?
They are not related, but it was done automatically I guess

ice1080: Oh I didn't even see that the messages existed before, I just assumed it was new messaging

PR #5690, opened 2026-05-11
SW-8349 Adding new annotations to a splat causes the model to disappear

constanzauanini:
VirtualWalkthroughViewer was re-rendering on every keystroke in the Title field. That cascades to SplatModel which was re-rendering too. The existing useMemo around the <SplatModel /> JSX keeps the element reference stable, but that alone doesn't stop the SplatModel component from re-rendering. React still calls the function component on parent re-render.

With React.memo inSplatModel VirtualWalkthroughViewer still re-renders on every keystroke but React.memo stops the cascade so SplatModel no longer re-renders


ice1080 (approved):
Nice find!

PR #5696, opened 2026-05-12
Allow enabling features by org id

constanzauanini:
This will allow features to be enabled by setting the `allowedOrganizationIds` prop. If the `allowedOrganizationIds` prop is set, the feature will not appear on the feature flag opt-in page and will only be enabled for those organizations.

### DIFF
--- a/src/features.ts
+++ b/src/features.ts
@@ -15,6 +15,8 @@ export type Feature = {
   allowInternalProduction: boolean;
   description: string[];
   disclosure: string[];
+  // if set, the feature is enabled only for these organization IDs and cannot be toggled by users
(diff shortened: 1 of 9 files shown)


src/features.ts
```diff
@@ -43,6 +45,7 @@ export const OPT_IN_FEATURES: Feature[] = [
     allowInternalProduction: false,
     description: ['Support for virtual monitoring plots'],
     disclosure: ['This is a WIP'],
+    allowedOrganizationIds: [243, 123],
```
tommylau523, line 48: I think this makes it difficult for test/staging. I also think there may be some concerns about pushing hard-coded organization IDs to a puiblic repo.

PR #5704, opened 2026-05-14
SW-8350 Plants Dashboard: Show Average Survival Rate for Project based on AWASSR

constanzauanini:


ice1080 (approved):
So we don't display any of the Highest/Lowest when doing weighted survival rates?

constanzauanini:
> So we don't display any of the Highest/Lowest when doing weighted survival rates?

We have to display them, but I don't think we have the data for that yet

PR #5722, opened 2026-05-19
SW-5197 Create tab in Accessions List View for Species-Aggregated List View

constanzauanini:


src/components/seeds/database/AccessionsBySpeciesTable.tsx
```diff
@@ -0,0 +1,256 @@
...
+            data={speciesRows}
+            enableEditing={false}
+            enableSorting={true}
+            enableGlobalFilter={true}
+            enableColumnFilters={true}
+            enableColumnOrdering={true}
+            storageKey={TABLE_STATE_STORAGE_KEY}
+            enablePagination={true}
+            enableTopToolbar={true}
+            enableBottomToolbar={true}
```
ice1080: Are any of these defaults that we can omit, or are we preferring to be explicit with tables?

src/components/seeds/database/AccessionsBySpeciesTable.tsx
```diff
@@ -0,0 +1,256 @@
...
+            }}
+            sx={{ padding: 0 }}
+          />
+        ) : (
+          <Box sx={{ padding: '32px', textAlign: 'center' }}>{strings.GENERIC_ERROR}</Box>
+        )
+      ) : (
+        <Box sx={{ display: 'flex', justifyContent: 'center', padding: '32px' }}>
+          <CircularProgress />
+        </Box>
```
ice1080: The table component is so long that I had to scroll up to see what state would cause the error or `CircularProgress`. Perhaps we can move those to above the main return:
```tsx
if (searchResults === undefined) {
  return <Box sx={{ display: 'flex', justifyContent: 'center', padding: '32px' }}>
    <CircularProgress />
  </Box>
} else if (searchResults === null) {
  ...
}

return ...
```

ice1080 (commented):
Comments apply to both tables, though I see now that the same basic code was there before this PR

ice1080 (approved):

PR #5739, opened 2026-05-26
SW-8395 Add planting season modal

constanzauanini:


src/strings/csv/en.csv
```diff
@@ -387,6 +387,8 @@ COOKIES_ACCEPT,Accept Cookies,
 COOKIES_DECLINE,Decline Cookies,
 COOKIES_DESCRIPTION,Terraware uses performance cookies to analyze and improve the quality of the site and enhance your user experience.,
 COOKIES_LEARN_MORE,Learn more about our cookie policy.,
+COPY_PREVIOUS_SEASON,Copy previous season,
+COPYING_SPECIES_ACROSS_SUBSTRATUM,"Copying {0} species across {1} substratum, with all targets set to 0",
```
ice1080: This should be plural: `across {1} substratum` -> `across {1} substrata`

src/scenes/NurseryRouter/PlantingSeasons/index.tsx
```diff
@@ -56,12 +58,21 @@ const PlantingSeasons = (): JSX.Element => {
...
+  const [addModalOpen, setAddModalOpen] = useState(false);
+
   const onAddPlantingSeason = useCallback(() => {
-    // TODO
+    setAddModalOpen(true);
+  }, []);
+
+  const onCloseAddModal = useCallback(() => {
+    setAddModalOpen(false);
   }, []);
```
ice1080: You might be able to change this to `useBoolean` so that you don't have to create separate `useCallbacks` for them:
```ts
const [addModalOpen, , onAddPlantingSeason, onCloseAddModal] = useBoolean(false);
```

src/scenes/NurseryRouter/PlantingSeasons/AddPlantingSeasonModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+  const [copyPrevious, setCopyPrevious] = useState<boolean>(false);
+  const [seasonToCopyId, setSeasonToCopyId] = useState<number | undefined>();
+  const [seasonName, setSeasonName] = useState<string>('');
+  const [startDate, setStartDate] = useState<string | undefined>();
+  const [endDate, setEndDate] = useState<string | undefined>();
+
+  const [plantingSiteError, setPlantingSiteError] = useState<string>('');
+  const [seasonNameError, setSeasonNameError] = useState<string>('');
+  const [startDateError, setStartDateError] = useState<string>('');
+  const [endDateError, setEndDateError] = useState<string>('');
```
ice1080: Don't we normally handle this with a single `validate` flag that checks for null values in the render block?

ice1080 (changes requested):

ice1080 (approved):
I _think_ the graphite comment is incorrect since we're showing what the new season will look like after the copy, but I'll let you confirm @constanzauanini . Approving for now to unblock

constanzauanini:
> I _think_ the graphite comment is incorrect since we're showing what the new season will look like after the copy, but I'll let you confirm @constanzauanini . Approving for now to unblock

I think the graphite comment is correct. Because when selecting the planting season to copy we only know that data, we don't have data from this new season. I will update it to use the count from the target as suggested

PR #5749, opened 2026-05-29
Remove use of old planting season

constanzauanini:


src/hooks/useOrganizationPlantingSites.ts
```diff
@@ -8,6 +8,11 @@ type UseOrganizationPlantingSitesProps = {
   organizationId?: number;
 };
 
+export type OrganizationPlantingSite = Omit<PlantingSitePayload, 'plantingSeasons'>;
```
ice1080: The BE now has this removed so it shouldn't need a separate type, but that can be cleaned up in a later PR

ice1080 (approved):

PR #5770, opened 2026-06-03
SW-8445 Inventory Planning view

constanzauanini:


src/queries/extensions/plantingSeasons.ts
```diff
@@ -47,17 +47,20 @@ api.enhanceEndpoints({
       invalidatesTags: (_result, _error, arg) => [
         { type: QueryTagTypes.PlantingSeasons, id: `${arg.plantingSeasonId}-targets` },
         { type: QueryTagTypes.PlantingSeasons, id: `${arg.plantingSeasonId}-summary` },
+        { type: QueryTagTypes.PlantingSeasons, id: 'INVENTORY_PLANNING' },
```
ice1080: This feels slightly weird to use this id instead of something like 
```ts
{ type: QueryTagTypes.InventoryPlanning, id: 'LIST' },
```

Every other instance uses either an actual `id` or `LIST`

src/queries/search/inventoryPlanning.ts
```diff
@@ -0,0 +1,276 @@
...
+                  field: 'batches.id',
+                  values: [null],
+                },
+              },
+            ],
+          },
+          count: 0,
+        },
+      }),
+      providesTags: [{ type: QueryTagTypes.NurseryWithdrawals }],
```
ice1080: I think this tag needs to change

src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -138,7 +151,14 @@ export default function NavBar({
       />
     );
 
-    return showNurseryWithdrawals ? [inventoryMenu, withdrawalLogMenu] : [inventoryMenu];
+    const items = [inventoryMenu];
+    if (isPlantingSeasonsEnabled) {
```
ice1080: I think this menu item is only supposed to be available if there is at least one planting season with species targets

src/scenes/InventoryPlanningRouter/InventoryPlanningView.tsx
```diff
@@ -0,0 +1,489 @@
...
+
+  const { data: plantingSitesData } = useListPlantingSitesQuery(
+    { organizationId: organizationId ?? 0 },
+    { skip: !organizationId }
+  );
+  const [listPlantingSeasons, { data: plantingSeasonsData }] = useLazyListPlantingSeasonsQuery();
+
+  useEffect(() => {
+    if (organizationId) {
+      void listPlantingSeasons({ organizationId });
```
ice1080: This needs `true` as the second arg to use caching

src/scenes/InventoryPlanningRouter/InventoryPlanningView.tsx
```diff
@@ -0,0 +1,489 @@
...
+        ) : (
+          <>
+            <Typography fontSize='16px' color={allocatedColor} fontWeight={600}>
+              {season.allocated.toLocaleString(activeLocale || undefined)}
+            </Typography>
+            <Button
+              icon='iconEdit'
+              onClick={() => {
+                setDraftValue(season.allocated.toString());
+                setEditing(true);
```
ice1080: This might be coming in a later PR, but I think this table isn't supposed to be editable if the planting season is closed.

src/queries/search/inventoryPlanning.ts
constanzauanini: Yes, probably we don't need a tag here because NurseryInventoryService is not in rtkQuery

src/scenes/InventoryPlanningRouter/InventoryPlanningView.tsx
constanzauanini: Oh, thanks. I will add it in this PR

src/queries/extensions/plantingSeasons.ts
```diff
@@ -47,20 +47,20 @@ api.enhanceEndpoints({
       invalidatesTags: (_result, _error, arg) => [
         { type: QueryTagTypes.PlantingSeasons, id: `${arg.plantingSeasonId}-targets` },
         { type: QueryTagTypes.PlantingSeasons, id: `${arg.plantingSeasonId}-summary` },
-        { type: QueryTagTypes.PlantingSeasons, id: 'INVENTORY_PLANNING' },
+        { type: QueryTagTypes.PlantingSeasons, id: 'LIST' },
```
ice1080: Oh sorry, this will invalidate the list of planting seasons whenever a target changes. I was saying that I think we need a new query tag type such as `InventoryPlanning` in addition to `PlantingSeasons` and these endpoints should invalidate that with an id of `LIST`, that way we're not overfetching other tables

constanzauanini: Oh, ok! I can change that!

src/queries/search/inventoryPlanning.ts
```diff
@@ -0,0 +1,281 @@
...
+            sortOrder: [{ field: 'startDate', direction: 'Ascending' }],
+            count: 0,
+          },
+        };
+      },
+      providesTags: [{ type: QueryTagTypes.InventoryPlanning, id: 'LIST' }],
+      transformResponse: (response: InventoryPlanningSeasonSearchResponse) => response.results,
+    }),
+
+    listInventoryPlanningSpeciesAvailable: build.query<Map<number, number>, number>({
```
tommylau523, line 103: This one should have a tag of SpeciesInventory I think.

src/queries/search/inventoryPlanning.ts
```diff
@@ -0,0 +1,281 @@
...
+            search: {
+              operation: 'and',
+              children: filters,
+            },
+            sortOrder: [{ field: 'startDate', direction: 'Ascending' }],
+            count: 0,
+          },
+        };
+      },
+      providesTags: [{ type: QueryTagTypes.InventoryPlanning, id: 'LIST' }],
```
tommylau523: I think we can use PlantingSeasons tag for this

src/queries/extensions/plantingSeasons.ts
```diff
@@ -47,17 +47,20 @@ api.enhanceEndpoints({
       invalidatesTags: (_result, _error, arg) => [
         { type: QueryTagTypes.PlantingSeasons, id: `${arg.plantingSeasonId}-targets` },
```
tommylau523: Let's simplify to just the IDs. We may over-fetch in some cases but that is better than the extra complexity.

Right now, if you invalidate `{ type: QueryTagTypes.PlantingSeasons, id: plantingSeasonId }` maybe by deleting a planting season, the species target will stay because the target ID is not invalidatesd.

If we want more granular control, we should have a new tag called PlantingSeasonsTargets as well.

src/queries/extensions/plantingSeasons.ts
```diff
@@ -47,17 +47,20 @@ api.enhanceEndpoints({
...
+        { type: QueryTagTypes.InventoryPlanning, id: 'LIST' },
       ],
     },
     upsertAllocatedSpecies: {
       invalidatesTags: (_result, _error, arg) => [
         { type: QueryTagTypes.PlantingSeasons, id: `${arg.plantingSeasonId}-summary` },
+        { type: QueryTagTypes.InventoryPlanning, id: 'LIST' },
       ],
     },
     getScheduledPlantingDates: {
```
tommylau523, line 66: <p class='graphite__hidden'><i>[Re: line +73]</i></p>
These should be a seprate set of tags, PlantingSeasonDates 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/terraware/terraware-web/5770?utm_source=unchanged-line-comment">Graphite</a>.</p>

src/queries/search/inventoryPlanning.ts
constanzauanini: I will add it, but not sure what should invalidate this. I think we don't have batches/inventory in RTK query yet

ice1080: I think it probably makes more sense to make it just `{ type: QueryTagTypes.InventoryPlanning, id: 'LIST' }` since those should provide and invalidate together

ice1080 (approved):
Just one minor comment to make sure the species search dropdown invalidates at the same time as the rest of InventoryPlanning

src/queries/search/inventoryPlanning.ts
constanzauanini: Yes, I think that makes more sense

PR #5774, opened 2026-06-03
Patch RTK Query multi-form by detecting Blobs and using form data

tommylau523:
APIs with content type `multipart/form-data` fails due to lack of codegen support. In these cases, we intercept payloads containing a Blob and casts those to be FormData.

PR #5775, opened 2026-06-03
Migrate accessions to RTK Query and fetch via useAccession

tommylau523:
Replace the remaining AccessionService calls in the accessions component
tree with the generated RTK Query hooks (accessionsV1/V2). Mutations use the
generated hooks with .unwrap() + try/catch; reads use auto-firing queries.

Add a shared useAccession hook so the modals/panels fetch the accession via
RTK Query (cache-keyed by id, so no duplicate requests) instead of having it
prop-drilled down from Accession2View. Drop the now-redundant reload prop —
mutations invalidate the Accessions cache tag, so the shared query
auto-refetches.

Remove the now-unused AccessionService functions, keeping only the pure
utilities (getTransitionToStates, getParsedCoords) and transferToNursery,
which is still used by the batches Redux thunk.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/hooks/useAccession.ts
+++ b/src/hooks/useAccession.ts
@@ -1,10 +1,6 @@
 import { useGetAccessionQuery } from 'src/queries/generated/accessionsV2';
 import { Accession } from 'src/types/Accession';
 
-/**
- * Fetches a single accession via RTK Query. The query is cache-keyed by id, so multiple components
- * requesting the same accession share one cached result (no duplicate requests).
- */
 const useAccession = (accessionId: number) => {
   const { currentData, isFetching, isSuccess, refetch } = useGetAccessionQuery(accessionId, {
     skip: !accessionId,
--- a/src/hooks/useAccession.ts
+++ b/src/hooks/useAccession.ts
@@ -1,0 +1,18 @@
+import { useGetAccessionQuery } from 'src/queries/generated/accessionsV2';
+import { Accession } from 'src/types/Accession';
+
+/**
+ * Fetches a single accession via RTK Query. The query is cache-keyed by id, so multiple components
+ * requesting the same accession share one cached result (no duplicate requests).
+ */
+const useAccession = (accessionId: number) => {
+  const { currentData, isFetching, isSuccess, refetch } = useGetAccessionQuery(accessionId, {
+    skip: !accessionId,
+  });
+
+  const accession: Accession | undefined = currentData?.accession;
+
+  return { accession, refetch, isLoading: isFetching, isSuccess };
+};
+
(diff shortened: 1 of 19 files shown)


tommylau523:
* **#5775** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/5775?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-web/5775?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5774** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/5774?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/hooks/useAccession.ts
```diff
@@ -0,0 +1,18 @@
+import { useGetAccessionQuery } from 'src/queries/generated/accessionsV2';
+import { Accession } from 'src/types/Accession';
+
+/**
+ * Fetches a single accession via RTK Query. The query is cache-keyed by id, so multiple components
+ * requesting the same accession share one cached result (no duplicate requests).
+ */
```
ice1080: I think this comment is probably unnecessary since that's how we're already doing every other rtk query

src/scenes/AccessionsRouter/Accession2View.tsx
```diff
@@ -102,41 +110,25 @@ export default function Accession2View(): JSX.Element {
...
-        }
-      }
-    };
-
-    if (accessionId !== undefined) {
-      void populateAccession();
-    } else {
-      setAccession(undefined);
+    if (accessionId) {
+      void refetchAccession();
```
ice1080, line 115: You probably already know this but we should be able to get rid of this refetch stuff once `projectAssign` uses rtk query (assuming we setup invalidation properly)

tommylau523: Good catch! I will at least add a comment for now. After projects API are updated to use RTK Query, this can be removed.

ice1080 (approved):

tommylau523:
### Merge activity

* **Jun 9, 3:14 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-web/5775).
* **Jun 9, 3:14 PM UTC**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-web/5775).

PR #5776, opened 2026-06-04
SW-8339: FE: Observations in Activity Log

joshbuchea:
This PR implements observations in the activity log.

Co-authored by Claude.


src/components/ActivityLog/ActivityDetailsForm.tsx
```diff
@@ -390,10 +482,18 @@ export default function ActivityDetailsForm({ activityId, projectId }: ActivityD
...
       navToActivityLog();
     } else if (syncActivityMediaRequest?.status === 'success') {
+      if (activityId !== undefined) {
+        dispatch(
+          baseApi.util.invalidateTags([
+            { type: QueryTagTypes.Activities, id: activityId },
+            { type: QueryTagTypes.Activities, id: 'LIST' },
+          ])
+        );
+      }
```
ice1080, line 492: Huh, I didn't realize redux could dispatch an invalidation of rtk query. It makes sense obviously, I just hadn't see that before.
I think eventually we'll want everything to be rtk query and `syncActivityMedia` will invalidate these tags in the config, but for now this works well

src/types/Activity.ts
ice1080, line 1: These functions seem to belong in a separate class and be unit tested instead of in the `types` file

ice1080 (approved):
Just one comment that you can choose if you want to address or ignore

src/types/Activity.ts
joshbuchea: Good flag, updated ✅

ice1080 (approved):

PR #5779, opened 2026-06-05
Notify when saving a planting date

constanzauanini:


src/scenes/NurseryRouter/PlantingSeasons/PlantingDatesTab.tsx
```diff
@@ -283,23 +283,28 @@ const PlantingDateForm = ({
...
+  const performSave = async (notifyOptions?: { note: string }) => {
     if (!date) {
       setValidate(true);
       return false;
     }
     const speciesPayload = buildPayloadSpecies();
+    const payload = {
+      date,
+      species: speciesPayload,
+      ...(notifyOptions ? { createNurseryRequest: true, nurseryRequestNotes: notifyOptions.note } : {}),
```
ice1080: Does this still set `createNurseryRequest` if the user doesn't add the optional note?

constanzauanini: No, because there is no notifyOptions, so the notify options object is empty. Do we need to set it to false?

constanzauanini: Oh, I see the issue now. We want to set it to true when no optional note but wanting to notify! Will fix it

ice1080 (approved):

PR #5786, opened 2026-06-08
SW-8456 Record withdrawal for planting date

constanzauanini:


src/scenes/NurseryRouter/PlantingDateRequestsTabContent.tsx
```diff
@@ -70,6 +72,13 @@ const PlantingDateRequestsTabContent = (): JSX.Element => {
   );
 
   const rows = requests ?? [];
+  const [withdrawRequest, setWithdrawRequest] = useState<PlantingDateRequestRow | undefined>(undefined);
+
+  const onWithdrawn = () => {
+    if (organizationId) {
+      void listPlantingDateRequests({ organizationId, plantingSiteId, plantingSeasonId, speciesId }, false);
+    }
+  };
```
ice1080: Will this one happen automatically with RTK query if the tags are correct?

src/queries/search/batchesForWithdraw.ts
```diff
@@ -0,0 +1,84 @@
...
+                type: 'Exact',
+                values: args.speciesIds.map(String),
+              },
+            ],
+          },
+          sortOrder: [{ field: 'batchNumber', direction: 'Ascending' }],
+          count: 0,
+        },
+      }),
+      providesTags: [{ type: QueryTagTypes.NurseryWithdrawals, id: 'BATCHES_FOR_WITHDRAW' }],
```
ice1080: Can we use a different query tag type instead of custom id strings? Up until now the only id string we use is `LIST` and I think it makes sense to separate to a new tag whenever a new id type is needed

src/scenes/NurseryRouter/WithdrawFromBatchesModal.tsx
```diff
@@ -0,0 +1,611 @@
...
+      cancelButton,
+      backButton,
+      <Button
+        key='withdraw'
+        label={strings.WITHDRAW}
+        onClick={() => void onSubmit()}
+        disabled={isSaving || !canGoNextFromStep2}
+      />,
+    ];
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ice1080: Can you add a comment explaining why we're don't need all the dependencies?

src/scenes/NurseryRouter/WithdrawFromBatchesModal.tsx
```diff
@@ -0,0 +1,611 @@
...
+            <Typography fontSize='14px' fontWeight={600} textAlign='right'>
+              {strings.COVERAGE}
+            </Typography>
+          </Box>
+          {requestSpecies.map((s, index) => {
+            const ready = readyBySpecies.get(s.speciesId) ?? 0;
+            const coverage =
+              ready === 0 ? 'NOT_COVERED' : ready >= s.requestedQuantity ? 'COVERED' : 'PARTIALLY_COVERED';
+            const coverageColor =
+              coverage === 'COVERED'
```
ice1080: This isn't required, just noting that this file is pretty long and perhaps the RequestSpecies box could be a separate component. I'm not sure how you feel about moving components like that to separate files, but at least a separate component might help with readability of the file. Totally up to you though

src/scenes/NurseryRouter/PlantingDateRequestsTabContent.tsx
constanzauanini: You are right. I think it was not happening automatically because I was missing a tag invalidation

src/queries/extensions/nurseryWithdrawals.ts
```diff
@@ -4,7 +4,12 @@ import { QueryTagTypes } from '../tags';
 api.enhanceEndpoints({
   endpoints: {
     createBatchWithdrawal: {
-      invalidatesTags: [{ type: QueryTagTypes.NurseryWithdrawals, id: 'LIST' }, { type: QueryTagTypes.PlantingSites }],
+      invalidatesTags: [
+        { type: QueryTagTypes.NurseryWithdrawals, id: 'LIST' },
+        { type: QueryTagTypes.PlantingSites },
+        { type: QueryTagTypes.PlantingSeasonDates, id: 'REQUESTS' },
```
ice1080: Doesn't have to be in this PR, but I think I'll again flag this as a non-standard ID (though I recognize that I missed it in the last PR). Because PlantingSeasonDates don't directly invalidate PlantingDateRequests, I think it makes sense to break them into a separate tag type (a planting season date only overwrites its corresponding request if the notify is set to true. Otherwise the date and the request can have different data)

ice1080 (approved):
Approving, but marking one last thing

PR #5792, opened 2026-06-09
SW-8466: Disable link to observation if the console user is not in the project Org

joshbuchea:
This PR includes a change to only enable a link to an observation from an observation activity, when the user is a member of the project org.

Co-authored by Claude.


ice1080 (approved):
This works, though I'm wondering if it belongs in `acl.ts` instead. I recognize that we've only done functions that pass in an org, but we could do one that passes in both the org and project. Approving it either way, so up to you where you think it belongs

joshbuchea:
@ice1080 moved the logic to a function in `acl.ts` and added some tests.

ice1080:
I was more talking about using the existing `isAllowed` functionality in `acl.ts` like the rest of the functions in there. Usage would be something like this:
```ts
const hasOrgAccess = isAllowed('VIEW_ORG_ACTIVITIES', { organization, project, isAcceratorRoute });
```

Again I'll leave it up to you on if you want to change or not

PR #5794, opened 2026-06-10
SW-8482 Update batch withdrawals to use new modals

constanzauanini:
Replaced all batches withdrawal full-page flow with the new modal flow. A lot of code was taken from the old flow. Did it with Claude and manually tested


src/scenes/InventoryRouter/BatchWithdrawModal/SeedlingBatchBox.tsx
```diff
@@ -0,0 +1,106 @@
...
+                      ...(prev[batch.batchId] ?? emptyQuantities()),
+                      readyQuantityWithdrawn: Math.max(0, Number(v ?? 0)),
+                    },
+                  }))
+                }
+                min={0}
+                max={batch.readyQuantity}
+                errorText={
+                  exceeds ? strings.formatString(strings.EXCEEDS_READY_TO_PLANT, batch.readyQuantity).toString() : ''
+                }
```
ice1080, line 96: Is `exceeds` possible with this max value?

src/strings/csv/en.csv
```diff
@@ -1662,6 +1668,7 @@ PUBLISHED_REPORT_CONSOLE_WARNING,You are viewing the latest published report as
 PURPOSE,Purpose,
 PURPOSE_AND_DESTINATION,Purpose & Destination,
 PURPOSE_REQUIRED,Purpose *,
+PURPOSE_REQUIRED,Purpose *,
```
ice1080: Duplicate

src/scenes/InventoryRouter/BatchWithdrawModal/PurposeAndDestinationStep.tsx
```diff
@@ -0,0 +1,313 @@
...
+      void listPlantingSites({ organizationId }, true);
+      void listPlantingSeasons({ organizationId }, true);
+    }
+  }, [organizationId, listPlantingSites, listPlantingSeasons]);
+
+  useEffect(() => {
+    if (draft.plantingSiteId) {
+      void getPlantingSite({ id: draft.plantingSiteId }, true);
+    }
+  }, [draft.plantingSiteId, getPlantingSite]);
```
ice1080: It seems like `listPlantingSites` provides a tag for every site id in the org, so `getPlantingSite` is always going to use the result from that. Can you test commenting out the second `useEffect` to see if it can be removed?

src/scenes/InventoryRouter/BatchWithdrawModal/PurposeAndDestinationStep.tsx
```diff
@@ -0,0 +1,313 @@
...
+    }
+    return seasons.filter((s) => s.plantingSiteId === draft.plantingSiteId && s.status === 'Active');
+  }, [plantingSeasonsData, draft.plantingSiteId]);
+
+  // The "Planting Season (optional)" selector is shown only when the org has
+  // any planting season with species targets (regardless of status).
+  const showPlantingSeasonSelector = useMemo(
+    () => (plantingSeasonsData?.seasons ?? []).some((s) => s.speciesTargets.length > 0),
+    [plantingSeasonsData]
+  );
```
ice1080, line 89: Should this exclude `Closed` seasons?

src/scenes/InventoryRouter/BatchWithdrawModal/PurposeAndDestinationStep.tsx
```diff
@@ -0,0 +1,313 @@
...
+  // Default From: Nursery to the batches' nursery if they all share one.
+  useEffect(() => {
+    if (draft.fromFacilityId === undefined && batches.length > 0) {
+      const uniqueIds = Array.from(new Set(batches.map((b) => b.facilityId)));
+      if (uniqueIds.length === 1) {
+        onChange({ fromFacilityId: uniqueIds[0] });
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [batches]);
```
ice1080: I think claude likes to do this and it doesn't understand why. Can you add a comment explaining why `exhaustive-deps` is disabled, or update the dependencies to be exhaustive?

src/scenes/InventoryRouter/BatchWithdrawModal/SeedlingBatchBox.tsx
constanzauanini: Yes, the max value only limits the arrows, but not what can be typed

src/scenes/InventoryRouter/BatchWithdrawModal/PurposeAndDestinationStep.tsx
constanzauanini: You are right. Looks like this second useEffect is not needed

src/scenes/InventoryRouter/BatchWithdrawModal/PurposeAndDestinationStep.tsx
constanzauanini: Probably yes. I wil exclude them

src/scenes/InventoryRouter/BatchWithdrawModal/PurposeAndDestinationStep.tsx
constanzauanini: Yes, I'm not sure why it keeps doing this. I will try to include all dependencies

ice1080 (approved):

PR #5818, opened 2026-06-12
SW-8463: Needed to reload page to see caption updates (FOLLOW-UP)

joshbuchea:
This PR fixes an issue that resulted in stale observation photo captions in Activity Log.

Co-authored by Claude Code.

## Context

After editing observation photo captions, the activity log showed stale data until the page was reloaded.

**Root cause:** Observation mutations invalidated `{ type: Activities, id: 'LIST' }`, which only busts the list query. The activity detail view uses `getActivity(id)`, cached under a numeric id tag that `'LIST'` never matches.

**Fix:** Switch to type-only invalidation tags (no `id`) on observation mutations that affect activity data, which invalidates all cached entries for that tag type regardless of id.


ice1080 (approved):
The AI description seems a little unnecessary/verbose

PR #5822, opened 2026-06-12
SW-8480 Allocation number appears to be incorrect

constanzauanini:


src/scenes/NurseryRouter/PlantingSeasons/PlantingDatesTab.tsx
```diff
@@ -551,6 +576,7 @@ const SubstratumDraftSection = ({
             substratumSpecies={substratumSpecies}
             species={species}
             targetsBySpecies={targetsBySpecies}
+            allocatedBySpecies={allocatedBySpecies}
```
ice1080, line 579: I thought that I only had `allocatedBySpecies` in the response payload. Are you seeing it required in the new/update payloads too? If so I might need to change the BE for that

constanzauanini: No, I don't think so. These values can't be changed from here

ice1080: Oops I put this on the wrong line. It You're passing in `allocatedQuantity` on line 526 and line 648, that's what I meant to ask about?

constanzauanini: No, we don't need to update allocatedQuantity from any part of PlantingDatesTab. In fact, we are totally removing it per Jennifer's request.

ice1080 (approved):

ice1080 (approved):

PR #5823, opened 2026-06-12
SW-8475 Remove substratum total number if no substrata

constanzauanini:


ice1080 (approved):
NIT - PR title should technically be `if no substrata`

PR #5832, opened 2026-06-12
Sort organization picker options alphabetically

joshbuchea:
This PR implements alphabetical sorting of the organization picker options in the site header.

This has been driving my crazy for a while 😬, finally decided to do something about it 😄

### BEFORE

staging terraware io_home_organizationId=139

### AFTER

localhost_3000_home_organizationId=139

### DIFF
--- a/src/components/OrganizationsDropdown.tsx
+++ b/src/components/OrganizationsDropdown.tsx
@@ -64,8 +64,7 @@ export default function OrganizationsDropdown(): JSX.Element {
         anchor={<p style={{ fontSize: '16px' }}>{selectedOrganization?.name}</p>}
         menuSections={[
           organizations
-            ?.slice()
-            .sort((a, b) => a.name.localeCompare(b.name, activeLocale || undefined))
+            ?.toSorted((a, b) => a.name.localeCompare(b.name, activeLocale || undefined))
             .map((organization) => ({ label: organization.name, value: organization.id.toString() })) || [],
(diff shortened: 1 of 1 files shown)


src/components/OrganizationsDropdown.tsx
```diff
@@ -62,7 +63,10 @@ export default function OrganizationsDropdown(): JSX.Element {
       <PopoverMenu
         anchor={<p style={{ fontSize: '16px' }}>{selectedOrganization?.name}</p>}
         menuSections={[
-          organizations?.map((organization) => ({ label: organization.name, value: organization.id.toString() })) || [],
+          organizations
+            ?.slice()
+            .sort((a, b) => a.name.localeCompare(b.name, activeLocale || undefined))
```
ice1080: This works, though I think you can also do `organizations?.toSorted(...)` which returns a copy instead of modifying the list

ice1080 (approved):

src/components/OrganizationsDropdown.tsx
joshbuchea: I like that better, will update.

joshbuchea: And I should have used that initially, great flag.

joshbuchea: `.slice` created a copy so it wasn't modifying the original list, but using `.toSorted` definitely feels cleaner than using both `.slice` and `.sort`.

PR #5843, opened 2026-06-15
SW-8494 Mobile web: Numbers in season cards flowing off screen

constanzauanini:


src/scenes/NurseryRouter/PlantingSeasons/PlantingSeasonBox.tsx
```diff
@@ -87,75 +196,67 @@ const PlantingSeasonBox = ({
       }}
     >
       <Card style={{ width: '100%' }} radius={theme.spacing(1)}>
-        <Box display='flex' flexDirection={isMobile ? 'column' : 'row'} gap={theme.spacing(2)}>
-          <Box flex={1}>
-            <Box display='flex' alignItems='center' gap={theme.spacing(1)}>
+        {isMobile ? (
+          <Box display='flex' flexDirection='column'>
```
ice1080, line 195: it's a bummer that desktop and mobile are so different they can't share any common components, but I can't think of a better way to do this

ice1080 (approved):

src/scenes/NurseryRouter/PlantingSeasons/PlantingSeasonBox.tsx
constanzauanini: Yes, I couldn't find a better way either

PR #5853, opened 2026-06-15
Sort organization picker options alphabetically (mobile)

joshbuchea:
This PR includes a change that was missed in a [previous PR](https://github.com/terraware/terraware-web/pull/5832): alphabetically sort organization picker options on mobile.

Note: This PR also includes a change to memoize organization sorting for the non-mobile menu.

### DIFF
--- a/src/components/SmallDeviceUserMenu.tsx
+++ b/src/components/SmallDeviceUserMenu.tsx
@@ -20,7 +20,7 @@ import { APP_PATHS } from 'src/constants';
 import { useDocLinks } from 'src/docLinks';
 import { useSyncNavigate } from 'src/hooks/useSyncNavigate';
 import { useUser } from 'src/providers';
-import { useOrganization } from 'src/providers/hooks';
+import { useLocalization, useOrganization } from 'src/providers/hooks';
 import strings from 'src/strings';
 import { Organization } from 'src/types/Organization';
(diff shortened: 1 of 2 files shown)


src/components/SmallDeviceUserMenu.tsx
```diff
@@ -281,25 +282,27 @@ export default function SmallDeviceUserMenu({
...
-                                  },
-                                ]}
-                              >
-                                {org.name}
-                              </MenuItem>
-                            );
-                          })}
+                          {organizations
+                            ?.toSorted((a, b) => a.name.localeCompare(b.name, activeLocale || undefined))
+                            .map((org, index) => {
```
ice1080: This is more of a curiosity question, but do you know if this will resort on every rerender? Or does the react compiler treat it basically as a `useMemo` when `organizations` doesn't change?

ice1080 (approved):

src/components/SmallDeviceUserMenu.tsx
joshbuchea: That's a good question, I'm not sure I fully understand when the react compiler will and won't automatically memoize a value. And I know you didn't request a change, but I'll go ahead and memoize the sorted value here.

ice1080 (approved):

PR #5873, opened 2026-06-16
SW-8470 Remove add/delete buttons and automatically save

constanzauanini:


src/App.tsx
ice1080, line 1: Was this file supposed to be part of this PR?

constanzauanini: Yes. In this ticket there is also a requirement of Gradient background should extend to full height of browser. Sorry I didn't mentioned that

ice1080: Oh no worries. I glanced at the ticket, but didn't read it fully

ice1080 (approved):

PR #5882, opened 2026-06-17
SW-8470 Add new row to table when adding a new target on species target tab

constanzauanini:


ice1080 (approved):
This seems much nicer, thanks Constanza!

PR #5910, opened 2026-06-19
SW-8550 Planting Season column appearing on wrong side and lacking ability to filter

constanzauanini:


src/scenes/NurseryRouter/NurseryWithdrawalsTable.tsx
```diff
@@ -164,6 +275,21 @@ export default function NurseryWithdrawalsTable(): JSX.Element {
...
+  }, [plantingSeasonsResponse.currentData?.seasons]);
+
+  const plantingSeasonNameToIds = useMemo(() => {
+    const idsByName = new Map<string, number[]>();
+    (plantingSeasonsResponse.currentData?.seasons ?? []).forEach((season) => {
+      idsByName.set(season.name, [...(idsByName.get(season.name) ?? []), season.id]);
+    });
+
+    return idsByName;
+  }, [plantingSeasonsResponse.currentData?.seasons]);
```
ice1080: You might be able to combine these so that `plantingSeasonNames` is just `plantingSeasonNameToIds.values()`

ice1080 (approved):
Just one potential improvement. If it doesn't make sense just ignore it

PR #5921, opened 2026-06-22
SW-8558 Withdrawn plants not showing up in request card

constanzauanini:


ice1080 (approved):
This seemed slightly complicated to get the scheduled planting date. Is there a way I could update the BE to simplify?

constanzauanini:
> This seemed slightly complicated to get the scheduled planting date. Is there a way I could update the BE to simplify?

I couldn't get the scheduledPlantingDateRequestId in the plantingDateRequests search. So that's why I did that in the modal. Maybe if we include it in the plantingDateRequests the id we can simplify this

constanzauanini:
> > This seemed slightly complicated to get the scheduled planting date. Is there a way I could update the BE to simplify?
> 
> I couldn't get the scheduledPlantingDateRequestId in the plantingDateRequests search. So that's why I did that in the modal. Maybe if we include it in the plantingDateRequests the id we can simplify this

I will merge it for now, and I can update it if we update BE

PR #5928, opened 2026-06-23
SW-8450 Use BE calculated project level average strata survival rate on plants dashboard

constanzauanini:


tommylau523 (approved):
Looks correct! TrackingStats tag is a bit too general but we will need ways to get a planting site’s org or project IDs to have better control. Okay for now!

PR #5931, opened 2026-06-23
SW-8521 Show Message when there are no requests

constanzauanini:


src/scenes/NurseryRouter/PlantingDateRequestsTabContent.tsx
```diff
@@ -100,6 +100,12 @@ const PlantingDateRequestsTabContent = (): JSX.Element => {
   );
```
tommylau523, line 103: <p class='graphite__hidden'><i>[Re: line +47]</i></p>
This may be better as a useMemo
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/terraware/terraware-web/5931?utm_source=unchanged-line-comment">Graphite</a>.</p>

tommylau523 (approved):

PR #5935, opened 2026-06-24
SW-8441 Change history section

constanzauanini:
Reused EventLog component from Observation View

### DIFF
--- a/src/queries/plantingSeasons/events.ts
+++ b/src/queries/plantingSeasons/events.ts
@@ -1,0 +1,33 @@
+import { baseApi as api } from '../baseApi';
+import { EventLogEntryPayload, ListEventLogEntriesApiResponse } from '../generated/events';
+import { QueryTagTypes } from '../tags';
+
+export type ListPlantingSeasonEventsArgs = {
+  organizationId: number;
+  plantingSeasonId: number;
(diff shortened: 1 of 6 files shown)


src/queries/plantingSeasons/events.ts
```diff
@@ -0,0 +1,33 @@
...
+        url: '/api/v1/events/list',
+        method: 'POST',
+        body: {
+          organizationId,
+          plantingSeasonId,
+        },
+      }),
+      providesTags: (_results, _error, { plantingSeasonId }) => [
+        { type: QueryTagTypes.PlantingSeasons, id: plantingSeasonId },
+        { type: QueryTagTypes.PlantingSeasonDates, id: plantingSeasonId },
```
ice1080, line 23: Not sure if `PlantingDateRequests` should also be in here or if that's will be on a different page (or perhaps that one doesn't have a change log mentioned in the PRDs)

ice1080 (approved):

src/queries/plantingSeasons/events.ts
constanzauanini: I'm not sure, I will not add it for now, we can add it later if necessary

PR #5936, opened 2026-06-24
SW-8576 Show banner for notifications

constanzauanini:


src/components/common/PlantingSeasonNotificationBanners.tsx
```diff
@@ -0,0 +1,253 @@
...
+};
+
+type NotificationGroupBodyProps = {
+  group: PlantingSeasonNotificationGroupPayload;
+  activeLocale: string | null;
+  showPageContext: boolean;
+};
+
+type PlantingSeasonNotificationPayloadWithRequestDetails = PlantingSeasonNotificationPayload & {
+  date?: string;
```
constanzauanini: @ice1080 I added this here because I think we are missing in the API the dates and notes that we want to show in some notifications. Let me know if that is correct or if I should get that data in a different way

ice1080: @constanzauanini you're correct, I missed the 2 notifications that have `[dates]` and `[any note added by the PM]`. I think those are the only 2 missed, but your code has `date`, `dates`, `nurseryRequestNotes`, `notes`, and `scheduledPlantingDates`. Do we need all 5 or just `dates` and `notes`?

src/components/common/PlantingSeasonNotificationBanners.tsx
```diff
@@ -0,0 +1,253 @@
...
+  );
+
+const SpeciesNamesList = ({ names }: { names: string[] }): JSX.Element => (
+  <Box component='span' sx={{ fontStyle: 'italic', fontWeight: 600 }}>
+    {names.map((name, index) => (
+      <React.Fragment key={name}>
+        {index > 0 ? ', ' : ''}
+        {name}
+      </React.Fragment>
+    ))}
```
ice1080: Can't this be done with `{names.join(', ')}`?

ice1080 (approved):
Did AI write this one? It seems like claude really likes to put const functions at the top of a file. `PlantingSeasonNotificationBanners` had 7 functions at the top, many were only used once. I'm not saying it's better or worse, just commenting that it's interesting

ice1080:
I approved, but you can also wait to merge till the BE fix is done for `dates` and `notes` if you'd prefer

src/components/common/PlantingSeasonNotificationBanners.tsx
ice1080: Jennifer just updated the PRD so that it only includes `dates` for this one type. I'm working on the BE for it now; it should display as a comma separated sorted list of LocalDates

constanzauanini: Perfect. Just dates will be enough. I will make the change when BE is ready!

ice1080: Ok the BE just merged, so should be deployed in a bit, unless the staging issues cause that to not work correctly.

constanzauanini:
> Did AI write this one? It seems like claude really likes to put const functions at the top of a file. `PlantingSeasonNotificationBanners` had 7 functions at the top, many were only used once. I'm not saying it's better or worse, just commenting that it's interesting

I used Codex for this one, but yes, it's true it generate variables that are not re-used later

ice1080 (approved):
Looks good, though I think the graphite comment is correct and it needs one more `?`

PR #5937, opened 2026-06-24
SW-8568 Tablet View: Updates

constanzauanini:


src/scenes/NurseryRouter/PlantingSeasons/PlantingSeasonDetailsView.tsx
```diff
@@ -42,6 +42,8 @@ const PlantingSeasonDetailsView = (): JSX.Element => {
   const theme = useTheme();
   const { activeLocale } = useLocalization();
   const { isMobile, isTablet } = useDeviceInfo();
+  const isBelowLargeDesktop = useMediaQuery(theme.breakpoints.down('lg'));
+  const showTabletLayout = isTablet || (!isMobile && isBelowLargeDesktop);
```
ice1080, line 46: Is there a reasoning from product for why we're changing the definition of a tablet width just for this one page?

constanzauanini: Just because it looks bad for small desktop, there is not enough room for all the elements, so we are rendering the tablet view for those

ice1080 (approved):

PR #5945, opened 2026-06-25
SW-8578 After new inventory of species is added Inventory Planning page should refresh

constanzauanini:


src/redux/features/batches/batchesAsyncThunks.ts
```diff
@@ -80,6 +82,7 @@ export const requestSaveBatch = createAsyncThunk(
     }
 
     if (response && response.requestSucceeded && responseQuantities.requestSucceeded) {
+      dispatch(baseApi.util.invalidateTags([{ type: QueryTagTypes.InventoryPlanning, id: 'LIST' }]));
```
constanzauanini, line 85: I have to manually do it here since adding inventory batches it's not in rtkQuery yet

ice1080: I've also had to do that a few times

ice1080 (approved):

PR #5946, opened 2026-06-25
SW-8579 Inventory Planning: Closed Season still shows in table

constanzauanini:


src/queries/search/inventoryPlanning.ts
```diff
@@ -9,6 +9,7 @@ export type InventoryPlanningArgs = {
 };
 
 export type PlantingSeasonStatus = 'Active' | 'Upcoming' | 'Past End Date' | 'Closed';
+export const inventoryPlanningPlantingSeasonStatuses: PlantingSeasonStatus[] = ['Active', 'Upcoming'];
```
ice1080: Does this need to also include `Past End Date`?

constanzauanini: I'm not sure. I excluded because I thought we wanted active and upcoming only, but I can ask Jennifer

ice1080 (approved):

PR #5947, opened 2026-06-25
SW-8581 Delete Species Goal: Update dialog copy

constanzauanini:


ice1080 (approved):
I was wondering if using several words for the same concept would end up causing more confusion

PR #5949, opened 2026-06-25
SW-8587 Planting Season card "Withdrawn for Planting" number not a link to the season withdrawals

constanzauanini:


src/scenes/NurseryRouter/PlantingSeasons/PlantingSeasonDetailsView.tsx
```diff
@@ -192,15 +206,27 @@ const PlantingSeasonDetailsView = (): JSX.Element => {
...
-          fontWeight={600}
-          lineHeight={valueLineHeight}
-          color={highlight && value !== undefined ? theme.palette.TwClrTxtBrand : theme.palette.TwClrTxt}
-          sx={highlight && value !== undefined ? { textDecoration: 'underline' } : undefined}
-        >
-          {value === undefined ? '-' : value.toLocaleString(activeLocale || undefined)}
-        </Typography>
+        {to && value !== undefined ? (
+          <Link
+            to={to}
```
ice1080: Could this be simplified to only a `<Link>` with an undefined `to` if `value` is undefined?

ice1080: Err I was meaning can we remove the need for a separate `<Typography>` as well as the separate variables for `valueColor`, `valueText`, and `valueStyle`

ice1080 (approved):

PR #5958, opened 2026-06-26
SW-8591 Add change history in Inventory Planning View

constanzauanini:
Filter 'PlantingSeasonAllocatedSpecies' events from Planting Season page history section, and include a new history log section in Inventory Planning view that only shows those events


src/queries/plantingSeasons/events.ts
```diff
@@ -22,12 +53,32 @@ const injectedRtkApi = api.injectEndpoints({
         { type: QueryTagTypes.PlantingSeasons, id: plantingSeasonId },
         { type: QueryTagTypes.PlantingSeasonDates, id: plantingSeasonId },
       ],
-      transformResponse: (results: ListEventLogEntriesApiResponse) =>
-        results.events.map((event) => ({ ...event })).reverse(),
+      transformResponse: (results: ListEventLogEntriesApiResponse, _meta, { excludedSubjects }) =>
+        transformEvents(results, excludedSubjects),
```
ice1080: I think it might be better to use the `subjects` field in the body payload to whitelist the subjects we care about instead of filtering out data being returned. That way we're not returning 20 events we don't care about and only 1 event we do care about

src/scenes/InventoryPlanningRouter/InventoryPlanningEventLog.tsx
```diff
@@ -28,13 +39,69 @@ const InventoryPlanningEventLog = ({
...
+    [strings.NONE, theme.palette.TwClrTxtSuccess]
+  );
+
+  const renderEventDescription = useCallback(
+    (event: EventLogEntryPayload): ReactNode => {
+      if (event.subject.type === 'PlantingSeasonAllocatedSpecies' && event.action.type === 'FieldUpdated') {
+        const season = plantingSeasonsById.get(event.subject.plantingSeasonId);
+        const plantingSite = plantingSitesById.get(event.subject.plantingSiteId);
+        const seasonName = season?.name ?? event.subject.shortText;
+        const plantingSiteName = plantingSite?.name ?? event.subject.fullText;
```
ice1080, line 72: I think this works, but I will see if I can update the BE to include the planting season name and planting site name in the subjects so that the list of sites and seasons doesn't have to get passed to the component

src/scenes/InventoryPlanningRouter/InventoryPlanningEventLog.tsx
```diff
@@ -28,13 +39,69 @@ const InventoryPlanningEventLog = ({
...
+          >
+            {seasonName}
+          </Link>
+        );
+
+        return strings.formatString(
+          strings.PLANTS_OF_SPECIES_ALLOCATED_TO_SEASON_FOR_SITE_CHANGED,
+          <>{event.subject.scientificName ?? event.subject.shortText}</>,
+          seasonLink,
+          <>{plantingSiteName}</>,
```
ice1080: Do these need the react fragment or can they just be the string like `plantingSiteName`?

src/queries/plantingSeasons/events.ts
```diff
@@ -29,11 +29,13 @@ const transformInventoryPlanningEvents = (
   speciesId?: number
 ): EventLogEntryPayload[] => {
   return orderEvents(
-    results.events.filter(
-      (event) =>
-        speciesId === undefined ||
-        (event.subject.type === 'PlantingSeasonAllocatedSpecies' && event.subject.speciesId === speciesId)
-    )
+    results.events
+      .filter((event) => event.action.type !== 'Created')
```
ice1080, line 33: Why are we filtering out created events?

ice1080 (approved):
I think a BE change to include the planting site name and planting season name should simplify this a bit, but the rest looks good overall!

src/queries/plantingSeasons/events.ts
constanzauanini: This in only for PlantingSeasonAllocatedSpecies, because there is always a creation event and then an update event that changes the values from 0 to {number}, and we are only showing that second event in the history

src/scenes/InventoryPlanningRouter/InventoryPlanningEventLog.tsx
constanzauanini: They can be strings

PR #5967, opened 2026-06-29
SW-8593 Add configurable and filterable "Planting Date" column, hidden by default

constanzauanini:


src/queries/search/nurseries.ts
```diff
@@ -303,6 +321,7 @@ const injectedRtkApi = api.injectEndpoints({
             'batchWithdrawals.batch_species_scientificName',
             'batchWithdrawals.destinationBatchProjectName',
             'plantingSeason_name',
+            'plantingDateRequest.date',
```
ice1080: I think if you change this to `plantingDateRequest_date` with `_` instead of `.` then you won't have to do the `isArray` check on line 337?

src/scenes/NurseryRouter/NurseryWithdrawalsTable.tsx
```diff
@@ -122,22 +129,41 @@ const normalizeMenuColumnOrder = (columnOrder: MRT_ColumnOrderState): MRT_Column
   ];
 };
 
-const movePlantingSeasonBeforeMenu = (columnOrder: MRT_ColumnOrderState): MRT_ColumnOrderState => {
+const isHiddenByDefaultColumn = (columnId: string): boolean =>
+  HIDDEN_BY_DEFAULT_COLUMN_IDS.some((hiddenColumnId) => hiddenColumnId === columnId);
```
ice1080: NIT this can be simplified to `HIDDEN_BY_DEFAULT_COLUMN_IDS.includes(columnId);`

ice1080 (approved):
A couple suggestions but nothing real important

PR #5968, opened 2026-06-29
SW-8589 More Planting Dates and Planting Dates UI updates

constanzauanini:


src/scenes/NurseryRouter/PlantingSeasons/PlantingDatesTab.tsx
```diff
@@ -1004,6 +1030,8 @@ const AddSpeciesRow = ({
           <Icon name='iconSubtract' size='medium' fillColor={theme.palette.TwClrIcn} />
         </IconButton>
       </Box>
+      <Box />
+      <Box />
```
ice1080: Were these empty boxes intentional? Should bottom padding be used instead of this?

ice1080 (approved):

src/scenes/NurseryRouter/PlantingSeasons/PlantingDatesTab.tsx
constanzauanini: Hmm those were there as placeholders for the grid. I think I can remove them now

PR #5974, opened 2026-06-30
SW-8595 Don't include species with 0 requested quantity

constanzauanini:


src/scenes/NurseryRouter/WithdrawFromBatchesModal.tsx
```diff
@@ -453,6 +453,18 @@ const Step1Content = ({
   readyBySpecies,
 }: Step1ContentProps): JSX.Element => {
   const theme = useTheme();
+  const { activeLocale } = useLocalization();
+  const requestedSpecies = useMemo(
+    () =>
+      requestSpecies
```
ice1080: A minor NIT that `requestedSpecies` and `requestSpecies` are very similar and might be confusing, but I'm not sure I have a much better suggestion. Perhaps `sortedSpecies`? Or `sortedRequestSpecies`?

ice1080 (approved):

PR #5976, opened 2026-06-30
Revert "Update dependency js-yaml to v5"

joshbuchea:
Reverts terraware/terraware-web#5964

**NOTE: The `js-yaml` update from v4 to v5 broke API types generation. Reverting this change fixes it.**

A recent dependency update added "js-yaml": "5.1.0" to the resolutions field in `package.json`, which forces `js-yaml` 5.x on all packages in the dependency tree — including `@redocly/openapi-core` (a transitive dependency of `openapi-typescript`, used by the `generate-types` script). `js-yaml` 5.x is a breaking API rewrite that removed the types export (e.g. `types.merge`, `types.binary`), which `@redocly/openapi-core` relies on. Even the latest `@redocly/openapi-core@1.34.6` still requires `js-yaml@^4.1.0`, so there is no upstream fix available yet.

Fix: Revert the resolution from "js-yaml": "5.1.0" back to "js-yaml": "^4.1.0". The project does not import `js-yaml` directly anywhere in source, so this has no impact on application code — it only affects how the version is resolved for transitive dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -238,7 +238,7 @@
     "brace-expansion": "^2.0.3",
     "fork-ts-checker-webpack-plugin": "^9.0.0",
     "immutable": "^5.1.5",
-    "js-yaml": "5.1.0",
+    "js-yaml": "4.2.0",
     "picomatch": "^4.0.4",
     "serialize-javascript": "^7.0.5",
     "simple-git": "^3.32.3",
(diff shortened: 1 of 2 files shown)


constanzauanini:
Oops, I merged that one because the app was working as expected and all the test passing. Thanks for the revert

constanzauanini (approved):

joshbuchea:
@constanzauanini no worries at all, I didn't notice anything until I tried to regenerate the API types. Thanks for the quick review 🙌 💯

PR #5979, opened 2026-07-01
SW-8603 Fix Editing allocated quantity on Inventory Planning page causes console error

constanzauanini:
I asked Claude about the error and it explained and fixed it:

The [Immer] MapSet error (Immer error #18) happens when a Map or Set gets written into Immer-produced state without the enableMapSet() plugin. In this case:

- listInventoryPlanningSpeciesAvailable's transformResponse returned a Map<number, number>.
- RTK Query stores every query result in the Redux store via Immer's producer, and Immer can't handle a Map in state
  by default.
- It surfaced specifically on editing allocated quantity because the upsertAllocatedSpecies mutation invalidates the
  InventoryPlanning tag → that query refetches → RTK Query tries to write the Map back into the store → Immer throws.

The fix

Changed the query to return a plain serializable object (Record<number, number>) instead of a Map, which Immer stores without issue.

### DIFF
--- a/src/queries/search/inventoryPlanning.ts
+++ b/src/queries/search/inventoryPlanning.ts
@@ -106,7 +106,7 @@ const injectedRtkApi = api.injectEndpoints({
       transformResponse: (response: InventoryPlanningSeasonSearchResponse) => response.results,
     }),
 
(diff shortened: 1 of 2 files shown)


ice1080 (approved):
Very interesting 🤔

PR #5980, opened 2026-07-01
SW-7736 Activities RTK Query

tommylau523:
Add API_PATHS enum with activity media file paths

Use API_PATHS for activity media file URLs

Delete unused ActivityService

Migrate ProjectView funder activities to RTK Query

Delete unused FunderActivityService and funder activities redux

PR #5985, opened 2026-07-01
Moved api-paths to constants

tommylau523:


tommylau523:
* **#5985** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/5985?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-web/5985?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5980** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/5980?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

constanzauanini (approved):
Are we doing this so that we can re-use them in different parts of the app?

tommylau523:
> Are we doing this so that we can re-use them in different parts of the app?

​Yes in some cases, but I find it easier to maintain. For example if we change an API path, we won’t have to dig through the app.

PR #5994, opened 2026-07-02
SW-8604 Fix Search API IllegalArgumentException: Value Project Lead not recognized for role

constanzauanini:
Adding Accept-Language: en header to search api calls that are called from accelerator console and sends localized enum values


joshbuchea (approved):
I think we're hardcoding 'en' as the locale in this case because these requests are for the console and we don't need to localize there.

constanzauanini:
> I think we're hardcoding 'en' as the locale in this case because these requests are for the console and we don't need to localize there.

Yes

PR #6006, opened 2026-07-07
Enable Graphite CI Optimization

alexTerraformation:
Since we use AWS secrets we can't use the Graphite Buildkite plugin. The plugin requires secrets to be filled at pipeline load time but AWS secrets are not inserted until runtime. 

Instead we copy the plugin functionality into a script and run it ourselves.

Co-authored by Claude


constanzauanini (approved):
Looks good to me

PR #6014, opened 2026-07-08
SW-8606 Hide login buttons if logged in

constanzauanini:


src/scenes/LearnMore/LearnMoreView.tsx
```diff
@@ -40,6 +40,26 @@ const LearnMoreView = (): JSX.Element => {
...
+      })
+      .catch(() => {
+        if (active) {
+          setIsLoggedIn(false);
+        }
+      });
+    return () => {
+      active = false;
+    };
+  }, []);
```
ice1080, line 61: If we eventually have any other pages that should display in a non-logged in state, perhaps it would be worth it to update `UserProvider` to have an `isLoggedIn` flag so that we can use `useUser` and we wouldn't have to refetch the user data. It might require some refactoring of that hook/provider though, so I'm just flagging in case we want to follow up or add any others that need that

ice1080 (approved):

src/scenes/LearnMore/LearnMoreView.tsx
constanzauanini: Yes, the UserProvider right now redirects to login page if not logged in, that's why I added this "logic" here. But I agree it would be worth updating it if we need to reuse it later

PR #6050, opened 2026-07-13
SW-8633 Show tooltip for Planting Purpose at all times

constanzauanini:


ice1080 (approved):
Playwright failures seem valid

PR #6060, opened 2026-07-14
SW-8496 Add breaking space after slash in Germination/Establishment label

constanzauanini:


ice1080 (approved):
I'm not sure what this is addressing, so not sure if the graphite comment is valid

constanzauanini:
> I'm not sure what this is addressing, so not sure if the graphite comment is valid

We just want to break the Germination/Establishment label, and without a space it doesn't break correctly, so the graphite comment doesn't make a lot of sense here

PR #6123, opened 2026-07-23
SW-8679 Display icons for annotations

ice1080:
Display either a menu, camera, or video icon depending on if the annotation has media or not.

Co-authored-by: Claude Opus 4.8 <noreply@anthropic.com>

(waiting on https://github.com/terraware/web-components/pull/772 to change the menu icon to a text icon)

---

### DIFF
--- a/src/components/GaussianSplat/TfAnnotationManager.ts
+++ b/src/components/GaussianSplat/TfAnnotationManager.ts
@@ -9,7 +9,7 @@ import { AnnotationIconType } from './Annotation';
 
 // Maps each annotation icon type to a @terraware/web-components icon.
 const ANNOTATION_ICONS: Record<AnnotationIconType, IconName> = {
-  menu: 'iconMenu',
+  text: 'iconDocument',
   image: 'iconPhoto',
   video: 'iconVideo',
 };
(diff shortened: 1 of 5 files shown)


src/components/GaussianSplat/TfAnnotationManager.ts
```diff
@@ -104,6 +160,85 @@ export class TfAnnotationManager extends PcAnnotationManager {
...
+   */
+  _createHotspotMaterial(texture: any, options?: any) {
+    const material = super._createHotspotMaterial(texture, options);
+    material.emissive.set(1, 1, 1);
+    material.update();
+    return material;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _setAnnotationHover(_annotation: any, _hover: boolean) {
```
constanzauanini, line 233: Looks like this function and the one in line 238 don't do anything

ice1080: Correct, that's intentional. These are both overriddes for the playcanvas `AnnotationManager`, and set to a no-op so that the default behavior is not executed

constanzauanini (approved):

PR #6140, opened 2026-07-24
SW-8699 Create planting plans list

constanzauanini:


src/scenes/PlantingPlansRouter/PlantingPlanMapThumbnail.tsx
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  const closed = ensureClosed(dedupeConsecutive(sampled));
+  if (closed.length < 4 || ringArea(closed) === 0) {
+    return null;
+  }
+  return closed;
+};
+
+const simplifyToBudget = (boundary: MultiPolygon, pointBudget: number): number[][][][] => {
```
tommylau523: Can we move this to a helper function file and add test? maybe src/utils/geometry.

src/scenes/PlantingPlansRouter/PlantingPlanMapThumbnail.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const last = ring[ring.length - 1];
+  return first[0] === last[0] && first[1] === last[1] ? ring : [...ring, first];
+};
+
+const cleanRing = (ring: Ring, maxPoints: number): Ring | null => {
+  const step = Math.max(1, Math.ceil(ring.length / maxPoints));
+  const sampled: Ring = [];
+  for (let i = 0; i < ring.length; i += step) {
+    sampled.push([roundCoord(ring[i][0]), roundCoord(ring[i][1])]);
+  }
```
tommylau523: can this part become something like const sampled = ring.map(...)? And we can not call roundCoord twice as well.

src/scenes/PlantingPlansRouter/PlantingPlanMapThumbnail.tsx
```diff
@@ -0,0 +1,150 @@
...
+// remaining points to fit a global budget. This keeps even pathological boundaries (dozens of polygons) within the URL
+// limit while looking identical at thumbnail scale.
+const COORD_DECIMALS = 5;
+const MAX_URL_LENGTH = 7500;
+const MIN_POLYGON_AREA_FRACTION = 0.0004;
+const MIN_POINTS_PER_RING = 4;
+
+type Ring = number[][];
+
+const roundCoord = (value: number): number => Number(value.toFixed(COORD_DECIMALS));
```
tommylau523: This function seems like it is rounding a number more than a coordinate. But maybe we can change this function signature to be `roundCoord: [number, number] => [number, number]` instead,

src/scenes/PlantingPlansRouter/PlantingPlanMapThumbnail.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const significant = withArea
+    .filter(({ area }) => area >= maxArea * MIN_POLYGON_AREA_FRACTION)
+    .sort((a, b) => b.area - a.area);
+
+  const maxPolygons = Math.max(1, Math.floor(pointBudget / MIN_POINTS_PER_RING));
+  const kept = significant.slice(0, maxPolygons);
+  const pointsPerRing = Math.max(MIN_POINTS_PER_RING, Math.floor(pointBudget / kept.length));
+
+  return kept
+    .map(({ ring }) => cleanRing(ring, pointsPerRing))
```
tommylau523: Should we distribute points to the rings uniformly? I suspect in some sites, we will have one relatively complex outer ring vs simpler inner rings. Would it make more sense to scale down every ring by a factor instead?

tommylau523 (changes requested):
Some changes requested around geometry simplification functions.

src/scenes/PlantingPlansRouter/PlantingPlanMapThumbnail.tsx
constanzauanini: I initially did that (scale down every ring by a factor) but it was not good because:

- A global factor = ceil(total / budget) decimates small rings too. A 5-point zone at factor = 3 → ~2 points → cleanRing returns null (fewer than 4 points / zero area) → that polygon is dropped.
- Dropping small polygons changes the shape and shifts the auto fit extent (making a whole cluster of a site disappear for example).

So a factor optimizes detail on the big ring at the cost of silently deleting the small ones. Not a good trade for a thumbnail whose whole job is conveying overall shape/extent.

Current behaviour: simplifyToBudget only ever looks at polygon[0] (the outer ring of each polygon). Holes (inner rings) are already dropped entirely. So the "rings" are the outer rings of the separate polygons in the MultiPolygon (a site's zones/scattered parcels). The complexity variance is across polygons, not within one.

Please let me know if it makes sense to you or if I'm missing something

src/scenes/PlantingPlansRouter/PlantingPlanMapThumbnail.tsx
tommylau523: I am also now wondering, should we run the simplification and store the thumbnail GeoJson in the backend? That seems much mroe sensible to run once and store instead of having the client rerun it over and over. 

Bonus is we already have simplification implemnetaions in the backend.

constanzauanini: Yes, if BE already have that, I think it would be a good thing to move it to BE then! And the planting sites api can return the thumbnail url

tommylau523: We probably would still have the client build the mapbox url, but at least it will not have to do geojson construction.

constanzauanini:
@tommylau523 are we waiting for the planting site map image BE changes before merging this PR?

tommylau523 (approved):
Approving this for now! We can update the thumbnail logic when BE is ready.

PR #6141, opened 2026-07-24
SW-8700 Planting Plan Overview page

constanzauanini:


src/scenes/PlantingPlansRouter/PlantingPlanOverview.tsx
```diff
@@ -0,0 +1,90 @@
...
+        : strings.formatString(strings.X_HA, numberFormatter.format(plantingSite.areaHa, { decimals: 1 })).toString(),
+    [numberFormatter, plantingSite.areaHa]
+  );
+
+  const strataCount = useMemo(
+    () => numberFormatter.format(plantingSite.strata?.length ?? 0),
+    [numberFormatter, plantingSite.strata]
+  );
+
+  const tile = (label: string, value: string) => (
```
tommylau523: nit: Looks nicer if we break this into a component. Then we can use  
`<PlantingPlantTitle label={} value={} />` down below.  
  
Same for the vertical divider. Or it can be part of a prop in the title component too.

constanzauanini: I like this idea!

tommylau523 (approved):

PR #6161, opened 2026-07-27
Fix pipeline for different agents

ice1080:
The `Validate type generation` step needs spring security env to work properly on different agents. This ensures that if that step is run on a different agent than the previous step (`Run end-to-end tests`) then it correctly starts the BE server with the spring security env variables necessary.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -113,6 +113,10 @@ steps:
             - file
             - branch
     command: .buildkite/scripts/validate-types.sh
+    env:
+      SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_KEYCLOAK_ISSUERURI: https://auth.staging.terraware.io/realms/terraware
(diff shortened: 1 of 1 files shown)


sgrimm (approved):

.buildkite/pipeline.yml
```diff
@@ -113,6 +113,10 @@ steps:
             - file
             - branch
     command: .buildkite/scripts/validate-types.sh
+    env:
+      SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_KEYCLOAK_ISSUERURI: https://auth.staging.terraware.io/realms/terraware
```
tommylau523, line 117: Since this is a public repo technically, should we move these URLs to a private env?

ice1080: Not sure I'm following. These have been in this pipeline since it was created (in the previous step)

tommylau523: I do see that they already exist in the last step. Just a question I think, not a blocker.

ice1080: It looks like our CI has always had this, including on GHA, so I'll leave as-is for now

PR #6183, opened 2026-07-31
SW-8723 Planting Seasons section

constanzauanini:


src/scenes/PlantingPlansRouter/PlantingPlanSeasons.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const addButton = () => (
+    <Button
+      id='addPlantingSeason'
+      icon='plus'
+      label={strings.ADD_PLANTING_SEASON}
+      onClick={onAddSeason}
+      size='medium'
+    />
+  );
```
ice1080: I think this one can be removed

ice1080 (changes requested):

src/scenes/PlantingPlansRouter/PlantingPlanSeasons.tsx
constanzauanini: That's true. It's in the top now. I reused the empty state from the other planting season view, but it is not necessary here

ice1080 (approved):

PR #6193, opened 2026-08-04
SW-8732 Species check flow modal

constanzauanini:


src/scenes/Species/SpeciesCheck/NameCheckStep.tsx
```diff
@@ -0,0 +1,115 @@
...
+            padding={theme.spacing(1.5, 2)}
+            sx={{ backgroundColor: theme.palette.TwClrBgSecondary }}
+          >
+            <span />
+            <Typography sx={headerCell}>{strings.SPECIES}</Typography>
+            <Typography sx={headerCell}>{strings.ISSUE}</Typography>
+            <Typography sx={{ ...headerCell, textAlign: 'right' }}>{strings.SUGGESTION}</Typography>
+          </Box>
+          {speciesWithProblems.map((species) => {
+            const problem = species.problems?.[0];
```
ice1080, line 83: Can a species have more than 1 problem?

src/scenes/Species/SpeciesCheck/SetLocationStep.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onChange,
+}: SetLocationStepProps): JSX.Element => {
+  const theme = useTheme();
+
+  const countryOptions: DropdownItem[] = countries.map((country) => ({ label: country.name, value: country.code }));
+  const hasCountryMapping = botanicalCountries.some((bc) => bc.countryCode !== undefined);
+
+  return (
+    // Full width so the dividers between projects span wide; each project's fields sit in a narrower
+    // centered column.
```
ice1080: I think this comment can probably be removed; it seems like it's related to the prompts

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
```diff
@@ -0,0 +1,547 @@
...
+    });
+    setLocationEdits(initial);
+    setStep(0);
+    setNativeMode('list');
+    setNameSelected(new Set());
+    setNativeSelected(new Set());
+    setOverrideEdits({});
+    setShowCancel(false);
+    setForceLocation(false);
+    setLocationsSubmitted(false);
```
ice1080, line 117: It seems odd to set this many state items in a single useEffect, but the file (and PR) is so large that I don't know if there's a better way. Up to you for if you want to change it

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
```diff
@@ -0,0 +1,547 @@
...
+        pending: data.pending,
+      })),
+    [buildSummary, targetData]
+  );
+
+  const hasAnyPending = nativeSections.some((section) => section.pending.length > 0);
+
+  const goToStep = useCallback((key: StepKey) => setStep(stepKeys.indexOf(key)), [stepKeys]);
+
+  const handleClose = useCallback(() => onClose(), [onClose]);
```
ice1080: I think usages of `handleClose` can be replaced with `onClose`

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
ice1080, line 1: This file is pretty big, but I'm not sure if there are any natural places to break it up

src/scenes/Species/SpeciesCheck/types.ts
```diff
@@ -0,0 +1,42 @@
...
+export type LocationTarget = {
+  key: number;
+  projectId?: number;
+  name: string;
+  countryCode?: string;
+  botanicalCountryCode?: string;
+  isOrg: boolean;
+};
+
+export const ORG_TARGET_KEY = -1;
```
ice1080, line 33: What is this needed for?

src/scenes/Species/SpeciesListView.tsx
```diff
@@ -486,6 +563,29 @@ export default function SpeciesListView({ reloadData, species }: SpeciesListProp
...
+                return nativity ? getNativityLabel(nativity) : strings.NOT_SET;
+              },
+              enableEditing: false,
+              filterVariant: 'select' as const,
+              filterSelectOptions: [
+                strings.NATIVE,
+                strings.INTRODUCED,
+                strings.INVASIVE,
+                strings.UNKNOWN,
+                strings.NOT_SET,
```
ice1080, line 589: Will the values from the BE be in the user's locale, or will these be coming back as an enum? If as an enum, then double check that this works:
- change language from english
- reload page
- filter on one of these statuses
If it works, great; if not then these options might need to change to just the english versions

ice1080 (commented):
This PR was pretty large, making it difficult to review the whole thing in one pass

src/scenes/Species/SpeciesCheck/NameCheckStep.tsx
constanzauanini: Hmm, I'm not sure, but in the old implementation we were using like this: problems[0]

src/scenes/Species/SpeciesCheck/types.ts
constanzauanini: This is used for species that belong to the org, and not to a project

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
constanzauanini: This is "reseting" all the flow/selection state to a fresh session, that is why we set all those items

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
constanzauanini: Yes. It is already broken by steps in different components. It has the buttons that are common to the modal and the functions that we need to pass to each step. I know it became big, but not sure what other part could be separated

src/scenes/Species/SpeciesListView.tsx
constanzauanini: It works with the translated version!

ice1080 (approved):

PR #6210, opened 2026-08-07
Add achievements, challenges, financials, and comments back to the tabs

tommylau523:
Added existing text fields to report view.

Co-Authored-By: Claude Opus 5 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)


tommylau523:
* **#6217** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6217?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6216** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6216?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6215** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6214** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6214?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6213** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6212** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6211** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6210** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6210?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6210?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6209** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6209?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6208** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6207** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6207?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6206** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6206?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6205** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6205?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6204** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6204?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6203** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6202** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6201** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6200** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6199** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6199?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ice1080:
@codex review

src/components/AcceleratorReports/ChallengesMitigationBox.tsx
```diff
@@ -162,7 +166,7 @@ const ChallengesMitigationBox = (props: ReportBoxProps) => {
   }, [challengeMitigations]);
 
   const areFilteredChallengesDifferent = useMemo(() => {
-    return nonEmptyChallenges.length > 0 && JSON.stringify(nonEmptyChallenges) !== JSON.stringify(report?.challenges);
+    return JSON.stringify(nonEmptyChallenges) !== JSON.stringify(report?.challenges ?? []);
```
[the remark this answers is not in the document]
tommylau523: This fix is incoming.

src/scenes/Reports/AcceleratorReportTabV2.tsx
```diff
@@ -83,6 +87,18 @@ const AcceleratorReportTabV2 = (): JSX.Element => {
...
+
+          {projectId !== undefined && (
+            <>
+              <AchievementsBox projectId={projectId} report={selectedReport} />
+
+              <ChallengesMitigationBox projectId={projectId} report={selectedReport} />
+
+              <FinancialSummariesBox projectId={projectId} report={selectedReport} />
+
+              <AdditionalCommentsBox projectId={projectId} report={selectedReport} />
```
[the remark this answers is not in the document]
tommylau523: This is fixed in a downstream branch, where state management is clearer. Path {reportId} takes precedent, and sets the selectedProject to the report project.   
  
project switch -> first report of the project subseuqnetly.

src/components/AcceleratorReports/ChallengesMitigationBox.tsx
```diff
@@ -183,14 +187,10 @@ const ChallengesMitigationBox = (props: ReportBoxProps) => {
...
-      if (areFilteredChallengesDifferent) {
-        // only call onChange if the non-empty challenges are different, but call it with all to include the empty
-        // challenges; otherwise deleting characters can cause rows to disappear
-        onChange(challengeMitigations);
-      }
+    if (onChange && areFilteredChallengesDifferent) {
+      onChange(nonEmptyChallenges);
     }
-  }, [addRow, areFilteredChallengesDifferent, challengeMitigations, onChange]);
+  }, [addRow, areFilteredChallengesDifferent, challengeMitigations, nonEmptyChallenges, onChange]);
```
ice1080, line 193: Flagging this because of the previous comment. It mentions that deleting characters in challenges can cause rows to disappear (I believe specifically if you add a row and type some content, then delete all the content, the row used to disappear). Can you confirm that is no longer the case since you changed to only passing `nonEmptyChallenges`? It seems like this bug would regress with this change

tommylau523: I had also discovered a bug where you cannot remove the last row, because there is a default row and empty rows are not counted as a "change". 

The new behavior is that, if you deleted all text in a row, it will treat it as if it is removed. In addition, if you first open the edit page, it will automatically add an empty row if none exists, but if you delete the last row, an empty row will not be automatically created.

This should clarify these cases.

ice1080: The original reason we added that comment and changed the functionality in the first place was if a user added a row, typed a letter, then hit backspace. It felt like bad UX to make the user click the Add Row button again just because they had a typo. 
The original behavior should have already had the empty row when first opening the edit page, I wonder if that was a regression at some point.

tommylau523: In this interaction, the user will not have to hit "Add Row" again, the empty row persists. The user just has to type back whatever they wanted to, or hit crtl+z since the textbox is the same one. 

What we couldn't do now on staging is actually deleting a row. If I have one challenge and mitigation row, deleting row or erasing the content do not accomplish this.

This implementation satisfies these constrats but do let me know if I am missing something:

- User sees an empty row with textboxs if there is no existing row (this prevents users from having to hit add row once)
- User sees existing ones and can edit/delete as required.
    - One change here is that, if user deleted every text within the row, this is equivalent as deleting the row.
- User can still hit Add Row/Delete Row, which updates the number of text boxes they see. 
    - Slight difference is that, if the user deletes the last row, currently on staging, a row with empty text boxes is immediately added back, which feels clunky
- Whatever the user ends up submitting in the form, empty rows are filtered out, then they are what get saved.
    - This differs from existing implementation very slightly. The bug comes from `nonEmptyChallenges.length > 0 `. So I could actually never submit an empty list, which is why deleting the last row does nothing.

ice1080: Deleting the last row does feel clunky, I'm glad we fixed that. I think at one point it was a requirement so that users didn't have to click Add again; essentially every field was supposed to have an editable box, but I agree that it was an odd experience and think we're moving in the right direction. 

I think we might have just been misunderstanding each other. You had originally said
> if you deleted all text in a row, it will treat it as if it is removed

And I assumed that meant that the row was removed, but I believe you were saying that if the user saves after that then what gets saved is as if the row was removed. I definitely agree with this behavior and again am glad we're fixing some of the odd behavior that probably just was not tested sufficiently before!

ice1080 (approved):

src/components/AcceleratorReports/ChallengesMitigationBox.tsx
tommylau523: Hahaha "it" does feel unclear now. The form will persists the empty row, but the API submit will treat the row as deleted. I should be better about ambigious pronouns in my comments going forward.

PR #6220, opened 2026-08-07
SW-8731 Show message when attempting to Save Planting Dates with species that don't have quantities set

constanzauanini:


src/scenes/NurseryRouter/PlantingSeasons/PlantingDatesTab.tsx
```diff
@@ -453,6 +454,7 @@ const PlantingDateForm = ({
   });
   const [validate, setValidate] = useState(false);
   const [notifyModalOpen, setNotifyModalOpen] = useState(false);
+  const [notSetWarningOpen, setNotSetWarningOpen] = useState(false);
```
ice1080, line 457: `setNotSetWarningOpen` is slightly confusing, but reading it a second time makes sense. Up to you if you want to change it, I can't think of a better name though

ice1080 (approved):

src/scenes/NurseryRouter/PlantingSeasons/PlantingDatesTab.tsx
constanzauanini: I think is ok since "Not set" is the term we are using

PR #6222, opened 2026-08-07
SW-8744 Move withdraw modal component out of the header to fix render issue

constanzauanini:


ice1080 (approved):
The playwright failure is related to withdraw so might need fixed

PR #6231, opened 2026-08-10
SW-8754 Add location to project/org pages

constanzauanini:


src/hooks/useBotanicalCountries.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const useBotanicalCountries = (skip = false): UseBotanicalCountriesResult => {
+  const [listBotanicalCountries, { data, isUninitialized }] = useLazyListBotanicalCountriesQuery();
+
+  useEffect(() => {
+    if (!skip) {
+      void listBotanicalCountries(undefined, true);
+    }
+  }, [listBotanicalCountries, skip, isUninitialized]);
```
ice1080, line 17: Should this only run if `isUninitialized` is true? It looks like this would get called twice right now in the network requests. 
The definition on their webpage states this:
> isUninitialized - When true, indicates that the query has not started yet.

Which seems like `isUninitialized` would change to false before the first request completed, meaning that the cache can't be used yet for the second request

constanzauanini: Yes, I think that's true. Will add isUninitialized to the condition

ice1080 (approved):

PR #6254, opened 2026-08-14
SW-8726 Connect FE components with BE api

constanzauanini:


ice1080:
Are there any natural places to split this one up? It's pretty large for reviewing

constanzauanini:
> Are there any natural places to split this one up? It's pretty large for reviewing

I created 3 separeted PRs. The last one is still pretty big, but couldn't find a better way

PR #6261, opened 2026-08-17
Remove unused type declarations

sgrimm:
Delete type declarations that aren't referenced anywhere.

This does not touch unused types in generated code; those types
will reappear every time the code generator is run.


ice1080 (approved):

sgrimm:
There's one unused type I intentionally didn't touch: the `SharedStringKeys` type in `src/strings/index.tsx`. I figured removing that one might mess up the effort to split web-components strings out of the code base.

PR #6286, opened 2026-08-18
Add a component test harness

alexTerraformation:
Create src/test-utils with a renderWithProviders that supplies the full stack needed for rendering components. Also add MSW for API mocking, and fixture builders for the common domain objects.

Four example tests are included, covering context, permissions, reads and writes.

Co-authored by Claude


src/test-utils/msw/setup.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Global MSW lifecycle, registered as an rstest setup file so no individual test has to start the
+ * server.
+ *
+ * `onUnhandledRequest: 'error'` is deliberate. Without it, a component that fires a request the
+ * test didn't anticipate hangs until the test times out and the failure points at the assertion
+ * rather than the missing mock. With it, you get the offending method and URL immediately.
+ */
+beforeAll(() => {
+  server.listen({ onUnhandledRequest: 'error' });
```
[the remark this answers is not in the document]
alexTerraformation: Confirmed and fixed

src/test-utils/renderWithProviders.tsx
```diff
@@ -0,0 +1,178 @@
...
+    ...overrides,
+  };
+};
+
+const buildLocalizationContext = (overrides: Partial<ProvidedLocalizationData> = {}): ProvidedLocalizationData => ({
+  bootstrapped: true,
+  activeLocale: 'en',
+  selectedLocale: 'en',
+  setSelectedLocale: () => undefined,
+  strings: defaultStrings,
```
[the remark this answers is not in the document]
alexTerraformation: Confirmed and fixed

src/scenes/InventoryRouter/view/InventorySummaryForNursery.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+  it('formats quantities using the active locale', async () => {
+    mockGet(SUMMARY_URL, { summary });
+
+    renderWithProviders(<InventorySummaryForNursery nurseryId={NURSERY_ID} />, {
+      localization: { activeLocale: 'fr', selectedLocale: 'fr' },
+    });
+
+    // French groups with a narrow no-break space rather than a comma. Match on the digits alone so
+    // the assertion doesn't depend on which space character the runtime's ICU data picks.
+    expect(await screen.findByText((content) => content.replace(/\s/g, '') === '12960')).toBeInTheDocument();
```
ice1080: I understand the comment and reasoning for removing the space character, but if the purpose is to check the active locale formatting, it seems kindof wrong to then only check the digits...

src/scenes/InventoryRouter/view/InventorySummaryForNursery.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  it('renders nothing until the summary arrives', () => {
+    mockGet(SUMMARY_URL, { summary });
+
+    renderWithProviders(<InventorySummaryForNursery nurseryId={NURSERY_ID} />);
+
+    expect(screen.queryByText(strings.TOTAL_QUANTITY)).not.toBeInTheDocument();
+  });
```
ice1080: I've seen too often where only testing a negative case actually hides an instance where the positive was never there. Let's add this to one of the other tests:
```ts
expect(screen.queryByText(strings.TOTAL_QUANTITY)).toBeInTheDocument();
```

AGENTS.md
```diff
@@ -29,6 +29,8 @@ See `package.json` for the latest list of dependencies and versions.
 - Prefer using existing components from `@terraware/web-components`
 - Use MUI components for UI only when no suitable `@terraware/web-components` option exists
 - Prefer existing utilities and helpers instead of introducing new ones
+- Write component tests with `renderWithProviders` from `src/test-utils`, and mock API responses
+  with its MSW helpers rather than stubbing hooks or services — see `src/test-utils/README.md`
```
ice1080: I think this should mention a new agent file instead of directly controlling the writing of component tests

rstest.config.ts
```diff
@@ -3,13 +3,31 @@ import path from 'path';
 
 export default defineConfig({
   testEnvironment: 'jsdom',
-  exclude: ['playwright'],
-  setupFiles: ['./src/setupTests.js'],
+  // Playwright specs use a different `test` global and can't run here. The globs are recursive
+  // because copies of the tree show up under agent worktrees and yalc links, and a bare 'playwright'
+  // only matches the top-level directory.
+  exclude: ['**/playwright/**', '**/.claude/**', '**/.yalc/**', '**/node_modules/**', '**/dist/**'],
+  setupFiles: ['./src/setupTests.js', './src/test-utils/setupStrings.ts', './src/test-utils/msw/setup.ts'],
```
ice1080: Because we already have a `setupTests.js`, I think it'd be better to move the calling of `setupStrings` and `msw/setup` to that instead of continuing to add more setupFiles to the list

src/test-utils/README.md
```diff
@@ -0,0 +1,176 @@
...
+`buildSeedBank`, `buildSpecies`, `buildSpeciesList`, `buildWithdrawalRow`. Each takes an overrides
+object and fills in plausible defaults, so a test names only what it cares about:
+
+```ts
+buildOrganization({ role: 'Contributor', facilities: [] });
+```
+
+Add a builder here the first time a second test needs the same shape.
+
+## Exemplars
```
ice1080: No need to change this, I just find it funny that I don't think I've ever seen a human use this word

src/test-utils/README.md
```diff
@@ -0,0 +1,176 @@
...
+Note that `onUnhandledRequest: 'error'` alone would _not_ do this. MSW turns an unhandled request
+into a rejected fetch, and RTK Query stores that as ordinary error state, so a test that doesn't
+assert on the affected query would stay green. The `afterEach` check is what makes the guarantee
+real: add an endpoint to a component and every test rendering it fails until the mock exists.
+
+## Fixtures
+
+`buildUser`, `buildAcceleratorAdmin`, `buildFunderUser`, `buildOrganization`, `buildFacility`,
+`buildSeedBank`, `buildSpecies`, `buildSpeciesList`, `buildWithdrawalRow`. Each takes an overrides
+object and fills in plausible defaults, so a test names only what it cares about:
```
ice1080: I think it'd be better to explain these at a high level and only provide a couple examples, since this line becomes out of date as soon as a new fixture is added

src/test-utils/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+| Option           | Purpose                                                                     |
+| ---------------- | --------------------------------------------------------------------------- |
+| `currentUser`    | Overrides for `useUser()`.                                                  |
+| `organization`   | Overrides for `useOrganization()`.                                          |
+| `localization`   | Overrides for `useLocalization()` — locale, countries, time zones.          |
+| `fundingEntity`  | Overrides for `useUserFundingEntity()`.                                     |
+| `preloadedState` | Initial redux state, for components reading a legacy feature slice.         |
+| `store`          | Supply your own store to dispatch before rendering or share across renders. |
+| `route` / `path` | Initial URL and route pattern, so `useParams()` resolves.                   |
```
ice1080: I think these are pretty self explanatory, and `renderWithProviders` should be the source of truth. Otherwise this list becomes out of date very quickly which causes the whole file's value to be thrown into question

ice1080 (changes requested):
Because we already have a `test-agent.md` for playwright tests, I think we should rename that to `playwright-test-agent.md` and create a new `component-test-agent.md` that knows how to write these tests. Both agent files should have the info for when they should be used as opposed to the other one. `component-test-agent.md` can reference `src/test-utils/README.md` since some of the info is already in there

src/test-utils/README.md
alexTerraformation: I literally asked Claude during development why it kept using that word.

src/test-utils/README.md
alexTerraformation: Done. Worth noting that our sort-imports plugin reorders these, so MSW's afterEach ended up ahead of the locale reset. Instead I moved the reset to beforeEach so each test starts in English regardless of how the last one ended.

src/test-utils/README.md
alexTerraformation: Done

rstest.config.ts
alexTerraformation: Good call. Done.

AGENTS.md
alexTerraformation: Re-worked the agent files to better align with your suggestion

src/scenes/InventoryRouter/view/InventorySummaryForNursery.test.tsx
alexTerraformation: Good call, added.

src/scenes/InventoryRouter/view/InventorySummaryForNursery.test.tsx
alexTerraformation: Good catch. Addressed.

.claude/agents/playwright-test-agent.md
```diff
@@ -1,22 +1,37 @@
...
+- Create, update, and debug unit tests for non-rendering code in `src/`
 - Reuse shared test utilities before adding new helpers
 - Read the source code and existing tests before making changes
 - Run the relevant test commands and `yarn format` after updates
 
 ## Project context
 
 - Unit tests use Rstest (`@rstest/core`), not Jest
-- Unit tests are co-located as `src/**/*.test.ts` and `src/**/*.test.tsx`
+- Unit tests are co-located as `src/**/*.test.ts` (some older files use `.spec.ts`)
```
ice1080, line 34: Noting this (not for this PR) - we should probably normalize these so there aren't two file endings

ice1080 (approved):

ice1080:
For the yarn.lock conflicts, just accept what's on main and then run `yarn install` again

PR #6287, opened 2026-08-19
SW-8795 Make updates to the Accept Changes flow

constanzauanini:


src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
```diff
@@ -95,6 +97,9 @@ const SpeciesCheckModal = ({
   const targetsRef = useRef(targets);
   targetsRef.current = targets;
 
+  const speciesRef = useRef(species);
+  speciesRef.current = species;
```
ice1080, line 101: It feels like claude has been using a lot of refs lately. Do you understand why the ref is needed here instead of just using `species`?

constanzauanini: It is needed. The use effect in the next line has on the dependency array the "open" value, so it "resets" everything when open status changes. If I use species in the reset (instead of the ref), I should include species in the array of dependencies and that will reset the modal when accepting the name suggestions or when overriding (because those api calls invalidates species), and then the modal will be reset when we don't want to (going back to first step for example)

ice1080 (approved):

PR #6294, opened 2026-08-19
SW-8710 Implemented report CSV export

tommylau523:


tommylau523:
* **#6306** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6306?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6294** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6294?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6294?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6283** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6283?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6282** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/scenes/AcceleratorRouter/AcceleratorProjects/Reports/FunderReportPageV2.tsx
```diff
@@ -25,6 +24,8 @@ type FunderReportPageV2Props = {
   projectId: number;
   report?: AcceleratorReportPayload | PublishedReportPayload;
   reportId: number;
+  /** page-level actions, e.g. the CSV export */
```
ice1080: NIT `rightComponent` is already a common prop and doesn't need a comment

src/components/AcceleratorReports/utils.ts
ice1080, line 1: Should any of these be unit tested since there's a decent amount of logic getting added to the FE to distinguish indicator classes?

src/components/AcceleratorReports/ReportExportMenu.tsx
ice1080, line 1: Comments seem unnecessary and/or fragile here

src/components/AcceleratorReports/useExportReportCsv.ts
```diff
@@ -0,0 +1,469 @@
...
+    [strings]
+  );
+
+  const makeIndicatorsCsv = useCallback(
+    (params: ReportCsvParams) => {
+      const columnKeys = getIndicatorCsvColumns(params.audience);
+      const columns: ColumnHeader[] = columnKeys.map((key) => ({ key, displayLabel: indicatorColumnLabels[key] }));
+
+      const rows = makeIndicatorCsvRows(params).map((row) =>
+        Object.fromEntries(columnKeys.map((key) => [key, localizeValue(row[key])]))
```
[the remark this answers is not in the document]
tommylau523: The sheet will not have other data for an attack to happen. Although this could be true.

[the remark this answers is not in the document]
tommylau523: In addition, CSV itself has no concepts of formulas.

src/components/AcceleratorReports/utils.ts
```diff
@@ -7,14 +7,200 @@ import {
...
+  };
+
+  return unpublishedProperties.map((property) => propertyLabels[property] ?? property).join(', ');
+};
+
+export const acceleratorReportStatusLabel = (
+  status: AcceleratorReportStatus | undefined,
+  strings: ILocalizedStrings,
+  isConsoleView?: boolean
+): string => {
```
ice1080, line 143: ReportStatus, IndicatorStatus, IndicatorCategory, and IndicatorLevel are all localizable enums in the BE, so I don't think these are necessary?

src/components/AcceleratorReports/utils.ts
```diff
@@ -7,14 +7,200 @@ import {
...
+    case 'Submitted':
+      return strings.SUBMITTED;
+    default:
+      return '';
+  }
+};
+
+export const metricStatusLabel = (status: MetricStatus | undefined, strings: ILocalizedStrings): string => {
+  switch (status) {
+    case 'Achieved':
```
ice1080: I think this should be renamed to IndicatorStatus

src/components/AcceleratorReports/utils.ts
tommylau523: This is true if we used search API to retrieve the enum names. But this CSV is built using the FE.

tommylau523: I remember why I didn't use the Search API. It is because we are still missing achievements, challenge and mitigations. In addition, a lot of indicators include runtime calculations such as YTD progress and previous year proresses.

ice1080: Probably makes sense to add achievements and challenges/mitigations to the search api eventually, but ya those calculations would add a lot of complexity (at the same time it would probably have better unit tests in the BE than in the FE)

tommylau523: I think so. BE does have unit tests for them actually. They are just calculated at runtime instead of stored in tables, which Search API accesses directly. 

I think we can move to using Search API for all fields except for indicators. But unless we have some dedicated endpoints to fetch only indicators, we ended up double fetching a lot of these values if we used a Search + GET approach.

src/components/AcceleratorReports/utils.ts
tommylau523: Making a change now. I will also add another followup PR to finish the metric -> indicator rename, since there are some copy texts + types leftovers.

ice1080 (approved):

tommylau523:
### Merge activity

* **Aug 21, 10:22 PM UTC**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-web/6294).

PR #6297, opened 2026-08-19
Copy viability observations array before mutating

alexTerraformation:
Prevents an error caused when trying to mutate RTK query cached data


ice1080 (approved):
Another option is changing .reversed to .toReversed which creates a copy

PR #6316, opened 2026-08-25
SW-8802 Get thumbnail planting site SVGs from BE

constanzauanini:
Removed all client-side Mapbox logic to generate thumbnail images and replaced it with the new thumbnail SVG endpoint


tommylau523:
I added the Vercel preview tag to check on Vercel.

tommylau523 (approved):

src/queries/extensions/siteThumbnail.ts
```diff
@@ -0,0 +1,16 @@
+import { api } from '../generated/siteThumbnail';
+
+api.enhanceEndpoints({
+  endpoints: {
+    getPlantingSiteThumbnail: {
```
tommylau523, line 5: Let's add a new tag to this endpoint. Maybe PlantingSiteGeometry. We can potentially only invalidate this tag if a site geometry is updated.

tommylau523: I just realized we don't actually have planting sites endpoints yet in RTK haha.

constanzauanini: We can add the tag later then, when implementing rtk in planting sites

PR #6317, opened 2026-08-25
SW-8815 Setting location during species check hides native/invasive changes

constanzauanini:


src/components/ProjectView/ProjectEditModal.tsx
```diff
@@ -46,13 +48,13 @@ export default function ProjectEditModal({ open, onClose, project, reload }: Pro
...
-      return;
-    }
+  const locationChanged =
+    showProjectLocation &&
+    !!project?.botanicalCountryCode &&
+    (countryCode !== project?.countryCode || botanicalCountryCode !== project?.botanicalCountryCode);
 
-    if (!name) {
-      setValidateFields(true);
+  const performSave = () => {
```
tommylau523: should this and saveProject be combined and also useCallback

src/scenes/OrganizationRouter/EditOrganizationModal.tsx
```diff
@@ -92,6 +94,23 @@ export default function EditOrganizationModal({
     setOrganizationTypeError('');
   };
 
+  const locationChanged =
+    showBotanicalCountry &&
+    !!organization.botanicalCountryCode &&
+    (organizationRecord.countryCode !== organization.countryCode ||
+      organizationRecord.botanicalCountryCode !== organization.botanicalCountryCode);
+
+  const performSave = async () => {
```
tommylau523: This should be a callback, and we have a RTK for update organization I think.

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
```diff
@@ -294,7 +328,8 @@ const SpeciesCheckModal = ({
       if (targets.some((target) => target.isOrg) && selectedOrganization) {
         await reloadOrganizations(selectedOrganization.id);
       }
-      reloadSpecies();
+      await reloadSpecies();
```
tommylau523, line 331: I think we have species RTK setup so we shouldn't need explicity reloadSpeceis calls. but worthwhile to check if we need some tag invalidations.

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
```diff
@@ -294,7 +328,8 @@ const SpeciesCheckModal = ({
       if (targets.some((target) => target.isOrg) && selectedOrganization) {
         await reloadOrganizations(selectedOrganization.id);
```
tommylau523, line 329: don't need this if we use updateOrganization RTK mutations

src/scenes/Species/SpeciesListView.tsx
```diff
@@ -112,7 +112,7 @@ const ProblemsCellComponent = ({ row, reloadData, onRowClick }: ProblemsCellProp
 };
 
 type SpeciesListProps = {
-  reloadData: () => void;
+  reloadData: () => Promise<void>;
```
tommylau523: Don't need this if we use RKT query

src/scenes/Species/SpeciesListView.tsx
```diff
@@ -436,7 +436,7 @@ export default function SpeciesListView({ reloadData, species }: SpeciesListProp
```
tommylau523, line 381: <p class='graphite__hidden'><i>[Re: line +302]</i></p>
We should move this search function to RTK Query. 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6317?utm_source=unchanged-line-comment">Graphite</a>.</p>

tommylau523 (changes requested):
I think if we fix up the SearchSpecies and be consistent with RTK Query usages, we can likely get rid of a lot of the reload props.

src/scenes/Species/SpeciesListView.tsx
```diff
@@ -112,7 +112,7 @@ const ProblemsCellComponent = ({ row, reloadData, onRowClick }: ProblemsCellProp
 };
 
 type SpeciesListProps = {
-  reloadData: () => void;
+  reloadData: () => Promise<void>;
   species: Species[];
 };
```
tommylau523: <p class='graphite__hidden'><i>[Re: line +163]</i></p>
We pass in species here from the organization species hook directly, so we can likely use the hook directly instead. 

But also, we actully use the Search API when applying the filters. It seems like we should either commit to using Search, or just use FE filtering. 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6317?utm_source=unchanged-line-comment">Graphite</a>.</p>

src/scenes/Species/SpeciesListView.tsx
tommylau523: Or not use search but use FE filtering directly. We sort of have the worst of both worlds here, where we load in all species for the organization, then still make search API requests per param change.

src/scenes/OrganizationRouter/EditOrganizationModal.tsx
constanzauanini: I didn't find rtkQuery for organization update

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
constanzauanini: This is true. The invalidations are already there. I will remove this manual reloads. For "onSetLocation" I will keep it since updateOrganization is not in rtk

src/scenes/Species/SpeciesListView.tsx
constanzauanini: I will keep it just for the org update

src/scenes/Species/SpeciesCheck/SpeciesCheckModal.tsx
constanzauanini: We can add rtkQuery for this later, not implemented yet

tommylau523 (approved):

PR #6333, opened 2026-08-29
People list view component test

alexTerraformation:


src/test-utils/msw/server.ts
```diff
@@ -1,10 +1,35 @@
...
+ * worse than no test.
+ *
+ * What qualifies instead is infrastructure noise: a request that every page makes, that carries no
+ * domain data, and whose response no assertion should ever depend on.
+ *
+ * - `/build-version.txt` — `PageSnackbar` renders `DetectAppVersion`, which polls the deployed build
+ *   version once a minute to offer an "upgrade available" banner. 36 components mount PageSnackbar,
+ *   so essentially every scene-level test would otherwise have to mock it. An empty body reads as
+ *   "no newer build" (`useAppVersion` treats a falsy version as not stale), so the banner stays out
+ *   of the way. A test that wants to exercise the banner overrides this with `server.use(...)`.
```
ice1080: This seems to have some implementation-specific verbiage in it. Perhaps we can just say that this endpoint is used often so should be mocked

src/scenes/PeopleRouter/PeopleListView.test.tsx
```diff
@@ -0,0 +1,430 @@
...
+  renderWithProviders,
+  server,
+} from 'src/test-utils';
+
+const ORG_ID = 1;
+const SEARCH_URL = '/api/v1/search';
+const ORG_ROLES_URL = `/api/v1/organizations/${ORG_ID}/roles`;
+const ORG_URL = `/api/v1/organizations/${ORG_ID}`;
+const ORG_USER_URL = `/api/v1/organizations/${ORG_ID}/users/:userId`;
+const userUrl = (userId: string) => `/api/v1/organizations/${ORG_ID}/users/${userId}`;
```
ice1080, line 26: Could alternatively add these to `constants.ts`. Optional, not required for approval

ice1080 (changes requested):

src/scenes/PeopleRouter/PeopleListView.test.tsx
alexTerraformation: Leaving these local for now since they don't quite match the pattern we have in constants.ts currently (MSW routes vs media paths). Happy to move them to src/test-utils if other tests start needing the same ones.

ice1080 (approved):

PR #6344, opened 2026-09-01
Species detail view component test

alexTerraformation:
Pulled out a couple common utilities into shared files

### DIFF
--- a/src/scenes/Species/SpeciesDetailView.test.tsx
+++ b/src/scenes/Species/SpeciesDetailView.test.tsx
@@ -10,6 +10,7 @@ import {
   buildOrganization,
   buildSpecies,
   captureRequests,
+  dialogTitled,
   mockError,
   mockGet,
   renderWithProviders,
@@ -47,7 +48,16 @@ type RenderOptions = {
 const renderDetailView = ({ inUse = [], organization }: RenderOptions = {}) => {
   mockGet(PROJECTS_URL, { projects: [] });
   mockGet(`${SPECIES_URL}/${SPECIES_ID}`, { species: SPECIES });
-  mockGet(SPECIES_URL, { species: inUse });
+
+  // Matching on the query keeps this fixture from also answering a plain species list, which would
+  // surface as an unhandled request instead — nothing here asks for one today.
+  server.use(
+    http.get(SPECIES_URL, ({ request }) =>
+      new URL(request.url).searchParams.get('inUse') === 'true'
+        ? HttpResponse.json({ status: 'ok', species: inUse })
+        : undefined
+    )
+  );
 
   const reloads: boolean[] = [];
 
@@ -85,18 +95,7 @@ const optionsMenuButton = (): HTMLElement => {
   return button as HTMLElement;
 };
 
-/**
- * The open dialog with this title, so its Delete button is told apart from the destructive Delete
- * item in the options menu behind it. `DialogBox` renders no ARIA dialog role to scope by.
- */
-const dialogTitled = (title: string): HTMLElement => {
-  const box = screen.getByText(title).closest('.dialog-box');
-  if (!box) {
-    throw new Error(`No open dialog titled "${title}"`);
-  }
-  return box as HTMLElement;
-};
-
+/** Scoped so the dialog's Delete button is told apart from the Delete menu item behind it. */
 const deleteDialog = () => dialogTitled(strings.DELETE_SPECIES);
 
 const currentPath = () => screen.getByTestId('location').textContent;
@@ -106,12 +105,6 @@ const openDeleteDialog = async (user: ReturnType<typeof renderDetailView>['user'
   await user.click(await screen.findByRole('menuitem', { name: strings.DELETE }));
   await screen.findByText(strings.DELETE_SPECIES);
 };
-
-// PageSnackbar mounts the app-version poller on every page. An empty body means "no newer build",
-// which keeps the upgrade banner out of the way of what these tests are looking at.
-beforeEach(() => {
-  server.use(http.get('/build-version.txt', () => HttpResponse.text('')));
-});
 
 describe('SpeciesDetailView', () => {
   describe('permission gating', () => {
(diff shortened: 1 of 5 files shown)


src/test-utils/README.md
```diff
@@ -121,6 +121,14 @@ into a rejected fetch, and RTK Query stores that as ordinary error state, so a t
 assert on the affected query would stay green. The `afterEach` check is what makes the guarantee
 real: add an endpoint to a component and every test rendering it fails until the mock exists.
 
+## Queries
+
+`src/test-utils/queries.ts` holds queries for markup Testing Library can't reach on its own, where a
+component renders without the roles or accessible names a normal query needs.
+
+See the file for what exists. Add one there the first time a second test needs the same handle, and
+export it from `index.ts`.
```
ice1080: I think this line is unnecessary

src/test-utils/queries.ts
```diff
@@ -0,0 +1,19 @@
+import { screen } from '@testing-library/react';
+
+/**
+ * The open dialog with this title, to scope queries to it. `DialogBox` renders no ARIA dialog role,
+ * so its class is the only handle a test has — which matters when the same label appears both inside
+ * the dialog and in the page behind it.
```
ice1080: We should probably use this comment to indicate that we need to add an ARIA dialog role to `DialogBox`. That would make this function's implementation better without needing to change its usages

src/scenes/Species/SpeciesDetailView.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+  inUse?: Species[];
+  organization?: Organization;
+};
+
+/**
+ * Renders the detail view for {@link SPECIES}.
+ *
+ * The routes are declared here rather than through the `path` render option because deleting
+ * navigates back to the species list: the probe has to outlive the detail route to report where the
+ * user ended up, and the list route has to exist for the navigation to land somewhere.
```
ice1080: NIT clean this comment up to remove references to "rather than" which usually indicates historical code or references to the agent conversation

src/scenes/Species/SpeciesDetailView.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+ */
+const renderDetailView = ({ inUse = [], organization }: RenderOptions = {}) => {
+  mockGet(PROJECTS_URL, { projects: [] });
+  mockGet(`${SPECIES_URL}/${SPECIES_ID}`, { species: SPECIES });
+
+  // Matching on the query keeps this fixture from also answering a plain species list, which would
+  // surface as an unhandled request instead — nothing here asks for one today.
```
ice1080: Seems like "nothing here asks for one today" is unnecessary

src/scenes/Species/SpeciesDetailView.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+      await waitFor(() => expect(screen.queryByText(strings.SELECTED_SPECIES_UNUSED)).not.toBeInTheDocument());
+      expect(deletes).toHaveLength(0);
+      expect(currentPath()).toBe(`/species/${SPECIES_ID}`);
+    });
+
+    /**
+     * Documents what the view does today, which is not obviously what it should do: the error toast
+     * is the only signal that anything went wrong. The modal closes and the user is returned to the
+     * species list exactly as if the delete had succeeded, and the species is still listed there.
```
ice1080: Comment should be cleaned up from "today"

src/scenes/Species/SpeciesDetailView.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+const currentPath = () => screen.getByTestId('location').textContent;
+
+const openDeleteDialog = async (user: ReturnType<typeof renderDetailView>['user']) => {
+  await user.click(optionsMenuButton());
+  await user.click(await screen.findByRole('menuitem', { name: strings.DELETE }));
+  await screen.findByText(strings.DELETE_SPECIES);
+};
+
+describe('SpeciesDetailView', () => {
+  describe('permission gating', () => {
```
ice1080, line 110: I'm always on the fence about tests like this for permissions. To me it seems like we would be better with thorough unit tests on the permissions code, but I could be convinced that these are needed

ice1080 (changes requested):

src/test-utils/queries.ts
alexTerraformation: Are you saying that you would like me to add the ARIA dialog role to `DialogBox` as a part of this PR?

src/scenes/Species/SpeciesDetailView.test.tsx
alexTerraformation: In this case I think we're actually testing that the component uses the ACL to do the right thing (e.g. hiding the edit button while still showing the menu button). Can't test that at the ACL level.

src/test-utils/queries.ts
alexTerraformation: For now I updated the comment and added a Jira ticket to add the ARIA dialog role

src/scenes/Species/SpeciesDetailView.test.tsx
ice1080: Yes and no. You're referring to how the ACLs are used inside the component, but we have functions like this in acl.ts `isAllowedCreatePlantingSite`, and we can test those functions. 

In this specific case, I see that we directly used the `isContributor` function in `SpeciesDetailView`, which I think we should move away from. 

I'll approve this PR for now, but I think we'd be better off moving all ACL usage to the functions in acl.ts and unit testing those. Otherwise we end up basically duplicate testing the same permission functions in all their usages which seems like an unnecessary slowdown of tests

src/test-utils/queries.ts
ice1080: Perfect, that's what I was gonna suggest!

ice1080 (approved):

src/scenes/Species/SpeciesDetailView.test.tsx
alexTerraformation: Fair point. Moving all of these to the ACL would make sense. I can create a ticket for that.

PR #6345, opened 2026-09-01
Planting Goals Tab: Show spinner only on first load

constanzauanini:
This is to avoid showing the spinner in every density edit

### DIFF
--- a/src/hooks/usePlantingSite.ts
+++ b/src/hooks/usePlantingSite.ts
@@ -23,7 +23,8 @@ const usePlantingSite = (plantingSiteId: number | undefined) => {
 
   return {
     plantingSite,
-    isLoading: getPlantingSiteResponse.isFetching,
+    isLoading: getPlantingSiteResponse.isLoading,
+    isFetching: getPlantingSiteResponse.isFetching,
     isSuccess: getPlantingSiteResponse.isSuccess,
   };
 };
--- a/src/scenes/PlantingPlansRouter/PlantingPlanDetailsView.tsx
+++ b/src/scenes/PlantingPlansRouter/PlantingPlanDetailsView.tsx
@@ -32,7 +32,7 @@ const PlantingPlanDetailsView = (): JSX.Element => {
 
   const params = useParams<{ plantingSiteId: string }>();
   const plantingSiteId = Number(params.plantingSiteId);
-  const { plantingSite, isLoading } = usePlantingSite(plantingSiteId);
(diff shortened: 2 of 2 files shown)


src/scenes/PlantingPlansRouter/PlantingPlanDetailsView.tsx
```diff
@@ -93,7 +93,7 @@ const PlantingPlanDetailsView = (): JSX.Element => {
     [description, plantingSite, segment, segments, selectSegment, strings, theme]
   );
 
-  if (isLoading || !plantingSite) {
+  if (!plantingSite) {
```
tommylau523: depending on whether usePlantingSite actually uses `data` vs `currentData`, this may not actually change anything. My recommendation is actually to use both `isLoading` and `isFetching` from the RTK Query accordingly. I think `usePlantingSite` right now exposes `isFetching` as `isLoading`.  
  
The difference is that, under RTK Query, `isFetching` is `true` whenever there is a request in flight, so any invalidations or reloads will cause this to flip. `isLoading` is only `true` for the first load.

constanzauanini: Interesting, the isLoading from usePlantingSite hook internally uses isFetching. I will fix it to use it the real isLoading then

src/hooks/usePlantingSite.ts
```diff
@@ -23,7 +23,8 @@ const usePlantingSite = (plantingSiteId: number | undefined) => {
 
   return {
     plantingSite,
-    isLoading: getPlantingSiteResponse.isFetching,
+    isLoading: getPlantingSiteResponse.isLoading,
```
tommylau523, line 26: This change is okay with me. We likely should change existing usages of isLoading to be isFetching to ensure correct behavior.

For context, we often want a loading state to appear if isFetching is true. For example, if I updated a planting site's name, the site detail page wants to finish getting the new name before displaying it, to avoid showing the old name before the new request completes.

I do see that planting goals need different behaviors and therefore, functionally, this change is okay. Perhaps we just need to have a conversation with the team regarding what each api request state should be named, so that we can be consistent going forward.

tommylau523 (approved):

src/hooks/usePlantingSite.ts
constanzauanini: I checked the other uses of isLoading and I thought we wanted to use isLoading on those too, but I can change them to isFetching

PR #6347, opened 2026-09-01
SW-8835 Planting Seasons tab: Withdrawn for Planting number in seasons card list is missing link

constanzauanini:


src/scenes/NurseryRouter/PlantingSeasons/PlantingSeasonBox.tsx
```diff
@@ -75,9 +87,11 @@ const PlantingSeasonBox = ({
     label: string,
     value: number | undefined,
     align: 'left' | 'center' | 'right' = 'right',
-    highlight = false
+    highlight = false,
+    to?: string
```
tommylau523: I am not sure if this is the best prop name. Perhaps url or even destination is a better name.

tommylau523 (approved):

PR #6368, opened 2026-09-03
Remove species intelligence feature flag

constanzauanini:


src/scenes/Species/SpeciesDetailView.tsx
```diff
@@ -126,10 +124,7 @@ export default function SpeciesDetailView({ reloadData }: SpeciesDetailViewProps
     [gridSize, theme]
   );
 
-  const dataSource = useCallback(
-    (source?: SpeciesDataSourcePayload) => (speciesIntelligenceEnabled ? source : undefined),
-    [speciesIntelligenceEnabled]
-  );
+  const dataSource = useCallback((source?: SpeciesDataSourcePayload) => source, []);
```
ice1080: It seems like this function doesn't do anything anymore and should be removed

ice1080 (changes requested):
I couldn't tell if the codex comments were actually valid or not. They seemed to be addressing something that already existed

constanzauanini:
> I couldn't tell if the codex comments were actually valid or not. They seemed to be addressing something that already existed

I will fix them since they are kind of releated

ice1080 (approved):

PR #6369, opened 2026-09-03
Remove planting goals feature flag

constanzauanini:


src/scenes/OrgRouter/NavBar.tsx
```diff
@@ -412,7 +410,7 @@ export default function NavBar({
             id='plants-dashboard'
           />
 
-          {isEnabled('Planting Goals') && hasPlantingSites === true && hasSpecies === true && (
+          {hasPlantingSites === true && hasSpecies === true ? (
```
ice1080: I think we can leave the `&&` and remove the `<></>` here right?

ice1080 (approved):

PR #6389, opened 2026-09-09
SW-8636 New KB articles for new Planting Seasons & Goals and Inventory Planning features

constanzauanini:
TODO: There is still one KB article pending to add


ice1080 (approved):
What's the purpose of having separate `es` and `fr` sections if all of the links are always identical?

constanzauanini:
> What's the purpose of having separate `es` and `fr` sections if all of the links are always identical?

I think the idea was to have different links at some point, but not sure

PR #6404, opened 2026-09-11
SW-8883 Table Type Size Increase: 16px

constanzauanini:
When no font value was specified in tables, the font was 14px. Set it to 16px


tommylau523 (approved):
Changes look good. I am wondering if there is a MUI style we can set for the entire app instead.

constanzauanini:
> MUI style we can set for the entire app instead

Probably yes, but I wasn't sure if do it or no. Links and Texttruncated would still need an specific override

PR #6421, opened 2026-09-14
SW-8874 Fix reseting when unchecking withdraw all

constanzauanini:


src/scenes/AccessionsRouter/withdraw/WeightWithdrawal.tsx
```diff
@@ -151,19 +151,18 @@ export default function WeightWithdrawal(props: WeightWithdrawalProps): JSX.Elem
...
-      if (withdrawAll) {
-        if (!withdrawAllWeight) {
-          return;
-        }
-        onUnitsUpdate(withdrawAllWeight.units);
-        onChangeAmount(withdrawAllWeight.quantity, withdrawAllWeight.units);
-      } else {
-        onChangeAmount(undefined, units);
+      if (!withdrawAll) {
+        setWithdrawAllSelected(false);
```
ice1080, line 155: Does this need to call `onChangeAmount(undefined...)`?

constanzauanini: No, that resets the estimate count to 0, and that is what we don't want to do

ice1080 (approved):

PR #6423, opened 2026-09-14
SW-8884 Add Collector(s) as an optional column for the accessions table

constanzauanini:


src/hooks/useTableState.ts
```diff
@@ -43,7 +43,10 @@ const useTableState = (storageKey: string, options?: UseTableStateOptions) => {
   const [columnVisibility, setColumnVisibility] = useState<MRT_VisibilityState>(() => {
     try {
       const saved = localStorage.getItem(`${storageKey}-columnVisibility`);
-      return saved ? (JSON.parse(saved) as MRT_VisibilityState) : defaultColumnVisibility;
+      if (!saved) {
+        return defaultColumnVisibility;
+      }
+      return { ...defaultColumnVisibility, ...(JSON.parse(saved) as MRT_VisibilityState) };
```
ice1080, line 49: Doesn't this change mean that the default columns can't be unselected? It seems like with this change, if columns A, B, and C are defaults, and the user adds column D and unchecks B, then the table should show A, C, D. With this change it seems like if the user then navigates away from this page and comes back, it would show B again even though the user unselected it. Is that intentional?

constanzauanini: No, defaultColumnVisibility contains the hidden columns by default only (not the visible). So if A, B and C are defaults (visible) they don't appear in defaultColumnVisibility. Then user adds D and remove B, the saved will contains: saved = { D: true, B: false }.

The merge would be:

```
 { ...defaultColumnVisibility, ...saved }

= { ...{ D: false }, ...{ D: true, B: false } }

= { D: true, B: false }        // B stays hidden ✓
```

A column with no entry is treated as visible.

ice1080: That makes total sense; these are objects with column visibilities, not just lists of visible columns. Thanks for explaining!

ice1080 (approved):

PR #6425, opened 2026-09-14
SW-8878 Show suggestions count in NameCheck step

constanzauanini:


ice1080 (approved):
Minor typo in title - Sueggestions -> Suggestions

PR #6433, opened 2026-09-14
Remove user info when changing email address

ice1080:
The console Add Person page allows global roles to be added to users, and new users to be invited to be part of the console. 
After typing in an existing user's email address, their info gets added to the form. 
If the email address is then changed to a non-existant email, the info never gets removed. 
Update this to remove their info when changing emails.

Also change this from a debounce to an on-blur effect for populating the info.

Co-Authored-By: GPT-6 <noreply@openai.com>

---


joshbuchea (commented):
This looks good, but there's un-addressed comment from Codex that seems valid.

ice1080:
> This looks good, but there's un-addressed comment from Codex that seems valid.

Thanks Josh! I had missed that comment. Resolved now

src/scenes/AcceleratorRouter/People/NewView.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+const expectEmptyPerson = () => {
+  expect(textbox('firstName')).toHaveValue('');
+  expect(textbox('lastName')).toHaveValue('');
+  expect(textbox('globalRole')).toHaveValue('');
+  expect(screen.queryByText(strings.GIS)).not.toBeInTheDocument();
+};
+
+describe('NewView', () => {
+  it.each(['user details', 'internal interests'])(
+    'keeps Save disabled until the existing account’s %s finish loading',
```
joshbuchea, line 58: Not sure if `%s` is intentional here, seems like it's an artifact from AI.

joshbuchea (approved):
Just one non-blocking comment, otherwise 🚢

src/scenes/AcceleratorRouter/People/NewView.test.tsx
ice1080: `%s` is a string formatter. This test checks that the save stays disabled until both of those 2 types of requests finish loading by creating 2 different tests - hence the `it.each(['user details', 'internal interests'])(...`. For each of those it's creating a separate test and running it, and `%s` is replaced with either 'user details' or 'internal interests'.

joshbuchea: I'm familiar with its usage, but didn't realize the test framework would replace it in the test name. I thought AI had left a placeholder behind. Nice!

PR #6434, opened 2026-09-14
Add error message when project name is in use

ice1080:
When a project name is already in use, the UI lets the user go all the way through the Create Project flow and then just displays "An error occurred". 
Add a (useful) error message to the first page of the flow if the user enters an extant name, and block the user from continuing the flow.

Co-Authored-By: GPT-6 <noreply@openai.com>

---


joshbuchea (commented):
Not sure if the comment from Codex is valid, it seems like it could be. I'll re-review the PR once that's been addressed.

ice1080:
> Not sure if the comment from Codex is valid, it seems like it could be. I'll re-review the PR once that's been addressed.

Thanks, addressed

joshbuchea (approved):

PR #6437, opened 2026-09-15
SW-8792 Add multi-step Withdraw Seeds dialog (behind feature flag)

constanzauanini:
Add the new 3-step (Purpose/Quantities/Photos) Withdraw Accession dialog supporting single and bulk same-species accession withdrawals


ice1080 (commented):
Request to split this up, it's over 2000 lines right now

PR #6452, opened 2026-09-17
Rank stratum survival by each stratum's own latest observation

tommylau523:
Each stratum is now resolved against its own latest observation, via two new search endpoints — one for strata, one for substrata (the map shades both). Each issues a single `POST /api/v1/search` covering every stratum of every site passed in, then collapses the rows to the most recently completed one per stratum.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tommylau523:
* **#6457** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6457?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6452** <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6452?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-web/6452?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/queries/search/strata.ts
```diff
@@ -115,9 +115,104 @@ const injectedRtkApi = api.injectEndpoints({
...
+              },
+              {
+                operation: 'field',
+                field: 'observation_state',
+                values: ['Completed', 'Abandoned'],
+              },
+            ],
+          },
+          sortOrder: [{ field: 'observation_completedTime' }],
+          count: 0,
```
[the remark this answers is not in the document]
tommylau523: Practically this is a non-issue.

src/queries/search/strata.ts
```diff
@@ -115,9 +115,104 @@ const injectedRtkApi = api.injectEndpoints({
...
+              plantDensity: plantDensity === undefined ? undefined : Number(plantDensity),
+              survivalRate: survivalRate === undefined ? undefined : Number(survivalRate),
+            });
+          }
+        });
+
+        return [...latestByStratumId.values()];
+      },
+      providesTags: (_results, _error, args) => [
+        { type: QueryTagTypes.Observation, id: 'LIST' },
```
ice1080, line 174: Won't this mean that the observations page won't show all results if this gets queried first because it's setting the `Observation` `LIST` tag? I think we might want this to be a new tag type of strictly `ObservationStratumResults` or something like that

src/queries/search/strata.ts
```diff
@@ -115,9 +115,104 @@ const injectedRtkApi = api.injectEndpoints({
...
+        ...args.plantingSiteIds.map((plantingSiteId) => ({
+          type: QueryTagTypes.PlantingSiteSurvivalRate,
```
ice1080, line 176: A similar question, this isn't querying the planting site survival rate, yet we're caching against that tag which seems wrong

src/queries/search/substrata.ts
```diff
@@ -31,9 +31,94 @@ const injectedRtkApi = api.injectEndpoints({
...
+          }
+        });
+
+        return [...latestBySubstratumId.values()];
+      },
+      providesTags: (_results, _error, args) => [
+        { type: QueryTagTypes.Observation, id: 'LIST' },
+        ...args.plantingSiteIds.map((plantingSiteId) => ({
+          type: QueryTagTypes.PlantingSiteSurvivalRate,
+          id: plantingSiteId,
```
ice1080, line 93: Same notes on the tags

src/scenes/PlantsDashboardRouter/components/PlantingDensityCard.tsx
```diff
@@ -40,9 +40,9 @@ export default function PlantingDensityCard({
           <Box flexBasis='100%' marginTop={isDesktop ? 0 : 4}>
             <Box display={'flex'} alignItems={'center'}>
               <Typography fontSize={'20px'} fontWeight={600} marginRight={1}>
-                {strings.OBSERVED_DENSITY}
+                {strings.SITE_PLANT_DENSITY}
```
ice1080: These changes feel unrelated, why are they in this PR?

tw-empty2.mjs
ice1080, line 1: What is this script?

ice1080 (commented):
I'm only about 1/4 through this PR, but wanted to post comments now while I change tasks to something else

tw-empty2.mjs
tommylau523: Removed.

src/queries/search/strata.ts
tommylau523: I don't think so. The tag will not prevent other queries from being fetched. Adding ObservationStratumResults won't help with much unless we are invalidating that tag. Although I think a blanket Observation tag should be enough in all these queries.

src/scenes/PlantsDashboardRouter/components/PlantingDensityCard.tsx
tommylau523: I split out hte plant density card changes, since it has the same issue/symptoms. I also want to be specific about using Platn Density vs Observed Density since the definitions are different.

tommylau523:
Superseded: replacing the search-based approach with the new /tracking/observations/results/stats endpoint. Reopening as a fresh 4-PR stack.

PR #6466, opened 2026-09-18
SW-8910 Add Species

constanzauanini:
Apply edit/save buttons pattern in Add species. Use Page component for that


src/components/Page/index.tsx
```diff
@@ -67,7 +67,12 @@ export default function Page({
   stickyHeaderElevated,
   subHeader,
 }: PageProps): JSX.Element {
-  const contentRef = useRef(null);
+  // A callback ref backed by state: when the content element mounts, Page re-renders and hands the
+  // real element to PageHeaderWrapper. A plain ref would leave nextElement null on the first paint,
+  // so a permanently fixed (sticky) header could not reserve its height and would overlap the first
+  // row until some unrelated re-render happened.
+  const [contentEl, setContentEl] = useState<HTMLElement | null>(null);
+  const contentRef = useCallback((node: HTMLElement | null) => setContentEl(node), []);
```
ice1080, line 75: I feel like I've seen other solutions for dynamic element height that were easier to understand than this, but I can't think of them now. I'll approve this but with the ask that we ask AI to come up with a different solution just to compare and see which one is simpler

ice1080 (approved):

src/components/Page/index.tsx
constanzauanini: Let me check if there is a simpler one

constanzauanini: According to AI this is the recommended solution. It says that we have other components passing nextElement={contentRef.current} with a plain useRef but that has the bug codex flagged. The other option is to make PageHeaderWrapper reserve its own space with an in-flow spacer <div style={{ height }} /> instead of mutating nextElement.style.marginTop, but that it has some risk to the scroll/animation/collapsible behavior.

ice1080: And just confirming, without the change you saw the issue that codex flagged, that on refresh the header overlaps the first row?

constanzauanini: Not in a typical use of the app, but I saw it when changing from one branch with out the change to this one with the new header without refreshing the page

